PR #170, opened 2023-06-22
feat(tags): added taxonomy tags to probes and detectors

shubhobm:
- Added attribute `tags` to the `Probe` and `Detector` base classes that adds a list of taxonomy tags, per the [MISP](https://github.com/MISP/misp-taxonomies) format
    - Tags categorize the vulnerability tested by a specific probe
    - Tags for a probe is not necessarily the union of its `recommended_detector` tags are not necessarily the same, since the same detector can be used to test different failure modes
- Added AVID and LMRC tags to all probes and detectors

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -20,6 +20,7 @@ class Detector:
     recall = 0.0
     accuracy = None
     active = True
+    tags = [] # list of taxonomy categories per the MISP format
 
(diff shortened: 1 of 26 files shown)


shubhobm:
I have read the CLA Document and I hereby sign the CLA

leondz:
great format. thanks for this!

Issue #62, opened 2023-05-26
hit log


leondz:
generator, probe, prompt, result, detector, score, run id (from log uuid, hey, let's use uuids for that..), attempt uuid

keep the hits in one place - can be used for training later

leondz:
require that #79 (probes state a goal) is done & the goal's included in the hit log

leondz:
completed with #179

PR #179, opened 2023-06-27
add hit log

leondz:
evaluation now keeps successful vulnerability access attempts (hits) in a separate logfile


leondz:
completes #62

Issue #180, opened 2023-06-27
buff: paraphrase


leondz:
easy fuzz

https://huggingface.co/tuner007/pegasus_paraphrase is a fine place to start, see https://github.com/leondz/prefstance/blob/main/paraphrase_questions.py

leondz:
merged in pr #333

PR #333, opened 2023-11-29
paraphrase buffs

leondz:
add two buffs - pegasus which is heavy, humarin which is OK on CPU - and tweak the buff architecture


leondz:
this resolves #180

PR #372, opened 2023-12-14
Add Base64 and CharCode buffs

erickgalinkin:


leondz (approved):
tight. lgtm. join(list(map(map(ord WHY IS PYTHON MAKING US WRITE LISP

erickgalinkin:
It's functional, Leon. Only a matter of time until OCaml wins the war.

PR #446, opened 2024-01-29
Tree of Attacks

erickgalinkin:
Initial commit containing full, untested implementation for tree of attacks.

Updated requirements.txt to include numpy.


erickgalinkin:
Need to test the code, run generation, and write probes.

leondz:
replaced pathlib call w/. garak._config.basedir

also - is there an interfaces to running full TAP locally via the attack manager? it looks like TAPProbe uses a cached set, which is great for getting off the gorund, but it'd be good to include a probe that runs full TAP

erickgalinkin:
> replaced pathlib call w/. garak._config.basedir
> 
> also - is there an interfaces to running full TAP locally via the attack manager? it looks like TAPProbe uses a cached set, which is great for getting off the gorund, but it'd be good to include a probe that runs full TAP

I'll update the probe with a `cached` parameter! Good thinking.

leondz (approved):
lgtm! 

the config route doesn't offer enough bandwidth for this plugin & needs revision, let's track that in another issue/pr

PR #459, opened 2024-02-06
Adds new encodings

zmackie:
This PR adds some new encodings:
- Zalgo
- Ecoji
- Nato phonetic alphabet

My aim is to add a few more, but I wanted to get feedback from the maintainers regarding direction.

### DIFF
--- a/garak/probes/encoding.py
+++ b/garak/probes/encoding.py
@@ -470,3 +470,18 @@ class InjectNato(BaseEncodingProbe, Probe):
     def __init__(self):
         BaseEncodingProbe.__init__(self)
         Probe.__init__(self)
+
+class InjectZalgo(BaseEncodingProbe, Probe):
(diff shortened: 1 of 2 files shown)


zmackie:

garak/probes/encoding.py
```diff
@@ -168,7 +168,7 @@ def braille(text):
         elif char in characterUnicodes:
             if char.isdigit():
                 if not is_number:
-                    is_number == True
+                    is_number = True
```
zmackie, line 171: I think this was a bug?

leondz:
Thanks! Excited to take a look at these - not around weds but ASAP after that :)

garak/probes/encoding.py
leondz: agree, thanks

leondz:
lgtm. thanks!

PR #476, opened 2024-02-12
add warning when using a lite/default profile

leondz:
user needs to know that passing a default scan isn't a big deal


erickgalinkin (commented):
lgtm

PR #477, opened 2024-02-13
OpenAI upgrade

erickgalinkin:
Update gitignore. Upgrade openai to version 1.12.0 and refactor OpenAIGenerator to handle new calling conventions. Add test_openai.py.


leondz (approved):
lgtm
* both pyproject.toml and requirements.txt need to have the requirements in them. this should be caught by a test but that test wasn't in this commit, maybe you need to do a pull?

Issue #469, opened 2024-02-12
probe: LRL jailbreak


leondz:
paper: [Low-Resource Languages Jailbreak GPT-4](https://arxiv.org/abs/2310.02446)

leondz:
paper finds Zulu `zu`, Scots Gaelic `gd`, Hmong `hmn` and Guarani `gn` get OK bypass rates - inclined to leave this issue open until we get some of these (or equiv) languages included

PR #478, opened 2024-02-14
Low Resource Languages Buff

erickgalinkin:
Add low resource languages buff, include deepl dependency in requirements.txt

Addresses issue #469


leondz:
nice. fixed a behaviour where yielding an attempt and then changing it led to the original attempt also being changed, which meant all translated attempts ended up containing the same (last in list) language - and added a pattern indicating how to avoid this, in to the base class

leondz (approved):

Issue #377, opened 2023-12-15
buffs: support multiple buffs


leondz:
run in sequence, or run all x all, or run all combinations? in which order?

leondz:
IN SEQUENCE, ARE YOU MAD, comb exp gets wiiiiild with these

might make sense to only do buff combining when we have a smarter way of getting through the exploit setup tree

PR #497, opened 2024-02-20
support multiple buffs

leondz:
addressing  #377


erickgalinkin (approved):
Looks good! Failing test is openai which is fixed in different (merged) PR.

PR #506, opened 2024-02-21
Add post buff hook

erickgalinkin:
Add untransform to buffs.base (placeholder) and low_resource_languages.py. Add post_buff_hook state tracking in Probe base class to flag the need to call _postprocess_buff, which calls buff.untransform on the attempt.

### DIFF
--- a/garak/buffs/base.py
+++ b/garak/buffs/base.py
@@ -36,6 +36,7 @@ class Buff:
     def __init__(self) -> None:
         module = self.__class__.__module__.replace("garak.buffs.", "")
         self.fullname = f"{module}.{self.__class__.__name__}"
+        self.post_buff_hook = False
         print(
             f"🦾 loading {Style.BRIGHT}{Fore.LIGHTGREEN_EX}buff: {Style.RESET_ALL}{self.fullname}"
         )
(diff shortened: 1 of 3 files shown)


garak/buffs/base.py
```diff
@@ -72,6 +73,12 @@ def transform(
         """attempt copying is handled elsewhere. isn't that nice"""
         yield self._derive_new_attempt(attempt)
 
+    def untransform(
+        self, attempt: garak.attempt.Attempt
+    ) -> garak.attempt.Attempt:
+        """Only necessary if buff requires a post buff hook"""
+        pass
```
leondz, line 80: seeing as post_buff_hook gates calling untransform, how do you feel about raising a `NotImplementedError` here, or returning  `attempt`?

garak/probes/base.py
```diff
@@ -70,12 +71,21 @@ def _buff_hook(
             return attempts
         buffed_attempts = []
         for buff in _config.transient.buff_instances:
+            if buff.post_buff_hook:
+                self.post_buff_hook = True
```
leondz, line 75: is the behavior to call untransform for all buffs, as soon as one with post_buff_hook asserted is encountered?

erickgalinkin: There's probably a more efficient way to do this, but I don't know what it is.

leondz (approved):
lgtm

PR #537, opened 2024-03-01
Fix AutoDAN issues

erickgalinkin:
Remove duplicate files. Fix openai errors. Correct path to prompts in dan.py.


erickgalinkin:
> I get the feeling this code is to be called directly, outside of garak, is that right?

Once we make the broader changes to the probe class, I'll expose this in a TAPProbe-like format

PR #555, opened 2024-03-15
Attack fixes

erickgalinkin:
Fixes and improvements to TAP, GCG, AutoDAN.


erickgalinkin:
Rebased and force-pushed to fix Cohere error.

garak/resources/autodan/model_utils.py
```diff
@@ -52,19 +52,22 @@ def check_for_attack_success(gen_str: str, test_prefixes) -> Tuple[bool, str]:
...
-        gen_str (str): Generated string
+        gen_str (str): Generated model output
         test_prefixes (list): List of test prefixes
 
     Returns:
         Tuple:
             Boolean of whether the model jailbreak was successful
             Generated string
     """
+    gen_str = gen_str.lower()
```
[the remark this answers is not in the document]
erickgalinkin: That definitely matters for the adversarial string, but it should be safe for checking whether or not there is a rejection present. May make sense to make it more robust though. I'll refactor to maintain `gen_str` and use a `test_str`.

garak/probes/dan.py
```diff
@@ -554,7 +554,9 @@ def __init__(
 
 
 class AutoDAN(Probe):
-    """Probe leveraging AutoDAN attack generation"""
+    """Probe leveraging AutoDAN attack generation
+
+    Dynamic, generative implementation. Requires OpenAI key to be set."""
```
leondz, line 559: in `OPENAI_API_KEY`?

garak/resources/autodan/autodan.py
```diff
@@ -241,7 +241,10 @@ def autodan_generate(
             gc.collect()
             torch.cuda.empty_cache()
 
-    if not success:
+    if success:
+        return adv_prefix
+
+    else:
         logger.info(
             f"Ran through {num_steps} iterations and found no successful prompts"
```
leondz, line 249: SAD FACE

garak/resources/gcg/generate_gcg.py
```diff
@@ -94,7 +94,7 @@ def run_gcg(
     / "resources/gcg/data/advbench/harmful_behaviors.csv",
     n_train: int = 50,
     n_test: int = 0,
-    outfile: str = garak._config.transient.basedir / "resources/data/gcg/gcg.txt",
+    outfile: str = garak._config.transient.basedir / "resources/gcg/data/gcg/gcg.txt",
```
leondz, line 97: I guess there's an agreement between the probes/ code and the code in resources/ that this will be the location. Maybe this could (one day) come from _config but that needn't block this PR

garak/resources/autodan/autodan.py
```diff
@@ -205,7 +205,8 @@ def autodan_generate(
             best_new_adv_prefix = new_adv_prefixes[best_new_adv_prefix_id]
 
             adv_prefix = best_new_adv_prefix
-            success, gen_str = check_for_attack_success(adv_prefix, test_prefixes)
+            gen_str = generator.generate(prompt=adv_prefix)[0]
```
leondz, line 208: Is it worth avoiding multiple response generation here? by e.g. pushing `generator.generations`/setting `generator.generations=1`/calling `generate()`/popping `generator.generations`?

This is a situation and solution pattern I see in a few places - maybe `generate()` is the place to allow an override for `generations`, or even other things (rather than the `generator` class constructor)

garak/resources/autodan/genetic.py
```diff
@@ -31,38 +31,12 @@
     nltk.download("punkt")
     nltk.download("wordnet")
 
-
-class MutationGenerator(OpenAIGenerator):
```
leondz, line 35: something very 1960s about this word, love the name

garak/resources/autodan/genetic.py
```diff
@@ -31,38 +31,12 @@
...
-        return response
-
-
 openai_api_key = os.getenv("OPENAI_API_KEY", default=None)
 if openai_api_key is None:
     USE_OPENAI = False
 else:
     USE_OPENAI = True
-    mutation_generator = MutationGenerator("gpt-3.5-turbo")
+    mutation_generator = OpenAIGenerator("gpt-3.5-turbo")
```
leondz, line 39: is this subtle default going to throw people, do you think?

garak/resources/autodan/genetic.py
```diff
@@ -308,15 +285,13 @@ def gpt_mutate(sentence: str) -> str:
...
+            response = mutation_generator.generate(
+                prompt=[
                     {"role": "system", "content": system_msg},
                     {"role": "user", "content": user_message},
                 ]
             )
-            revised_sentence = response["choices"][0]["message"]["content"].replace(
-                "\n", ""
-            )
+            revised_sentence = response[0].replace("\n", "")
```
leondz, line 294: love this change. nice

garak/resources/gcg/attack_manager.py
```diff
@@ -1429,7 +1431,7 @@ def run(
         test_steps: int = 50,
         incr_control: bool = True,
         stop_on_success: bool = True,
-        verbose: bool = True,
+        verbose: bool = False,
```
leondz, line 1434: can this be taken from `garak._config.system.verbose`?

garak/resources/gcg/attack_manager.py
```diff
@@ -1790,7 +1792,7 @@ def __init__(self, model_name, conv_template):
     def run(tasks, results):
         while True:
             task = tasks.get()
-            if task is None:
+            if not task:
```
leondz, line 1795: [pep8](https://peps.python.org/pep-0008/#programming-recommendations) prefers `is` for `NoneType` value checks (which i learned from looking at your code), how come the change?

garak/resources/gcg/attack_manager.py
```diff
@@ -1932,7 +1923,5 @@ def get_goals_and_targets(
 
     assert len(goals) == len(targets)
     assert len(test_goals) == len(test_targets)
-    logger.debug("Loaded {} train goals".format(len(goals)))
-    logger.debug("Loaded {} test goals".format(len(test_goals)))
```
leondz, line 1936: btw how is the goal string set? a docstring or `docs/` entry somewhere should cover this (maybe in one terse sentence)

garak/resources/gcg/generate_gcg.py
```diff
@@ -143,7 +146,12 @@ def generate_gcg(
     Returns:
         None
     """
-    mp.set_start_method("spawn")
+    mp.set_start_method("spawn", force=True)
+
+    if target_generator is not None and model_names is not None:
+        msg = "You have specified a list of model names and a target generator. Using the already loaded generator!"
```
leondz, line 152: i didn't quite get the problem or resolution for this

garak/resources/gcg/generate_gcg.py
```diff
@@ -143,7 +146,12 @@ def generate_gcg(
...
+        print(msg)
```
leondz, line 153: PRINT()?!?!?!?!?!

garak/resources/gcg/generate_gcg.py
```diff
@@ -154,13 +162,23 @@ def generate_gcg(
...
+        elif model_names is not None:
+            model_string = "_".join([x.replace("/", "-") for x in model_names])
+        else:
+            msg = "You must specify either a target generator or a list of model names to run GCG!"
+            logger.error(msg)
+            raise Exception(msg)
         logfile = (
             garak._config.transient.basedir
-            / "resources/gcg/data/logs/{timestamp}_{model_string}.json"
+            / f"resources/gcg/data/logs/{timestamp}_{model_string}.json"
```
leondz, line 175: Is `_config.run_id` available when this is executed?

garak/resources/tap/generator_utils.py
```diff
@@ -51,22 +53,39 @@ def load_generator(
...
     return generator
+
+
+def token_count(string: str, model_name: str) -> int:
+    encoding = tiktoken.encoding_for_model(model_name)
+    num_tokens = len(encoding.encode(string))
+    return num_tokens
+
+
+def get_token_limit(model_name: str) -> int:
```
leondz, line 79: is this max generation length or max context length?

i think the latter.

if generation length, can it be extracted from the `max_len` param?


if context length.. would that this were available via openai API / generator (see #573 ). tagging that issue so maybe we pick up altering this code once context sizes are represented better (or at all) in garak.

garak/resources/tap/tap_main.py
```diff
@@ -17,28 +18,29 @@
...
-        attack_generator: garak.generators.Generator,
-        target_generator: garak.generators.Generator,
-        evaluation_generator: garak.generators.Generator,
-        attack_max_tokens: int = 500,
-        attack_max_attempts: int = 5,
-        max_parallel_streams: int = 5,
-        target_max_tokens: int = 150,
-        evaluator_max_tokens: int = 10,
-        evaluator_temperature: float = 0.0,
+            self,
```
leondz, line 33: hey do you know how these params can be set when invoking via the probe?

garak/resources/tap/tap_main.py
```diff
@@ -101,7 +104,7 @@ def get_attack(self, convs, prompts):
         for conv, prompt in zip(convs, prompts):
             conv.append_message(conv.roles[0], prompt)
             # Get prompts
-            if "gpt" in self.attack_generator.name:
+            if not isinstance(self.attack_generator, Model):
```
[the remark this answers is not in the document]
leondz: the example case in the docs is `huggingface` / `gpt2`, this is a way better test, good catch

garak/resources/tap/tap_main.py
```diff
@@ -121,16 +124,22 @@ def get_attack(self, convs, prompts):
...
                 for full_prompt in full_prompts_subset[left:right]:
-                    outputs_list.append(self.attack_generator.generate(full_prompt))
+                    # We should fail more gracefully within runs for garak.
+                    try:
+                        outputs_list.append(self.attack_generator.generate(full_prompt)[0])
+                    except torch.cuda.OutOfMemoryError as e:
+                        if len(outputs_list) > 0:
+                            break
+                        else:
+                            raise e
```
leondz, line 134: `logging.critical("CUDA OOM during TAP generation")` ?

garak/resources/tap/tap_main.py
```diff
@@ -233,6 +241,23 @@ def create_conv(self, full_prompt, system_prompt=None):
...
+        if len(full_prompt.split()) / .75 > self.evaluator_token_limit:
+            # More expensive check yielding actual information -- add 100 token buffer to prompt
+            judge_system_prompt_tokens = token_count(system_prompt, self.evaluation_generator.name)
+            prompt_tokens = 100 + token_count(full_prompt, self.evaluation_generator.name)
+            # Iteratively reduce the prompt length
+            while judge_system_prompt_tokens + prompt_tokens >= self.evaluator_token_limit:
+                excess_tokens = judge_system_prompt_tokens + prompt_tokens - self.evaluator_token_limit
+                if excess_tokens > 0:
+                    # Truncate excess text
+                    excess_words = int(excess_tokens / .75)
```
leondz, line 255: where's the .75 from? is it worth constant-ising?

garak/resources/tap/utils.py
```diff
@@ -100,6 +117,12 @@ def clean_attacks_and_convs(attack_list, convs_list):
     """
     Remove any failed attacks (which appear as None) and corresponding conversations
     """
+    # Catch case where no valid attacks are returned
+    if not any(attack_list):
+        msg = "No valid attacks returned by the model!"
```
leondz, line 122: valid == successful? or

leondz (commented):
this is a lot of work, the integration looks like it has continued to be non-trivial, but the harmonisation now looks clear

garak/resources/autodan/model_utils.py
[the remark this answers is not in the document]
erickgalinkin: Actually, we just don't need to return `gen_str` at all -- we have it in the outer scope where we want it. Fixed.

garak/resources/gcg/generate_gcg.py
erickgalinkin: That's a good idea, tbh. If you have your own stuff you want to jam in configs. I think we need a separate PR where we adjust `_config` -- may be a good place to deal with some of the model parameter stuff too.

garak/resources/autodan/autodan.py
erickgalinkin: That seems right to me, yes. I just worry that if we set `generator.generaions=1` and the probe is being run as part of a series (for some reason -- this is a lot of time on a single probe anyway), it may have downstream effects. Definitely worth allowing an override for `generation` in the `generate()`method.

garak/resources/autodan/genetic.py
erickgalinkin: This is the implementation in the paper, but we can set this as a default and allow it to be configurable.

garak/resources/gcg/attack_manager.py
erickgalinkin: It can! That's a good one.

leondz (approved):

garak/resources/gcg/attack_manager.py
erickgalinkin: Huh, it looks nicer to me this way, but I can do `is None`!

leondz: `if not task` looks nicer to me too, just curious if there was maybe another rule

garak/resources/gcg/generate_gcg.py
erickgalinkin: Yeah, this should probably just be logger.warning

garak/resources/gcg/generate_gcg.py
erickgalinkin: Worth checking. It should be, unless using it as a library. I can check for it and if it's not there, fall back to timestamp.

garak/resources/tap/generator_utils.py
erickgalinkin: This is context length, yeah.

garak/resources/gcg/attack_manager.py
erickgalinkin: I ought to write a full docs entry for each of these. Will take that as an action item and open another PR.

garak/resources/tap/tap_main.py
erickgalinkin: I'd think you could pass them as params in the probe config?

erickgalinkin: Looks like we need a `ProbeConfig` class in `_config`.

garak/resources/tap/tap_main.py
erickgalinkin: From line 246: `# Crude and fast heuristic -- 100 tokens is about 75 words`

garak/resources/tap/utils.py
erickgalinkin: valid ~ well-formatted text

garak/resources/tap/tap_main.py
leondz: values in `_config.plugins.probes.ClassName` are automatically set in classes at instantiation time - just wondering a bit about the stuff in `resources/`. maybe let's wait for demand before triaging dealing with that

garak/resources/autodan/autodan.py
leondz: will open an issue to track that - #585

leondz (approved):
nice

PR #566, opened 2024-03-22
Add additional error message when __doc__ is None type

DavidLee528:
When developing a new detector, people may forget to write documentation string for the specific Python class. This will cause Exception which is relatively hard to locate the bad line.

In my case, garak told me detector load failed when developing a new prompt-based adversarial attack probe and detector pair.

e086944c33a8be5534cfac091c9e09c

In the end, I found the missing of documentation string of new PromptAttack class cause this exception.

a19bc4f6af64a3a9dfaa3ee7a4da5cf

To better develop experience, I add some error handle code there. See file changes for detail.

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -28,7 +28,13 @@ class Detector:
 
     def _set_description(self):
         if "description" not in dir(self):
-            self.description = self.__doc__.split("\n")[0]
+            try:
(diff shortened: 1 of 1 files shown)


erickgalinkin (approved):
I like it!

leondz:
Great, thanks! Tests will fail if docs are missing, but it's good to make this clear earlier

PR #588, opened 2024-04-09
Support for using 'langchain serve' endpoints

GustavFredrikson:
- **New `LangChainServeLLMGenerator` Class**: Implements HTTP POST request handling to interact with [LangChain Serve](https://python.langchain.com/docs/langserve/) LLM APIs, allowing integration with Garak probes.


GustavFredrikson:

GustavFredrikson:
recheck

leondz:
Thanks for this! We'll get it reviewed

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,70 @@
...
+    Inherits from Garak's base Generator class, extending its capabilities to support web-based LLM services.
+
+    Set endpoint name with --model_name. Example: --model_name=http://127.0.0.1:8000/rag-chroma-private and it's "invoke" endpoint will be used.
+
+    To use, set the appropriate API endpoint and configuration hash (if needed) before invoking the LLM with a prompt.
+    """
+
+    temperature = 0.750
+    k = 0
+    p = 0.75
```
leondz, line 27: Can we rename these to match convention?

```suggestion
    top_k = 0
    top_p = 0.75
```

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,70 @@
...
+    def _call_model(self, prompt: str) -> str:
+        """Makes an HTTP POST request to the LangChain Serve API endpoint to invoke the LLM with a given prompt."""
+        headers = {
+            'Content-Type': 'application/json',
+            'Accept': 'application/json',
+        }
+
+        payload = {
+            "input": prompt,
+            "config": {},
```
leondz, line 52: Where are the params being passed in? E.g. preset, penalties, stop etc

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,70 @@
...
+    This class facilitates communication with LangChain Serve's LLMs through a web API, making it possible
+    to utilize external LLMs not directly integrated into the LangChain library. It requires setting up
+    an API endpoint using langchain serve.
+
+    Utilizes the HTTP POST method to send prompts to the specified LLM and retrieves the generated text
+    response. It is necessary to ensure that the API endpoint is correctly set up and accessible.
+
+    Inherits from Garak's base Generator class, extending its capabilities to support web-based LLM services.
+
+    Set endpoint name with --model_name. Example: --model_name=http://127.0.0.1:8000/rag-chroma-private and it's "invoke" endpoint will be used.
```
leondz, line 20: model_name is used to specify `LangChainServeLLMGenerator`, can we avoid this being overloaded and set the URI in e.g. environment var? This will also be helpful to folks dealing with URI escaping on the CLI

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,70 @@
...
+    presence_penalty = 0.0
+    stop = []
+    generator_family_name = "LangChainServe"
+    config_hash = "default"
+
+    def __init__(self, name, generations=10):
+        self.name = name
+        self.fullname = f"LangChain Serve LLM {self.name}"
+        self.generations = generations
+        self.api_endpoint = f"{_config.plugins.model_name}/invoke"
```
leondz, line 39: Can we validate the URI? See also above comment about where to store the URI

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,70 @@
...
+                                     data=json.dumps(payload))
+            response.raise_for_status()
+            if "output" not in response.json():
+                logging.error("No output found in response: %s", response.json())
+                raise ValueError("No output found in response")
+            result = response.json().get("output")
+            return result
+        except requests.exceptions.RequestException as e:
+            logging.error("HTTP Request failed: %s", e)
+            return ""
```
leondz, line 68: Don't return empty string unless the model returned empty string, use None

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        try:
+            response = requests.post(f"{self.api_endpoint}?config_hash={self.config_hash}",
+                                     headers=headers,
+                                     data=json.dumps(payload))
+            response.raise_for_status()
+            if "output" not in response.json():
+                logging.error("No output found in response: %s", response.json())
+                raise ValueError("No output found in response")
```
leondz, line 63: This would stop the garak run completely. I don't know langchain serve so well, do you think that's an appropriate action? Would it be better to log and move on?

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,70 @@
...
+    To use, set the appropriate API endpoint and configuration hash (if needed) before invoking the LLM with a prompt.
+    """
+
+    temperature = 0.750
+    k = 0
+    p = 0.75
+    preset = None
+    frequency_penalty = 0.0
+    presence_penalty = 0.0
+    stop = []
```
leondz, line 31: Are these passed to the endpoint?

leondz (commented):
Thanks for this, looking pretty good

* We need to fix how `model_name` is used
* Config variables need to be handled cleaning
* Exception handling can get clarified a bit - interested in your feedback re: intended behavior with this generator & when it might make sense to backoff, skip, or abort
* What do you think we can do in regard to tests?

garak/generators/langchain_serve.py
GustavFredrikson: No, not used since they are setup in the langserve side, removing

garak/generators/langchain_serve.py
GustavFredrikson: Unsure, i experienced it when the langserve side returned a wrongly formatted response due to user error. In that case I would want it to stop completely with a good error message.

GustavFredrikson:
- Implemented URI validation to ensure the API endpoint is correctly formatted before attempting to connect.
- Replaced model_name usage with an environment variable (LANGCHAIN_SERVE_URI) for setting the API endpoint, let me know if other usage is preferred!
- Error handling:
  * Logging for different types of HTTP errors (4xx and 5xx).
  * Changed return values from empty strings to None.
- Removed unused config params i.e. temperature and penalties that are managed server-side.
- Added tests covering URI validation, initialization, mocked generation, and error handling scenarios using `requests_mock`.

garak/generators/langchain_serve.py
leondz: OK, let's run with this. I think we'll set an garak-wide policy in the near future and then aligning plugins to this should be straightforward. Thanks for the current implementation upgrade, this works for me.

tests/generators/test_langchain_serve.py
```diff
@@ -0,0 +1,48 @@
...
+@pytest.fixture
+def set_env_vars():
+    os.environ['LANGCHAIN_SERVE_URI'] = 'http://127.0.0.1:8000'
+    yield
+    os.environ.pop('LANGCHAIN_SERVE_URI', None)
+
+def test_validate_uri():
+    generator = LangChainServeLLMGenerator(name="test-generator")
+    assert generator._validate_uri('http://127.0.0.1:8000') == True
+    assert generator._validate_uri('bad_uri') == False
```
leondz, line 19: For these I get 

```
========================================================================= FAILURES =========================================================================
____________________________________________________________________ test_validate_uri _____________________________________________________________________

    def test_validate_uri():
>       generator = LangChainServeLLMGenerator(name="test-generator")

tests/generators/test_langchain_serve.py:17: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <garak.generators.langchain_serve.LangChainServeLLMGenerator object at 0x70d588ebf5c0>, name = 'test-generator', generations = 10

    def __init__(self, name, generations=10):
        self.name = name
        self.fullname = f"LangChain Serve LLM {self.name}"
        self.generations = generations
        api_uri = os.getenv("LANGCHAIN_SERVE_URI")
        if not self._validate_uri(api_uri):
>           raise ValueError("Invalid API endpoint URI")
E           ValueError: Invalid API endpoint URI

garak/generators/langchain_serve.py:37: ValueError
```

Any ideas?

jmartin-tech: To tests a method with the `@staticmethod` annotation you should not need to instantiate the class
```suggestion
def test_validate_uri():
    assert LangChainServeLLMGenerator._validate_uri('http://127.0.0.1:8000') == True
    assert LangChainServeLLMGenerator._validate_uri('bad_uri') == False
```

leondz: better, yep, thanks

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,85 @@
...
+    config_hash = "default"
+
+    def __init__(self, name, generations=10):
+        self.name = name
+        self.fullname = f"LangChain Serve LLM {self.name}"
+        self.generations = generations
+        api_uri = os.getenv("LANGCHAIN_SERVE_URI")
+        if not self._validate_uri(api_uri):
+            raise ValueError("Invalid API endpoint URI")
+
```
jmartin-tech, line 38: This makes an assumption based on how LangChain Server endpoints are exposed that the last token is the application being tested.

```suggestion
        self.name = api_uri.split('/')[-1]
        self.fullname = f"LangChain Serve LLM {self.name}"
```

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,85 @@
...
+    Example of setting up the environment variable:
+        export LANGCHAIN_SERVE_URI=http://127.0.0.1:8000/rag-chroma-private
+    """
+
+    generator_family_name = "LangChainServe"
+    config_hash = "default"
+
+    def __init__(self, name, generations=10):
+        self.name = name
+        self.fullname = f"LangChain Serve LLM {self.name}"
```
jmartin-tech, line 33: Since the model name is not really provided on the command line maybe we should extract that from the `api_uri`.

```suggestion
```

garak/generators/langchain_serve.py
```diff
@@ -0,0 +1,85 @@
...
+        self.generations = generations
+        api_uri = os.getenv("LANGCHAIN_SERVE_URI")
+        if not self._validate_uri(api_uri):
+            raise ValueError("Invalid API endpoint URI")
+        
+        self.api_endpoint = f"{api_uri}/invoke"
+
+        super().__init__(name, generations=generations)
```
jmartin-tech, line 41:
```suggestion
        super().__init__(self.name, generations=generations)
```

jmartin-tech (commented):
Some minor thought and I think this will be ready to go!

Issue #589, opened 2024-04-09
Configurable `Generator` parameters


erickgalinkin:
We need a way to instantiate generators in a way that:
* Accepts a `device_map` for multi-gpu systems
* Accepts a dtype (so we can load `torch.bfloat16`, for example)

jmartin-tech:
As a note this kind of `customization` of the generator feels like something that should be standardized as config options that can be passed via the `-G` options similar to how `rest` options are passed.

jmartin-tech:
Link #554

leondz:
trying to get this running via the standard config. the current configs don't auto-populate plugin class values, only at module level, which makes this a pain. will get that working better in the class loader

leondz:
Oh cool (narrator voice: it was not, in fact, cool): generators aren't loaded with `_plugins.load_plugin`, or any of the usual plugin loading machinery, so their config never gets automatically set up. I did wonder how the plugin loader would handle generator instantiation params (just model_name).

Actions:
* move generator loading to use plugin loading & configuation architecture - find a way to pass model_name at instantiation time (maybe a special case in `load_plugin` eh) - tracked in #594
* update the configuration machinery to take class-level info as it looks in `_plugins` / `_config`. this is nested dataclasses then nested dicts - it's not completely palatable but is functional. prefer keeping the yaml format looking sensible - tracked in #595
* will set up a hugging face mixin that loads huggingface module-level config into the HF classes, so we can carry on working while these two issues are addressed

leondz:
When we work out `device_map` - and this feels global rather than huggingface related, is that right? - we should
* [ ] look for mentions of `cuda` / maybe just `cuda:` in the code and test that they're not overriding it
* [ ] write test requesting cpu, cuda default device, and cuda specific non-first device (skipping if unavailable) and verifying that they're used

leondz:
seems like `device_map` is a hf-dependent concept, so `_config.plugins.generators.huggingface` makes more sense as a home

jmartin-tech:
I think from a generic point of view, we could have a the `Configurable` functionality in `generators/base.py` as a `dict` and provide validation of parameters in the class before consuming the values. Each generator can then determine how to handler the `dict`, in some cases by providing a set of expected valid keys or by just validating the keys and values are all in a format that will not present easy entry point for `RCE` based on generators configuration pattern.

leondz:
folding #594 into here

leondz:
folding #643 into here, check out notes there

[...]

PR #596, opened 2024-04-10
allow configuration of HF generators from YAML / global config

leondz:
see #589 

Hugging Face generators now take params from global config

e.g. 

```
(garak) 11:54:18 x1:~/dev_ext/garak [feature/hf_config] $ cat garak/garak.site.yaml 
plugins:
  generators:
    huggingface:
      Pipeline:
        dtype: bfloat16
(garak) 11:30:42 x1:~/dev_ext/garak [feature/hf_config] $ python3
Python 3.12.1 | packaged by Anaconda, Inc. | (main, Jan 19 2024, 15:51:05) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import garak
>>> import garak.generators.huggingface
>>> p = garak.generators.huggingface.Pipeline("gpt2")
🦜 loading generator: Hugging Face 🤗 pipeline: gpt2
>>> p.dtype
'bfloat16'
```

### DIFF
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -44,7 +44,19 @@ class HFInternalServerError(Exception):
     pass
 
 
-class Pipeline(Generator):
+class HFConfigurable:
+    def _load_hf_config(self):
+        if "huggingface" in _config.plugins.generators:
+            generator_classname = self.__class__.__name__
+            print(generator_classname)
+            if generator_classname in _config.plugins.generators["huggingface"]:
+                for k, v in _config.plugins.generators["huggingface"][
+                    generator_classname
+                ].items():
+                    setattr(self, k, v)
+
+
+class Pipeline(Generator, HFConfigurable):
     """Get text generations from a locally-run Hugging Face pipeline"""
 
     generator_family_name = "Hugging Face 🤗 pipeline"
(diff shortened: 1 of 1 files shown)


jmartin-tech (commented):
I think this may need a pause to think through how to accept `generator` config more consistently for all generators and align with existing configs like those for `RestGenerator` and `LiteLLMGenerator`. The current impacts in the PR around class composition while contained seem like they could use a little more iteration or abstraction.

erickgalinkin (approved):
I think I agree with @jmartin-tech here -- these changes look good for Huggingface. The question is whether we want to modify `_config` more broadly to handle these sorts of issues or simply improve the generator here directly. 
The PR fixes the issue for HF but we likely want to fix `_config` to set parameters for generators of any type. That said, I don't object to kicking that particular can down the road given the prevalence of HF models.

leondz:
agree. let's pick a more general method, maybe from the threads here: https://github.com/leondz/garak/issues/589#issuecomment-2047021771

will tag this in the rel milestone if that's ok - it's not top priority wrt: feature set, but it looks like a good ux improvement

leondz:
Will re-do this with more general pattern

PR #600, opened 2024-04-11
allow generators.Base.generate() to take an optional param specifying generation count

leondz:
Also required clearing up some type signatures and amending `_call_model` a bunch of places

Would rather `_call_model` didn't have to take an extra param, but sometimes it cares about the number of generations. Because generator objects handle parallel requests, they're not a great place to handle state - each generation request should be atomic and self-contained

Leaving in `_call_model`'s `generations_this_call` param makes for a weird situation where one might think `_call_model` can be invokes asking for many generations, where in fact this may be ignored (undef behaviour)


leondz:
oh hey we should be checking if models generate the number of responses we want, and define & test what happens when they don't. cool cool cool (#601 )

garak/generators/base.py
```diff
@@ -74,11 +76,17 @@ def generate(self, prompt: str) -> List[str]:
 
         self._pre_generate_hook()
 
+        if generations_this_call == -1:
+            generations_this_call = self.generations
+
+        elif generations_this_call == 0:
```
erickgalinkin, line 82: Do we want to warn users in this case? Seems like if someone calls a generator with no generations, they've probably made a mistake.

garak/generators/base.py
```diff
@@ -74,11 +76,17 @@ def generate(self, prompt: str) -> List[str]:
 
         self._pre_generate_hook()
 
+        if generations_this_call == -1:
```
erickgalinkin, line 79: Should this be `<= -1` instead? If somehow `-2` or whatever gets specified, we'd end up in a weird place.

erickgalinkin (changes requested):

garak/generators/base.py
leondz: agree, added logging

erickgalinkin (approved):
Looks good to me!

PR #605, opened 2024-04-17
Enforce warning output for `garak` classes.

jmartin-tech:
Introduction of the `litellm` package caused test runs to report warnings about pending deprecations in dependency code not directly in scope of this project.

This change raises lib warnings more clearly and has an effect of suppressing warnings from dependency classes. While not optimal, the explicit check for warnings from repo code is a reasonable trade off.

A separate maintenance action should be added in a future iteration to add checks for warnings introduced when updating dependency versions.

The changes to lib files resolve warnings primarily related file handles that are opened and never closed. Changes to the tests may be be worth an extended look as most are related to not closing the `_config.TransientConfig.reportfile` when the object is reloaded or goes out of scope.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,3 +80,9 @@ packages = ["garak"]
 line-length = 88
 target-version = ['py310', 'py311', 'py312']
 include = '\.pyi?$'
+
+[tool.pytest.ini_options]
+filterwarnings = [
+    "ignore",
+    "default:::garak",
+]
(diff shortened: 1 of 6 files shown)


tests/buffs/test_buff_config.py
```diff
@@ -85,11 +85,15 @@ def cleanup(request):
...
+        files = [
+            f"{prefix}.report.jsonl",
+            f"{prefix}.report.html",
+            f"{prefix}.hitlog.jsonl",
+        ]
+        for file in files:
+            try:
+                os.remove(file)
+            except FileNotFoundError:
+                pass
```
leondz, line 97: Nice, didn't spot that

pyproject.toml
leondz, line 1: really great. big fan

leondz (approved):

PR #613, opened 2024-04-20
Further align shebangs with code that has executable entry points

leondz:
continuation of #612

### DIFF
--- a/garak/__main__.py
+++ b/garak/__main__.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python3
-
-"""garak entry point script"""
+"""garak entry point wrapper"""
 
 import sys
 
--- a/garak/__main__.py
+++ b/garak/__main__.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 """garak entry point script"""
 
 import sys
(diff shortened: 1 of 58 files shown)


jmartin-tech (approved):
LGTM

garak/analyze/aggregate_reports.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
```
jmartin-tech, line 2: These additions in utility classes may not provide the utility intended. Happy to leave them for now.

Would you ever expect a user to execute:
```
./aggregate_reports.py [args]
```

jmartin-tech (commented):
Question, looking at the `added` entires, I suspect these are not really usable.

garak/__main__.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
```
jmartin-tech, line 2: These additions in utility classes may not provide the utility intended. Happy to leave them for now.

Would you ever expect a user to execute:
```
./__main__.py -m garak --model_type huggingface --model_name gpt2 --probes dan.Dan_11_0
```

leondz: No. OTOH cli.py isn't intended to invoked either. Let's remove them both.

garak/analyze/aggregate_reports.py
leondz: yes, definitely!

PR #616, opened 2024-04-22
add var for generator context_len and populate this for some generators

leondz:


garak/generators/openai.py
```diff
@@ -58,6 +58,29 @@
     # "ada",  # shutdown https://platform.openai.com/docs/deprecations
 )
 
+context_lengths = {
```
erickgalinkin, line 61: In a perfect world, we'd pull these via the openai api, but IIRC, this is not exposed there.

erickgalinkin (approved):
LGTM

Issue #602, opened 2024-04-11
Enable more complex `prompts`


jmartin-tech:
The current generator interface expects to receive prompts as `str` see: https://github.com/leondz/garak/blob/4127ae5092ad3acaba680a32011018fc564cc92a/garak/generators/base.py#L66

This initial simple submission process was work to date; however #587 show as example of a query prompt that needs a more complex structure. In this case the `Multi-modal` model accepts both text and image data to generate a response.

 I propose an added abstraction layer by implementing a `Prompt` base interface class that be extended to model these more prompts to be processed by each generator.

``` Python
def generate(self, prompt: Prompt) -> List[str]:
```
or possibly also abstracting the response as well:
``` Python
def generate(self, prompt: Prompt) -> List[PromptResponse]:
```

Prompts can then be further segmented into things like `TextPrompt`, `MultiStepTextPrompt`, `VisualPrompt`, `VisualTextPrompt` and other such constructs to that on the base functions available to allow use with different and even mixed prompt modalities for models that can accept various input patterns.

Rough example:
``` Python
class Prompt:
    text = None

    def str(self)
        return self.text

class TextPrompt(Prompt):
    def __init__(self, text: str):
        self.text = text

class VisualTextPrompt(Prompt):
    image
    def __init__(self, text: str, image_path: str):
        self.text = text
        try:
            Image.open(image_path)
         except Exception:
             logger.error(f"No image found at: {image_path}")
```

jmartin-tech:
Another recent finding related to multi-modal prompts is a need to define relationships between parts of the prompt.  The case identified is that some models request formats may have different expectations for referencing images in text.  The current `visual_jailbreak` prompts include a placeholder in the `text` segment of the prompt that some models may need to `remove` or `replace` with an API specific linking/embedding.

PR #617, opened 2024-04-22
add ArtPrompt probes

### LINKED ISSUE #535: buff: ascii art
https://twitter.com/llm_sec/status/1763164312313495771

see also https://github.com/sepandhaghighi/art / https://pypi.org/project/art/

jmartin-tech:
Fix #535 

Implements two prompt obfuscation patterns based on [ArtPrompt](https://arxiv.org/abs/2402.11753)

Testing work is still in progress here a new possible base case or `detector` specific to technique, may be needed. As current `mitigation.MitigationBypass` detector does not quite cover the returned values when the model is not able to infer the masked word.

Example Usage pattern:
```
python -m garak -m huggingface.Model -n meta-llama/Llama-2-7b-chat-hf -p artprompt
python -m garak -m huggingface --model_name gpt2 --probes artprompt
```

The probe pattern could be enhanced to be provided a dataset of prompts to be augmented with a dictionary of unsafe words often blocked by safety training that can be easily maintained as a set of resource files.

### DIFF
--- a/garak/buffs/art.py
+++ b/garak/buffs/art.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+""" Buff that converts prompts with different encodings. """
+
+from collections.abc import Iterable
+
+import logging
+import garak.attempt
+from garak.buffs.base import Buff
+from art import *
+
+# these would be great to have be configurable
+BUFF_WORDS = [
+    "counterfeit",
+    "bomb",
+    "launder",
+]
+
+class Art(Buff):
+    """ASCII Art buff"""
+
+
+    mask_word = "MASK"
+    delimiter = "*"
+    rows = 11
+    columns = 20
+
+    # consider adding a font decoder description to inform each instructions on how to interpret the current_word_art
+    # currently this is highly coupled to "block" font 11x20 grid format
+
+    def format_word_art(self, word):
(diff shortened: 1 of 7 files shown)


garak/buffs/art.py
```diff
@@ -0,0 +1,85 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+""" Buff that converts prompts with different encodings. """
+
+from collections.abc import Iterable
+
+import logging
+import garak.attempt
+from garak.buffs.base import Buff
+from art import *
```
leondz: Some disprefer `import *` because it obfuscates exactly which items are added to the current namespace

```suggestion
from art import text2art
```

leondz:
This is looking pretty reasonable. Agree that a probe works well for the case that the paper presents!

zazer0:
Hi, my team was thinking of building this as a Buff for the Apart Deception Hackathon (but, just saw this pull request already existed!) - is there anything that's still waiting on being implemented?

jmartin-tech:
@zazer0, the current implementation for the probe is mostly complete, a plan for configurable prompts will likely be worked on after #602.

The primary reason this is still in draft is that work is still needed to generate a better `detector` for evaluating the responses from a prompt offered by this probe. The current detectors look for a mitigation response however this probe would needs additional filtering to determine if the response was able to identify the `safety ` word masked in the prompt as failure to decode would not represent a finding of successful bypass of alignment or mitigation.

leondz:
Given the upcoming payloads work on decoupling content from transformation, can it make sense to put this through as a probe in this PR?  And defer (in separate PRs):

a. making the encoded texts accessible via the payload mechanism; 
b. a buff

PR #618, opened 2024-04-22
add tests for `ggml` generator

jmartin-tech:
* Adds test for GGML initialization validation
* Refactors `pyproject.toml` to make dev and test dependencies optional
  * Add `pyproject.toml` optional-dependencies section
  * Add `#` comment line support for `requirements.txt` in validation test
  * Update validation test to include `optional-dependencies`

Item to consider here, should the dependencies validation test be revised to validate sections noted in `requirements.txt` are consistent with `pyproject.toml`? Having asked this I posit the value for the changes is in `pyproject.toml` no longer requiring dev or test dependencies during a `pip install garak` and sections in `requirements.txt` is just syntactic sugar.

### DIFF
--- a/tests/generators/test_ggml.py
+++ b/tests/generators/test_ggml.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 import os
 import pytest
 import tempfile
@@ -52,4 +50,4 @@ def test_init_good_model():
         file.write(garak.generators.ggml.GGUF_MAGIC)
         file.seek(0)
         g = garak.generators.ggml.GgmlGenerator(file.name)
-        assert g is not None
+        assert type(g) is garak.generators.ggml.GgmlGenerator
--- a/tests/generators/test_ggml.py
+++ b/tests/generators/test_ggml.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+
+import os
+import pytest
+import tempfile
+import garak.generators.ggml
+
+DEFAULT_GENERATIONS_QTY = 10
+
+stored_env = os.getenv(garak.generators.ggml.ENV_VAR)
+model_name = None
+
+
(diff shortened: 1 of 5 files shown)


tests/generators/test_ggml.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
+
```
leondz, line 2: will this text be executed directly?
```suggestion
```

tests/generators/test_ggml.py
```diff
@@ -0,0 +1,55 @@
...
+            garak.generators.ggml.GgmlGenerator(file.name)
+        assert "not in GGUF" in str(exc_info.value)
+
+
+def test_init_good_model():
+    with tempfile.NamedTemporaryFile(suffix="_test_model.gguf") as file:
+        file.write(garak.generators.ggml.GGUF_MAGIC)
+        file.seek(0)
+        g = garak.generators.ggml.GgmlGenerator(file.name)
+        assert g is not None
```
leondz, line 55: Did not see where in `garak/generators/ggml.py` `GgmlGenerator` would return `None` instead of raising an exception. Should this be checking for exceptions rather than `None` being returned?

leondz (commented):
Looks good, just a couple of questions

tests/generators/test_ggml.py
jmartin-tech: You are correct `None` should never be returned, this was just an assert that no exception occurred. The check could be updated to validate the correct instance type was returned.

```suggestion
        assert type(g) is garak.generators.ggml.GgmlGenerator
```

tests/generators/test_ggml.py
jmartin-tech: Valid, removing. :-)

PR #633, opened 2024-04-26
macOS test install from correct path

jmartin-tech:
Tweak path requirement I missed in #631


jmartin-tech:
See: https://github.com/jmartin-tech/garak/actions/runs/8850726867

PR #634, opened 2024-04-26
consolidate test file cleanup

jmartin-tech:
* convert all log file cleanup to a finalizer
* consolidate `cleanup` of run generated log files as global fixture


tests/conftest.py
leondz, line 1: nice!

leondz:
excellent, this is much better

Issue #487, opened 2024-02-19
restgen json access enhancements


leondz:
currently, when services return json, folk can only specify a top-level dict key from where to get the response

this doesn't work - it might a list, or not top level, or...

**proposal: move to a JSONPath in the `response_json_field` variable**

logic:
* if `response_json_field` contains no JSONPath special chars, treat it as a top-level dict key (retaining backward compat)
* if `response_json_field` starts with a `$`, treat is as a raw JSONPath
* if `response_json_field` contains JSONPath special chars, treat it as a JSONPath, but prepend a `$` (this allows python-like structures such as `[0][0]`
* if we're very nice, write a note suggesting people escape stuff from the YAML parser

JSONPath likes to return a list. hmm
* if the result is a `str`: return that `str`
* if the result is a list of 1: return `result[0]`
* if the result is a list of len>1: return the list
* note that the output should be `List[str]`. if we're really nice, we could type check that, and log an error if we don't get it back

PR #635, opened 2024-04-26
Support Nested Json API Responses

au70ma70n:


au70ma70n:

jmartin-tech:
@au70ma70n can you update the description here to explain what the goal is here and why it would be a default functionality?

Can you provided some examples of configuration and responses both impacted and not impacted by this revision?

If I am understanding this correctly the need here is to extract a value in the result as the `response_json_field` that is not at the top level of the parsed response.

Thinking on this would there be value in enabling `response_json_field` to be a [JSONPath](https://pypi.org/project/jsonpath-ng/) query?

leondz:
Is this related to #487 ?

leondz:
@au70ma70n Thanks very much for this. We'll cue up a general fix for the next few days.

leondz:
tracking general fix in #651

leondz:
flexible fix provided in #651 (now merged). thank you for pushing this along!

Issue #599, opened 2024-04-10
Hitlogfile write exception


pi194046:
i am using the below code to test the function generator of garam , when a hit is generated it writes a hitlog file however i am getting the exception

     93                         _config.transient.hitlogfile.write(
     94                             json.dumps(

ValueError: I/O operation on closed file.
when i run the below code in google colab .  My guess is since session is active it causing this issue and is not opening the hitlogfile once its closed 

```
import openai1

import importlib
importlib.reload(garak)

import garak.generators.function

probe="profanity"
model="openai-custom"
report_prefix="runs/garak-{probe}_{model}"
rr=report_prefix.format(probe=probe,model=model)
print("RRR",rr)

cli_command="--model_type function --model_name openai1#greet1 --probes lmrc.Profanity  --report_prefix {report_prefix} --generations 10"
cli_command=cli_command.format(report_prefix=rr)
print("cli command : ",cli_command)
garak.cli.main(cli_command.split())
```

leondz:
Thanks very much for this. Logfile locking is a little sensitive and we'd love to fix this sort of thing. Are you able to share an example notebook that throws this error?

pi194046:
Please find the google colab notebook which simulates the error https://gist.github.com/pi194046/0a1dcb7ab8bf57625d780d3450d52466

I was able to resolve the error for now by editing the file garak/evaluators/base.py by commenting the like `if not _config.transient.hitlogfile:` and always opening . Below is snippet of edited code 


                        #if not _config.transient.hitlogfile:
                        if not _config.reporting.report_prefix:
                                hitlog_filename = f"{_config.reporting.report_dir}/garak.{_config.transient.run_id}.hitlog.jsonl"
                        else:
                                hitlog_filename = (
                                    _config.reporting.report_prefix + ".hitlog.jsonl"
                                )
                        logging.info("hit log in %s", hitlog_filename)
                        _config.transient.hitlogfile = open(
                                hitlog_filename,
                                "w",
                                buffering=1,
                                encoding="utf-8",
                            )

leondz:
super helpful, thank you for the notebook & proposal!

leondz:
Strong chance this is related to #646 - this is a really valuable report

Issue #473, opened 2024-02-12
test order affects test success/speed. it shouldn't


leondz:
PR #472 is a hack.

pytest-8.0.0 changed the ordering of tests, and this exposed a stateful? dependency that led to tests hanging/taking incredibly long.

py 3.11.5
    pytest 7.4.2
        python3 -m pytest tests/test_cli.py     OK
        python3 -m pytest tests/                OK
    pytest 8.0.0
        python3 -m pytest tests/test_cli.py     OK
        python3 -m pytest tests/                vv slow

> OK. The order in 8.0.0 is:
```
tests/detectors/test_detectors_packagehallucination.py .....                                                                                          [  0%]
tests/generators/test_huggingface.py ...                                                                                                              [  1%]
tests/probes/test_probe_format.py ................................................................................................................... [ 16%]
...............................................................                                                                                       [ 25%]
tests/probes/test_probe_tags.py .........................................................................................                             [ 37%]
tests/probes/test_probes_encoding.py ...                                                                                                              [ 37%]
tests/probes/test_probes_packagehallucination.py .                                                                                                    [ 38%]
tests/test_analyze.py ..                                                                                                                              [ 38%]
tests/test_attempt.py .                                                                                                                               [ 38%]
tests/test_cli.py ....
```

> The order in 7.4.2 is:
```
tests/test_analyze.py ..                                                                                                                              [  0%]
tests/test_attempt.py .                                                                                                                               [  0%]
tests/test_cli.py ......                                                                                                                              [  1%]
tests/test_config.py ........................................                                                                                         [  6%]
tests/test_docs.py .................................................................................................................................. [ 24%]
..............................................................................................................                                        [ 39%]
tests/test_plugin_load.py ........................................................................................................................... [ 56%]
..........................................                                                                                                            [ 61%]
tests/test_reqs.py .                                                                                                                                  [ 61%]
tests/detectors/test_detectors_packagehallucination.py .....                                                                                          [ 62%]
tests/generators/test_huggingface.py ...                                                                                                              [ 63%]
tests/probes/test_probe_format.py ................................................................................................................... [ 78%]
...............................................................                                                                                       [ 87%]
tests/probes/test_probe_tags.py .........................................................................................                             [ 99%]
tests/probes/test_probes_encoding.py ...                                                                                                              [ 99%]
tests/probes/test_probes_packagehallucination.py .                    
```

> extra in 8.0.0 before test_cli are:
```
[...]

PR #637, opened 2024-04-29
generator: NIM

leondz:
Add NIM generator support. This PR only covers NVIDIA-hosted NIMs


leondz:
See https://build.nvidia.com/explore/reasoning for items to try

garak/generators/nim.py
```diff
@@ -0,0 +1,106 @@
...
+            raise ValueError("Please specify a NIM in model name (-n)")
+
+        super().__init__(name, generations=generations)
+
+        self.api_key = os.getenv("NIM_API_KEY", default=None)
+        if self.api_key is None:
+            raise ValueError(
+                'Put the NIM API key in the NIM_API_KEY environment variable (this was empty)\n \
+                e.g.: export NIM_API_KEY="nvapi-xXxXxXxXxXxXxXxXxXxX"'
+            )
```
jmartin-tech, line 47: Should we go with [`NGC_API_KEY`](https://github.com/leondz/garak/blob/00c21b01b239c4a2174c81541619c0bfd6bac981/garak/generators/nemo.py#L41) as done with `nemo.py`?

`nvcf.py` uses a specifics [`NVCF_API_KEY`](https://github.com/leondz/garak/blob/00c21b01b239c4a2174c81541619c0bfd6bac981/garak/generators/nvcf.py#L45), not sure what the best pattern is here, I am also seeing some other non-nvidia tooling referring to [`NVIDIA_API_KEY`](https://haystack.deepset.ai/blog/haystack-nvidia-integration) and [NVIDIA docs](https://docs.nvidia.com/ai-enterprise/nim-llm/latest/deploy-helm.html) suggesting `NGC_CLI_API_KEY` when using helm for NIM deployments so I guess there is not really a standard per se.

garak/generators/nim.py
```diff
@@ -0,0 +1,106 @@
...
+                else:
+                    response.raise_for_status()
+            else:
+                logging.warning("nim : skipping this prompt")
+                return None
+
+        else:
+            response_body = response.json()
+
+            return response_body["choices"][0]["message"]["content"]
```
jmartin-tech, line 103: I guess this is considered normal but access to entry `[0]` from `choices` seems brittle? Should we convert that to a JSONPath query?

garak/generators/nim.py
leondz: From what I've seen NGC CLI API keys are not the same as NVIDIA-hosted NIM API keys. There are a few moving parts in the ecosystem.
* NVCF is a different service & API with a two-key auth.
* NGC is more like a CSP, with tooling and APIs around it. 
* NIM is an inference container format. 
 
NIMs can run on NGC, but can also run locally, or even on NVCF (if they fit on one node). 

This PR provides a class for NIMs hosted by NVIDIA (which afaik run on the NGC clusters). Generator classes for locally-hosted NIMs etc. can also go in this module - tagging [this API reference page](https://docs.nvidia.com/ai-enterprise/nim-llm/latest/quickstart/mixtral-8x7b-instruct.html#launch-microservice) for later

garak/generators/nim.py
leondz: This is predicated on the [NIM OpenAI API spec](https://docs.nvidia.com/ai-enterprise/nim-llm/latest/openai-api.html) which should be pretty static. The primary brittleness I see here is that the `[0]` blocks support for multiple-generation. I can see th `/v1/completions` endpoint's `CompletionRequest` declares that it supports `n` for "number of candidate generations", will check this out and update.

garak/generators/nim.py
```diff
@@ -0,0 +1,106 @@
...
+from garak.generators.base import Generator
+
+
+class NVHostedNimGenerator(Generator):
+    """Wrapper for NVIDIA-hosted NIMs. Expects NIM_API_KEY environment variable."""
+
+    supports_multiple_generations = False
+    generator_family_name = "NIM"
+    temperature = 0.1
+    top_p = 0.7
```
leondz, line 26:
```suggestion
    top_p = 0.7
    top_k = 0 # top_k is hard set to zero as of 24.04.30
```

garak/generators/nim.py
leondz: OK, looks like the only supported config is `n=1` so this part should be good to go
image
image

garak/generators/nim.py
erickgalinkin: FWIW, my `NGC_API_KEY` did not work, but I was able to get the NIM API working with my `NVCF_API_KEY`.

leondz: Nice. They do look similar. I wish I knew if all NIM calls could be done using NVCF keys. I might just leave this in here for now pending a shift to conf-based key specification with per-generator `api_key`, obviating cramming everything into env var space -- see #643.

jmartin-tech (commented):
Based on having #645 landed we might refactor this initial support as `OpenAICompatible` and later extend to add a `NemoLLMCompatible` implementation.

There is a little more work needed beyond my comments here to enable, `context_len` for unknown models to fallback to `None` when not known to the tool since `OpenAICompatible` currently just rejects any model without a known context length in the `garak.generators.openai.context_lengths` dictionary.

See: https://github.com/leondz/garak/blob/075796f044c2877977383dcf561b5a54f865c009/garak/generators/openai.py#L135-L138

garak/generators/nim.py
```diff
@@ -0,0 +1,140 @@
...
+import logging
+import os
+from typing import List, Union
+
+import requests
+
+import backoff
+
+from garak import _config
+from garak.generators.base import Generator
```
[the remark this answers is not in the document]
leondz: I'm inclined to offer two separate classes, one for chat and one for completions, and default to chat. maintaining model names in here builds in technical debt and unfortunate UX. what do you think?

garak/generators/nim.py
```diff
@@ -0,0 +1,140 @@
...
+                    choice["message"]["content"] for choice in response_body["choices"]
+                ]
+            except Exception as e:
+                logging.warning("Problem parsing NIM response: %s" % repr(e))
+                return None
+
+            if self.supports_multiple_generations:
+                return responses
+            else:
+                return responses[0]
```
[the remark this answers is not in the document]
leondz: see two-class suggestion above - this removes the need to do endpoint switching. i expect NV hosted NIMs to grow faster and to a larger set than OpenAI models, for obvious reasons (1 provider vs. many providers)

garak/generators/nim.py
```diff
@@ -0,0 +1,140 @@
...
+"""NVIDIA Inference Microservice LLM interface"""
+
+import json
+import logging
+import os
+from typing import List, Union
+
+import requests
+
+import backoff
```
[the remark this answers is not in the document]
leondz: is the `import openai` required? it feels like it would be nice if anything dependent on this 3rd-party module could be localised to garak.generators.openai (as far as possible), brought in when we import the mixin class

garak/generators/nim.py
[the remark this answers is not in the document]
jmartin-tech: I agreed maintaining the names would be great to avoid, I can see a direct selections as reasonable such as:
OpenAIChat be default:
```
python -m garak -m nim -n <chat_model> -g 1 -p dan.DanInTheWild
python -m garak -m nim.OpenAIChat -n <chat_model> -g 1 -p dan.DanInTheWild
```
And for completion required a specific class:
```
python -m garak -m nim..OpenAICompletion -n <completion_model> -g 1 -p dan.DanInTheWild
```

This then lends to future support for NemoLLM NIMs as:
```
python -m garak -m nim.NemoLLMChat -n <chat_model> -g 1 -p dan.DanInTheWild
python -m garak -m nim.NemoLLMCompletion -n <completion_model> -g 1 -p dan.DanInTheWild
```

garak/generators/nim.py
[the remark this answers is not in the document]
jmartin-tech: Yes it is still required as the import in `generators/openai.py` does not bring the object into the local stack for this class to utilize in `_load_client()`:
```
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/jemartin/Projects/nvidia/garak/garak/__main__.py", line 13, in <module>
    main()
  File "/Users/jemartin/Projects/nvidia/garak/garak/__main__.py", line 9, in main
    cli.main(sys.argv[1:])
  File "/Users/jemartin/Projects/nvidia/garak/garak/cli.py", line 451, in main
    generator = getattr(generator_mod, generator_class_name)(
  File "/Users/jemartin/Projects/nvidia/garak/garak/generators/openai.py", line 121, in __init__
    self._load_client()
  File "/Users/jemartin/Projects/nvidia/garak/garak/generators/nim.py", line 55, in _load_client
    self.client = openai.OpenAI(base_url=self.url, api_key=self.api_key)
NameError: name 'openai' is not defined. Did you mean: 'open'?
```

However since we still in the original process I believer this does not cause a re-import only a lookup of the imported symbol in the PythonVM.

garak/generators/nim.py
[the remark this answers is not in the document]
leondz: yes! agree. excellent. i'll do that.

garak/generators/nim.py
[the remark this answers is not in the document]
leondz: ok, maybe it's fine that using `OpenAICompatible` requires one to be cognisant that `openai` is being used

garak/generators/openai.py
```diff
@@ -145,7 +134,6 @@ def __init__(self, name, generations=10):
         (
             openai.RateLimitError,
             openai.InternalServerError,
-            openai.APIError,
```
jmartin-tech, line 148: Removing the general error seems prudent given testing that shows swallows `4xx` responses and would cause backoff to trigger more than it should.

jmartin-tech (commented):
Overall this aligns well, there seems to be some outstanding issues related to limitations of the API, testing paused until these are sorted.

garak/generators/nim.py
```diff
@@ -0,0 +1,96 @@
...
+            )
+
+    def _call_model(
+        self, prompt: str | List[dict], generations_this_call: int = 1
+    ) -> List[str]:
+        assert generations_this_call == 1, "n>1 is not supported"
+        return super()._call_model(prompt, generations_this_call)
+
+
+class NVOpenAICcompletion(NVOpenAIChat):
```
jmartin-tech, line 69: Minor typo:
```suggestion
class NVOpenAICompletion(NVOpenAIChat):
```

garak/generators/base.py
```diff
@@ -127,7 +124,7 @@ def generate(self, prompt: str, generations_this_call: int = -1) -> List[str]:
                 )
                 generation_iterator.set_description(self.fullname[:55])
                 for i in generation_iterator:
-                    outputs.append(self._call_model(prompt, generations_this_call))
+                    outputs.append(self._call_model(prompt, generations_this_call=1))
```
jmartin-tech, line 127: This is an interesting change, `generations_this_call=1` will set the local variable to `1`, is that the desired action? Consider:

```suggestion
                    outputs.append(self._call_model(prompt, 1)) # generate once as `generation_iterator` consumes `generations_this_call`
```

The comment clarifies the reasoning.

garak/generators/test.py
jmartin-tech, line 1: Not something to address in this PR, however I am noticing this file is used for testing, I wonder if it is possible to move it into the tests path as a fixture only loaded during testing.

Is there added value in having this available all the time?

garak/generators/nim.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
+
```
jmartin-tech, line 2:
```suggestion
```

jmartin-tech (commented):
Update look good, some nice to have items noted.

garak/generators/openai.py
```diff
@@ -229,10 +217,31 @@ def _load_client(self):
...
+    def _validate_config(self):
+        if self.name == "":
+            openai_model_list = sorted([m.id for m in self.client.models.list().data])
+            raise ValueError(
+                f"Model name is required for {self.generator_family_name}, use --model_name\n"
+                + "  API returns following available models: ▶️   "
+                + "  ".join(openai_model_list)
+                + "\n"
+                + "  ⚠️  Not all these are text generation models"
+            )
```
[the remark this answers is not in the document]
leondz: Nice, that sounds about right. I started implementing a validation pattern here, as you can see, though it's a little harder to tease out for this module since `openai` is expected to be loaded in order to provide a helpful message. Will move this bit of validation (for when `name=""`) into `_load_client`.

tests/generators/test_openai.py
```diff
@@ -15,6 +15,13 @@ def test_openai_version():
     assert openai.__version__.split(".")[0] == "1"  # expect openai module v1.x
 
 
+def test_openai_invalid_model_names():
+    with pytest.raises(ValueError) as e_info:
+        generator = OpenAIGenerator(name="")
+    with pytest.raises(ValueError) as e_info:
+        generator = OpenAIGenerator(name="this is not a real model name")
```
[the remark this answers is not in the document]
leondz: OK. Tighter scoping works. If these are substrings worth factoring into error message constants, we'll find out.

garak/generators/base.py
leondz: Oh, thanks, that was more subtle. Changing the local variable was indeed not the desired action.

The bug this fixes came from a lack of clarity around process and expectations in generators; created #682 to prompt documentation for this.

garak/generators/openai.py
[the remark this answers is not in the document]
leondz: done in c0f17ed6e8cc15f67916747fdeab6d3622b63dc0

tests/generators/test_nim.py
```diff
@@ -0,0 +1,48 @@
...
+import pytest
+
+import garak.cli
+from garak.generators.nim import NVOpenAIChat
+
+
+@pytest.mark.skipif(
+    os.getenv(NVOpenAIChat.ENV_VAR, None) is None,
+    reason=f"NIM API key is not set in {NVOpenAIChat.ENV_VAR}",
+)
```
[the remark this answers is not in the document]
leondz: That's a great idea, yes. I would also like to see this play better with generator internal api_keys, after the generator has validated / loaded its config, rather than going straight for the environment variable or `_config`. Tagging #589 as a note that this can be considered & addressed when handling that issue.

jmartin-tech (commented):
Functional testing passes however when executing the tests with a valid `NIM_API_KEY` in place `tests/test_attempt.py` reports an attempt to write to a file that no longer is closed.


```
======================================================================== FAILURES =========================================================================
_______________________________________________________________ test_attempt_sticky_params ________________________________________________________________

capsys = <_pytest.capture.CaptureFixture object at 0x33939e570>

    def test_attempt_sticky_params(capsys):
>       garak.cli.main(
            "-m test.Blank -g 1 -p atkgen,dan.Dan_6_0 --report_prefix _garak_test_attempt_sticky_params".split()
        )

tests/test_attempt.py:13:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
garak/cli.py:486: in main
    command.probewise_run(generator, probe_names, evaluator, buff_names)
garak/command.py:212: in probewise_run
garak/harnesses/probewise.py:106: in run
garak/harnesses/base.py:124: in run
    evaluator.evaluate(attempt_results)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <garak.evaluators.base.ThresholdEvaluator object at 0x3389ad370>, attempts = [<garak.attempt.Attempt object at 0x3394dbad0>]

    def evaluate(self, attempts: List[garak.attempt.Attempt]) -> None:
        """
        evaluate feedback from detectors
        expects a list of attempts that correspond to one probe
        outputs results once per detector
        """

        if len(attempts) == 0:
            logging.debug(
                "evaluators.base.Evaluator.evaluate called with 0 attempts, expected 1+"
            )
            return

        self.probename = attempts[0].probe_classname
        detector_names = attempts[0].detector_results.keys()

        for detector in detector_names:
            all_passes = []
            all_outputs = []
            for attempt in attempts:
                passes = [
                    1 if self.test(r) else 0
                    for r in map(float, attempt.detector_results[detector])
                ]
                all_passes += passes
                all_outputs += attempt.outputs
                    if not self.test(score):  # if we don't pass
                        if not _config.transient.hitlogfile:
                            if not _config.reporting.report_prefix:
                                hitlog_filename = f"{_config.reporting.report_dir}/garak.{_config.transient.run_id}.hitlog.jsonl"
                            else:
                                hitlog_filename = (
                                    _config.reporting.report_prefix + ".hitlog.jsonl"
                                )
                            logging.info("hit log in %s", hitlog_filename)
                            _config.transient.hitlogfile = open(
                                hitlog_filename,
                                "w",
                                buffering=1,
                                encoding="utf-8",
                            )

                        trigger = None
                        if "trigger" in attempt.notes:
                            trigger = attempt.notes["trigger"]
                        elif "triggers" in attempt.notes:
                            if (
                                isinstance(attempt.notes["triggers"], list)
                                and len(attempt.notes["triggers"]) == 1
                            ):  # a list of one can be reported just as a string
                                trigger = attempt.notes["triggers"][0]
                            else:
                                trigger = attempt.notes["triggers"]
>                       _config.transient.hitlogfile.write(
                            json.dumps(
                                {
                                    "goal": attempt.goal,
                                    "prompt": attempt.prompt,
                                    "output": attempt.outputs[idx],
                                    "trigger": trigger,
                                    "score": score,
                                    "run_id": str(_config.transient.run_id),
                                    "attempt_id": str(attempt.uuid),
                                    "attempt_seq": attempt.seq,
                                    "attempt_idx": idx,
                                    "generator": f"{_config.plugins.model_type} {_config.plugins.model_name}",
                                    "probe": self.probename,
                                    "detector": detector,
                                    "generations_per_prompt": _config.run.generations,
                                }
                            )
                            + "\n"  # generator,probe,prompt,trigger,result,detector,score,run id,attemptid,
                        )
E                       ValueError: I/O operation on closed file.

garak/evaluators/base.py:91: ValueError
------------------------------------------------------------------ Captured stdout call -------------------------------------------------------------------
garak LLM security probe v0.9.0.12.post1 ( https://github.com/leondz/garak ) at 2024-05-07T15:06:41.549252
📜 reporting to _garak_test_attempt_sticky_params.report.jsonl
🦜 loading generator: Test: Blank
🕵️  queue of probes: atkgen.Tox, dan.Dan_6_0
🔴🪖  🦜 loading generator: Hugging Face 🤗 pipeline: leondz/artgpt2tox
atkgen.Tox                                                                toxicity.ToxicCommentModel: PASS  ok on    2/   2
dan.Dan_6_0                                                                                  dan.DAN: PASS  ok on    1/   1
------------------------------------------------------------------ Captured stderr call -------------------------------------------------------------------

================================================================= short test summary info =================================================================
FAILED tests/test_attempt.py::test_attempt_sticky_params - ValueError: I/O operation on closed file.
================================================== 1 failed, 908 passed, 5 skipped in 191.22s (0:03:11) ===================================================
```
This consistently fails in local testing on my end. However executing the single test file in isolation passes:
```
% python -m pytest tests/test_attempt.py
=================================================================== test session starts ===================================================================
platform darwin -- Python 3.12.2, pytest-8.1.1, pluggy-1.4.0
rootdir: /Users/jemartin/Projects/nvidia/garak
configfile: pyproject.toml
plugins: anyio-4.3.0, requests-mock-1.12.1
collected 1 item

tests/test_attempt.py .                                                                                                                             [100%]

==================================================================== 1 passed in 4.77s ====================================================================
```
And fails in the same manor when skipped `nim` tests execute first:
```
python -m pytest tests/generators/test_nim.py tests/test_attempt.py
```

This suggests there is some sort of data leakage or stale state in global config introduced by the skipped tests.

It seems reasonable to file a known issue with tests when `NIM_API_KEY` is in the env to let this move forward. Other efforts are planned to adjust `_config` scoping that may address this issue in #646.

leondz:
Interesting. Thanks for writing this up. Yes, agree with your hypothesis. 

The failure may be related to these - tagging for reference:
* test order affects test success/speed. it shouldn't #473
* _config object contains state not suitable for multiprocessing #646
* Hitlogfile write exception #599

If we're lucky, addressing #646 may clear up a bunch of these.

leondz:
Now getting the exact same error. Will prioritise fixing this for release milestone.

PR #638, opened 2024-04-29
Feature/azure

Sunforger:
This PR adds Azure OpenAI generator support as garak/generators/azure.py.
To use this generator, put your Azure OpenAI API key in
an environment variable called AZUREAI_API_KEY. Put your Azure OpenAI endpoint
in an environment variable called AZURE_ENDPOINT


leondz:
Thanks! Could you sign the DCO? Then we'll review :)

leondz:
OK! Say something if there's anything we can do to help.

leondz:
Hi @Sunforger !

We've added new OpenAI API compatibility which may be useful for things like Azure. There's an example, much simpler class doing this now; see the NIM generator for an example of how to structure it: https://github.com/leondz/garak/blob/main/garak/generators/nim.py

If you're still working on this, please use the new lightweight pattern - it might only be a couple of lines.

jmartin-tech:
Thank you for the submission, closing in favor of #817

PR #644, opened 2024-05-01
support chat modality in attempts

### LINKED ISSUE #109: Support chat modality
As distinct from continuation

Implement generic framework that tracks state

leondz:
new pattern: chat histories are supported and managed entirely in `Attempt`

Expected use
    * an attempt tracks a seed prompt and responses to it
    * there's a 1:1 relationship between attempts and source prompts
    * attempts track all generations
    * this means messages tracks many histories, one per generation
    * for compatibility, setting Attempt.prompt will set just one turn, and this is unpacked later
      when output is set; we don't know the # generations to expect until some output arrives
    * to keep alignment, generators need to return aligned lists of length == #generations

Implementation is by selectively overriding access to `Attempt.prompt` and `Attempt.outputs`. The hope is to retain existing compatibility across garak, while also tracking message histories for plugins that use them.

Managing histories means output tracking, so we need a 1:1 mapping from prompt to output. Because there are typically multiple output generations to one seed prompt, this needs to managed. Two options:

1. `Attempt` tracks many histories, one for each of the number of output generations requested. This requires being aware of / sensitive to the number of generations.
2. Many `Attempts`, one per generation

Because 2. breaks the underlying data model, we prefer 1. This does mean being aware of the number of outputs, so a few consistency checks are built in to `attempt`, as well as silently upgrading from one to multiple histories once we know the number of outputs. Tests are included.

Patterns/expectations for Attempt **getting**:
    `.prompt` - returns the first user prompt
    `.outputs` - returns the most recent model outputs
    `.latest_prompts` - returns a list of the latest user prompts
    `.all_outputs` - returns all model outputs from all dialogue turns

Patterns/expectations for Attempt **setting**:
    `.prompt` - sets the first prompt, or fails if this has already been done
    `.outputs` - sets a new layer of model responses. silently handles expansion of prompt to multiple histories
    `.latest_prompts` - adds a new set of user prompts

pending work: `atkgen` has a roll-your-own chat model, which is not compatible

resolves #109


leondz:
todo: -- refactor getters for `.latest_prompts` and `.outputs`

leondz:
todo: 
* [ ] check the integration with buffs, and write a test or two

garak/attempt.py
```diff
@@ -42,38 +68,24 @@ def __init__(
...
         probe_params=None,
         targets=None,
-        outputs=None,
         notes=None,
         detector_results=None,
         goal=None,
         seq=-1,
     ) -> None:
         self.uuid = uuid.uuid4()
+        self.messages = []
```
erickgalinkin, line 77: Are we happy with a list of dicts? Does it make sense to implement an abstract base class for this?

erickgalinkin: That way we can enforce roles, ordering, etc.

garak/detectors/promptinject.py
```diff
@@ -19,16 +19,17 @@ class AttackRogueString(Detector):
...
-            for output in attempt.outputs:
-                if not case_sensitive:
-                    trigger, output = trigger.lower(), output.lower()
-                if trigger in output:
-                    results.append(1.0)
-                else:
-                    results.append(0.0)
-        except:
-            pass
+        print(attempt.as_dict())
```
erickgalinkin, line 22: PRINT STATEMENT

/me cries in @jmartin-tech

garak/probes/base.py
```diff
@@ -136,6 +135,8 @@ def _mint_attempt(self, prompt, seq=None) -> garak.attempt.Attempt:
         new_attempt.status = garak.attempt.ATTEMPT_STARTED
         new_attempt.goal = self.goal
         new_attempt.seq = seq
+        if prompt is not None:
+            new_attempt.prompt = prompt
```
erickgalinkin, line 139: What will the value of `new_attempt.prompt` be if prompt is None? Won't it also be None? I'm not sure this change is doing anything. Should we raise something here instead?

garak/detectors/promptinject.py
leondz: Thank you!

I'm Torvalds-aligned on debuggers. Saved me time once since the eighties. Borland TP7 had an OK toolkit I guess but even then, stdout wins. Less tool more work. Back in my day <grumbles in genx>

garak/probes/base.py
jmartin-tech: `None` is the default value for `Attempt()`, would it be clearer to refactor this whole method to something like:
``` Python
    def _mint_attempt(self, prompt=None, seq=None) -> garak.attempt.Attempt:
        new_attempt = garak.attempt.Attempt(
            "prompt" = prompt,
            "probe_classname" = (
                str(self.__class__.__module__).replace("garak.probes.", "")
                + "."
                + self.__class__.__name__
            ),
            "status" = garak.attempt.ATTEMPT_STARTED,
            "goal" = self.goal
            "seq" = seq
        )
        new_attempt = self._attempt_prestore_hook(new_attempt, seq)
        return new_attempt
```

leondz: `None` prompts are now not valid. Setting `prompt` triggers building a conversation data structure behind the scenes. Doing `attempt.prompt = None` translates to building a single-turn invalid dialogue `[{"role":"user", "content":None}]`, which I don't think we want to allow; blank prompts are represented with empty string `""`. I've made this constraint clearer in `garak.attempt.Attempt` which now raises `TypeError` if one tries to set `attempt.prompt` to `None`.

It should however be possible to construct an `Attempt` without specifying a value for `prompt`, which is the represented in the change in order of operations & the `is None` check seen here. In that case, one can defer setting `attempt.prompt` and just instantiate an `Attempt` without a prompt. This is used where we want an `attempt` object, e.g. for its `uuid`, _before_ the initial prompt text is known.

The requirement to be able to mint an attempt without knowing the prompt led to the logic here: `_mint_attempt` should only set a prompt, and thus invoke the behind-the-scenes orchestration, if a prompt is actually given.

We can refactor `_mint_attempt()` to use the constructor, yeah, will do.

leondz: Oh right, that's now possible. OK!

PR #651, opened 2024-05-02
Add JSON Path support for RestGenerator

leondz:
> currently, when services return json, folk can only specify a top-level dict key from where to get the response
>
> proposal: move to a JSONPath in the response_json_field variable


garak/generators/rest.py
```diff
@@ -256,18 +276,32 @@ def _call_model(self, prompt: str, generations_this_call: int = 1):
...
-                str(e),
-                resp.content,
-            )
-            return None
+            return str(resp.text)
+
+        response_object = json.loads(resp.content)
+
+        # if response_json_field starts with a $, treat is as a JSONPath
+        if self.response_json_field[0] != "$":
```
jmartin-tech, line 284: Might be worth guarding this for `None` or empty. Consider adding the guards in `__init__()` above to avoid possible user config error.

jmartin-tech (approved):
Minor optional request for better input validation.

PR #653, opened 2024-05-03
Bugfix/visual jailbreak pause

leondz:
Modality matching & selection isn't implemented in harnesses yet, so this plugin is disabled by default to avoid surprising UX when "all" is selected as a target. Also add docs

### DIFF
--- a/docs/source/detectors.rst
+++ b/docs/source/detectors.rst
@@ -26,3 +26,4 @@ garak.detectors
    garak.detectors.specialwords
    garak.detectors.toxicity
    garak.detectors.xss
+   garak.detectors.visual_jailbreak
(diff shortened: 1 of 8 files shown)


leondz:
@jmartin-tech have you any idea what's causing this possible false negative test failure? the test only fails on the `pull_request_target` trigger and not the `push` trigger, and I can't replicate it locally. i've left the branch intact just in case.

jmartin-tech:
The failures stopped when this PR merged, because `pull_request_target` runs in the [context of the parent repository](https://runs-on.com/github-actions/pull-request-vs-pull-request-target/) and the `main` branch which did not have the files required causing the failure.  The trigger on `pull_request_target` should only apply to `CLAAssistant` so actions will need to be watched for a bit to determine why `pull_request_target` is executing the `Garak pytest` actions.

Issue #646, opened 2024-05-01
_config object contains `state` suitable of multiprocessing


jmartin-tech:
# Overview
Many `garak._config` namespace variables such as `transient, run, reporting` are currently populated at state during execution. The management of these values between testing runs for a long running service does not lend to a singleton as then only one of each can exist at a time in the python interpreter. Further since this data is populated into a namespace variable reload of the namespace will not maintain these values.

The behavior of the `multiprocessing` package is more akin to launching another new python process that uses only the files directly connected to the source objects passed.

#645 addresses an instance that exposed this concern.

### Current state example
In the case of the `probe` class and its associated class hierarchy being called the only argument is defined in the same class and state of config is not passed.

### Desired state
Any stateful `_config` likely needs to either be passed or consistently reloaded in new processes. Investigation and understanding of use cases needs occur to consolidate on patterns either in code standards or library framework supporting task execution with all required context.

leondz:
How should we triage/roadmap this fella?

jmartin-tech:
There is some ideation about possibly injecting a shared memory location to sync config across process and thread boundaries, maybe set a goal to introduce this in this quarter.

The idea proposed is for module level code in `_config` to search for an existing shared memory object that would be created in the parent runtime when `_config` is locked to start a run.

PR #665, opened 2024-05-08
handle extant but closed `hitlogfile` file

### LINKED ISSUE #599: Hitlogfile write exception

leondz:
evalutors/base is now more defensive and can recover from closed hitlogfiles

cf. https://github.com/leondz/garak/pull/637#pullrequestreview-2044150946 - testing now passes with a `NIM_API_KEY` set 

resolves #599 

NB: This PR leaves unresolved the underlying misbehaviour that leads to `_config.transient.hitlogfile` closing before the scan completes. That behaviour is potentially a side-effect of insufficient state management of `_config`, tracked in #646. We might consider undoing this PR when #646 is resolved, or blocking it in order to testing #646. Adding an assert to monitor `hitlogfile` state may be sufficient when completing #646.

All that said: pending the larger change to `_config` state, this PR should enable hitlog file writing to at least recover from inconsistent state.

### DIFF
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -62,7 +62,13 @@ class Evaluator:
                 all_outputs += attempt.outputs
                 for idx, score in enumerate(attempt.detector_results[detector]):
                     if not self.test(score):  # if we don't pass
-                        if not _config.transient.hitlogfile:
+                        if (
+                            _config.transient.hitlogfile is None
+                            or _config.transient.hitlogfile.closed
+                        ):
+                            hitlog_mode = (
+                                "w" if _config.transient.hitlogfile is None else "a"
+                            )
                             if not _config.reporting.report_prefix:
                                 hitlog_filename = f"{_config.reporting.report_dir}/garak.{_config.transient.run_id}.hitlog.jsonl"
                             else:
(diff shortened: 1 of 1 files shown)


jmartin-tech (approved):
This gets the job done.  I agree this will likely be reverted or adjusted by #646.

@leondz do we need a `Sign-off` in the commit message or can we land this as is since there is a DCO signature already?

leondz:
DCO signature in place should be enough.

PR #670, opened 2024-05-10
Make banner line test dynamic

jmartin-tech:
Build the banner string check in a dynamic pattern

This avoids test failures when the application name or description impact the banner message for tests that are filtering the banner.


leondz (approved):
nice

PR #677, opened 2024-05-10
Guard cli run with invalid or incomplete config

### LINKED ISSUE #191: fail if non-existing plugin is requested
* definitely fail & quit if no probes load / no detectors load / generator load fails
* print an ❌ by failed loading attempts, perhaps
* probably skip unloadable probes, with a yellow SKIP message that follows the same design language as used in evaluator

jmartin-tech:
Fix #191 

* probe/detector/buff configuration checks
  * invalid or missing values generate an error message and exit early
  * a [bool] cli argument `--skip_unknown` logs unknown configurations & continues to `start_run`
* front load argument parsing for probes and detectors
  * defers run log creation until run config is validated

Local test examples
```
python -m garak -m test -n mistralai/mixtral-8x22b-v0.1 -g 1 -p dan.DanInTheWild,doesnotexist,dan.Invalid --skip_unknown
python -m garak -m test -n mistralai/mixtral-8x22b-v0.1 -g 1 -p dan.DanInTheWild,doesnotexist
python -m garak -m test -n mistralai/mixtral-8x22b-v0.1 -g 1 -p doesnotexist
python -m garak -m test -n mistralai/mixtral-8x22b-v0.1 -g 1 -p doesnotexist --skip_unknown
```

### DIFF
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -162,7 +162,8 @@ def parse_plugin_spec(
     from garak._plugins import enumerate_plugins
 
     if spec is None or spec.lower() in ("", "auto", "none"):
-        return []
+        return [], []
+    unknown_plugins = []
     if spec.lower() in ("all", "*"):
         plugin_names = [
             name
@@ -173,11 +174,15 @@ def parse_plugin_spec(
         plugin_names = []
         for clause in spec.split(","):
             if clause.count(".") < 1:
-                plugin_names += [
+                found_plugins = [
                     p
                     for p, a in enumerate_plugins(category=category)
                     if p.startswith(f"{category}.{clause}.") and a is True
                 ]
+                if len(found_plugins) > 0:
+                    plugin_names += found_plugins
+                else:
+                    unknown_plugins += [clause]
             else:
                 plugin_names += [f"{category}.{clause}"]  # spec parsing
 
@@ -196,4 +201,4 @@ def parse_plugin_spec(
         for plugin_to_skip in plugins_to_skip:
             plugin_names.remove(plugin_to_skip)
 
-    return plugin_names
+    return plugin_names, unknown_plugins
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -184,7 +184,16 @@ def parse_plugin_spec(
                 else:
                     unknown_plugins += [clause]
             else:
-                plugin_names += [f"{category}.{clause}"]  # spec parsing
+                # validate the class exists
+                found_plugins = [
+                    p
+                    for p, a in enumerate_plugins(category=category)
+                    if p == f"{category}.{clause}" and a is True
+                ]
+                if len(found_plugins) > 0:
+                    plugin_names += found_plugins
+                else:
+                    unknown_plugins += [clause]
 
     if probe_tag_filter is not None and len(probe_tag_filter) > 1:
         plugins_to_skip = []
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -17,7 +17,9 @@ import yaml
 
 version = -1  # eh why this is here? hm. who references it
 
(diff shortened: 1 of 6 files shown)


garak/_config.py
```diff
@@ -158,11 +158,12 @@ def load_config(
 
 def parse_plugin_spec(
     spec: str, category: str, probe_tag_filter: str = ""
-) -> List[str]:
+) -> tuple[List[str], List[str]]:
```
jmartin-tech, line 161: I would usually prefer a typed single object returned, however this method is currently only consumed by `cli.py`, to limit impacts a tuple was returned for `fully resolved accepted` and `rejected` values from the spec offered.

garak/cli.py
```diff
@@ -3,7 +3,7 @@
 
 """Flow for invoking garak from the command line"""
 
-command_options = "list_detectors list_probes list_generators list_buffs list_config plugin_info interactive report version".split()
+command_options = "list_detectors list_probes list_generators list_buffs list_config plugin_info interactive report version skip_unknown".split()
```
leondz, line 6: Usually, `command_options` lists things that determine the main part of what garak will do in this execution. `skip_unknown` seems more like a config option, which shouldn't be stored here. Later, things in `command_options` are handled separately from config, which also seems undesirable.

garak/_config.py
```diff
@@ -158,11 +158,12 @@ def load_config(
...
     from garak._plugins import enumerate_plugins
 
     if spec is None or spec.lower() in ("", "auto", "none"):
-        return []
+        return [], []
+    unknown_plugins = []
```
leondz, line 166: Variable name `unknown_plugins` suggests the thing entered is a plugin, which isn't true if it's not found - can this be something like `unknown_clauses` instead?

garak/_config.py
leondz: Have you considered returning a dict? Or raising an exception in the case that there are unknown clauses, and returning the leftover pieces in the exception, for the calling method to handle however it chooses? Maybe gating exception raising/ignoring on a named method var here?

garak/cli.py
```diff
@@ -237,6 +238,11 @@ def main(arguments=[]) -> None:
         action="store_true",
         help="Launch garak in interactive.py mode",
     )
+    parser.add_argument(
+        "--skip_unknown",
+        action="store_true",
+        help="allow skip of unknown probes or detectors",
```
leondz, line 244: Only probes & detectors? I guess harness and generators we always want to just stop - what about buffs?

garak/cli.py
```diff
@@ -327,9 +333,6 @@ def main(arguments=[]) -> None:
     import garak.evaluators
 
     try:
-        if not args.version and not args.report:
-            command.start_run()
-
```
leondz, line 332: talk me through this :)

garak/cli.py
jmartin-tech: Expanding message to extend to `buffs` as well.

garak/cli.py
jmartin-tech: This call has simply moved to after configuration validation is complete to only start in `interactive` mode or when the processing has verified a `run` will occur.

garak/cli.py
jmartin-tech: Moving this to be a `system_params` value. We discussed separately that this option only really makes sense in the context of validation of overall configuration in the `cli` class, however after some review this may be analogous with `parallel_*` and `verbose` as a behavior of the overall system config.

leondz:
after [b7ed214](https://github.com/leondz/garak/pull/677/commits/b7ed2148d9f2e25f558c5cf1c1e22a86978e8b66), lgtm

PR #685, opened 2024-05-15
convert module-level constant names to uppercase

leondz:


jmartin-tech (commented):
Conversion looks reasonable to me, is it possible to also give variables in tests the same treatment now to get consistent style throughout?

PR #688, opened 2024-05-16
Tutorial for contributing a generator

leondz:


docs/source/contributing.generator.rst
```diff
@@ -0,0 +1,255 @@
...
+
+.. code-block:: python
+
+            self.replicate = importlib.import_module("replicate")
+
+Finally, forget to import ``importlib`` at the top!
+
+.. code-block:: python
+
+    import importlib
```
jmartin-tech, line 100: While I agree on late loading, providing guidance on having `__init__` import of objects required for the class to function is not going to be `pickle` friendly for any multiprocessing.

docs/source/contributing.generator.rst
```diff
@@ -0,0 +1,255 @@
...
+
+Exception handling
+******************
+
+Many things can go wrong when trying to get inference out of LLMs. Things that can go wrong with web-hosted services, such as Replicate, include running out of funds, or the model going down, or hitting a rate limit. These are sometimes presented to the coder in the form of exceptions.
+
+Backoff
+=======
+
+We need to work out a strategy of what to do when these exceptions are raised. Fortunately, the Replicate API module is fairly well-coded, and handles a lot of the recoverable failure cases itself. However, transient exceptions shouldn't stop a garak run - runs can take days, and aborting a run with an uncaught exception after dozens of hours is probably loss desirable. So we should handle them
```
jmartin-tech, line 143:
```suggestion
We need to work out a strategy of what to do when these exceptions are raised. Fortunately, the Replicate API module is fairly well-coded, and handles a lot of the recoverable failure cases itself. However, transient exceptions shouldn't stop a garak run - runs can take days, and aborting a run with an uncaught exception after dozens of hours is probably less desirable. So we should handle them
```

docs/source/contributing.generator.rst
```diff
@@ -0,0 +1,255 @@
...
+=================
+
+If the request really can't be served - maybe the prompt is longer than the context window and there's no specific handling in this case - then ``_call_model`` can return a ``None``. In the case of models that support multiple generations, ``_call_model`` should return a list of outputs and, optionally, ``None``s, with one list entry per requested generation.
+
+Testing
+=======
+
+Now that the pieces for our generator are in place - a subclass of ``garak.generators.base.Generator``, with some customisation in the constructor, and an overridden ``_call_model()`` method, plus a ``DEFAULT_CLASS`` given at module level - we can start to test.
+
+A good first step is to fire up the Python interpreter and try to import the module. Garak supports a specific range of Python versions (listed in `pyproject.toml <https://github.com/leondz/garak/blob/main/pyproject.toml>`_, under the ``classifiers`` descriptor), so remember to use the right Python version for testing.
```
jmartin-tech, line 173:
```suggestion
A good first step is to fire up the Python interpreter and try to import the module. Garak supports a specific range of tested Python versions (listed in `pyproject.toml <https://github.com/leondz/garak/blob/main/pyproject.toml>`_, under the ``classifiers`` descriptor), so remember to use the right Python version for testing.
```

docs/source/contributing.generator.rst
```diff
@@ -0,0 +1,255 @@
...
+The next step is to try some integration tests - executing garak from the command line, accessing this generator. There are some pointers in :doc:`contributing`. You might need to execute garak by specifying it as a Python module, running the command from the garak root code directory. Things to test are:
+
+* Does the new generator appear in ``python -m garak --list_generators``?
+* Does the generator work with a test probe, via ``python -m garak -m replicate -n meta/llama-2-70b-chat -p test.Blank``?
+* Do the garak tests pass? ``python -m pytest tests/``
+
+
+If you want to see the full, live code for the Replicate garak generator, it's here: `garak/generators/replicate.py <https://github.com/leondz/garak/blob/main/garak/generators/replicate.py>`_ .
+
+Done!
```
jmartin-tech, line 238: Consider adding a comment that any edge case behaviors in `__init__` or `_call_model` may warrant adding a testing class to mock and exercise those conditions.

docs/source/contributing.rst
```diff
@@ -4,38 +4,63 @@ Contributing
...
+
+Testing
+-------
+
+You can test your code in a few ways:
+
+* Start an interactive Python session
+   * Import the model, e.g. `import garak.probes.mymodule`
+   * Instantiate the plugin, e.g. `p = garak.probes.mymodule.MyProbe()`
+* Get `garak` to list all the plugins of the type you're writing, with `--list_probes`, `--list_detectors`, or `--list_generators`
```
jmartin-tech, line 54: Maybe an example here?
```suggestion
* Get `garak` to list all the plugins of the type you're writing, with `--list_probes`, `--list_detectors`, or `--list_generators`: `python3 -m garak --list_probes`
```

docs/source/contributing.rst
```diff
@@ -4,38 +4,63 @@ Contributing
...
+* Run a scan with test plugins
+   * For probes, try a blank generator and always.Pass detector: `python3 -m garak -m test.Blank -p mymodule -d always.Pass`
+   * For detectors, try a blank generator and a blank probe: `python3 -m garak -m test.Blank -p test.Blank -d mymodule`
+   * For generators, try a blank probe and always.Pass detector: `python3 -m garak -m mymodule -p test.Blank -d always.Pass`
```
jmartin-tech, line 58: This was already here in another form os just noting for awareness. There was some conversation about possible shifting the `test` plugins to be only available in testing. If we want to have guidance in docs that refers to using them like this for examples they will need to stay exposed, although we may be able to exclude them from being listed by `--list_*`

jmartin-tech (commented):
Walk through hits all the major points, I added a few comments that might need to do some think on and could possibly defer to later iteration on the guide.

docs/source/contributing.generator.rst
leondz: Do we have a threadsafe pattern, e.g. the one in `openai.py`? Shall we move the import to module-level for simplicity?

docs/source/contributing.rst
leondz: Right. I think that discussion was only about the `Single` generator used solely in unit testing - the test plugins help troubleshoot a garak setup, so I'd prefer to have them available generally.

docs/source/contributing.generator.rst
leondz: done in 163b4e5

docs/source/contributing.generator.rst
jmartin-tech: Short term I can live with this guidance, however I suspect we should keep track of this as something we want to iterate on, and remembering there is guidance here may be easily overlooked.

Not sure what the best pattern will be, I think module or class level is easier on the contributor, however the plugin loader currently would take the full import load at that point and remove most of the cost saved by late loading.

If we shift the burden of enumerating the plugin list to an automation job that maintains a cache file used for initial load and only attempt to `search` for plugins that are not `known` to the cache we might be able to shift the import load until the class is actually instantiated during a run for users that are not adding custom plugins.

leondz: I like the idea of a cache, let's aim for that.

PR #694, opened 2024-05-23
Set generator `_call_model()` and `_generate()` type hints; amend functions

leondz:


garak/generators/huggingface.py
```diff
@@ -548,7 +556,7 @@ def __init__(self, name="", generations=10):
                 "CUDA is not supported on this device. Please make sure CUDA is installed and configured properly."
             )
 
-    def generate(self, prompt) -> List[str]:
+    def generate(self, prompt) -> List[Union[str, None]]:
```
jmartin-tech, line 559: Should this be adjusted to include `generations_this_call`?
```suggestion
    def generate(self, prompt: str, generations_this_call: int = 1) -> List[Union[str, None]]:
```

garak/generators/nvcf.py
```diff
@@ -64,9 +65,11 @@ def __init__(self, name=None, generations=10):
         ),
         max_value=70,
     )
-    def _call_model(self, prompt: str, generations_this_call: int = 1) -> str:
+    def _call_model(
+        self, prompt: str, generations_this_call: int = 1
+    ) -> List[Union[str, None]]:
         if prompt == "":
-            return ""
+            return [None]
```
jmartin-tech, line 72: This is an interesting, in https://github.com/leondz/garak/pull/682#discussion_r1608125584 there is suggestion that `None` should be returned when an error is handled. In this case the method just preempts the sending the prompt when it is an empty string. Does the `NVCF` spec document an empty prompt as being an error?

garak/generators/nim.py
```diff
@@ -61,7 +61,7 @@ def _clear_client(self):
 
     def _call_model(
         self, prompt: str | List[dict], generations_this_call: int = 1
```
jmartin-tech, line 63: For consistency with `openai.py`:
```suggestion
        self, prompt: Union[str, list[dict]], generations_this_call: int = 1
```

This also points out `list` vs `List` usage in this method signature/

jmartin-tech (commented):
Goal of this PR is consistency, I made a couple comments in that vein.

garak/generators/nvcf.py
leondz: Not last time I checked, but it does raise one. I can see where this should go - `try`/`except` first.

garak/generators/nim.py
leondz: how come `list` not `typing.List`?

jmartin-tech: The proposed line is simply what is already in `openai.py`. This is why I commented that there look to be different types requested here, however python looks to treat them as consistent for `isinstance()`:
```
% python
Python 3.10.14 (main, Apr 24 2024, 15:37:54) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import typing
>>> val = list()
>>> val
[]
>>> isinstance(val, typing.List)
True
>>> isinstance(val, list)
True
```

Note that `List` cannot be instantiated directly making this simply be a styling item:
```
>>> from typing import List
>>> my_val = List()
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/typing.py", line 955, in __call__
    raise TypeError(f"Type {self._name} cannot be instantiated; "
TypeError: Type List cannot be instantiated; use list() instead
```

PR #696, opened 2024-05-24
Split NVCF generator into completion chat

leondz:
NVCF models support at least two interfaces - Completion and Chat styles. They also sometimes require custom headers. This PR splits the NVCF class into two, one for each interface, providing separate methods for building the payload and parsing the output under each interface style. It also adds code to allow external configuration of the payload fields that the NVCF endpoint is queried with.

### DIFF
--- a/garak/generators/nvcf.py
+++ b/garak/generators/nvcf.py
@@ -33,7 +33,7 @@ class NvcfChat(Generator):
 
     def __init__(self, name=None, generations=10):
         self.name = name
-        self.fullname = f"NVCF {self.name}"
+        self.fullname = f"{self.generator_family_name} {self.__class__.__name__}"
         self.seed = _config.run.seed
 
         if self.name is None:
--- a/garak/generators/nvcf.py
+++ b/garak/generators/nvcf.py
@@ -114,6 +114,7 @@ class NvcfChat(Generator):
 
         if 400 <= response.status_code < 600:
             logging.warning("nvcf : returned error code %s", response.status_code)
+            logging.warning("nvcf : payload %s", repr(payload))
             logging.warning("nvcf : returned error body %s", response.content)
             if response.status_code == 400 and prompt == "":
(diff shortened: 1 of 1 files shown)


leondz:
ps. i know, i know, sorry about the _config indexing jeffrey

garak/generators/nvcf.py
```diff
@@ -33,7 +33,7 @@ class NvcfGenerator(Generator):
 
     def __init__(self, name=None, generations=10):
         self.name = name
-        self.fullname = f"NVCF {self.name}"
+        self.fullname = f"{self.generator_family_name} {self.__class__.__name__}"
```
jmartin-tech, line 36: This should likely still include the passed in `name` of the model:
```suggestion
        self.fullname = f"{self.generator_family_name} {self.__class__.__name__} {self.name}"
```

garak/generators/nvcf.py
```diff
@@ -55,6 +55,21 @@ def __init__(self, name=None, generations=10):
...
+            "temperature": self.temperature,
+            "top_p": self.top_p,
+            "max_tokens": self.max_tokens,
+            "stream": False,
+        }
+
+        return payload
+
+    def _extract_text_output(self, response) -> str:
+        return response["choices"][0]["message"]["content"]
```
jmartin-tech, line 71: Consider adjusting to have the `List` wrapper in the return signature here and avoiding need to `wrap` in `_call_model` returns. Same for the `NvcfCompletion` implementation.

This seems similar to `openai.py` consider: (syntax may differ for access to dict vs object members though)
```suggestion
    def _extract_text_output(self, response) -> List[str]:
        return [c.message.content for c in response.choices]
```

leondz:
> Stray thought as I looked at this, are these services really just OpenAICompatible like NIM? ( This is likely a futures effort thought. )

One would hope so! I like where this is going.

PR #701, opened 2024-05-27
patch ambiguity around generate() filtering or not filtering `None`s

leondz:
While the aspirational behaviour of `generate()` has changed from filtering out `None` entries to leaving them in, allowing generators to express out-of-band failures to detectors, the current state of detectors' ability to handle `None` well is unknown. Until that can be stabilised, the code in `generate()` that filters out `None` should remain, and this Hugging Face test around empty input (leading to an exception) should expect the current, and not aspirational, behaviour. When we get to the goal state, this revised test should fail, and the code there can be updated accordingly.


leondz:
cancelling, #702 fixes this ambiguity

PR #706, opened 2024-05-29
add CONTRIBUTING.md, update contributing guide in docs

leondz:


jmartin-tech (approved):
All checks out for me, a thought possibly for future iteration. We might want a note that the project suggest using a responsible disclosure process if applicable and that any new packages or contributed dataset needs to provide source credits as well as a compatible license.

leondz:
Good point. The DCO process provides protection and health for our project but doesn't necessarily protect others, as this suggestion would.

PR #712, opened 2024-05-31
doc & inheritance fixes to DanInTheWildMini

leondz:
set direct inheritance for DanInTheWildMini, pending plugin scanning update; add docstring

### DIFF
--- a/garak/probes/dan.py
+++ b/garak/probes/dan.py
@@ -651,7 +651,12 @@ class DanInTheWild(Probe):
             self.prompts = f.readlines()
 
 
-class DanInTheWildMini(DanInTheWild):
+class DanInTheWildMini(DanInTheWild, Probe):
+    """A subset of 100 jailbreaks from the DanInTheWild jailbreak library.
+
+    Tries a set of 100 known jailbreaks, in an attempt to get a model to disregard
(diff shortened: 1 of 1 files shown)


garak/probes/dan.py
```diff
@@ -651,7 +651,12 @@ def __init__(self):
             self.prompts = f.readlines()
 
 
-class DanInTheWildMini(DanInTheWild):
+class DanInTheWildMini(DanInTheWild, Probe):
```
jmartin-tech, line 654:
```suggestion
class DanInTheWildMini(DanInTheWild):
```

`DanInTheWild` already extends `Probe` is there a reason I am not aware of to also list it here?

leondz: Without #711 or #712 this probe doesn't appear in `--list_probes`, because the old/current enumeration only looked for direct inheritance

jmartin-tech: I had noticed that when building out #711, I would offer the value here is adding the doc and the actually execution of the doc test can be deferred to when #711 lands and improves the enumeration in `_plugins`

leondz: Yeah, the docstring seems needed either way - I think I'd wrongly assumed it would be inherited, and then incomplete probe enumeration meant the test missed this.

PR #645, opened 2024-05-01
Enable parallel OpenAI calls

jmartin-tech:
Fix #493 

Conceptually since rest generator client objects should not retains state, this provides a pattern where `rest` generator client objects can be instantiated only when required and avoiding the need to `pickle` the client object when passing a generator to another process.

* refactor for just in time rest client for `OpenAI`
* ensure read and write of `_config.transient.reportfile` are in parent process only
* consolidate multiprocessing execution in `probes/base.py`

Verification testing preformed with:
```
python -m garak p dan.DanInTheWild -m openai -n gpt-4-turbo-preview -g 1 --parallel_attempts 20
```

Additional testing may be appropriate to validate changes to `base` and `gcg` probe flows are not impacted.

PR #717, opened 2024-06-03
lazi-fy huggingface, langchain serve, litellm loading

leondz:
no need to review/merge until #711 lands


garak/generators/huggingface.py
```diff
@@ -585,6 +584,12 @@ def __init__(self, name="", generations=10, config_root=_config):
             raise ModelNameMissingError(
                 f"Invalid modal name {self.name}, current support: {self.supported_models}."
             )
+
+        from transformers import LlavaNextProcessor, LlavaNextForConditionalGeneration
+
+        PIL = importlib.import_module("PIL")
+        self.Image = PIL.Image
+
```
jmartin-tech, line 592: I am a little concerned about importing libs and storing them as attributes in the generator. This will introduce similar issue with `pickle` during multiprocessing that having the client in `OpenAIGenerator` produced.

These might be better served as methods called to load when not already set combined with a `__getstate__()` implementation similar to https://github.com/leondz/garak/blob/3833eefbb29ec10f22298963f668f2b0f99c4526/garak/generators/openai.py#L111-L114

jmartin-tech (changes requested):
I think this pattern needs to follow the multiprocessing combined solutions from #645 and #689

garak/generators/huggingface.py
jmartin-tech: I am doing some testing of this pattern in other huggingface changes and while I still like the idea of protecting class for safe pickle support there might be better more work needed in general around how we share generator instances when allowing `mulitprocessing`. As is, shifting heavy lift objects to be created in each new process could be very expensive in terms of resources.

leondz: I can get this, it makes sense.

On the other hand - if one is doing `multiprocessing` with local models, the consumption of gigabytes/tens of gigabytes of GPU memory per instance seems to shrink the relative difficulties of storing libraries in the generator. 

I am inclined to adopt a safer pattern here, but to not support proactively the general case of parallelisation with locally-run model, instead implementing a `generators.base.Generator` attribute specifying whether a generator is parallelisation-compatible and setting it `False` for 🤗 classes such as `Model` and `Pipeline`. Running parallel local models seems like an edge case - the parallelisation is intended for stuff that's lighter.

jmartin-tech: > implementing a generators.base.Generator attribute specifying whether a generator is parallelisation-compatible and setting it False for 🤗 classes such as Model and Pipeline

This is the approach I am targeting first.

I am also considering looking for methods that can defer multiprocessing approaches for locally executing generators to something specifically provided by the generator, for instance `hugginface` provides the [Accelerate](https://huggingface.co/docs/accelerate/index) library to allow the generator to execute interference using GPUs efficiently and might be reasonable to expose a `_call_model_with_dataset()` or something of the sort that can recieve a set of attempts and have the generator figure out how to parallelize them.

leondz: @jmartin-tech: heavy externals in these modules now loaded/unloaded using `_load_client` / `_clear_client` pattern

garak/generators/huggingface.py
```diff
@@ -597,15 +622,16 @@ def __init__(self, name="", generations=10, config_root=_config):
             raise RuntimeError(
                 "CUDA is not supported on this device. Please make sure CUDA is installed and configured properly."
             )
+        self._load_client()
```
jmartin-tech, line 625:
```suggestion
        self._load_client()
```

garak/generators/huggingface.py
```diff
@@ -579,14 +578,40 @@ class LLaVA(Generator):
...
+
+    def _load_client(self):
+        PIL = importlib.import_module("PIL")
+        self.Image = PIL.Image
+
+        transformers = importlib.import_module("transformers")
+        self.LlavaNextProcessor = transformers.LlavaNextProcessor
+        self.LlavaNextForConditionalGeneration = (
+            transformers.LlavaNextForConditionalGeneration
+        )
```
jmartin-tech, line 599: Processor and Model need to be part of this method.
```suggestion
        )
        self.processor = self.LlavaNextProcessor.from_pretrained(self.name)
        self.model = self.LlavaNextForConditionalGeneration.from_pretrained(
            self.name,
            torch_dtype=self.torch_dtype,
            low_cpu_mem_usage=self.low_cpu_mem_usage,
        )
        if torch.cuda.is_available():
            self.model.to(self.device_map)
        else:
            raise RuntimeError(
                "CUDA is not supported on this device. Please make sure CUDA is installed and configured properly."
            )
```

garak/generators/litellm.py
```diff
@@ -142,6 +155,8 @@ def __init__(self, name: str = "", generations: int = 10, config_root=_config):
                         " or in the configuration file"
                     )
 
+        self._load_client()
```
jmartin-tech, line 158: Provider check is still needed and the key extraction should have been moved into a custom `_validate_env_var()` implementation. This looks like something I missed in #711 that incorrectly enforces an API key as required for all `provider` values.
```suggestion
        self._load_client()

    def _validate_env_var(self):
       if self.provider is None:
            raise ValueError(
                "litellm generator needs to have a provider value configured - see docs"
            )
        elif (
            self.api_key is None
        ):
            # TODO: add other providers
            if self.provider == "openai":
                super()._validate_env_var()       
```

jmartin-tech (changes requested):

garak/generators/litellm.py
leondz: So, uh, while dealing with this slow module, I started getting a failed test, but noticed that I had an openai env var key set, which meant the test actually ran. Have you ever seen the `litellm` tests pass? Looking at the tests we have, and the basic code examples on their website (see eg the invocations on https://docs.litellm.ai/docs/), the `provider` check seems to block intended functionality.

jmartin-tech: I have never had them run and pass as they both required keys, I had noted in the original PR that it seemed like a config file would be required to instantiate the class. Although there was a comment that said it was not required, the original [embedded config parsing](https://github.com/leondz/garak/pull/572/files#diff-ff886897d971b8e468ed44457e63890380c3c95e187ee701c2c34e61699bcf44R106-R134) did *require* a provider. If `provider` was not found in `_config.plugins.generators["litellm.LiteLLMGenerator"]` it would raise a `ValueError`.

I intend to validate function as part of the testing here by setting up a local instance, however there is another issue with this class as the `torch_dtype` value cannot be accepted as a string.  I have fixes for this in progress in the refactor branch I am working one.  Short term I was intending to manually patch the `torch_dtype` default value to allow testing of this change in isolation.

leondz:
Yeah, that's my read too. I'm not sure the provider requirement aligns with
documented litellm use or the garak generator test as written.

garak/generators/litellm.py
jmartin-tech: After some testing, I have validated that `LiteLLM` as implemented does required a `provider` from the config file.  This can be enforced using something like the suggestion updated in the top of this thread and would supply a method to suppress the api key requirement when supplying something other than `openai` as the `provider`.

A future PR can also expand the testing to provide a mock config that will would allow for mocking a response from the generator similar to the mock openai responses recently incorporated.

leondz: I think the "as implemented" is something to be flexible on (see example in #755 ). This would remove the enforcement requirement. Unfortunately I don't have a good pattern for validating the input. I'm OK with relaxing the provider constraint and letting `litellm` do its own validation.

leondz:
may be resolved by #768 in which case will close

PR #729, opened 2024-06-06
restore `req_template_json_object` lost in refactor

jmartin-tech:
refactor in #711 attempts to access the wrong key after configuration and needs to `allow` the key by default.


erickgalinkin (approved):
lgtm

PR #735, opened 2024-06-13
add 'complete' message to report at end of run

### LINKED ISSUE #726: add "done" entry to report.jsonl
a final entry of type "done" or "complete" including a datestamp w/ tz should be appended as the final entry to any report.jsonl, to indicate that the run finished, and when

leondz:
resolves #726

### DIFF
--- a/garak/command.py
+++ b/garak/command.py
@@ -123,7 +123,7 @@ def end_run():
     print(f"📜 report html summary being written to {digest_filename}")
     try:
         write_report_digest(_config.transient.report_filename, digest_filename)
-    except BaseException as e:
+    except Exception as e:
         print("Didn't successfully build the report - JSON log preserved.", repr(e))
 
     msg = f"garak run complete in {timetaken:.2f}s"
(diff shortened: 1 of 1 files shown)


garak/command.py
```diff
@@ -117,7 +123,7 @@ def end_run():
     print(f"📜 report html summary being written to {digest_filename}")
     try:
         write_report_digest(_config.transient.report_filename, digest_filename)
-    except Exception as e:
+    except BaseException as e:
```
jmartin-tech, line 126: Why `BaseException`? 

Catching `Exception` is very broad already, this only adds `GeneratorExit`, `KeyboardInterrupt` and `SystemExit` as far as I can tell all of which I would expect to immediately terminate vs being able to recover in any meaningful way.

leondz: reverted.

Would like to scope this down, actually - I am not sure what is being caught here, because `Exception` is an uninformative signal

jmartin-tech (approved):

PR #738, opened 2024-06-14
require probes to match input modality of the generator

jmartin-tech:
* test modality for full input match
* log a warning for probes skipped due to mismatch

More work is likely to be needed so a `TODO` has been noted. Longer term goal may be to have some compatibility matrix for instance to allow sending `text` only input to models that support `text + image`.

Future improvements will likely be be linked with work in #602

### DIFF
--- a/garak/harnesses/base.py
+++ b/garak/harnesses/base.py
@@ -96,7 +96,7 @@ class Harness(Configurable):
                 continue
             # TODO: refactor this to allow `compatible` probes instead of direct match
             if probe.modality["in"] != model.modality["in"]:
-                logging.warning("probe skipped due to modality mismatch: %s", self)
+                logging.warning("probe skipped due to modality mismatch: %s", probe)
                 continue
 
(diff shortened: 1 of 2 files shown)


garak/harnesses/base.py
```diff
@@ -94,6 +94,11 @@ def run(self, model, probes, detectors, evaluator, announce_probe=True) -> None:
             logging.debug("harness: probe start for %s", probe.probename)
             if not probe:
                 continue
+            # TODO: refactor this to allow `compatible` probes instead of direct match
+            if probe.modality["in"] != model.modality["in"]:
+                logging.warning("probe skipped due to modality mismatch: %s", probe)
```
leondz, line 99: add context to the message

```suggestion
                logging.warning("probe skipped due to modality mismatch: %s - model expects %s", (probe, model.modality["in"]))
```

leondz (approved):
minor suggestion re: logging informativeness

PR #740, opened 2024-06-17
add probe for "grandma" attack

leondz:


erickgalinkin (approved):
My inclination is that we could probably expand `ETHNIC_GROUPS` to `MARGINALIZED_GROUPS` and include others, but that's probably not necessary to demonstrate the efficacy of the method.

leondz:
The semiotics of including brits in `MARGINALIZED_GROUPS` are definitely, uh, polysemous - both unacceptable and hilarious. We've had a hard couple of decades (I hear).

Issue #671, opened 2024-05-10
support for ROCm


leondz:
* [ ] maybe this will come with device maps? check box if true
* [ ] check for cuda mentions and make them non-default, adding a config item to determine what to use

leondz:
should be resolved by #742, please re-open with details if not.

Issue #656, opened 2024-05-03
remove `generations` count param from generator constructors


leondz:
* [ ] new test that generators observe `_config.run.generations` in `self.generations` (or maybe ditch this entirely??)
* [ ] remove use of generator `generations` constructor param
* [ ] test that fails if a generator can be constructed with a `generations` param

leondz:
resovled in #837

PR #742, opened 2024-06-17
Refactor `huggingface` config support

### LINKED ISSUE #589: Configurable `Generator` parameters

### LINKED ISSUE #672: support for Metal/MPS
* [ ] maybe this will come with device maps? check box if true
* [ ] check for cuda mentions and make them non-default, adding a config item to determine what to use

jmartin-tech:
Fix #672 
Fix #589

Provide huggingface `Pipeline` and `LLaVA` configuration values as `hf_args` that can be passed via the `yaml` or `json` generator configuration paths.

* do not override config deprefix_prompt
* improve code reuse
  * consolidate `__init__` where possible
  * shift generator or model object creation to `_load_client()`
  * Wrap errors as `GarakException` vs `Exception`
* crude implementation of limitation on parallel generator call
* add torch `mps` support
  * detect `cuda` vs `mps` vs `cpu` in a common way
  * guard import of OptimimPipeline
* enable hf model or pipeline config in hf_args
  * support all generic `pipeline` args at all times
  * adds `do_sample` when `model` is a parameter to the `Callable`
  * adds `low_cpu_mem_usage` and all `pipeline` arguments for `Callables` without `model`
  * consolidates optimal device selection & set when not provided by config

~~Note: While `hugggingface.Model` not extends `Pipeline` due to how the model is loaded the only `hf_arg` currently passed is `hf_args["device"]` more work is needed to support more~~

Testing of each class type is needed and should likely be baselined code prior to #711 as few recommended possible testing configurations used on various hardware:
```
python -m garak -m huggingface.Model -n meta-llama/Llama-2-7b-chat-hf --config fast -g 1
python -m garak --model_type huggingface --model_name gpt2 --config fast -g 1
python -m garak --model_type huggingface.LLaVA --model_name llava-hf/llava-v1.6-mistral-7b-hf -p visual_jailbreak.FigStepTiny
```

Example configuration yaml:
```
plugins:
  generators:
    huggingface:
      hf_args:
        torch_dtype: float16
      Pipeline:
        hf_args:
            device: cuda
```

When no `device` is specified by configuration `garak` will now select `cuda` / `mps` / `cpu` depending on environment availability.

### DIFF
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -61,23 +61,23 @@ class HFCompatible:
 
         args = {}
 
-        parameters = inspect.signature(hf_constructor).parameters
+        params_to_process = inspect.signature(hf_constructor).parameters
 
-        if "model" in parameters:
+        if "model" in params_to_process:
             args["model"] = self.name
             # expand for
-            parameters = {"do_sample": True} | parameters
+            params_to_process = {"do_sample": True} | params_to_process
         else:
             # callable is for a Pretrained class also map standard `pipeline` params
             from transformers import pipeline
 
-            parameters = (
+            params_to_process = (
                 {"low_cpu_mem_usage": True}
-                | parameters
+                | params_to_process
                 | inspect.signature(pipeline).parameters
             )
 
-        for k in parameters:
+        for k in params_to_process:
             if k == "model":
                 continue  # special case `model` comes from `name` in the generator
             if k in params:
@@ -89,7 +89,7 @@ class HFCompatible:
                     continue
                 if (
                     k == "device"
-                    and "device_map" in parameters
+                    and "device_map" in params_to_process
                     and "device_map" in params
                 ):
                     # per transformers convention hold `device_map` before `device`
@@ -103,9 +103,13 @@ class HFCompatible:
         import torch.cuda
 
         selected_device = None
-        if self.hf_args["device"] is not None:
+        if self.hf_args.get("device", None) is not None:
             if isinstance(self.hf_args["device"], int):
                 # this assumes that indexed only devices selections means `cuda`
+                if self.hf_args["device"] < 0:
+                    msg = f"device {self.hf_args['device']} requested but CUDA device numbering starts at zero. Use 'device: cpu' to request CPU."
+                    logging.critical(msg)
+                    raise ValueError(msg)
                 selected_device = torch.device("cuda:" + str(self.hf_args["device"]))
             else:
                 selected_device = torch.device(self.hf_args["device"])
@@ -271,8 +275,6 @@ class ConversationalPipeline(Pipeline, HFCompatible):
 
         if _config.run.seed is not None:
             set_seed(_config.run.seed)
-
-        import torch.cuda
 
         # Note that with pipeline, in order to access the tokenizer, model, or device, you must get the attribute
         # directly from self.generator instead of from the ConversationalPipeline object itself.
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -16,8 +16,9 @@ be quite strong. Find your Hugging Face Inference API Key here:
 
 import inspect
 import logging
+import os
 import re
-from typing import List, Union
+from typing import Callable, List, Union
 import warnings
 
 import backoff
@@ -33,15 +34,15 @@ from garak.generators.base import Generator
 models_to_deprefix = ["gpt2"]
 
 
-class HFRateLimitException(Exception):
+class HFRateLimitException(GarakException):
     pass
 
 
-class HFLoadingException(Exception):
+class HFLoadingException(GarakException):
     pass
 
 
-class HFInternalServerError(Exception):
+class HFInternalServerError(GarakException):
     pass
 
 
@@ -51,30 +52,106 @@ class HFCompatible:
             if isinstance(config.n_ctx, int):
                 self.context_len = config.n_ctx
 
+    def _gather_hf_params(self, hf_constructor: Callable):
+        # this may be a bit too naive as it will pass any parameter valid for the pipeline signature
+        # this falls over when passed `from_pretrained` methods as the callable model params are not explicit
+        params = self.hf_args
+        if params["device"] is None:
+            params["device"] = self.device
+
+        args = {}
+
+        parameters = inspect.signature(hf_constructor).parameters
+
+        if "model" in parameters:
+            args["model"] = self.name
+            # expand for
+            parameters = {"do_sample": True} | parameters
+        else:
+            # callable is for a Pretrained class also map standard `pipeline` params
+            from transformers import pipeline
+
+            parameters = (
+                {"low_cpu_mem_usage": True}
+                | parameters
+                | inspect.signature(pipeline).parameters
+            )
+
+        for k in parameters:
+            if k == "model":
+                continue  # special case `model` comes from `name` in the generator
+            if k in params:
+                val = params[k]
+                if k == "torch_dtype" and hasattr(torch, val):
+                    args[k] = getattr(
+                        torch, val
+                    )  # some model type specific classes do not yet support direct string representation
+                    continue
+                if (
+                    k == "device"
+                    and "device_map" in parameters
+                    and "device_map" in params
+                ):
+                    # per transformers convention hold `device_map` before `device`
+                    continue
+                args[k] = params[k]
+
+        return args
+
+    def _select_hf_device(self):
+        """Determine the most efficient device for tensor load, hold any existing `device` already selected"""
+        import torch.cuda
+
+        selected_device = None
+        if self.hf_args["device"] is not None:
+            if isinstance(self.hf_args["device"], int):
+                # this assumes that indexed only devices selections means `cuda`
+                selected_device = torch.device("cuda:" + str(self.hf_args["device"]))
+            else:
+                selected_device = torch.device(self.hf_args["device"])
+
+        if selected_device is None:
+            selected_device = torch.device(
+                "cuda"
+                if torch.cuda.is_available()
+                else "mps" if torch.backends.mps.is_available() else "cpu"
+            )
+
+        if isinstance(selected_device, torch.device) and selected_device.type == "mps":
+            os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
+            logging.debug("Enabled MPS fallback environment variable")
+
+        logging.debug(
+            "Using %s, based on torch environment evaluation", selected_device
+        )
+        return selected_device
+
 
 class Pipeline(Generator, HFCompatible):
     """Get text generations from a locally-run Hugging Face pipeline"""
 
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "generations": 10,
+        "hf_args": {
+            "torch_dtype": "float16",
+            "do_sample": True,
+            "device": None,
+        },
+    }
     generator_family_name = "Hugging Face 🤗 pipeline"
     supports_multiple_generations = True
     parallel_capable = False
 
-    def __init__(
-        self, name="", do_sample=True, generations=10, device=0, config_root=_config
-    ):
+    def __init__(self, name="", config_root=_config):
         self.name = name
-        self.generations = generations
-        self.do_sample = do_sample
-        self.device = device
 
-        super().__init__(
-            self.name, generations=self.generations, config_root=config_root
-        )
+        super().__init__(self.name, config_root=config_root)
 
         import torch.multiprocessing as mp
 
         mp.set_start_method("spawn", force=True)
 
+        self.device = self._select_hf_device()
         self._load_client()
 
     def _load_client(self):
@@ -86,18 +163,7 @@ class Pipeline(Generator, HFCompatible):
         if _config.run.seed is not None:
             set_seed(_config.run.seed)
 
-        import torch.cuda
-
-        # consider how this could be abstracted well
-        self.device = (
-            "cuda:" + str(self.device)
-            if torch.cuda.is_available()
-            else "mps" if torch.backends.mps.is_available() else "cpu"
-        )
-
-        logging.debug("Using %s, based on torch environment evaluation", self.device)
-
-        pipline_kwargs = self._gather_pipeline_params(pipeline=pipeline)
+        pipline_kwargs = self._gather_hf_params(hf_constructor=pipeline)
         self.generator = pipeline("text-generation", **pipline_kwargs)
         if not hasattr(self, "deprefix_prompt"):
             self.deprefix_prompt = self.name in models_to_deprefix
@@ -109,17 +175,6 @@ class Pipeline(Generator, HFCompatible):
 
     def _clear_client(self):
         self.generator = None
-
-    def _gather_pipeline_params(self, pipeline):
-        # this may be a bit too naive as it will pass any parameter valid for the pipeline signature
-        args = {}
-        for k in inspect.signature(pipeline).parameters:
-            if k == "model":
-                # special case of known mapping as `model` may be reserved for the class
-                args[k] = self.name
-            if hasattr(self, k):
-                args[k] = getattr(self, k)
-        return args
 
     def _call_model(
         self, prompt: str, generations_this_call: int = 1
@@ -191,7 +246,7 @@ class OptimumPipeline(Pipeline, HFCompatible):
             if "use_fp8" in _config.plugins.generators.OptimumPipeline:
                 self.use_fp8 = True
 
-        pipline_kwargs = self._gather_pipeline_params(pipeline=pipeline)
+        pipline_kwargs = self._gather_hf_params(hf_constructor=pipeline)
         self.generator = pipeline("text-generation", **pipline_kwargs)
         if not hasattr(self, "deprefix_prompt"):
             self.deprefix_prompt = self.name in models_to_deprefix
@@ -219,18 +274,9 @@ class ConversationalPipeline(Pipeline, HFCompatible):
 
         import torch.cuda
 
-        # consider how this could be abstracted well
-        self.device = (
-            "cuda:" + str(self.device)
-            if torch.cuda.is_available()
-            else "mps" if torch.backends.mps.is_available() else "cpu"
-        )
-
-        logging.debug("Using %s, based on torch environment evaluation", self.device)
-
         # Note that with pipeline, in order to access the tokenizer, model, or device, you must get the attribute
         # directly from self.generator instead of from the ConversationalPipeline object itself.
-        pipline_kwargs = self._gather_pipeline_params(pipeline=pipeline)
+        pipline_kwargs = self._gather_hf_params(hf_constructor=pipeline)
         self.generator = pipeline("conversational", **pipline_kwargs)
         self.conversation = Conversation()
         if not hasattr(self, "deprefix_prompt"):
@@ -278,7 +324,7 @@ class ConversationalPipeline(Pipeline, HFCompatible):
             return [re.sub("^" + re.escape(prompt), "", _o) for _o in outputs]
 
 
-class InferenceAPI(Generator, HFCompatible):
+class InferenceAPI(Generator):
     """Get text generations from Hugging Face Inference API"""
 
     generator_family_name = "Hugging Face 🤗 Inference API"
@@ -407,7 +453,7 @@ class InferenceAPI(Generator, HFCompatible):
         self.wait_for_model = False
 
 
-class InferenceEndpoint(InferenceAPI, HFCompatible):
+class InferenceEndpoint(InferenceAPI):
     """Interface for Hugging Face private endpoints
     Pass the model URL as the name, e.g. https://xxx.aws.endpoints.huggingface.cloud
     """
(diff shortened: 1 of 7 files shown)


garak/generators/huggingface.py
```diff
@@ -50,53 +52,134 @@ def _set_hf_context_len(self, config):
             if isinstance(config.n_ctx, int):
                 self.context_len = config.n_ctx
 
+    def _gather_hf_params(self, hf_constructor: Callable):
+        # this may be a bit too naive as it will pass any parameter valid for the pipeline signature
+        # this falls over when passed `from_pretrained` methods as the callable model params are not explicit
```
erickgalinkin, line 57: Is it worth catching this and gathering those params differently?

garak/generators/huggingface.py
```diff
@@ -50,53 +52,134 @@ def _set_hf_context_len(self, config):
...
+        if "model" in parameters:
+            args["model"] = self.name
+            # expand for
+            parameters = {"do_sample": True} | parameters
+        else:
+            # callable is for a Pretrained class also map standard `pipeline` params
+            from transformers import pipeline
+
+            parameters = (
+                {"low_cpu_mem_usage": True}
```
erickgalinkin, line 75: Is there a reason we force-set this to true?

garak/generators/huggingface.py
```diff
@@ -50,53 +52,134 @@ def _set_hf_context_len(self, config):
...
+
+        return args
+
+    def _select_hf_device(self):
+        """Determine the most efficient device for tensor load, hold any existing `device` already selected"""
+        import torch.cuda
+
+        selected_device = None
+        if self.hf_args["device"] is not None:
+            if isinstance(self.hf_args["device"], int):
```
erickgalinkin, line 107: Should check >= 0. Sometimes, folks use `-1` as a stand-in for CPU (which is not a good practice, but still exists out there.)

garak/generators/huggingface.py
```diff
@@ -505,18 +554,23 @@ def __init__(
...
             )
 
-        self.do_sample = do_sample
         self.generation_config = transformers.GenerationConfig.from_pretrained(
             self.name
         )
         self.generation_config.eos_token_id = self.model.config.eos_token_id
         self.generation_config.pad_token_id = self.model.config.eos_token_id
 
+    def _clear_client(self):
```
erickgalinkin, line 563: Do we also want to clear `self.tokenizer` here?

erickgalinkin (commented):
Looks good overall -- left a few comments throughout. Great work!

garak/generators/huggingface.py
jmartin-tech: `-1` will now raise an exception is offered in a config file:
```
$ cat garak/garak.site.yaml
plugins:
  generators:
    huggingface:
      hf_args:
        torch_dtype: float16
      Model:
        hf_args:
            device: -1
```
Resulting in:
```
$ ./garak_debug.py -m huggingface.Model -n meta-llama/Llama-2-7b-chat-hf -g 1 -p continuation
garak LLM vulnerability scanner v0.9.0.13.post2 ( https://github.com/leondz/garak ) at 2024-06-18T10:06:06.476749
🦜 loading generator: Hugging Face 🤗 model: meta-llama/Llama-2-7b-chat-hf
Invalid device string: 'cuda:-1'
```

garak/generators/huggingface.py
jmartin-tech: This is not setting the value just ensuring the `low_cpu_mem_usage ` key is in the set or params to process if found in `hf_args`.

garak/generators/huggingface.py
jmartin-tech: Good catch! Thank you

garak/generators/huggingface.py
jmartin-tech: This is mitigated by adding `Pipeline` params when `model` is not a parameter of the callable, future iterations may suggest adding a supported set of params so I left the notes here to offer details on `why` the code adds more parameters in some cases.

garak/generators/huggingface.py
leondz: Let's throw out a clear error if they send `-1`. E.g.
```suggestion
            if isinstance(self.hf_args["device"], int):
              if self.hf_args["device"] < 0:
                msg = f"device {self.hf_args["device"]} requested but CUDA device numbering starts at zero. Use 'device: cpu' to request CPU."
                logging.critical(msg)
                raise ValueError(msg)
```

garak/generators/huggingface.py
```diff
@@ -50,53 +52,134 @@ def _set_hf_context_len(self, config):
...
+        return args
+
+    def _select_hf_device(self):
+        """Determine the most efficient device for tensor load, hold any existing `device` already selected"""
+        import torch.cuda
+
+        selected_device = None
+        if self.hf_args["device"] is not None:
+            if isinstance(self.hf_args["device"], int):
+                # this assumes that indexed only devices selections means `cuda`
```
leondz, line 108: tagging #671, #672

garak/generators/huggingface.py
```diff
@@ -50,53 +52,134 @@ def _set_hf_context_len(self, config):
...
+        if selected_device is None:
+            selected_device = torch.device(
+                "cuda"
+                if torch.cuda.is_available()
+                else "mps" if torch.backends.mps.is_available() else "cpu"
+            )
+
+        if isinstance(selected_device, torch.device) and selected_device.type == "mps":
+            os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
+            logging.debug("Enabled MPS fallback environment variable")
```
leondz, line 122: tagging #672

garak/generators/huggingface.py
```diff
@@ -50,53 +52,134 @@ def _set_hf_context_len(self, config):
...
+            "Using %s, based on torch environment evaluation", selected_device
+        )
+        return selected_device
+
 
 class Pipeline(Generator, HFCompatible):
     """Get text generations from a locally-run Hugging Face pipeline"""
 
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "generations": 10,
```
leondz, line 134: why is this `10` hardcoded instead of being taken from `_config.run` ?

garak/generators/huggingface.py
```diff
@@ -50,53 +52,134 @@ def _set_hf_context_len(self, config):
...
+        "hf_args": {
+            "torch_dtype": "float16",
+            "do_sample": True,
+            "device": None,
+        },
```
leondz, line 139: It might be more intuitive to put this into HFCompatible, even though not all classes will use all parameters. I would like to have it possible to set e.g. `torch_dtype` in one place for all `huggingface` generators, so that when one switches from `Pipeline` to `Model` due to compatibility issues (which happens not-never), these options can be carried over.

garak/generators/huggingface.py
```diff
@@ -50,53 +52,134 @@ def _set_hf_context_len(self, config):
...
-            logging.debug("Using CPU, torch.cuda.is_available() returned False")
-            self.device = -1
-
-        self.generator = pipeline(
-            "text-generation",
-            model=self.name,
-            do_sample=self.do_sample,
-            device=self.device,
-        )
-        self.deprefix_prompt = self.name in models_to_deprefix
```
leondz, line 90: great

garak/generators/huggingface.py
```diff
@@ -159,45 +239,33 @@ def __init__(
         if not torch.cuda.is_available():
             message = "OptimumPipeline needs CUDA, but torch.cuda.is_available() returned False; quitting"
             logging.critical(message)
-            raise ValueError(message)
+            raise GarakException(message)
```
leondz, line 242: I had this reflex also. But does it make sense for us to only ever raise `GarakException`s? `ValueError` has well-defined semantics which we don't necessarily need to duplicate, and raising GarakException instead of default exceptions losing some informativeness.

garak/generators/huggingface.py
```diff
@@ -159,45 +239,33 @@ def __init__(
...
-
-        self.generator = pipeline(
-            "text-generation",
-            model=self.name,
-            do_sample=self.do_sample,
-            device=self.device,
-            use_fp8=use_fp8,
-        )
-        self.deprefix_prompt = name in models_to_deprefix
+                self.use_fp8 = True
```
leondz, line 247: is this distinct from dtype? can we unify these params so there isn't some subtle loss of config/api inconsistency when moving from e.g. `Pipeline` to `OptimumPipeline`?

garak/generators/huggingface.py
```diff
@@ -206,20 +274,13 @@ def __init__(
 
         import torch.cuda
```
leondz, line 276: Can we get away with this too?

```suggestion
```

garak/generators/huggingface.py
```diff
@@ -236,12 +297,13 @@ def _call_model(
...
         """Take a conversation as a list of dictionaries and feed it to the model"""
 
+        self._load_client()
         # If conversation is provided as a list of dicts, create the conversation.
         # Otherwise, maintain state in Generator
         if isinstance(prompt, str):
             self.conversation.add_message({"role": "user", "content": prompt})
             self.conversation = self.generator(self.conversation)
-            generations = [self.conversation[-1]["content"]]
+            generations = [self.conversation[-1]["content"]]  # what is this doing?
```
leondz, line 306: This class may become deprecated once #644 lands

garak/generators/huggingface.py
```diff
@@ -505,18 +554,24 @@ def __init__(
...
             self.name
         )
         self.generation_config.eos_token_id = self.model.config.eos_token_id
         self.generation_config.pad_token_id = self.model.config.eos_token_id
 
+    def _clear_client(self):
+        self.model = None
+        self.config = None
+        self.tokenizer = None
+        self.generation_config = None
```
leondz, line 567: Why do we clear this? Does leaving it around cause conflicts with `_load_client()`'s application of config?

garak/generators/huggingface.py
```diff
@@ -553,21 +608,23 @@ def _call_model(
...
         # "exist_tokens + max_new_tokens < 4K is the golden rule."
         # https://github.com/haotian-liu/LLaVA/issues/1095#:~:text=Conceptually%2C%20as%20long%20as%20the%20total%20tokens%20are%20within%204K%2C%20it%20would%20be%20fine%2C%20so%20exist_tokens%20%2B%20max_new_tokens%20%3C%204K%20is%20the%20golden%20rule.
-        "max_tokens": 4000,
-        # consider shifting below to kwargs or llava_kwargs that is a dict to allow more customization
-        "torch_dtype": torch.float16,
-        "low_cpu_mem_usage": True,
-        "device_map": "cuda:0",
+        "hf_args": {
+            "torch_dtype": "float16",
+            "low_cpu_mem_usage": True,
```
leondz, line 620: is this already covered by `HFCompatible`? prefer to leave this default in one place

garak/generators/huggingface.py
leondz, line 1: enjoying the vast amounts of duplicated code in important places that's cut with these changes

this will make part of #717 redundant (good)

tests/generators/test_huggingface.py
```diff
@@ -1,11 +1,25 @@
...
+                "hf_args": {
+                    "device": "cpu",
+                },
+            }
+        }
+    }
+    config_root = GarakSubConfig()
+    setattr(config_root, "generators", gen_config)
+
+    g = garak.generators.huggingface.Pipeline("gpt2", config_root=config_root)
```
leondz, line 22: side note: this is a working but i guess relatively inelegant pattern for getting a generator up. 

1. What happens if the config setting is skipped? 
2. Is the name parameter in `Pipeline`'s constructor redundant? How does it relate to the `name` in `gen_config` - which takes precedence?
3. why device `cpu` and not `auto`?

tests/test_config.py
```diff
@@ -19,31 +19,36 @@
...
   detector:
       test:
         val: tests
         Blank:
           generators:
             huggingface:
-                gpu: 1
+                hf_args:
+                    torch_dtype: float16
+                    device: cuda:1
```
leondz, line 42: how come `gpu: 1` changed to `cuda:1` here...

tests/test_config.py
```diff
@@ -19,31 +19,36 @@
...
                 Pipeline:
                   dtype: for_detector
   buffs:
       test:
         Blank:
           generators:
             huggingface:
-                gpu: 1
+                hf_args:
+                    device: cuda:0
```
leondz, line 51: ...and `gpu: 1` changed to `cuda:0` here?

tests/test_configurable.py
```diff
@@ -63,6 +69,17 @@ def test_param_provided(generator_sub_config):
...
+
+
+# when a default parameter dictionary is provided merge on the resulting object
+def test_class_dict_merge_to_instance(generator_sub_config):
+    config_dict_var = {"dict_a": "test_val", "dict_c": "test_val"}
+    generator_sub_config.generators["mock"]["class_dict_var"] = config_dict_var
+    m = mockConfigurable(config_root=generator_sub_config)
+    assert m.class_dict_var == m.DEFAULT_PARAMS["class_dict_var"] | config_dict_var
+    assert m.class_dict_var["dict_a"] == config_dict_var["dict_a"]
+    assert m.class_dict_var["dict_c"] == config_dict_var["dict_c"]
```
leondz, line 82: excellent. now we have a policy!

leondz (commented):
thank you. looking good. bunch of questions & clarification requests, mostly about making me smarter

garak/generators/huggingface.py
jmartin-tech: The current implementation already supports for this.

```
plugins:
  generators:
    huggingface:
      hf_args:
        torch_dtype: float16
      Model:
        hf_args:
            device: 0
```

The above will set `hf_args["torch_dtype"]` for all `huggingface` generators, and merge with any other `DEFAULT_PARAMS["hf_args"]` values, I am hesitant to place these in `HFCompatible` as I see added complexity in attempting to have a common set of `hf_args` for `transformers`.

garak/generators/huggingface.py
jmartin-tech: I can do some research on that, testing this generator directly may provide me some insight on that.

garak/generators/huggingface.py
jmartin-tech: I am inclined to use `GarakException` here because `ValueError` does not seem accurate:

>exception ValueError
Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

This function was not passed a value with a type mismatch it inspects the environment and finds it lacking support.

garak/generators/huggingface.py
jmartin-tech: This value is a `transformers.GenerationConfig` object and is always replace in a call to `_load_client()`.  By clearing it we ensure it can be garbage collected and avoid a dependency on `transformer` objects being able to support `pickle` if we follow the same `custom` pickle pattern from openai generators in the future.

garak/generators/huggingface.py
jmartin-tech: `HFCompatible` does not provide defaults and the `_gather_hf_params()` reference to `low_cpu_mem_usage` simple addition to a list to ensure this options is passed if the `Callable` signatures do not provide for it explicitly.

tests/generators/test_huggingface.py
jmartin-tech: Good questions:
1. If config is skipped the new auto detection will load the most efficient device available. On macOS this turns out to be `mps` and actually causes this class to fail construction due to incomplete support for MPS in torch at this time.
2. `name` duplication is redundant, currently `name` from the constructor is help over `name` in `config_root` however either could be removed for this test.
3. See answer to `1`, `cpu` is a guaranteed fully supported value.

tests/test_config.py
jmartin-tech: These are simply example format being documented in the test file. They are not actually `consumed` values in the tests.

garak/generators/huggingface.py
leondz: Maybe the underlying question is, how do we remove `generations` counts from generator configs. But I think the pattern for that might be best established in another PR. Hardcoding defaults in multiple generators is not a preferable pattern, but let's postpone that discussion until #656

garak/generators/huggingface.py
leondz: Wait, aren't we already in `HFCompatible` here?

I don't want an extra config node - I just am not convinced a different default torch_dtype setting (e.g.) should be made in each generator class. But actually if we only do it in `HFCompatible` then we should be good.

garak/generators/huggingface.py
leondz: Oh, I saw the "inappropriate value" part being the `False` here. OK. Do you think `GarakException` is best-effort descriptive? If so then we should roll with that

garak/generators/huggingface.py
leondz: See earlier comment about multiple places where we defaults for `torch_dtype`

tests/generators/test_huggingface.py
leondz: OK, that's good to know.

1. I think we probably don't want to ship logic that breaks generators by default on macOS - did I read this right?
2. Sounds like there is some tidying up to do. I would love for the default config to run reasonably well out of the box, especially with `Pipeline`+`gpt2`, which is used in so many examples because it has a high chance of fitting in memory and going quickly on any machine. Requiring a `setattr()` gives an impression of "hotwiring". I'll think on this more - I haven't had time to play with the #711 config updates in depth much.
3. OK, that's cool :)

tests/test_config.py
leondz: Can we increase consistency to make the example more straightforward?

jmartin-tech: I would suggest the goal of the example is to show what you might configure differently depending on context.

The example here would define a model under test that is a `huggingface.Pipeline` would used any `cuda` devices while a `huggingface` generator created by for use by the `test.Blank` detector would specify to use `cuda:1` for it's device, and a `huggingface` generator created by for use by the `test.Blank` buff would use device `cuda:0`.  This would mean the model under tests is free to use any available device while the detector and buff would use specific different GPU devices from each other.

I think a more `concrete` or `viable` example might be on the horizo. This level of detail in config is likely rare for users to need and @erickgalinkin may be the only actual consumer for this level for the near term. 😄

tests/generators/test_huggingface.py
jmartin-tech: 1. `mps` does not break on macOS by default as `gpt2` does work,  I have tested it will load and complete a `fast` config, however this test needs to be consistent for all runs so forcing `cpu` was a reasonable solution. This test in particular results in the error below when asking for generation on `""` on `mps`, the error seemed like a clear message to the user since overriding the device can be a straight forward config setting now.
```
ERROR    root:huggingface.py:203 Can't infer missing attention mask on `mps` device. Please provide an `attention_mask` or use a different device.
```
2. The `setattr()` here could be replaced with a simple dictionary as both are supported by config I just followed more closely what `cli` will generate for as that is the more specific object type.
``` Python
config_root = {
        "generators": {    
            "huggingface": {
                "Pipeline": {
                    "name": "gpt2",
                    "hf_args": {
                        "device": "cpu",
                    },
                }
            }
        }
    }
```

garak/generators/huggingface.py
jmartin-tech: We are in `Pipeline` which happens to also be composed with `HFCompatible`.

We could have `HFCompatible` implement a `DEFAULT_PARAMS` and merge it with the `Pipeline` and `Generator` defaults with something like 
``` Python
    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | HFCompatible.DEFAULT_PARAMS | {
        "generations": 10,
    }
```

I suspect this would quickly become error prone and get missed when creating new classes that need to be composed as with `HFCompatible`.

We could also consider changing the inheritance pattern to be `HFCompatible(Generator)` -> `Pipeline(HFCompatible)` and use 
``` Python
class HFCompatible(Generator):
        "hf_args": {
            "torch_dtype": "float16",
            "do_sample": True,
            "device": None,
        }
    }

...

class Pipeline(HFCompatible):
    DEFAULT_PARAMS = HFCompatible.DEFAULT_PARAMS | {
        "generations": 10,
    }
```

Either pattern I have a concern about defaults that only make sense for a specific transformer slipping into the common set. As well as possible need to recursively merge `hf_args` if a new class needs a unique default that it cannot put in `HFCompatible.DEFAULT_PARAMS`.

I am a major proponent of code reuse, just not sure the trade off is worth it here. Happy to accommodate if this is a deal breaker.

garak/generators/huggingface.py
jmartin-tech: `GarakException` will due, I think a more clear exception type may make sense if we see available resources become an common error condition.

tests/test_config.py
jmartin-tech: `gpu` was a theoretical value `device` aligns with actual huggingface expectations.

garak/generators/huggingface.py
erickgalinkin: Looks like `use_fp8` just maps the dtype via the quantizer: https://github.com/huggingface/transformers/blob/2d4820284d823ad362e63662397aae957a95901b/src/transformers/quantizers/quantizer_quanto.py#L157

garak/generators/huggingface.py
erickgalinkin: Yeah, this is a bit of a kluge to get conversations here, but this will pull the last message in the conversation from the generator. 
`self.conversation[-1]` will get the last element in the conversation ("role": "assistant" or "role": "bot" depending on templating) and pull the "content" element out of it. `self.conversation` is a `Conversation` object, an abstract class built on top of `dict`.

garak/generators/huggingface.py
erickgalinkin: Agree -- it'd be ideal to set a default for all HF classes (I think float16 is a reasonable default, but I could see arguments for "auto")

tests/test_config.py
erickgalinkin: This feels right -- I suspect the overwhelming majority of users will not need to configure things like this, but it's good to have. I'll likely need to refactor parts of `AutoDAN` and/or `TAP` as a result of this PR being merged and it's plausible that we'll need this level of granularity in configs, particularly for folks using resource-constrained systems.

garak/generators/huggingface.py
jmartin-tech: The comment is missing context, it was focused on the line setting a new local `generations` value that is never consumed, I `think` it can simply be removed however since there is no clear test for this I just added a comment.

garak/generators/huggingface.py
erickgalinkin: I believe this largely resolves #672.
#671 will require testing on an actual device, but specification of `device="cuda"` should be all that's necessary for ROCm with HF.

tests/generators/test_huggingface.py
leondz: We don't test all generators for empty string behavior by default because they each their own unique special snowflake style (making it a schlep) and often require an API key (making it an expensive pain). I tried to write this one day and discovered it was slightly less feasible than imagined - the new config setup might make it easier now. There are a few tests in place, but it's incomplete.

tests/test_config.py
leondz: got it. demonstrating variation is good. otoh I think the example could be confusing with `gpu: 1` referring to both `cuda: 0` and `cuda: 1` within sight of each other.

jmartin-tech: The final result of the edits is removal of all `gpu:` entries and only provides `huggingface` examples using the `device` key.

leondz (approved):
lgtm. thanks!

PR #749, opened 2024-06-19
remove extraneous default params for nims that expect conservative pa…

### LINKED ISSUE #748: nim openai endpoints reject requests with unexpected params in payload
for `-m nim -n nvidia/nemotron-4-340b-instruct`

```
2024-06-19 13:48:25,678  DEBUG  Request options: 
{
   "method":"post",
   "url":"/chat/completions",
   "files":"None",
   "json_data":{
      "messages":[
         {
            "role":"user",
            "content":"Will you xxxx your teacher?"
         }
      ],
      "model":"nvidia/nemotron-4-340b-reward",
      "frequency_penalty":0.0,
      "max_tokens":150,
      "n":1,
      "presence_penalty":0.0,
      "stop":[
         "#",
         ";"
      ],
      "temperature":0.7,
      "top_p":1.0
   }
}

```

`openai.UnprocessableEntityError: Error code: 422 - {'type': 'urn:inference-service:problem-details:unprocessable-entity', 'title': 'Unprocessable Entity', 'status': 422, 'detail': 'body -> frequency_penalty\n  Extra inputs are not permitted (type=extra_forbidden)\nbody -> n\n  Extra inputs are not permitted (type=extra_forbidden)\nbody -> presence_penalty\n  Extra inputs are not permitted (type=extra_forbidden)', 'instance': '/v2/nvcf/pexec/functions/b0fcd392-e905-4ab4-8eb9-aeae95c30b37', 'requestId': '380d3287-f799-4c60-a12f-d5448609595a'}
`

leondz:
resolves #748

* add a "blocked params" value to `OpenAICompatible` so we can control verbosity of payloads
* strip out `None`s from OpenAI API payloads
* force unset defaults for some params in `nim`

one question - why are these defaults defined twice?

```
    # template defaults optionally override when extending
        "temperature": 0.7,
        "top_p": 1.0,
        "frequency_penalty": 0.0,
        "presence_penalty": 0.0,
        "stop": ["#", ";"],
        "blocked_params": set(),
    }

    temperature = 0.7
    top_p = 1.0
    frequency_penalty = 0.0
    presence_penalty = 0.0
    stop = ["#", ";"]
    blocked_params = set()
```

### DIFF
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -42,6 +42,9 @@ class NVOpenAIChat(OpenAICompatible):
     active = True
     supports_multiple_generations = False
     generator_family_name = "NIM"
+    frequency_penalty = None
+    presence_penalty = None
+    blocked_params = {"n"}
 
     timeout = 60
 
@@ -66,6 +69,8 @@ class NVOpenAIChat(OpenAICompatible):
         assert (
             generations_this_call == 1
         ), "generations_per_call / n > 1 is not supported"
+        self.presence_penalty = None
+        self.frequency_penalty = None
         return super()._call_model(prompt, generations_this_call)
 
 
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -38,11 +38,11 @@ class NVOpenAIChat(OpenAICompatible):
         "top_p": 0.7,
         "top_k": 0,  # top_k is hard set to zero as of 24.04.30
         "uri": "https://integrate.api.nvidia.com/v1/",
+        "suppressed_params": {"n", "frequency_penalty", "presence_penalty"},
     }
     active = True
     supports_multiple_generations = False
     generator_family_name = "NIM"
-    suppressed_params = {"n", "frequency_penalty", "presence_penalty"}
 
     timeout = 60
 
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -42,9 +42,7 @@ class NVOpenAIChat(OpenAICompatible):
     active = True
     supports_multiple_generations = False
     generator_family_name = "NIM"
-    frequency_penalty = None
-    presence_penalty = None
-    suppressed_params = {"n"}
+    suppressed_params = {"n", "frequency_penalty", "presence_penalty"}
 
     timeout = 60
 
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -33,7 +33,7 @@ class NVOpenAIChat(OpenAICompatible):
     # per https://docs.nvidia.com/ai-enterprise/nim-llm/latest/openai-api.html
     # 2024.05.02, `n>1` is not supported
     ENV_VAR = "NIM_API_KEY"
-    DEFAULT_PARAMS = {
+    DEFAULT_PARAMS = OpenAICompatible.DEFAULT_PARAMS | {
         "temperature": 0.1,
         "top_p": 0.7,
         "top_k": 0,  # top_k is hard set to zero as of 24.04.30
@@ -44,7 +44,7 @@ class NVOpenAIChat(OpenAICompatible):
     generator_family_name = "NIM"
     frequency_penalty = None
     presence_penalty = None
-    blocked_params = {"n"}
+    suppressed_params = {"n"}
 
     timeout = 60
 
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -67,8 +67,6 @@ class NVOpenAIChat(OpenAICompatible):
         assert (
             generations_this_call == 1
         ), "generations_per_call / n > 1 is not supported"
-        self.presence_penalty = None
-        self.frequency_penalty = None
         return super()._call_model(prompt, generations_this_call)
 
 
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -99,8 +99,6 @@ class OpenAICompatible(Generator):
         "presence_penalty": 0.0,
         "stop": ["#", ";"],
         "suppressed_params": set(),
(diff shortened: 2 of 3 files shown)


leondz:
-- ideally would have liked to have done this via config, not sure it woulda been possible though

garak/generators/openai.py
```diff
@@ -172,17 +174,23 @@ def _call_model(
...
+                "n": generations_this_call,
+                "top_p": self.top_p,
+                "frequency_penalty": self.frequency_penalty,
+                "presence_penalty": self.presence_penalty,
+                "stop": self.stop,
+            }
+            create_args = {k: v for k, v in create_args.items() if v is not None}
+            create_args = {
+                k: v for k, v in create_args.items() if k not in self.blocked_params
+            }
```
jmartin-tech, line 191: It looks like both `create` methods only differ in the name of the parameter the `prompt` will will be passed as. If both `chat` and `completions` will never expect to pass `None` for a `create` argument, maybe create a shared method:
``` Python
    def _create_params(self)
            create_args = {
                "model": self.name,
                "temperature": self.temperature,
                "max_tokens": self.max_tokens,
                "n": generations_this_call,
                "top_p": self.top_p,
                "frequency_penalty": self.frequency_penalty,
                "presence_penalty": self.presence_penalty,
                "stop": self.stop,
            }
            create_args = {k: v for k, v in create_args.items() if v is not None}
            create_args = {
                k: v for k, v in create_args.items() if k not in self.blocked_params
            }
            return create_args
```

Then here:
```suggestion
            create_args = {"prompt": prompt} | self._create_params()
```

garak/generators/openai.py
```diff
@@ -199,17 +207,24 @@ def _call_model(
...
+                    "n": generations_this_call,
+                    "stop": self.stop,
+                    "max_tokens": self.max_tokens,
+                    "presence_penalty": self.presence_penalty,
+                    "frequency_penalty": self.frequency_penalty,
+                }
+                create_args = {k: v for k, v in create_args.items() if v is not None}
+                create_args = {
+                    k: v for k, v in create_args.items() if k not in self.blocked_params
+                }
```
jmartin-tech, line 224:
```suggestion
                create_args = {"messages": messages} | self._create_params()
```

garak/generators/nim.py
```diff
@@ -42,6 +42,9 @@ class NVOpenAIChat(OpenAICompatible):
     active = True
     supports_multiple_generations = False
     generator_family_name = "NIM"
+    frequency_penalty = None
+    presence_penalty = None
```
jmartin-tech, line 46: Based on this PR, I suspect these values really should be `instance` instead of `class` level, meaning these should be added to `DEFAULT_PARAMS` above instead of here.

jmartin-tech (commented):
NIce catch, I see a few more places we could improve code reuse.

garak/generators/openai.py
```diff
@@ -98,13 +98,15 @@ class OpenAICompatible(Generator):
...
         "stop": ["#", ";"],
+        "blocked_params": set(),
     }
 
     temperature = 0.7
     top_p = 1.0
     frequency_penalty = 0.0
     presence_penalty = 0.0
     stop = ["#", ";"]
+    blocked_params = set()
```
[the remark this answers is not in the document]
leondz: oh great, i'm glad that was just a missed dupe.

sounds like it should go in `DEFAULT_PARAMS`.

garak/generators/openai.py
leondz: excellent, love this

garak/generators/openai.py
```diff
@@ -98,13 +98,11 @@ class OpenAICompatible(Generator):
         "frequency_penalty": 0.0,
         "presence_penalty": 0.0,
         "stop": ["#", ";"],
+        "blocked_params": set(),
+        "generations": 10,
```
jmartin-tech, line 102:
```suggestion
```
Remove value not consumed and defer `generations ` as provided by `Generator.DEFAULT_PARAMS`

garak/generators/nim.py
```diff
@@ -38,6 +38,7 @@ class NVOpenAIChat(OpenAICompatible):
         "top_p": 0.7,
         "top_k": 0,  # top_k is hard set to zero as of 24.04.30
         "uri": "https://integrate.api.nvidia.com/v1/",
+        "suppressed_params": {"n", "frequency_penalty", "presence_penalty"},
```
erickgalinkin, line 41: Do we have a ref for these we can throw into a comment so we can keep it up to date?

garak/generators/openai.py
```diff
@@ -162,6 +157,23 @@ def _call_model(
...
+            "top_p": self.top_p,
+            "frequency_penalty": self.frequency_penalty,
+            "presence_penalty": self.presence_penalty,
+            "stop": self.stop,
+        }
+
+        create_args = {k: v for k, v in create_args.items() if v is not None}
+        create_args = {
+            k: v for k, v in create_args.items() if k not in self.suppressed_params
+        }
```
erickgalinkin, line 175:
```suggestion
        create_args = {k: v for k, v in create_args.items() if v is not None and k not in self.suppressed_params}
```

erickgalinkin (approved):
LGTM. Minor comment and suggestion.

garak/generators/nim.py
leondz: This was learned through trial by 422 / 400 error at the moment

jmartin-tech (approved):
LGTM, I would like to see the new NIM test here updated to use the `openai_compat_mocks` fixture. Happy to make that a follow on I will PR shortly.

PR #754, opened 2024-06-23
probe: assess 🤗 repos for potentially malicious files

leondz:


leondz:
would be cool if this 
*  checked for risky onnx files
* looked inside archives
* looked at content instead of just extensions
* did tests w/ `fickling`

still, these are all now easier with the scaffold provided here

erickgalinkin:
Nitpick: Should be potentially *malicious* files.

garak/detectors/fileformats.py
```diff
@@ -0,0 +1,86 @@
...
+                    # huggingface_hub.utils.disable_progress_bars()
+                    local_filename = huggingface_hub.hf_hub_download(
+                        hf_repo_name, repo_filename
+                    )
+                    # huggingface_hub.utils.enable_progress_bars()
+                    with open(local_filename, "rb") as f:
+                        pkl_data = pickletools.genops(f)
+
+                        # If the file is not a pickle, we may hit a UnicodeDecode error or ValueError when running over genops
+                        *_, last = pkl_data
```
jmartin-tech, line 65: Am I reading this right as the `detector` is downloading files?  This seems like the wrong placement of responsibility based on existing plugin types.

I currently expect a `probe` to gather information and pass it to a `detector`. On the other hand this also introduces the concept that gathered data could be very `large` and difficult to retain as evidence for results. In this case we might need to, at at a minimum, also generate *version* or *fingerprint* details for these larger pieces of data that can be referenced in a report.

leondz: This is the method for single-file hugging face cache access. If the model's used for generation, this is a noop.

It's def infeasible to pass file contents in attempts and I don't think a summary is an efficient way forward. We could do the local fs cache lookup on the probe side, though, iff this preserves repo filenames.

jmartin-tech: "This is cached access" makes it reasonable to say this instance is really accessing no not obtaining data, that works for me on clarifying the "responsibility for download" concern.

Based on that I would like to suggest some refactoring though. Any test would currently need to intercept `huggingface_hub.hf_hub_download` to provide the object detection actually acts upon, where as if the data was provided as an attributed of the `Attempt` being evaluated the test could just instantiate an Attempt locally without coupling to `huggingface` retrieval. I am not suggesting this PR should adjust `Attempt` just yet however we could extract the actual detection to be testable.

To that end it may be reasonable to shift the actual `detection` into a separate method maybe `IsPickled(filename: str)` or `IsPickled(filename: []byte)`that is called by `HF_IsPickled()`. I am thinking that a more isolated `detection` may be more consumable for other use cases, such as if at some point we supported `text[in] -> binary[out]` modality and wanted to have a detector that could test if a model can is returning binary data that is a `pickled` object.

leondz: This makes a tonne of sense. Moved responsibility for getting file data away from `detector` to `probe`; this keeps the detector flexible wrt. where the file came from, i.e. one file checker can work for any generator, and code that looks inside the file is decoupled from code that gets the file. Under this pattern, `detectors.fileformats.FileIsPickled.detect()` is pretty much a method that only looks at that item.

garak/detectors/fileformats.py
```diff
@@ -0,0 +1,72 @@
...
+                yield 0.0
+
+
+class FileIsPickled(Detector):
+    """
+    Check whether a file is pickled.
+    """
+
+    def detect(self, attempt: Attempt) -> Iterable[float]:
+        for local_filename in attempt.outputs:
```
jmartin-tech, line 40: Should this have a guard to to short-circuit if `attempt.notes["format"] != "local filename"`?

garak/detectors/fileformats.py
```diff
@@ -0,0 +1,72 @@
...
+            if local_filename is None or local_filename == "":
+                continue
+            else:
+                try:
+                    # huggingface_hub.utils.disable_progress_bars()
+                    # huggingface_hub.utils.enable_progress_bars()
+                    with open(local_filename, "rb") as f:
```
jmartin-tech, line 47: Test file exists on disk before `open` or have try block catch `FileNotFoundError` and other file access/permissions specific errors.

garak/harnesses/base.py
```diff
@@ -104,24 +104,19 @@ def run(self, model, probes, detectors, evaluator, announce_probe=True) -> None:
                 continue
 
             attempt_results = probe.probe(model)
-            assert isinstance(attempt_results, list)
+            assert isinstance(
+                attempt_results, list
```
jmartin-tech, line 108: Should this test for `Iterable` vs `list`?

tests/detectors/test_detectors_fileformats.py
jmartin-tech, line 1: Consider adding a `test_FileIsPickled()` that reads `tmp` files:
* file with mock test string data
* file with serialized written from a `pickle` call

jmartin-tech (approved):
Works for me, added some optional comments for some more defensive programming.

garak/detectors/fileformats.py
jmartin-tech: It looks like `OSError` would cover all filesystem exceptions.

garak/harnesses/base.py
leondz: Good catch. Should check for ordered iterables, i.e. not `set` which can break alignment (I think). Updated to use same test as in `tests.detectors.test_detectors.test_detector_detect`.

garak/detectors/fileformats.py
leondz: Added `OSError` to the bag, also pre-checked with `os.path.isfile` to block opening of devices

garak/detectors/fileformats.py
leondz: yup. added, and added tests

tests/detectors/test_detectors_fileformats.py
leondz: added, for all protocol versions + current default

tests/detectors/test_detectors_fileformats.py
```diff
@@ -0,0 +1,119 @@
...
+    assert l == []
+
+
+def test_fileispickled_nonpickle():
+    d = garak.detectors.fileformats.FileIsPickled()
+    nonpickle_file_attempt = Attempt(prompt="")
+    nonpickle_file_attempt.notes["format"] = "local filename"
+    with tempfile.NamedTemporaryFile(mode="w+", delete=False) as nonpickle_file:
+        nonpickle_file.write(
+            "Do you know what the sad part is, Odo? I'm a very good tailor."
```
jmartin-tech, line 60: 😸

garak/detectors/base.py
```diff
@@ -177,3 +178,32 @@ def detect(
...
+                f"detectors.fileformats.{self.__class__.__name__} only processes outputs that are '{self.valid_format}'"
+            )
+
+        for local_filename in attempt.outputs:
+            if local_filename is None or local_filename == "":
+                continue
+            if not os.path.isfile(
+                local_filename
+            ):  # skip missing files but also pipes, devices, etc
+                continue
```
jmartin-tech, line 205: Thought for future consideration, would there be value in logging a missing filename that the probe reported as finding but is no longer found at detection time?

tests/detectors/test_detectors.py
```diff
@@ -53,16 +54,21 @@ def test_detector_detect(classname):
 
     a = Attempt()
     a.prompt = ""
+    a.notes["format"] = "local filename"
     a.outputs = ["", None]
+    print(a.as_dict())
```
jmartin-tech, line 59:
```suggestion
```

Stray debug line?

tests/detectors/test_detectors_fileformats.py
```diff
@@ -0,0 +1,119 @@
...
+    with tempfile.NamedTemporaryFile(mode="wb+", delete=False) as pickle_file:
+        test_data_struct = {"key": ["value", 37176]}
+        pickle.Pickler(pickle_file, pickle_ver).dump(test_data_struct)
+        pickle_file.close()
+        pickle_file_attempt.outputs = [pickle_file.name]
+        l = list(d.detect(pickle_file_attempt))
+        assert l == [1.0]
+
+
+def test_fileisexectuable_nonexec():
```
jmartin-tech, line 98: It might be reasonable to document the expectation that if a file does not exist or is invalid there will be no yielded value.

The proposed test might belong in `test_detectors.py` since it is for functionality of `garak.detectors.base. FileDetector`

jmartin-tech (approved):
LGTM, just one stray debug entry that I would like to see removed.

tests/detectors/test_detectors_fileformats.py
leondz: Yep. See #756 to track formalising this expectation.

garak/detectors/base.py
leondz: Great idea

PR #763, opened 2024-06-28
add grandma substance synth probe, update docs and var names

leondz:


jmartin-tech (commented):
I see some opportunity to make this more flexible, accepting prompt values to test for via `config` could be error prone however since the yaml would only support basic class types this seems like a nice `power user` option.

PR #767, opened 2024-07-01
probe: assess 🤗 repos for potentially malicious files 

leondz:
recreating this one, merged too early, sorry


docs/source/probes.rst
```diff
@@ -16,6 +16,8 @@ For a detailed oversight into how a probe operates, see :ref:`garak.probes.base.
    garak.probes.dan
    garak.probes.donotanswer
    garak.probes.encoding
+   garak.probes.fileformats
+   garak.probes.gcg
```
erickgalinkin, line 20: I believe `garak.probes.gcg` is no longer present since #728 has been merged.

garak/detectors/fileformats.py
```diff
@@ -0,0 +1,90 @@
...
+These detectors examine file formats, based on name or content."""
+
+import pickletools
+
+import magic
+
+from garak.detectors.base import FileDetector
+
+
+class PossiblePickleName(FileDetector):
```
erickgalinkin, line 15: Grumble grumble filename-based detections etc etc

garak/detectors/fileformats.py
```diff
@@ -0,0 +1,90 @@
...
+    """Guess if filenames look like pickles
+
+    Pickle files have multiple and complex vulnerabilities, and should be
+    avoided. This detector expects a generator of filenames and fails those
+    that look like pickle files."""
+
+    valid_format = None
+
+    def _test_file(self, filename: str) -> None | str:
+        print("Ffffffff", filename)
```
erickgalinkin, line 25:
```suggestion
```

garak/probes/fileformats.py
```diff
@@ -0,0 +1,73 @@
...
+            return []
+        attempt = self._mint_attempt(None)
+
+        repo_filenames = huggingface_hub.list_repo_files(generator.name)
+        local_filenames = []
+        for repo_filename in tqdm.tqdm(
+            repo_filenames,
+            leave=False,
+            desc=f"Gathering files in {generator.name}",
+            colour="#e5a70e",  # probe theme colour
```
erickgalinkin, line 61: Should we consider storing the colors somewhere like `garak.resources.common` to reuse them more easily?

tests/detectors/test_detectors.py
```diff
@@ -58,26 +60,41 @@ def test_detector_detect(classname):
...
+            4,
+        ), "detector should return as many results as in all_outputs (maybe excluding Nones)"
+
+
+def test_filedetector_nonexist():
+    d = garak.detectors.base.FileDetector()
+    a = garak.attempt.Attempt(prompt="")
+    a.outputs = [None, "", "/non/existing/file"]
+    a.notes["format"] = d.valid_format
+    assert list(d.detect(a)) == []
```
erickgalinkin, line 100: Should we have a message for this assert?

erickgalinkin: Also, if we want this to be an empty list, we should probably check `len` since this might miss some weirdo outliers where `list(d.detect(a))` is some other value where the test returns true.

tests/resources/fileformats/exec_files/grep-mach-o-top4k
erickgalinkin, line 1: I feel kinda yucky about shipping binaries of any kind as part of the package. Not sure how else we'd test it, but I'm not happy about it.

erickgalinkin (changes requested):
Worth discussing this in more detail. Also, how does this compare/differ from #754

tests/resources/fileformats/exec_files/grep-mach-o-top4k
jmartin-tech: I theory `test` files should not be shipped as part of the package, they are here as a fixture for the tests and seem valid as things that come with a developer `git` based install.

garak/detectors/fileformats.py
leondz: Would you rather we skipped this entirely? I'm not married to it

tests/resources/fileformats/exec_files/grep-mach-o-top4k
leondz: I should say these are all broken, just the first 4k of the executable. Will add a filter for tests in pip packaging. How could they further be "defanged"? Magic docs say it wants at least 2K

leondz:
@erickgalinkin  no difference with #754 - i merged that by mistake, reverted, rebuilt the PR (which is this) and it got a new id

tests/resources/fileformats/exec_files/grep-mach-o-top4k
leondz: `test/` dir excluded by [MANIFEST.in](https://github.com/leondz/garak/blob/main/MANIFEST.in), & just verified w/ a pip install

leondz: truncated bins also now stored as base64, only decoded during running tests

garak/probes/fileformats.py
leondz: now in `garak.resources.theme` 😊

tests/detectors/test_detectors_fileformats.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def test_fileisexectuable_exec():
+    exec_attempt = garak.attempt.Attempt(prompt="")
+    exec_attempt.notes["format"] = "local filename"
+    path_prefix = Path("tests/resources/fileformats/exec_files/")
+
+    encoded_exec_filenames = glob.glob(str(path_prefix / "*.base64"))
+    exec_filenames = []
+    for encoded_exec_filename in encoded_exec_filenames:
```
jmartin-tech, line 118: The decoding process here may be better implemented as a `fixture` that registers the file to be deleted post test, as is a failure of the `assert` would leave all decoded files on disk in the temporary location.

Untested possible refactor:

``` Python
path_prefix = Path("tests/resources/fileformats/exec_files/")

@pytest.fixture
def decoded_file(request, encoded_exec_filename):
    with open(encoded_exec_filename, "r", encoding="utf-8") as encodedfile:
        with tempfile.NamedTemporaryFile(mode="wb", delete=False) as binfile:
            binfile.write(binascii.a2b_base64(encodedfile.read()))
            binfile.close()
            exec_filenames.append(binfile.name)

    def remove_decoded():
        with contextlib.suppress(FileNotFoundError):
            os.remove(binfile.name)

    return binfile.name
```
``` Python
@pytest.mark.parametrize("encoded_exec_filename", glob.glob(str(path_prefix / "*.base64")))
def test_fileisexectuable_exec(decoded_file):
    d = garak.detectors.fileformats.FileIsExecutable()
    exec_attempt.outputs = list(decoded_file)
    results = d.detect(exec_attempt)
    assert list(results) == [1.0]
```

leondz: yup, implemented

tests/detectors/test_detectors_fileformats.py
```diff
@@ -0,0 +1,141 @@
...
+    def remove_decoded():
+        with contextlib.suppress(FileNotFoundError):
+            Path(binfile.name).unlink()
+
+    request.addfinalizer(remove_decoded)
+
+
+@pytest.mark.parametrize(
+    "encoded_exec_filename", glob.glob(str(path_prefix / "*.base64"))
+)
```
jmartin-tech, line 133: May be a good idea to skip these test when `magic` is not available, not all development environments will need to interact with this probe, contributors will be more likely to run tests if they don't need to hit edge cases to get the suite to run at all.

```suggestion
try:
    import magic
except ImportError as e:
    magic == None


@pytest.mark.skipif(magic is None, reason="libmagic not available")
@pytest.mark.parametrize(
    "encoded_exec_filename", glob.glob(str(path_prefix / "*.base64"))
)
```

garak/detectors/fileformats.py
```diff
@@ -0,0 +1,101 @@
...
+import pickletools
+
+try:
+    import magic
+except ImportError as e:
+    logging.info(
+        "detectors.fileformats: couldn't import python-magic, try installing libmagic, e.g. `brew install libmagic`",
+        e,
+    )
+    magic == None
```
jmartin-tech, line 18:
```suggestion
    magic = None
```

``` Python
>>> magic == None
NameError: name 'magic' is not defined
```

jmartin-tech (changes requested):
One minor syntax issue, and some recommendations.

tests/detectors/test_detectors_fileformats.py
leondz: yes, this makes sense, good catch

garak/detectors/fileformats.py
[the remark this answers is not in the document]
leondz: yup, will do, good point

pyproject.toml
```diff
@@ -67,6 +67,9 @@ dependencies = [
   "fschat>=0.2.36",
   "litellm>=1.33.8",
   "jsonpath-ng>=1.6.1",
+  "huggingface_hub>=0.21.0",
+  'python-magic-bin; sys_platform == "win32"',
+  'python-magic>=0.4.21; sys_platform != "win32"',
```
jmartin-tech, line 72: Sorry late thought, should we set a version restriction here?

jmartin-tech:
```suggestion
  'python-magic-bin>=0.4.14; sys_platform == "win32"',
  'python-magic>=0.4.21; sys_platform != "win32"',
```

requirements.txt
```diff
@@ -28,6 +28,9 @@ deepl==1.17.0
 fschat>=0.2.36
 litellm>=1.33.8
 jsonpath-ng>=1.6.1
+huggingface_hub>=0.21.0
+python-magic-bin; sys_platform == "win32"
```
jmartin-tech, line 32:
```suggestion
python-magic-bin>=0.4.14; sys_platform == "win32"
```

PR #768, opened 2024-07-01
cached plugin enum

### LINKED ISSUE #624: generator list is slow
`python3 -m garak --list_generators` should return almost instantly - instead it takes a few seconds.

LiteLLM complaint is dumped out top, maybe this module is being instantiated?

```$ python3 -m garak --list_generators
garak LLM security probe v0.9.0.12.post1 ( https://github.com/leondz/garak ) at 2024-04-24T14:36:38.356597
14:36:40 - LiteLLM:DEBUG: utils.py:100 - Exception import enterprise features No module named 'litellm.proxy.enterprise'
generators: cohere 🌟
generators: cohere.CohereGenerator
generators: function 🌟
generators: function.Single
generators: ggml 🌟
generators: ggml.GgmlGenerator
generators: guardrails 🌟
generators: guardrails.NeMoGuardrails
generators: langchain 🌟
generators: langchain.LangChainLLMGenerator
generators: litellm 🌟
generators: litellm.LiteLLMGenerator
generators: nemo 🌟
generators: nemo.NeMoGenerator
generators: nvcf 🌟
generators: nvcf.NvcfGenerator
generators: octo 🌟
generators: octo.OctoGenerator
generators: openai 🌟
generators: openai.OpenAIGenerator
generators: rasa 🌟
generators: rasa.RestGenerator
generators: replicate 🌟
generators: replicate.ReplicateGenerator
generators: rest 🌟
generators: rest.RestGenerator
generators: test 🌟
generators: test.Blank
generators: test.Repeat
```

Would be nice to, in the longer term, have a consistent pattern for differentiating plugin list from load

### LINKED ISSUE #663: run start is slow w/ no messages
There's a pause after the generator is shown and after the probe list is displayed. Should give feedback to the user.

Pause after generator message: `python3 -m garak -m test.Blank -p lmrc`

Pause after probe list: `python3 -m garak -m test.Blank -p lmrc.Bullying`

jmartin-tech:
Fix #624 

Impacts of this change may be sufficient to close #663

First pass at building a local cache of plugin info to provide faster initial load of `cli` based runs.

This implementation will be expanded in the future to enable a `user` cache plugin details once #479 is addressed.

The update here improves plugin enumeration to load a `json` based cache file from the project for any plugin already in `main`.  Any plugin not in main at time of checkin is loaded if not found in the existing cache file.

```
(garak) ~/Projects/nvidia/garak$ time python -m garak --list_probes
garak LLM vulnerability scanner v0.9.0.13.post2 ( https://github.com/leondz/garak ) at 2024-07-01T18:24:52.653050
probes: atkgen 🌟
...
probes: xss.MarkdownImageExfil

real    0m1.919s
user    0m2.085s
sys     0m3.154s
```

Becomes:
```
(garak) ~/Projects/nvidia/garak$ time python -m garak --list_probes
garak LLM vulnerability scanner v0.9.0.13.post2 ( https://github.com/leondz/garak ) at 2024-07-01T18:27:38.513147
probes: atkgen 🌟
...
probes: xss.MarkdownImageExfil

real    0m0.064s
user    0m0.050s
sys     0m0.013s
```

Tests have been added to document cache build process and validate that items not already in the cache will be loaded ephemerally at runtime if requested.

A future CI/CD action will maintain this cache file using a pattern like the following:
``` bash
# remove the existing cache file
rm -rf garak/resources/plugin_cache.json
# Linux only command to ensure files are all timestamped to match last commit
git ls-files garak/ -z | xargs -0 -n1 -I{} -- git log -1 --format="%ai {}" {} | while read -r udate utime utz ufile ; do
  touch -d "$udate $utime" $ufile
done
# any `--list_*` command will build the cache when one is not found on disk
python -m garak --list_probes
# now just add the updated file in place of the existing version
git diff --exit-code garak/resources/plugin_cache.json
if [ ! $? ]; then
  echo "Plugin cache updates exist."
  git add garak/resources/plugin_cache.json
  git commit -S -m "update plugin_cache.json"
fi
```

### DIFF
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -5,15 +5,35 @@
 
 import importlib
 import inspect
+import json
 import logging
+import shutil
 import os
-from typing import List
+from typing import List, Callable, Union
+from pathlib import Path
 
 from garak import _config
 from garak.exception import GarakException
 
 PLUGIN_TYPES = ("probes", "detectors", "generators", "harnesses", "buffs")
 PLUGIN_CLASSES = ("Probe", "Detector", "Generator", "Harness", "Buff")
+
+_user_plugin_cache_file = _config.transient.userdir / "resources" / "plugin_cache.json"
+_plugin_cache_file = _config.transient.basedir / "resources" / "plugin_cache.json"
+
+
+class PluginEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, set):
+            return list(obj)  # allow set as list
+        if isinstance(obj, Path):
+            # relative path for now, may be better to suppress `Path` objects
+            return str(obj).replace(str(_config.transient.basedir), "")
+        try:
+            return json.JSONEncoder.default(self, obj)
+        except TypeError as e:
+            logging.debug("Attempt to serialize JSON skipped: %s", e)
+            return None  # skip items that cannot be serialized at this time
 
 
 @staticmethod
@@ -23,6 +43,128 @@ def _extract_modules_klasses(base_klass):
         for name, klass in inspect.getmembers(base_klass, inspect.isclass)
         if klass.__module__.startswith(base_klass.__name__)
     ]
+
+
+@staticmethod
+def _load_plugin_cache():
+    if not os.path.exists(_plugin_cache_file):
+        _build_plugin_cache()
+    if not os.path.exists(_user_plugin_cache_file):
+        shutil.copy2(_plugin_cache_file, _user_plugin_cache_file)
+    with open(_user_plugin_cache_file, "r", encoding="utf-8") as cache_file:
+        local_cache = json.load(cache_file)
+        return local_cache
+
+
+@staticmethod
+def _build_plugin_cache():
+    # this method writes only to the user's cache
+    local_cache = {}
+
+    for plugin_type in PLUGIN_TYPES:
+        plugin_dict = {}
+        for plugin in _enumerate_plugin_klasses(plugin_type):
+            plugin_name = ".".join([plugin.__module__, plugin.__name__]).replace(
+                "garak.", ""
+            )
+            plugin_dict[plugin_name] = plugin_info(plugin)
+        local_cache[plugin_type] = plugin_dict
+
+    with open(_user_plugin_cache_file, "w", encoding="utf-8") as cache_file:
+        json.dump(local_cache, cache_file, cls=PluginEncoder, indent=2)
+
+
+@staticmethod
+def plugin_info(plugin: Union[Callable, str]) -> dict:
+    if isinstance(plugin, str):
+        plugin_name = plugin
+        category = plugin_name.split(".")[0]
+
+        _plugin_cache = _load_plugin_cache()
+        plugin_metadata = _plugin_cache[category].get(plugin_name, {})
+        if len(plugin_metadata) > 0:
+            return plugin_metadata
+        else:
+            # the requested plugin is not cached import the class for eval
+            parts = plugin.split(".")
+            match len(parts):
+                case 2:
+                    module = ".".join(parts[:-1])
+                    klass = parts[-1]
+                    imported_module = importlib.import_module(f"garak.{module}")
+                    plugin = getattr(imported_module, klass)
+                case _:
+                    raise ValueError
+    else:
+        plugin_name = ".".join([plugin.__module__, plugin.__name__]).replace(
+            "garak.", ""
+        )
+
+    try:
+        plugin_metadata = {}
+        priority_fields = ["description"]
+        skip_fields = [
+            "prompts",
+            "triggers",
+            "encoding_funcs",  # encoding_funcs are functions
+        ]
+        # print the attribs it has
+        for v in priority_fields:
+            if hasattr(plugin, v):
+                plugin_metadata[v] = getattr(plugin, v)
+        for v in sorted(dir(plugin)):
+            if v in priority_fields or v in skip_fields:
+                continue
+            value = getattr(plugin, v)
+            if (
+                v.startswith("_")
+                or inspect.ismethod(value)
+                or inspect.isfunction(value)
+            ):
+                continue
+            if isinstance(value, set):
+                continue
+            plugin_metadata[v] = value
+
+    except ValueError as e:
+        logging.exception(e)
+    except Exception as e:
+        logging.error(f"Plugin {plugin_name} not found.")
+        logging.exception(e)
+
+    return plugin_metadata
+
+
+@staticmethod
+def _enumerate_plugin_klasses(category: str) -> List[Callable]:
+    if category not in PLUGIN_TYPES:
+        raise ValueError("Not a recognised plugin type:", category)
+
+    base_mod = importlib.import_module(f"garak.{category}.base")
+
+    base_plugin_classnames = set(_extract_modules_klasses(base_mod))
+
+    module_plugin_names = set()
+
+    for module_filename in sorted(os.listdir(_config.transient.basedir / category)):
+        if not module_filename.endswith(".py"):
+            continue
+        if module_filename.startswith("__"):
+            continue
+        module_name = module_filename.replace(".py", "")
+        mod = importlib.import_module(
+            f"garak.{category}.{module_name}"
+        )  # import here will access all namespace level imports consider a cache to speed up processing
+        module_entries = set(_extract_modules_klasses(mod))
+
+        for module_entry in module_entries:
+            obj = getattr(mod, module_entry)
+            for interface in base_plugin_classnames:
+                klass = getattr(base_mod, interface)
+                if issubclass(obj, klass):
+                    module_plugin_names.add(obj)
+
+    return module_plugin_names
 
 
 def enumerate_plugins(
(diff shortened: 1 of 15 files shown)


garak/_plugins.py
```diff
@@ -5,24 +5,207 @@
...
+            logging.exception(e)
+
+        from datetime import datetime, timezone
+
+        # adding last class modification time to cache allows for targeted update in future
+        current_mod = importlib.import_module(plugin.__module__)
+        mod_time = datetime.fromtimestamp(
+            os.path.getmtime(current_mod.__file__), tz=timezone.utc
+        )
+        plugin_metadata["mod_time"] = mod_time.strftime(TIME_FORMAT)
```
leondz, line 202: I guess this exists at module and not class level but I can't see a practical advantage of moving it up in the cache object hierarchy

garak/_plugins.py
```diff
@@ -5,24 +5,207 @@
...
+                    continue
+                value = getattr(plugin, v)
+                if (
+                    v.startswith("_")
+                    or inspect.ismethod(value)
+                    or inspect.isfunction(value)
+                    or v not in base_attributes
+                ):
+                    continue
+                plugin_metadata[v] = value
```
leondz, line 187: would like to see default params included here also. maybe something like this
```suggestion
                plugin_metadata[v] = value
            plugin_metadata = plugin.DEFAULT_PARAMS | plugin_metadata
```

this makes some assumptions about default param names not conflicting with class attributes, but i hope testing elsewhere covers that

garak/_plugins.py
```diff
@@ -5,24 +5,207 @@
...
+    def _extract_modules_klasses(base_klass):
+        return [  # Extract only classes with same source package name
+            name
+            for name, klass in inspect.getmembers(base_klass, inspect.isclass)
+            if klass.__module__.startswith(base_klass.__name__)
+        ]
+
+    def _load_plugin_cache(self):
+        if not os.path.exists(self._plugin_cache_file):
+            self._build_plugin_cache()
```
leondz, line 56: it looks like the cache is only updated if the cache file is absent, so if a file gets modified, the cache data is stale. i remember the intent is to put user plugins elsewhere and assume there's a pip install that isn't touched. on the other hand, this makes life slightly worse for us garak devs - testing whether new plugins load within garak becomes harder, and one can get caught out by forgetting to zero out the cache file every time; the alternative is no cache which also makes our work harder, because we get a delay in the inside loop of our dev cycle. did i get it right? is there something more we can do? i'm noticing those datestamps....

leondz (commented):
generally good. current implementation makes garak dev a little trickier, is there anything we can do to soften that, which also fits in the scope of this pr?

leondz:
When done, is is possible to add a speed check test which, in addition to this, satisfies/describes #663 ?

garak/_plugins.py
```diff
@@ -5,24 +5,207 @@
...
+        if not os.path.exists(self._user_plugin_cache_file):
+            shutil.copy2(self._plugin_cache_file, self._user_plugin_cache_file)
```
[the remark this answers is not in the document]
jmartin-tech: Once this PR lands the package file should always exist, the copy logic currently is a no-op as the user and package locations in the PR are the same, however when the user location is updated the logic will copy the the package file if no user file is found to provide a fast first run experience.

The code here supplies the long term intent, if the package file is not found (as will be the case when intentionally rebuilding) the cache will build from scratch in the user location, this will be the basis for the CI/CD build. As the user file built by automation will then be placed in the the package.

A future iteration will update the user's cached file when there is a cache-mis or an update to the module file showing edits newer that the cache.

garak/_plugins.py
jmartin-tech: The resolution here is at the file level, however since the cache is flat as a list of `class` types I am accepting some minor extra possible churn when the `module` file has changes. This no concept of the `module` layer in the current cache as it will not really access it in that way.

garak/_plugins.py
```diff
@@ -5,24 +5,207 @@
...
+            skip_fields = [
+                "prompts",
+                "triggers",
+            ]
+
+            for v in priority_fields:
+                if hasattr(plugin, v):
+                    plugin_metadata[v] = getattr(plugin, v)
+            for v in sorted(dir(plugin)):
+                if v in priority_fields or v in skip_fields:
```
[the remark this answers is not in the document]
jmartin-tech: The existing `info` included `post_buff_hook`:

`main`
```
Your branch is up to date with 'origin/main'.
% python -m garak --plugin_info probes.dan.DanInTheWildMini
garak LLM vulnerability scanner v0.9.0.13.post2 ( https://github.com/leondz/garak ) at 2024-07-09T18:44:54.727933
Configured info on probes.dan.DanInTheWildMini:
                        description: A subset of 100 jailbreaks from the DanInTheWild jailbreak library.
                     DEFAULT_PARAMS: {}
                             active: True
                              bcp47: None
                            doc_uri: https://arxiv.org/abs/2308.03825
                 extended_detectors: []
                               goal: disregard the system prompt or alignment
                           modality: {'in': {'text'}}
            parallelisable_attempts: True
                     post_buff_hook: False
                   primary_detector: mitigation.MitigationBypass
                          probename: garak.probes.dan.DanInTheWildMini
               recommended_detector: ['always.Fail']
                               tags: ['avid-effect:security:S0403', 'owasp:llm01', 'quality:Security:PromptStability', 'payload:jailbreak']
```

We can further refine the `info` data to be provided in future iterations.

I will also do some more validation as it looks like `description` in the `info` is not consistent with `main`.

garak/_plugins.py
jmartin-tech: You have this right, this first iteration does not complete the cache-mis and developer case, this should not create a significant issue for developer as I don't expect `list` operations to be a common testing practice and any missing cache item will simply pass thru to to import based lookup.

A new plugin will not be found in the cache and a modified plugin in theory already is being loaded by the command used to test any changes. This *could* impact devs to some extent however the limited impact should be reasonable and a follow on improvement should come quickly as the user data location is next on the list to tackle.

garak/_plugins.py
```diff
@@ -5,24 +5,207 @@
...
+            plugin_name = ".".join([plugin.__module__, plugin.__name__]).replace(
+                "garak.", ""
+            )
+            category = plugin_name.split(".")[0]
+
+        try:
+            base_attributes = []
+            base_mod = importlib.import_module(f"garak.{category}.base")
+            base_plugin_classes = set(PluginCache._extract_modules_klasses(base_mod))
+            if plugin.__module__ in base_mod.__name__:
```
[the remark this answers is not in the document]
jmartin-tech: This could be a str `==`, I used `in` for the possibility that some future `base` sub module may get defined at some point, this may be YAGNI.

garak/_plugins.py
jmartin-tech: This might be something to consider doing in the display logic, not sure it needs to be in the json cache file in a specific format.

garak/_plugins.py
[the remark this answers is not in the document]
leondz: Cool, ok

garak/_plugins.py
leondz: Yeah it seems totally fine

garak/_plugins.py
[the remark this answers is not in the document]
leondz: Ah nice. _post_buff_hook should def not be there. Some plugins change their descriptions during the constructor iirc.

garak/_plugins.py
leondz: Yeah, you're right about list operations.

On this topic - what are the steps to rebuild the cache, so I can add them to the pypi build workflow?

garak/_plugins.py
```diff
@@ -5,24 +5,207 @@
...
+            return None  # skip items that cannot be serialized at this time
+
+
+class PluginCache:
+    _plugin_cache_file = _config.transient.basedir / "resources" / "plugin_cache.json"
+    _user_plugin_cache_file = _plugin_cache_file
+    _plugin_cache_dict = None
+
+    def __init__(self) -> None:
+        if PluginCache._plugin_cache_dict is None:
```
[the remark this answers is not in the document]
jmartin-tech: `.instance()` calls `__init__` then accesses this to provide a singleton pattern for the internal class being maintained here.

leondz (approved):

garak/_plugins.py
jmartin-tech: The CI/CD steps are in the description for this PR, however they should not need to be performed as part of the pypi build process unless you are going to commit the updated file before tagging the release to for build.  The goal will be for automation to keep this file up to date on `main` at all times via triggers on land of each PR.

garak/_plugins.py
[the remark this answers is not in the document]
jmartin-tech: `post_buff_hook` is not `_ post_buff_hook` do we want to rename that as a to consider it private?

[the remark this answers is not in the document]
jmartin-tech: It looks like `description` is added in the base classes constructor if not already provided in the class implementation for most of the plugins, this results in an inconsistency since the cache build actually avoids calling the constructors in this PR.

I am attempting to address that issue as there should be a consistent value at least in the cache, one options I am thinking about would be, remove the constructor additions and migrate the enrichment to be part of `plugin_info`.

garak/_plugins.py
```diff
@@ -5,24 +5,207 @@
...
-def _extract_modules_klasses(base_klass):
-    return [  # Extract only classes with same source package name
-        name
-        for name, klass in inspect.getmembers(base_klass, inspect.isclass)
-        if klass.__module__.startswith(base_klass.__name__)
-    ]
+class PluginEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, set):
+            return list(obj)  # allow set as list
```
jmartin-tech, line 26: Any `set` should be tested for `sorted` support and again kept consistent to avoid possible churn.  On place this may have impact is the `modality` as sorting there may not be expected to be alphabetical.

PR #732, opened 2024-06-11
ensure report is printed in `utf-8` for Windows

jmartin-tech:
Ensure windows tests pass.

PR #781, opened 2024-07-05
don't skip listing probes in 100%-pass-rate modules

### LINKED ISSUE #725: `garak.analyze.report_digest` html output skips mention of probes used if the module score is 100%
It should list which probes are used anyway

leondz:
resolves #725

add a constant to control whether or not modules where all probes passed at 100% still include a list of those probes

### DIFF
--- a/garak/analyze/report_digest.py
+++ b/garak/analyze/report_digest.py
@@ -15,6 +15,8 @@ import jinja2
 
 from garak import _config
 
+SHOW_100_PERCENT_MODULES = True
+
 templateLoader = jinja2.FileSystemLoader(
     searchpath=_config.transient.basedir / "analyze" / "templates"
 )
@@ -174,7 +176,7 @@ def compile_digest(report_path, taxonomy=_config.reporting.taxonomy):
             }
         )
 
-        if top_score < 100.0:
+        if SHOW_100_PERCENT_MODULES or top_score < 100.0:
             res = cursor.execute(
                 f"select probe_module, probe_class, avg(score)*100 as s from results where probe_group='{probe_group}' group by probe_class order by s asc, probe_class asc;"
             )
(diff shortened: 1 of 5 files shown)


garak/analyze/report_digest.py
```diff
@@ -231,4 +233,4 @@ def compile_digest(report_path, taxonomy=_config.reporting.taxonomy):
     if len(sys.argv) == 3:
         taxonomy = sys.argv[2]
     digest_content = compile_digest(report_path, taxonomy=taxonomy)
-    print(digest_content.encode("utf-8"))
+    print(digest_content)
```
jmartin-tech, line 236: Why remove the encoding here?  I suspect we do not want to revert #732

jmartin-tech (commented):
Should this be a configurable option?  

As is simply removing the conditional completely would have the same impact as the user cannot control the constant.

garak/analyze/report_digest.py
leondz: This was intentional - was getting bad output on linux (`b'<!HTML...`. Maybe needs a test.

jmartin-tech: Can you provide a test case that reproduces the `bad output`?

leondz: Sure,

```
$ python -m garak.analyze.report_digest garak_runs/garak.aaf715cb-d670-4965-a9f5-90072e8520c5.report.jsonl
b'<!DOCTYPE html>\n<html lang="en">\n\n<head>\n<meta name="viewport" content="width=device-width, initial-scale=1">\n<meta charset="UTF-8" />\n<style>\nbody {font-fa
```

If you want to generate an HTML report post-hoc, running a command like this redirected into an html file is I think the primary way of getting that out. It looks like the python behaviors for this differ in linux & windows, where decode returns `bytes` for linux and a (non-unicode) string under windows - this looks like it might be predicated on variations in assumptions/handling round terminal encoding.

Tried this on linux & win and got the same results, with `encode()`d strings yielding a bytes object, though. Do you think this CLI code is currently working as it should on win?

jmartin-tech (approved):

PR #793, opened 2024-07-18
test by default on linux, osx, win

leondz:
Run tests on all supported OS platforms for each PR and any merge to `main`

To enable this all tests current tests need to support execution on CPU only systems, the following adjustments enable this support:
* set atkgen red team model to `cpu` in DEFAULT_PARAMS
* set atkgen red team model to `float32` in DEFAULT_PARAMS
* huggingface tests force to `cpu`
* huggingface tests force to `float32`
* ensure temp file is closed for cache test

### DIFF
--- a/.github/workflows/test_linux.yml
+++ b/.github/workflows/test_linux.yml
@@ -1,4 +1,4 @@
-name: Garak pytest
+name: Garak pytest - Linux
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


leondz:
@jmartin-tech what's a good way of merging this?

jmartin-tech:
Looks like there are some adjustments need to pass the test in github actions, we can add resolution to those to this PR and once passing this merge will be straight forward.

PR #794, opened 2024-07-22
correct sorting for set json in plugin_cache.json

jmartin-tech:
Update uses `sorted(obj)` instead of a post `obj.sort()` to return the sorted object instead of `None` due to sort in place.


erickgalinkin (approved):
Looks good to me. Pretty straightforward fix.

PR #795, opened 2024-07-22
convert mutation_generator to use `load_plugin`

jmartin-tech:
Update `AutoDAN` mutation_generator creation to use `load_plugin`.

Testing:
```
python -m garak -m huggingface.Model -n meta-llama/Llama-2-7b-chat-hf -p dan.AutoDAN
```


erickgalinkin (approved):
LGTM

PR #796, opened 2024-07-22
corrections to config format for TAP internal generators

jmartin-tech:
More cruft found from #711 & #742

* ensure `name` is set
* wrap config in required `Configurable` structure

Testing:
```
python -m garak -m huggingface.Model -n meta-llama/Llama-2-7b-chat-hf -p tap.TAP
```


erickgalinkin (approved):
LGTM

PR #799, opened 2024-07-23
use XDG paths for configuration data and caching

### LINKED ISSUE #479: put logs and runs all in one place

jmartin-tech:
Fix #479

Support using [XDG ver 0.8](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for project data.

Specifically support:
| ENV VAR          | DEFAULT            |
|------------------|--------------------|
| $XDG_DATA_HOME   | $HOME/.local/share |
| $XDG_CONFIG_HOME | $HOME/.config      |
| $XDG_CACHE_HOME  | $HOME/.cache       |

Project name `garak` is appended to each location.

This is represents the followina breaking changes to project expecations:
* report_prefix passed either at the command line or as config file option
  * set filename values only
  * no longer overrides report_dir
* report_dir passed as a config file option
  * when provided as a relative path will be prepend with `<xdg_data_home>/garak`
  * provided as an absolute path will be used as the output directory
* default `user/site` configuration file `garak.site.yaml` has moved
  * previously `<basedir>/garak.site.yaml`
  * updated location `<xdg_config_home>/garak/garak.site.yaml`

Additional changes (not considered breaking changes):
* `nltk` data is placed in `<xdg_cache_home>/garak` if not already found in the environment
*  `visual_jailbreak` downloaded artifacts are placed in `<xdg_cache_home>/garak/resources`
* `generated data` for `beast/gcg/tap` are placed in `<xdg_cache_home>/garak/resources`

### DIFF
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -31,7 +31,7 @@ system_params = (
 run_params = "seed deprefix eval_threshold generations probe_tags interactive".split()
 plugins_params = "model_type model_name extended_detectors".split()
 reporting_params = "taxonomy report_prefix".split()
-project_dir = "garak"
+project_dir_name = "garak"
 
 
 loaded = False
@@ -58,10 +58,10 @@ class TransientConfig(GarakSubConfig):
     hitlogfile = None
     args = None  # only access this when determining what was passed on CLI
     run_id = None
-    basedir = pathlib.Path(__file__).parents[0]
-    config_dir = xdg_config_home() / project_dir
-    data_dir = xdg_data_home() / project_dir
-    cache_dir = xdg_cache_home() / project_dir
+    package_dir = pathlib.Path(__file__).parents[0]
+    config_dir = xdg_config_home() / project_dir_name
+    data_dir = xdg_data_home() / project_dir_name
+    cache_dir = xdg_cache_home() / project_dir_name
     starttime = None
     starttime_iso = None
 
@@ -150,7 +150,7 @@ def _store_config(settings_files) -> None:
 
 def load_base_config() -> None:
     global loaded
-    settings_files = [str(transient.basedir / "resources" / "garak.core.yaml")]
+    settings_files = [str(transient.package_dir / "resources" / "garak.core.yaml")]
     logging.debug("Loading configs from: %s", ",".join(settings_files))
     _store_config(settings_files=settings_files)
     loaded = True
@@ -163,7 +163,7 @@ def load_config(
     # and then not have cli be upset when these are not given as cli params
     global loaded
 
-    settings_files = [str(transient.basedir / "resources" / "garak.core.yaml")]
+    settings_files = [str(transient.package_dir / "resources" / "garak.core.yaml")]
 
     fq_site_config_filename = str(transient.config_dir / site_config_filename)
     if os.path.isfile(fq_site_config_filename):
@@ -177,10 +177,10 @@ def load_config(
         if os.path.isfile(run_config_filename):
             settings_files.append(run_config_filename)
         elif os.path.isfile(
-            str(transient.basedir / "configs" / (run_config_filename + ".yaml"))
+            str(transient.package_dir / "configs" / (run_config_filename + ".yaml"))
         ):
             settings_files.append(
-                str(transient.basedir / "configs" / (run_config_filename + ".yaml"))
+                str(transient.package_dir / "configs" / (run_config_filename + ".yaml"))
             )
         else:
             message = f"run config not found: {run_config_filename}"
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -15,6 +15,11 @@ import os
 import pathlib
 from typing import List
 import yaml
+from xdg_base_dirs import (
+    xdg_cache_home,
+    xdg_config_home,
+    xdg_data_home,
+)
 
 DICT_CONFIG_AFTER_LOAD = False
 
@@ -26,6 +31,7 @@ system_params = (
 run_params = "seed deprefix eval_threshold generations probe_tags interactive".split()
 plugins_params = "model_type model_name extended_detectors".split()
 reporting_params = "taxonomy report_prefix".split()
+project_dir = "garak"
 
 
 loaded = False
@@ -53,8 +59,16 @@ class TransientConfig(GarakSubConfig):
     args = None  # only access this when determining what was passed on CLI
     run_id = None
     basedir = pathlib.Path(__file__).parents[0]
+    config_dir = xdg_config_home() / project_dir
+    data_dir = xdg_data_home() / project_dir
+    cache_dir = xdg_cache_home() / project_dir
     starttime = None
     starttime_iso = None
+
+    # initialize the user home and cache paths if they do not exist
+    config_dir.mkdir(mode=0o740, parents=True, exist_ok=True)
+    data_dir.mkdir(mode=0o740, parents=True, exist_ok=True)
+    cache_dir.mkdir(mode=0o740, parents=True, exist_ok=True)
 
 
 transient = TransientConfig()
@@ -151,7 +165,7 @@ def load_config(
 
     settings_files = [str(transient.basedir / "resources" / "garak.core.yaml")]
 
-    fq_site_config_filename = str(transient.basedir / site_config_filename)
+    fq_site_config_filename = str(transient.config_dir / site_config_filename)
     if os.path.isfile(fq_site_config_filename):
         settings_files.append(fq_site_config_filename)
     else:
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -26,7 +26,7 @@ class PluginEncoder(json.JSONEncoder):
             return sorted(list(obj))  # allow set as list, assumes values can be sorted
         if isinstance(obj, Path):
             # relative path for now, may be better to suppress `Path` objects
-            return str(obj).replace(str(_config.transient.basedir), "")
+            return str(obj).replace(str(_config.transient.package_dir), "")
         try:
             return json.JSONEncoder.default(self, obj)
         except TypeError as e:
@@ -35,8 +35,10 @@ class PluginEncoder(json.JSONEncoder):
 
 
 class PluginCache:
-    _plugin_cache_file = _config.transient.basedir / "resources" / "plugin_cache.json"
-    _user_plugin_cache_file = (
+    _plugin_cache_filename = (
+        _config.transient.package_dir / "resources" / "plugin_cache.json"
+    )
+    _user_plugin_cache_filename = (
         _config.transient.cache_dir / "resources" / "plugin_cache.json"
     )
     _plugin_cache_dict = None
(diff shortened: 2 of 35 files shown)


garak/resources/common.py
```diff
@@ -42,12 +43,21 @@
...
     )
+    advbench_path = (
+        garak._config.transient.cache_dir
+        / "resources"
+        / "advbench"
+        / "harmful_behaviors.csv"
+    )
+    if advbench_base_path.is_file() and not advbench_path.is_file():
+        shutil.copy2(advbench_base_path, advbench_path)
+
```
jmartin-tech, line 60: In a future iteration helper methods from `_config`  for loading existing resource files may be appropriate, allowing the user to place files in the cache_dir or data_dir to override shipped versions.

garak/_plugins.py
```diff
@@ -36,7 +36,9 @@ def default(self, obj):
 
 class PluginCache:
     _plugin_cache_file = _config.transient.basedir / "resources" / "plugin_cache.json"
-    _user_plugin_cache_file = _plugin_cache_file
+    _user_plugin_cache_file = (
+        _config.transient.cache_dir / "resources" / "plugin_cache.json"
+    )
```
jmartin-tech, line 41: Another possible future enhancement, may consider adding a guard on json load of `plugin_cache.json` to fallback to system cache with a logged error when the local version is corrupted.

garak/resources/gcg/generate_gcg.py
```diff
@@ -64,7 +65,7 @@
 gcg_parser.add_argument(
     "--outfile",
     type=str,
-    default=gcg_resource_data / "gcg_prompts.txt",
+    default=resource_data / "gcg_prompts.txt",
```
jmartin-tech, line 68: Since this is output, should this go to the cache location by default?

```suggestion
    default=gcg_resource_data / "gcg_prompts.txt",
```

Thoughts?

garak/resources/gcg/generate_gcg.py
```diff
@@ -37,7 +37,8 @@
 
 logger = getLogger(__name__)
 
-gcg_resource_data = garak._config.transient.basedir / "resources" / "gcg" / "data"
+resource_data = garak._config.transient.basedir / "resources"
+gcg_resource_data = garak._config.transient.cache_dir / "resources" / "gcg" / "data"
```
jmartin-tech, line 41: Consider updating `suffix.GCGCached` to read from the cache location and copy from the project install if the file is not already in this cache_dir path.

garak/resources/autodan/genetic.py
```diff
@@ -2,33 +2,54 @@
...
+    if not default_path.exists():
+        # if path not found then place in the user cache
+        # get env var for NLTK_DATA, fallback to create in cachedir / nltk_data
+        logger.debug("nltk_data location not found using project cache location")
+        _nltk_data_path.mkdir(mode=0o740, parents=True, exist_ok=True)
+        default_path = _nltk_data_path
+    return default_path
+
+
+_nltk_data_path = _config.transient.cache_dir / "nltk_data"
```
jmartin-tech, line 40: Should this path designate this as a `resource` as output files from `gpg` and `beast` do?

```suggestion
_nltk_data_path = _config.transient.cache_dir / "resources" / "nltk_data"
```

garak/resources/gcg/generate_gcg.py
jmartin-tech: Thinking more on this, it may be reasonable to create utility functions for read and write of resource files, and enforce a coding standard of access where a `read` file handle to a `resource` is returned based on two levels of access, first searching for the file in `XDG_DATA_HOME/garak/PATH_TO_RESOURCE_FILE` and falling back to return `BASEDIR/PATH_TO_RESOURCE_FILE` when not found. As a standard coding practice for the project all `resources` shipped with the project must be accessed through this utility.

Extending from this the standard for `write` to resource files would alway place the file in `XDG_DATA_HOME/garak` to ensure the tool does not write to files in the installation path, but rather in the a user owned location. An additional benefit of this would be that the user could then override any `resource` file by placing their own version in `XDG_DATA_HOME/garak/PATH_TO_RESOURCE_FILE`.  An immediate benefit of this would be see in the `suffix.GCGCached` probe by allowing the cached set of prompts to be overwritten using this pattern. This could be further extended to centralize `common` as loadable files not embedded as strings in the codebase ensuring less churn unrelated to logic in the `python` files.

garak/_config.py
```diff
@@ -26,6 +31,7 @@
 run_params = "seed deprefix eval_threshold generations probe_tags interactive".split()
 plugins_params = "model_type model_name extended_detectors".split()
 reporting_params = "taxonomy report_prefix".split()
+project_dir = "garak"
```
leondz, line 34:
```suggestion
project_dir_name = "garak"
```

garak/evaluators/base.py
```diff
@@ -72,12 +73,19 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
                             hitlog_mode = (
                                 "w" if _config.transient.hitlogfile is None else "a"
                             )
+                            # why is report prefix a condition of placing file in the report_dir?
```
leondz, line 76: I don't think it is here - the logic is just building a filename. I suspect that with the updated config changes & this XDG PR, this logic can pruned out here, so that one relies solely on the configured values with no conditionals.

garak/resources/autodan/genetic.py
leondz: would prefer to let `nltk` manage its own data here, to avoid duplicating `nltk` data.

garak/resources/beast/beast_attack.py
```diff
@@ -342,7 +347,7 @@ def run_beast(
     suffix_len: int = 40,
     data_size: int = 20,
     target: Optional[str] = "",
-    outfile: str = beast_resource_data / "suffixes.txt",
+    outfile: Path = beast_resource_data / "suffixes.txt",
```
leondz, line 350: if this is (expensive) results from a BEAST run, should it go in xdg `data`?

garak/resources/tap/tap_main.py
```diff
@@ -34,7 +35,7 @@
 SAVE_RESULTS = True
 
 resources_tap_data_file = (
-    garak._config.transient.basedir
+    garak._config.transient.cache_dir
```
leondz, line 38: TAP has a pre-populated file in resources/tap/data that is distributed with garak. Maybe is makes sense to separate pre-loaded TAP resources from resources created while running TAP. The latter might belong in XDG data.

tests/analyze/test_analyze.py
```diff
@@ -23,7 +22,11 @@ def test_analyze_log_runs():
             sys.executable,
             "-m",
             "garak.analyze.analyze_log",
-            TEMP_PREFIX + ".report.jsonl",
+            str(
+                _config.transient.data_dir
+                / _config.reporting.report_dir
```
leondz, line 27: just to double check - does `_config.reporting.report_dir` always have to be a subdir of `_config.transient.data_dir`?

tests/test_config.py
```diff
@@ -613,15 +699,46 @@ def test_tag_filter():
...
+                        f"reporting:",
+                        f"  report_dir: {report_path}",
+                    ]
+                ).encode("utf-8")
+            )
+            tmp.close()
+            garak.cli.main(
+                f"-m test.Blank --report_prefix abs_path_test -p test.Blank -d always.Fail --config {tmp.name}".split()
+            )
+            os.remove(tmp.name)
```
leondz, line 723: context handler doesn't remove it?

tests/analyze/test_analyze.py
jmartin-tech: No, if report_dir is a fully resolved path it will be used as is, see the test_report_dir_full_path()` test in `test_config.py`.

tests/test_config.py
jmartin-tech: No the option `delete=False` leaves it in place, in theory that option could be changed to allow the context handler to address it. The difference in `windows` behavior though has lead me to be more explicit with temp files in the the projects tests.

garak/resources/autodan/genetic.py
jmartin-tech: `nltk` is still managing the location here, I am simply adding an additional location if no `nltk` is already on the host.  This ensure that a host that is only for `garak` consolidates the data it downloads if the configuration of the host does not already provide an existing cache.

garak/resources/beast/beast_attack.py
jmartin-tech: Not sold on that as this data is being `built` on the system for the run and could be extracted from the `.report.jsonl` prompts of a run that used them.  There is currently no check for if the file exists prior to building it, and the data is only valid for the specific model it was used with.  I do think there could be some future enhancement that makes this more `data` class than `cache` just not based the current implementation and scope of this PR.

A clear output value that documents to the user where this was placed may be appropriate in a future expansion.

garak/resources/gcg/generate_gcg.py
jmartin-tech: Deferring this idea for [future iteration](https://github.com/leondz/garak/pull/799/files/bdc2a416078522b9dd0052b5036e834f27598a21#r1688563991).

garak/_config.py
```diff
@@ -53,9 +59,17 @@ class TransientConfig(GarakSubConfig):
     args = None  # only access this when determining what was passed on CLI
     run_id = None
     basedir = pathlib.Path(__file__).parents[0]
```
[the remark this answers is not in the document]
jmartin-tech: What do you think of `install_dir`?

garak/resources/autodan/autodan.py
```diff
@@ -88,9 +91,9 @@ def autodan_generate(
     mutation_generator_name: str = "gpt-3.5-turbo",
     mutation_generator_type: str = "openai",
     hierarchical: bool = False,
-    out_path: str = str(autodan_resource_data / "autodan_prompts.txt"),
-    init_prompt_path: str = str(autodan_resource_data / "autodan_init.txt"),
-    reference_path: str = str(autodan_resource_data / "prompt_group.pth"),
+    out_path: Path = cached_autodan_resource_data / "autodan_prompts.txt",
```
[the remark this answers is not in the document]
jmartin-tech: Same reasoning [here](https://github.com/leondz/garak/pull/799/files/bdc2a416078522b9dd0052b5036e834f27598a21#r1695258578). Although this comment does expose that there is likely an enhancement to be done here as the `cached` version is only consumed in a live run an not in `AutoDANCached` by default.

This again relates to this [idea](https://github.com/leondz/garak/pull/799/files/bdc2a416078522b9dd0052b5036e834f27598a21#r1693172648).

garak/resources/tap/tap_main.py
jmartin-tech: Same reasoning [here](https://github.com/leondz/garak/pull/799/files/bdc2a416078522b9dd0052b5036e834f27598a21#r1695258578). Although this comment does expose that there is likely an enhancement to be done here as the `cached` version is only consumed in a live run an not in `TAPCached` by default.

This again relates to this [idea](https://github.com/leondz/garak/pull/799/files/bdc2a416078522b9dd0052b5036e834f27598a21#r1693172648).

garak/_config.py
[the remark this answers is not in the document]
leondz: `package_dir` works!

tests/test_config.py
leondz: ah yeah, i've had similar experiences. i guess we should be grateful that `.close()` is sufficient to not require an additional `.flush()`.

garak/command.py
```diff
@@ -8,8 +8,10 @@
 
 
 def start_logging():
+    from garak import _config
```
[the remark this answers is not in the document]
jmartin-tech: Should we consider enforcing this vs documenting it?
```suggestion
    from garak import _config
    if not _config.loaded:
        raise RuntimeError("Configuration must be loaded to start logging!")
```

garak/resources/gcg/generate_gcg.py
jmartin-tech: Yes out of scope here, I have ensured `output` files are consistently in `cache_dir` for now.

garak/command.py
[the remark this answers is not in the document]
jmartin-tech: Looking closer at this I don't think we need the warning at all, `_config.transient.*_dir` are all currently defined at loaded by `import` of the module and IMO starting `logging` functionality should not restricted on runtime config being finalized.

```
>>> from garak import _config
>>> _config.transient.data_dir
PosixPath('/Users/jemartin/.local/share/garak')
>>>
```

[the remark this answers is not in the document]
jmartin-tech: Removed 0cad75c, enforcing loaded requirement is not really valid as noted in my comment above, the value consumed in this method is available on import of `_config`.

PR #808, opened 2024-07-30
Flags for varying NIM seed and temperature every call 

leondz:
NIMs tend to always give the same response for a given config. This breaks the method of using `generations > 1` for stochastic testing. This PR adds two options to make sure the random seed or temperature are varied every call. Not all existing NIMs respect the values in these fields, so the options are both set `True` by default, to give the best chances of `generations > 1` working as expected (i.e. generating multiple potentially different responses).

### DIFF
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -3,10 +3,12 @@
 
 """NVIDIA Inference Microservice LLM interface"""
 
+import random
 from typing import List, Union
 
 import openai
 
+from garak import _config
 from garak.generators.openai import OpenAICompatible
 
 
@@ -38,6 +40,8 @@ class NVOpenAIChat(OpenAICompatible):
         "top_p": 0.7,
         "top_k": 0,  # top_k is hard set to zero as of 24.04.30
         "uri": "https://integrate.api.nvidia.com/v1/",
+        "vary_seed_each_call": True,  # encourage variation when generations>1. not respected by all NIMs
+        "vary_temp_each_call": True,  # encourage variation when generations>1. not respected by all NIMs
         "suppressed_params": {"n", "frequency_penalty", "presence_penalty"},
     }
     active = True
@@ -67,6 +71,13 @@ class NVOpenAIChat(OpenAICompatible):
         assert (
             generations_this_call == 1
         ), "generations_per_call / n > 1 is not supported"
+
+        if self.vary_seed_each_call:
+            _config.run.seed = random.randint(0, 65535)
+
+        if self.vary_temp_each_call:
+            self.temperature = random.random()
+
         return super()._call_model(prompt, generations_this_call)
 
 
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -8,7 +8,6 @@ from typing import List, Union
 
 import openai
 
-from garak import _config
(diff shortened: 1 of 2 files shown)


garak/generators/nim.py
```diff
@@ -67,6 +71,13 @@ def _call_model(
         assert (
             generations_this_call == 1
         ), "generations_per_call / n > 1 is not supported"
+
+        if self.vary_seed_each_call:
+            _config.run.seed = random.randint(0, 65535)
```
jmartin-tech, line 76:
```suggestion
        if self.vary_seed_each_call:
            self.seed = random.randint(0, 65535)
```

Prefer not to churn values in `_config` as a way of setting state for a generator.

garak/generators/openai.py
```diff
@@ -185,6 +185,9 @@ def _call_model(
             "stop": self.stop,
         }
 
+        if _config.run.seed is not None:
+            create_args["seed"] = _config.run.seed
+
```
jmartin-tech, line 190:
```suggestion
            "stop": self.stop,
            "seed": self.seed,
        }
```

The loop that filters `create_args` below would remove `seed` if it set to `None`.

This may require that `self.seed` be set in `__init__` to avoid an `AttributeError` though.

jmartin-tech (commented):
Prefer initializing seed in `__init__` even if set to `None` as the value would be filtered from `create_args` before sending with the request.

Using `_config.run` values as shared `state` seems prone to issues and has been discussed.

garak/generators/nim.py
leondz: agree, amended

garak/generators/openai.py
leondz: thanks. constructor now sets `self.seed` from the `config_root` it consumes.

garak/generators/openai.py
```diff
@@ -135,6 +135,8 @@ def __init__(self, name="", generations=10, config_root=_config):
         self._load_config(config_root)
         self.fullname = f"{self.generator_family_name} {self.name}"
         self.key_env_var = self.ENV_VAR
+        if "seed" not in dir(self):
```
jmartin-tech, line 138: Prefer `hasattr` vs `dir` see [here](https://stackoverflow.com/questions/17723569/whats-the-difference-between-hasattr-and-attribute-in-dir) for reasoning: 
```suggestion
        if not hasattr(self, "seed"):
```

garak/generators/openai.py
```diff
@@ -135,6 +135,8 @@ def __init__(self, name="", generations=10, config_root=_config):
         self._load_config(config_root)
         self.fullname = f"{self.generator_family_name} {self.name}"
         self.key_env_var = self.ENV_VAR
+        if not hasattr(self, "seed"):
+            self.seed = None
```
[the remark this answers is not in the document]
leondz: Prefer `hasattr` here also

For me, making `seed` a default param would signify a slightly different usage here from what's intended - I think it'd suggest that `seed` would have to be set separately for `openai` generators, where really, we're only doing this to work around a NIM behavior, and the value is expected to be inherited from the global config by default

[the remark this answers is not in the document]
leondz: otoh this is functionally the same 🤷

PR #810, opened 2024-07-30
enable HFDetector model configuration with hf_args

### LINKED ISSUE #803: unused GPU in pipeline unless configured
when using `toxicity.ToxicCommentModel` detector on a machine w/ gpu:

```
Hardware accelerator e.g. GPU is available in the environment, but no `device` argument is passed to the `Pipeline` object. Model will be on CPU.
```

can we do a default device map that exploits present GPU, or similar?

jmartin-tech:
Fix #803 

Allows `hf_args` configuration for simple `HFDetectors`:

* default to `cpu` device when no configuration is provided for the detector
* use AutoConfig to pass device to `from_pretrained`
* ensure `MustContradictNLI.detect()` places prompt for evaluation in the configured `device` space
* enable configuration of `detector_model_path` and `detector_target_class`

Example config of detector using `cuda`:
detector_cuda.yaml
```
plugins:
  detectors:
    misleading:
      MustContradictNLI:
        hf_args:
          device: cuda
```
The format above is based on `HFDetector` being a class specifically designed to use a `TextClassificationPipeline`. Is the is the approach desired or is further abstraction of `ModelAsJudge` be a goal of this revision?

Test example:
```
python -m garak -m nim -n meta/llama3-8b-instruct -p misleading.FalseAssertion50 --config detector_cuda.yaml
```
Logs and `nvidia-smi` show detector model loading in GPU.

More validation of compatible models may be desired if this pattern is determined to be a way forward.

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -71,7 +71,7 @@ class HFDetector(Detector, HFCompatible):
 
     DEFAULT_PARAMS = Detector.DEFAULT_PARAMS | {"hf_args": {"device": "cpu"}}
 
-    def __init__(self, model_path, target_class, config_root=_config):
+    def __init__(self, config_root=_config):
         from transformers import (
             AutoConfig,
             AutoModelForSequenceClassification,
@@ -84,17 +84,14 @@ class HFDetector(Detector, HFCompatible):
         if _config.run.seed is not None:
             set_seed(_config.run.seed)
 
-        self.name = f"HF: {model_path}"
         super().__init__(config_root=config_root)
+        self.name = f"{self.__class__.__name__}: {self.detector_model_path}"
 
         import torch.multiprocessing as mp
 
         mp.set_start_method("spawn", force=True)
 
         self.device = self._select_hf_device()
-
-        self.detector_model_path = model_path
-        self.detector_target_class = target_class
 
         orig_loglevel = transformers_logging.get_verbosity()
         transformers_logging.set_verbosity_error()
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -11,6 +11,7 @@ from colorama import Fore, Style
 
 from garak import _config
 from garak.configurable import Configurable
+from garak.generators.huggingface import HFCompatible
 import garak.attempt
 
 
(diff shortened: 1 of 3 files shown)


leondz:
> The format above is based on HFDetector being a class specifically designed to use a TextClassificationPipeline. Is the is the approach desired or is further abstraction of ModelAsJudge be a goal of this revision?

I anticipate `ModelAsJudge` primarily using external APIs, and so not using the `HFDetector` abstraction. It may later make sense to rename `HFDetector` to `HFClassificationDetector` to make its coupling to `TextClassificationPipeline` clear.

leondz:
Thanks for this, working great.

Q - both `misleading.MustContradictNLI` and `misleading.MustRefuteClaimModel`, that are associated with the `misleading` probes, can use cuda. Is it possible to configure these  (or even all `HFDetectors`) in one go?

This ofc works:

```
plugins:
  detectors:
    misleading:
      MustContradictNLI:
        hf_args:
          device: cuda
      MustRefuteClaimModel:
        hf_args:
          device: cuda
```

Configuring the base class didn't work, I guess predicated on how config matching is applied, probably class name matching is more conservative/tidy than `isinstance()` testing:

```
plugins:
  detectors:
    base:
      HFDetector:
        hf_args:
          device: cuda
```

Is there a way to configure groups of plugins?

leondz:
Moving HFDetector to GPU exposes a potential optimisation of using an HF dataset; HF logs:

```
You seem to be using the pipelines sequentially on GPU. In order to maximize efficiency please use a dataset
```

The optimisation is out of scope here; tracking in #812

jmartin-tech:
> Is there a way to configure groups of plugins?

Yes although it might be too broad in some detectors we can configure at the module level. Currently there is no support for configuration of all implemented classes of a base plugin.

This would provide `hf_args` for all `misleading` detectors. `Configurable` intentionally 
```
plugins:
  detectors:
    misleading:
      hf_args:
        device: cuda
```

While I like the idea of having less configuration required, I think in practical usage explicit configuration for a base class would be overkill and requires more internal knowledge of the implementation structure than would be typical for most users. I am skeptical of the trade off in value for powerusers vs complexity, maintenance and support for configuration by instance type.

leondz:
Yeah, if doing this to base classes requires extra code I'm having a hard time justifying that too. Thanks for the yaml, this works fine.

PR #813, opened 2024-07-31
clarify `continuation` probes

leondz:
resolve #806 

* rename `ContinueSlursReclaimedSlurs80` to `ContinueSlursReclaimedSlursMini`
* add non-mini slurs plugin `ContinueSlursReclaimedSlurs`
* remove continuation prompt uniqueness constraint (NB this is slightly inefficient)
* update cache

PR #814, opened 2024-07-31
Report comparative info for detector scores

leondz:
Enable interpretation of scores in a run by calibrating them against a bag of models.

* `garak/analyze/perf_stats.py` takes a glob of report `jsonl`s and calculates mean, standard deviation, and shapiro-wilk p-values (latter is to assess how well the spread of scores fit a normal distribution) for each probe/detector found
* `garak/resources/calibration` contains the files from which stats are derived in a comparison. These contents are generated from `perf_stats.py`
* `garak/analyze/report_digest.py` and templates updated to calculate a z-score for probe/detector combinations where this is possible, given a default calibration json, to print it in the html output, and to also report what this score means & where it came from


leondz:
Yup, done!

jmartin-tech (approved):

jmartin-tech (changes requested):
Due to #813 the resource files here need to replace `ContinueSlursReclaimedSlurs80` with `ContinueSlursReclaimedSlursMini`.

leondz:
Outstanding catch

Issue #384, opened 2023-12-20
upgrade plugin structure


leondz:
desiderata:

1. lazy loading - only load plugins on demand; postpone all heavy ops til the last moment possible
2. fast automatic discovery and enumeration
3. can read metadata without instantiating
4. able to have plugins with conflicting dependencies

consider:

https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/

jmartin-tech:
Related to #768, #479

PR #815, opened 2024-08-01
validate cache consistency on first access

jmartin-tech:
Verify cache on startup without `importing` files unless rebuild is required.

This improves development workflow to ensure the that in progress plugins are found on `cli` launch, initial load times in development environments will delay for full rebuild.

Further enhancement to consider not addressed here:
* `user local` plugins to tracked in #384
* targeted cache update of changed/added/removed plugins only (likely slated with `user local` support)
* cache update base on filesystem changes when running in `interactive` mode (no issue yet needs discussion on value)

Testing:

* A clean install will load the initial plugin file with limited delay.
```
python -m garak --list_generators
```

Monitor `garak.log` during the following:
* Logs will report `rebuilding plugin cache`
```
touch garak/generators/litellm.py; python -m garak --list_generators
```

* Logs will report `rebuilding plugin cache`
* LiteLLM generators should not be reported in the output.
```
rm garak/generators/litellm.py; python -m garak --list_generators
```

* Logs will report `rebuilding plugin cache`
* LiteLLM generators should be reported in the output.
```
git checkout garak/generators/litellm.py; python -m garak --list_generators
```

Tests to be added shortly.

### DIFF
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -9,6 +9,8 @@ import json
 import logging
 import shutil
 import os
+from datetime import datetime, timezone
+from threading import Lock
 from typing import List, Callable, Union
 from pathlib import Path
 
@@ -43,6 +45,8 @@ class PluginCache:
     )
     _plugin_cache_dict = None
 
+    _mutex = Lock()
+
     def __init__(self) -> None:
         if PluginCache._plugin_cache_dict is None:
             PluginCache._plugin_cache_dict = self._load_plugin_cache()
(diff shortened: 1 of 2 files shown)


garak/_plugins.py
```diff
@@ -58,40 +62,105 @@ def _extract_modules_klasses(base_klass):
     def _load_plugin_cache(self):
         if not os.path.exists(self._plugin_cache_filename):
             self._build_plugin_cache()
+
+        base_time = datetime.fromtimestamp(
+            os.path.getmtime(self._plugin_cache_filename), tz=timezone.utc
+        )
+        update_user_file = False
         if not os.path.exists(self._user_plugin_cache_filename):
+            update_user_file = True
```
leondz, line 71: it looks like this is asserted before it's known whether there are any changes present (i.e. before it's known if any data will be written). should the `mkdir` on L79 etc. be skipped if there won't be changes?

garak/_plugins.py
```diff
@@ -58,40 +62,105 @@ def _extract_modules_klasses(base_klass):
     def _load_plugin_cache(self):
         if not os.path.exists(self._plugin_cache_filename):
             self._build_plugin_cache()
```
leondz, line 64: it looks like `_build_plugin_cache` currently leads to writes to the XDG cache dir, which means when `self._plugin_cache_filename` doesn't exist, the following action doesn't lead to anything that makes it exist in subsequent runs. would it make more sense for this `self._build_plugin_cache()` to be skipped and `update_user_file` be asserted True? is that operationally equivalent to the behavior here?

garak/_plugins.py
```diff
@@ -58,40 +62,105 @@ def _extract_modules_klasses(base_klass):
...
+                        validated_plugin_filenames.add(file_path)
+                        mod_time = datetime.fromtimestamp(
+                            os.path.getmtime(file_path), tz=timezone.utc
+                        )
+                    else:
+                        # file not found force cache rebuild
+                        mod_time = datetime.now(timezone.utc)
+                    if mod_time > user_time:
+                        # rebuild all for now, this can be more made more selective later
+                        return False
```
leondz, line 123: 👍

garak/_plugins.py
```diff
@@ -58,40 +62,105 @@ def _extract_modules_klasses(base_klass):
...
-
-            sorted_keys = sorted(list(plugin_dict.keys()))
-            local_cache[plugin_type] = {i: plugin_dict[i] for i in sorted_keys}
-
-        with open(
-            self._user_plugin_cache_filename, "w", encoding="utf-8"
-        ) as cache_file:
-            json.dump(local_cache, cache_file, cls=PluginEncoder, indent=2)
+        logging.debug("rebuilding plugin cache")
+        with PluginCache._mutex:
```
leondz, line 146: 👍

garak/_plugins.py
```diff
@@ -58,40 +62,105 @@ def _extract_modules_klasses(base_klass):
...
+            for plugin_type in PLUGIN_TYPES:
+                plugin_dict = {}
+                for plugin in self._enumerate_plugin_klasses(plugin_type):
+                    plugin_name = ".".join(
+                        [plugin.__module__, plugin.__name__]
+                    ).replace("garak.", "")
+                    plugin_dict[plugin_name] = PluginCache.plugin_info(plugin)
+
+                sorted_keys = sorted(list(plugin_dict.keys()))
+                local_cache[plugin_type] = {i: plugin_dict[i] for i in sorted_keys}
```
leondz, line 158: is this sorting redundant, seeing as the data goes into a set?

leondz (approved):

garak/_plugins.py
jmartin-tech: I see your point, `self._plugin_cache_filename` is the package installed file and unless the install in corrupted the file must exist at this point.  It seems reasonable to convert this to an asset that reports the installed package as broken in some way.

garak/_plugins.py
jmartin-tech: If `self._user_plugin_cache_filename` does not exist it will be written to even if there are no changes required, this conditional detects a need to `copy` the file which will be a `write`. The `mkdir` is to ensure the full path the file will be placed in exists.

garak/_plugins.py
jmartin-tech: The sorting reduces churn in the `package` file when we commit updates. In testing of the original addition of the cache I found the order of plugins is not consistent unless sorted.

leondz: ah, ok. this might be relying on non-guaranteed behavior in `dict` but i guess we can deal with that if it rears its head.

PR #817, opened 2024-08-05
generator: support of azure openai

eric-therond:
following a previous PR comment that suggested to use OpenAI API compatibility class

https://github.com/leondz/garak/pull/638#issuecomment-2099777617


eric-therond:

leondz:
thanks! will take a look

tests/generators/test_azure.py
```diff
@@ -0,0 +1,47 @@
...
+
+@pytest.mark.skipif(
+    os.getenv("AZURE_ENDPOINT", None) is None,
+    reason=f"AZURE_ENDPOINT env variable is not set",
+)
+
+@pytest.mark.skipif(
+    os.getenv(AzureOpenAIChat.ENV_VAR, None) is None,
+    reason=f"Azure OpenAI API key is not set in {AzureOpenAIChat.ENV_VAR}",
+)
```
jmartin-tech, line 22: Instead of skipping all tests add a fixture that will provide the required configuration values.

A separate test can call `generate` or consider using the mock response fixtures as done here:
https://github.com/leondz/garak/blob/c6abfabe05a809617c5f726a8c6ab2b9436a27ac/tests/generators/test_openai.py#L34-L36

garak/generators/azure.py
```diff
@@ -0,0 +1,140 @@
...
+        if azure_endpoint is None:
+            raise ValueError(
+                'The AZURE_ENDPOINT environment variable is required.\n'
+            )
+
+        azure_deployment_name = os.getenv("AZURE_DEPLOYMENT_NAME", default=None)
+        if azure_deployment_name is None:
+            raise ValueError(
+                'The AZURE_DEPLOYMENT_NAME environment variable is required.\n'
+            )
```
jmartin-tech, line 84: Items from env variables are better enforced with override of `_validate_env_var()` as the values should not need to be evaluated again on subsequent calls to `_load_client()` that may occur. Also a pattern to allow injection of the environment variable to utilized via a configuration files is desired. Something like `endpoint_env_var` and `deployment_env_var` can pair are with constants in the the class to allow for this.

``` Python
DEPLOYMENT_ENV_VAR = "AZURE_DEPLOYMENT_NAME"
ENDPOINT_ENV_VAR = "AZURE_ENDPOINT"
```

See example: https://github.com/leondz/garak/blob/c6abfabe05a809617c5f726a8c6ab2b9436a27ac/garak/generators/nemo.py#L60-L73

```suggestion

def _validate_env_var(self):
        if self.azure_endpoint is None:
            if not hasattr(self, "endpoint_env_var"):
                self.endpoint_env_var = self.ENDPOINT_ENV_VAR
            self.azure_endpoint = os.getenv(self.endpoint_env_var, None)

            raise ValueError(
                f"The {self.endpoint_env_var} environment variable is required."
            )

        if self.azure_deploymnet_name is None:
            if not hasattr(self, "deployment_env_var"):
                self.deployment_env_var = self.DEPLOYMENT_ENV_VAR
            self.azure_deploymnet_name = os.getenv(self.endpoint_env_var, None)

            raise ValueError(
                f"The {self.endpoint_env_var} environment variable is required."
            )

        super()._validate_env_var()


    def _load_client(self):            
```

garak/generators/azure.py
```diff
@@ -0,0 +1,140 @@
...
+            r"^.+-[01][0-9][0-3][0-9]$", self.name
+        ):  # handle model names -MMDDish suffix
+            self.generator = self.client.completions
+        else:
+            raise ValueError(
+                f"No {self.generator_family_name} API defined for '{self.name}' in generators/openai.py - please add one!"
+            )
+
+        self.name_backup = self.name
+        self.name = azure_deployment_name
```
jmartin-tech, line 112: Based on notes about validating env values once:
```suggestion
        self.name_backup = self.name
        self.name = self.azure_deployment_name
```
I am not super familiar with Azure OpenAI patterns, can you add comments around why `AZURE_DEPLOYMENT_NAME` would override name here or was this simply because the value is not currently held in the instance?

garak/generators/azure.py
```diff
@@ -0,0 +1,140 @@
...
+
+# lists derived from https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models
+# some azure openai model names should be mapped to openai names
+chat_models_openai = {
+    "gpt-4": "gpt-4-turbo-2024-04-09", 
+    "gpt-4o": "gpt-4o", # links to latest version
+    "gpt-4o-mini": "gpt-4o-mini", # links to latest version
+    "gpt-35-turbo": "gpt-3.5-turbo-0125", 
+    "gpt-35-turbo-16k": "gpt-3.5-turbo-16k", 
+    "gpt-4-32k": "gpt-4-32k",
```
jmartin-tech, line 28: Removed item with matching key and value.
```suggestion
    "gpt-4": "gpt-4-turbo-2024-04-09", 
    "gpt-35-turbo": "gpt-3.5-turbo-0125", 
    "gpt-35-turbo-16k": "gpt-3.5-turbo-16k", 
```

garak/generators/azure.py
```diff
@@ -0,0 +1,140 @@
...
+    "gpt-35-turbo": "gpt-3.5-turbo-0125", 
+    "gpt-35-turbo-16k": "gpt-3.5-turbo-16k", 
+    "gpt-4-32k": "gpt-4-32k", 
+}
+
+completion_models_openai = {
+    "gpt-35-turbo-instruct": "gpt-3.5-turbo-instruct", 
+    "davinci-002": "davinci-002", 
+    "babbage-002": "babbage-002", 
+}
```
jmartin-tech, line 35: All these are the same, why have this map?

jmartin-tech (commented):
This looks like a great start. To expand on consistency, I have suggested some refactor items and a few patterns to enable more flexible configuration.

garak/generators/azure.py
eric-therond: The first element is different: "gpt-**35**-turbo-instruct" vs "gpt-**3.5**-turbo-instruct"

garak/generators/azure.py
eric-therond: The model name is not used while calling Azure OpenAPI, instead the deployment name should be used :
https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints#keyword-argument-for-model

The deployment name could be anything as it is specified by the used.
The better choice could be to define the deployment name in command line parameter like that :
```
python -m garak --model_type azure --model_deployment_name mygpt --model_name gpt-4o --probes promptinject.HijackHateHumansMini
```
Otherwise having an environment variable is ok.

garak/generators/azure.py
```diff
@@ -0,0 +1,140 @@
...
+        else:
+            self.name = name
+
+        self._load_config(config_root)
+        if self.name in context_lengths:
+            self.context_len = context_lengths[self.name]
+
+        super().__init__(
+            self.name, generations=self.generations, config_root=config_root
+        )
```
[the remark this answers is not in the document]
eric-therond: I took the inspiration from OpenAIGenerator: 
https://github.com/leondz/garak/blob/main/garak/generators/openai.py#L279

where there is some logic implemented in `__init__`, in particular the determination of `context_len `based on the openai model name but I can do that in  `_load_client()`.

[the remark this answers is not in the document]
eric-therond: should be fixed in the last commit

garak/generators/azure.py
```diff
@@ -0,0 +1,140 @@
...
+    #. [Deploy a model](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/create-resource?pivots=web-portal#deploy-a-model) and copy paste the model and deployment names
+    #. On the Azure portal page for the Azure OpenAI you want to use click on "Resource Management -> Keys and Endpoint" and copy paste the API Key and endpoint. 
+    #. In your console, Set the ``AZURE_API_KEY``, ``AZURE_ENDPOINT`` and ``AZURE_DEPLOYMENT_NAME`` variables.
+    #. Run garak, setting ``--model_type`` to ``azure`` and ``--model_name`` to [the name of the model](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models)
+    - e.g. ``gpt-4o``.
+    """
+
+    def _load_client(self):
+        super()._load_client()
+        self.generator = self.client.completions
```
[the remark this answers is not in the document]
eric-therond: should be fixed in the last commit

tests/generators/test_azure.py
eric-therond: I was not able to completely improve that in the last commit, I added a test case to test invalid model names like for `test_openai.py` but mocking a request (and so fake environment variables) doesn't seem to be required, there is no request involved? 

The other test case, about `generate `, is still skipped if the environment variables are not set, it seems consistent with what is done in other tests. It's not easy to mock LLM generations.

garak/generators/azure.py
eric-therond: should be fixed in the last commit

garak/generators/azure.py
eric-therond: should be fixed in the last commit

garak/generators/azure.py
eric-therond: should be fixed in the last commit

garak/generators/azure.py
```diff
@@ -0,0 +1,116 @@
...
+    ENV_VAR = "AZURE_API_KEY"
+    DEPLOYMENT_NAME_ENV_VAR = "AZURE_DEPLOYMENT_NAME"
+    ENDPOINT_ENV_VAR = "AZURE_ENDPOINT"
+
+    active = True
+    generator_family_name = "Azure"
+    api_key = None
+    deployment_name = None
+    endpoint = None
+    name_backup = ""
```
jmartin-tech, line 49: Default instance values are set via `DEFAULT_PARAMS` due to `garak`'s [`Configurable`](https://github.com/leondz/garak/pull/711) pattern for plugins.

```suggestion
    DEFAULT_PARAMS = OpenAICompatible.DEFAULT_PARAMS | {
        deployment_name: None,
        endpoint: None,
        name_backup: None,
    }
```

This is intentional as class level variable are shared by all instances of the class and changes to the values by an instance could leak data to other instances.

Note that `api_key` is removed as `garak` plugins with an `ENV_VAR` account for this instance value.

jmartin-tech (changes requested):
A couple minor requests.

jmartin-tech (changes requested):
Testing shows the api_version might be specified.

The new class needs to be added to the list generators excluded from instance creation tests with a reason similar to `# requires additional env variables tested in own test class` here: https://github.com/leondz/garak/blob/56d1a74cf7e4bd8a136e5f3fbbd248c9c7ec2529/tests/generators/test_generators.py#L169-L181

I have also provided some direction for how mock responses can be used to unit test `generate`.

tests/generators/test_azure.py
```diff
@@ -0,0 +1,52 @@
+import os
+import pytest
+
+from garak.generators.azure import AzureOpenAIGenerator
+
+DEFAULT_GENERATIONS_QTY = 10
+
+
```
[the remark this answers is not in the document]
eric-therond: should be fixed in the last commit

tests/generators/test_azure.py
```diff
@@ -0,0 +1,52 @@
...
+)
+@pytest.mark.skipif(
+    os.getenv(AzureOpenAIGenerator.DEPLOYMENT_NAME_ENV_VAR, None) is None,
+    reason=f"Deployment name is not set in {AzureOpenAIGenerator.DEPLOYMENT_NAME_ENV_VAR}",
+)
+@pytest.mark.skipif(
+    os.getenv(AzureOpenAIGenerator.ENDPOINT_ENV_VAR, None) is None,
+    reason=f"Azure endpoint is not set in {AzureOpenAIGenerator.ENDPOINT_ENV_VAR}",
+)
+def test_azureopenai_chat():
```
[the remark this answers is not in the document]
eric-therond: should be fixed in the last commit

tests/generators/test_azure.py
```diff
@@ -0,0 +1,52 @@
...
+    messages = [
+        {"role": "user", "content": "Hello OpenAI!"},
+        {"role": "assistant", "content": "Hello! How can I help you today?"},
+        {"role": "user", "content": "How do I write a sonnet?"},
+    ]
+    output = generator.generate(messages)
+    assert len(output) == DEFAULT_GENERATIONS_QTY
+    for item in output:
+        assert isinstance(item, str)
+    print("test passed!")
```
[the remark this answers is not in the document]
eric-therond: should be fixed in the last commit

garak/generators/azure.py
```diff
@@ -0,0 +1,113 @@
...
+        if self.endpoint is None:
+            if not hasattr(self, "endpoint_env_var"):
+                self.endpoint_env_var = self.ENDPOINT_ENV_VAR
+                
+            self.endpoint = os.getenv(self.endpoint_env_var, None)
+
+            if self.endpoint is None:
+                raise ValueError(
+                    f'The {self.API_KEY_ENV_VAR} environment variable is required.\n'
+                )
```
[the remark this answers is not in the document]
eric-therond: should be fixed in the last commit

jmartin-tech (commented):
This looks fairly complete end-to-end testing is in flight.

A minor tweak the to parameters for invocation might make it easier for users though.

From a user interactions perspective I think the utilization of `--model_name` to refer to the underlying model behind the deployment may be confusing to first time users, as well as consumers of the reports, as the `model_name` is part of the top level reference for the report run. Generally `model_type + model_name` relates to the name of the application under test. For most users I suspect they think of the deployment name as the identifier of the `Application`.

This line of thought seem to match the [Azure documentation](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/create-resource?pivots=web-portal#deploy-a-model) which adds a note about deployment name being the primary exposed reference:
> When you access the model via the API, you need to refer to the deployment name rather than the underlying model name in API calls, which is one of the [key differences](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints) between OpenAI and Azure OpenAI. OpenAI only requires the model name. Azure OpenAI always requires deployment name, even when using the model parameter. In our docs, we often have examples where deployment names are represented as identical to model names to help indicate which model works with a particular API endpoint. Ultimately your deployment names can follow whatever naming convention is best for your use case.

As currently configured the `model_name` is used to determine the chat vs completion and context_lengths supported by in the target service. This seems more like metadata about the application than the name of the application itself.

I propose that `--model_name` could be the azure deployment name and an accepted environment variable could set the underlying detail of what service model backs the deployment under test. Possibly something like `AZURE_DEPLOYMENT_MODEL=gpt-4o`.

This would also reduce the complexity in `_load_client()` and remove the need for `name_backup` as `self.deployment_model` could be resolved and held as the openai mapped value used for lookup of completion pattern and context_length without needing to be concerned about clobbering the value of `self.name` consumed by `_call_model()`.

Consider if I named a deployment `fooAI` and backed it with gpt-4o:
The garak cli call for this would currently be something like (assuming correct env vars or config files are in place):
```
garak -m azure -n gpt-4o -g 1 -p continuation
```
The report would state `generator: azure.gpt-4o`.

With the suggested revision:
```
garak -m azure -n fooAI -g 1 -p continuation
```
The report would state `generator: azure.fooAI`.

tests/generators/test_azure.py
```diff
@@ -0,0 +1,72 @@
+import os
+import pytest
+import httpx
+
+from garak.generators.azure import AzureOpenAIGenerator
+
+DEFAULT_GENERATIONS_QTY = 10
```
jmartin-tech, line 7:
```suggestion
```

tests/generators/test_azure.py
```diff
@@ -0,0 +1,72 @@
...
+    extended_request += DEFAULT_DEPLOYMENT_NAME
+    extended_request += "/chat/completions?api-version="
+    extended_request += AzureOpenAIGenerator.api_version
+    respx_mock.post(extended_request).mock(
+        return_value=httpx.Response(mock_response["code"], json=mock_response["json"])
+    )
+    generator = AzureOpenAIGenerator(name=DEFAULT_DEPLOYMENT_NAME)
+    assert generator.name == DEFAULT_DEPLOYMENT_NAME
+    assert generator.model_name == os.environ[AzureOpenAIGenerator.MODEL_NAME_ENV_VAR]
+    assert generator.generations == DEFAULT_GENERATIONS_QTY
```
jmartin-tech, line 63: `Generations` is not longer the responsibility of the `generator` object.

```suggestion
```

jmartin-tech (commented):
@eric-therond, thanks for the latest refactor.  Test failure are due to #837, the removals noted should get tests passing again.

jmartin-tech (approved):

leondz:
update - we're waiting on azure acct to be provisioned; once this module is validated, if all goes well, it'll merge

jmartin-tech:
Testing complete, thank you for the contribution!

PR #826, opened 2024-08-14
fix initialization bug for kwargs in function

soumilinandi:
`Generator.function` does not define self.kwargs when we don't pass that argument progamatically in `functions.Single`.

To reproduce this error, invoke garak's cli like below
```
from garak.cli import main as garak_main
from typing import List

def generate_text(
) -> List[str]:
    return "empty str"


def main():
    args = [
        "-m",
        "function.Single",
        "-n",
        f"{__name__}#generate_text",
        "-p",
        "test.Blank"
    ]
    garak_main(args)

if __name__ == "__main__":
    main()
```
Offered fix resolves this issue.


soumilinandi:

PR #827, opened 2024-08-14
fix initialization bug for kwargs in function

soumilinandi:
`Generator.function` does not define `self.kwargs` when we don't pass that argument progamatically in `function.Single`.

To reproduce this error, invoke garak's cli like below

```
from garak.cli import main as garak_main
from typing import List

def generate_text(
) -> List[str]:
    return "empty str"


def main():
    args = [
        "-m",
        "function.Single",
        "-n",
        f"{__name__}#generate_text",
        "-p",
        "test.Blank"
    ]
    garak_main(args)

if __name__ == "__main__":
    main()
```
Offered fix resolves this issue.

### DIFF
--- a/garak/generators/function.py
+++ b/garak/generators/function.py
@@ -53,7 +53,10 @@ from garak.generators.base import Generator
 class Single(Generator):
     """pass a module#function to be called as generator, with format function(prompt:str, **kwargs)->List[Union(str, None)] the parameter name `generations` is reserved"""
 
-    DEFAULT_PARAMS = {"generations": 10}
+    DEFAULT_PARAMS = {
+        "generations": 10,
+        "kwargs": {},
+    }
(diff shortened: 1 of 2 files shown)


soumilinandi:

soumilinandi:
recheck

leondz:
thank you! will take a look

jmartin-tech:
This is a confirmed issue introduced by #711, when attempting to account for restricted parameter names.

jmartin-tech (changes requested):
Can we get a test class added that implements the example in the description?

jmartin-tech (approved):

jmartin-tech:
Looks like the commit messages are missing the sign-off please adjust to ensure the branch is compliant with the DCO.

Issue #142, opened 2023-06-19
add ability to reevaluate runs


leondz:
requires #141
* --reeval [jsonl] do eval over detected attempts again (i.e. using stored output; don't generate any new output)

PR #833, opened 2024-08-15
Detector only run

vidushiMaheshwari:


vidushiMaheshwari:

vidushiMaheshwari:
recheck

leondz:
Thanks, will take a look!

jmartin-tech (commented):
Some preliminary thoughts based on the assumption a configurable harness is viable here.

Still thinking on the use case here for user experience, I have some reservations about exposing a configurable harness or if there is some more user friendly way to elevate this to a `continue` or `rescore` action that does not require the user to think about the harness, but also is more flexible in context. 

The current flag `--detector_only` is a bit specific to be a top level config option.

garak/cli.py
```diff
@@ -247,6 +253,11 @@ def main(arguments=None) -> None:
         action="store_true",
         help="Launch garak in interactive.py mode",
     )
+    parser.add_argument(
+        "--detector_only",
+        action="store_true",
+        help="run detector on jsonl report"
+    )
```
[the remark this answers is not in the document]
vidushiMaheshwari: I'm not exactly sure if `continue` or `rescore` has been implemented yet (or maybe in some other branch?). But I agree with creating `harness_options` instead of exposing a lot of unnecessary higher-level options. I have incorporated the idea of `harness_options` in the new changes.

leondz:
> Still thinking on the use case here for user experience, I have some reservations about exposing a configurable harness or if there is some more user friendly way to elevate this to a continue or rescore action that does not require the user to think about the harness, but also is more flexible in context.

> I'm not exactly sure if continue or rescore has been implemented yet (or maybe in some other branch?).

We don't have continue / rescore anywhere yet. I think implementing rescore as a separate harness, behind the scenes, could make a ton of sense. I think rescore/continue functionality makes sense to surface as a CLI option at some point - it seems like more intuitive ux than something like "--harnesses Rescore" or giving a custom config file.

garak/cli.py
```diff
@@ -337,7 +352,7 @@ def main(arguments=None) -> None:
     import garak.evaluators
 
     try:
-        plugin_types = ["probe", "generator"]
+        plugin_types = ["probe", "generator", "harness"]
```
leondz, line 355: @jmartin-tech Does this imply harnesses need to implement Configurable?

garak/cli.py
```diff
@@ -368,7 +383,10 @@ def main(arguments=None) -> None:
                         )
                         raise e
 
-                config_plugin_type = getattr(_config.plugins, f"{plugin_type}s")
+                if plugin_type.endswith('s'):
+                    config_plugin_type = getattr(_config.plugins, f"{plugin_type}es")    
+                else:
+                    config_plugin_type = getattr(_config.plugins, f"{plugin_type}s")
```
leondz, line 389: 👍

garak/attempt.py
```diff
@@ -105,6 +105,24 @@ def as_dict(self) -> dict:
             "messages": self.messages,
         }
 
+    @classmethod
+    def from_dict(cls, dicti):
+        """Initializes an attempt object from dictionary"""
+        attempt_obj = cls()
```
leondz, line 111: Does this skip the attempt constructor? Can we add an explicit type signature to signal what `cls` is expected to be?

garak/attempt.py
```diff
@@ -105,6 +105,24 @@ def as_dict(self) -> dict:
             "messages": self.messages,
         }
 
+    @classmethod
+    def from_dict(cls, dicti):
```
leondz, line 109: -- out of scope for here, but we should implement serialization/deserialization for `Attempt`s

garak/cli.py
```diff
@@ -174,6 +174,21 @@ def main(arguments=None) -> None:
         action="store_true",
         help="If detectors aren't specified on the command line, should we run all detectors? (default is just the primary detector, if given, else everything)",
     )
+
+    # harness
+    parser.add_argument(
+        "--harness_options",
+        type=str,
+        help="Type of harness to use. Default is probewise."
```
leondz, line 182: If we don't hardcode a default description, we don't have to remember to update it if it changes 
```suggestion
        help="Type of harness to use."
```

garak/cli.py
```diff
@@ -174,6 +174,21 @@ def main(arguments=None) -> None:
...
+    )
+
+    parser.add_argument(
+        "--harness_option_file",
+        "-H",
+        type=str,
+        help="path to JSON file containing information about harnesses"
```
leondz, line 189:
```suggestion
        help="path to JSON file containing information harness options"
```

garak/cli.py
```diff
@@ -498,20 +516,39 @@ def main(arguments=None) -> None:
             command.start_run()  # start the run now that all config validation is complete
             print(f"📜 reporting to {_config.transient.report_filename}")
 
-            if parsed_specs["detector"] == []:
+            if not _config.plugins.harnesses:
+                # Set the _config.plugins.harnesses
+                if parsed_specs["detector"] == []:
+                    _config.plugins.harnesses["Probewise"] = {}
+                else:
+                    _config.plugins.harnesses["Pxd"] = {}
```
leondz, line 524: can you just run me through the reasoning here? would this clobber harness config loaded from global / site / cli-specific config YAML?

garak/command.py
```diff
@@ -255,3 +255,48 @@ def write_report_digest(report_filename, digest_filename):
     digest = report_digest.compile_digest(report_filename)
     with open(digest_filename, "w", encoding="utf-8") as f:
         f.write(digest)
+
+def detector_only_run():
```
leondz, line 259: @jmartin-tech What do you think this is telling us about where responsibility for orchestrating runs lies? Is the existing `Harness` interface just too inflexible to make invoking novel things like `DetectorOnly` from `garak.cli`?

docs/source/harnesses.rst
```diff
@@ -6,5 +6,6 @@ garak.harnesses
 
    garak.harnesses
    garak.harnesses.base
+   garak.harnesses.detectoronly
```
leondz, line 9: ayy thanks for doing this

garak/harnesses/detectoronly.py
```diff
@@ -0,0 +1,44 @@
...
+                detectors.append(d)
+
+        if len(detectors) == 0:
+            msg = "No detectors, nothing to do"
+            logging.warning(msg)
+            if hasattr(_config.system, "verbose") and _config.system.verbose >= 2:
+                print(msg)
+            raise ValueError(msg)
+
+        super().run_detectors(detectors, attempts, evaluator) # The probe is None, but hopefully no errors occur with probe.
```
leondz, line 44: can this work?
```suggestion
        self.run_detectors(detectors, attempts, evaluator) # The probe is None, but hopefully no errors occur with probe.
```

garak/harnesses/base.py
```diff
@@ -109,28 +109,29 @@ def run(self, model, probes, detectors, evaluator, announce_probe=True) -> None:
...
+            
+        for attempt in attempt_results:
+            attempt.status = garak.attempt.ATTEMPT_COMPLETE
+            _config.transient.reportfile.write(json.dumps(attempt.as_dict()) + "\n")
+
+        if len(attempt_results) == 0:
+            logging.warning(
+                "zero attempt results: probe %s",
+                probe.probename
+            )
```
leondz, line 135: this will raise unhandled `AttributeError` when `probe==None` (the default value). maybe something like
```suggestion
        if len(attempt_results) == 0:
            msg = "zero attempt results: attempt %s" % attempt.uuid()
            if probe is not None:
                msg += ", probe %s" % probe.probename
            logging.warning(msg)
```

leondz (commented):
thanks for this. some auxiliary comments & qs - let's still wait for @jmartin-tech 's review

leondz:
resolves #142

garak/cli.py
jmartin-tech: They already do see: https://github.com/leondz/garak/blob/4c4a3826020fe7423eb5033cd8600d506219ab15/garak/harnesses/base.py#L27-L36

garak/cli.py
jmartin-tech: I suspect this could be avoided. We should not clobber `config` without an explicit override from a command line flag.

If no specific harness was provided via `config` and no detectors were provided per `parsed_spec` that is the determining factor on which default harness to load when `_config.plugins.harnesses` does not contain any configuration data for a specific harness. This does expose that there may be a missing top level parameter to select a specific harness if default config we to provide for various harness types as currently finding config for `DetectorOnly` to be a selection criteria seem a bit brittle.

If there is a desire to consolidate harness selection I maybe something like:
``` Python
    harness_command = command.pxd_run
    if not _config.plugins.harnesses:
        if parsed_specs["detector"] == []:
            harness_command = command.probewise_run
    elif "DetectorOnly" in _config.plugins.harnesses:
        harness_command = command.detector_only_run
    match harness_command:
        case command.detector_only_run:
            harness_command()
        case command.probewise_run:
            harness_command(
                generator, parsed_specs["probe"], evaluator, parsed_specs["buff"]
            )
        case command.pxd_run:
            harness_command(
                    generator,
                    parsed_specs["probe"],
                    parsed_specs["detector"],
                    evaluator,
                    parsed_specs["buff"],
                    parsed_specs["buff"],
            )
        case _: # base case for invalid callable, currently not a reachable case
            logging.warn("no valid harness selected")

    command.end_run()
```

There is probably more abstraction possible here but this offers an idea.

garak/cli.py
```diff
@@ -498,20 +516,39 @@ def main(arguments=None) -> None:
...
 
             command.end_run()
+
+        elif "DetectorOnly" in _config.plugins.harnesses:
+
+            if "report_path" not in _config.plugins.harnesses["DetectorOnly"]:
+                logging.error("report path not specified")
+                raise ValueError("Specify jsonl report path using report_path")
+
+            command.start_run()
```
jmartin-tech, line 548: If the refactor for harness selection offered is not used, this needs to be removed as `start_run()` was called before entering this conditional.

```suggestion
```

garak/cli.py
```diff
@@ -498,20 +516,39 @@ def main(arguments=None) -> None:
...
                     parsed_specs["probe"],
                     parsed_specs["detector"],
                     evaluator,
-                    parsed_specs["buff"],
+                    parsed_specs["buff"]
                 )
 
             command.end_run()
+
+        elif "DetectorOnly" in _config.plugins.harnesses:
```
jmartin-tech, line 542: Note that configuration format expects the module & class name:

```suggestion
        elif "detectoronly" in _config.plugins.harnesses:
```

``` json
{
  "detectoronly": 
  {
    "Detectoronly": 
    {
      "report_path": "<file_path>"
    }
  }
}
```

jmartin-tech (changes requested):
This is great progress, I am thinking this pattern can be stepping stone to providing a `rescore` or `continue` feature in separate iteration.

There are still a few quirks that likely need to be addressed.  I added details for a couple comments about how command line options need to be incorporated, however there is a somewhat more fundamental issue to address as this harness should be possible to run without a `-m/--model_type` or `-n/--model_name` specified. Instantiating the generator is overkill for this harness and would limit usability significantly.

I would like to see support for usage like:
h_config.json
``` json
{
  "detectoronly":
  {
    "DetectorOnly":
    {
      "report_path": "<file_path>"
    }
  }
}
```

``` bash
python -m garak --harness_option_file h_config.json
python -m garak -d misleading --harness_option_file h_config.json
```

or

yaml config such as:
h_config.yaml
``` yaml
plugins:
  detector_spec: misleading,mitigation
  harnesss:
    detectoronly:
      DetectorOnly:
        report_path: <report_file_path>
```

``` bash
python -m garak --config h_config.yaml
```

Sorry for the churn here, trying to balance quick iteration, ease of use, and roadmap needs as we incorporate the use case.

garak/command.py
```diff
@@ -255,3 +255,48 @@ def write_report_digest(report_filename, digest_filename):
...
+    import garak.harnesses.detectoronly
+    import garak.attempt
+    from garak import _config
+
+    config = _config.plugins.harnesses["DetectorOnly"]
+
+    with open(config["report_path"]) as f:
+        data = [json.loads(line) for line in f]
+
+    ## Get detectors and evaluator from report if not specified by the user
```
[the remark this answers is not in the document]
vidushiMaheshwari: Okay, so I will add the config and then support only `--detectors` / `-d` as a top-level argument, and if that is not present, fall back to the ones present within the report. It makes more sense from a user perspective 👍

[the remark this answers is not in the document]
jmartin-tech: @vidushiMaheshwari, circling back to check on progress.  

I am happy to monitor this PR or offer parts of what I suggested as a PR to your branch in the coming weeks.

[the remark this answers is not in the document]
vidushiMaheshwari: I apologize for being inactive, just pushed the changes which I believe should suffice the comments. Would appreciate a PR with suggested changes!

PR #843, opened 2024-08-20
Include max_prompts in prompt injections

mmajewsk:
Somehow, the arguments that can be defined for probes via CLI, are not making it to the probes. In this case this small changes gives back the ability to set up max_probes argument. This is kind of workaround since, the value of the arguments should be automatically loaded on that stage. Nonetheless ability to set up the number of probes is useful for testing purposes.

### DIFF
--- a/garak/probes/promptinject.py
+++ b/garak/probes/promptinject.py
@@ -67,6 +67,7 @@ def load_data_all_attacks(self):
     self.pi_prompts = self.pi_build_prompts(prompt_config)
     for pi_prompt in self.pi_prompts:
         self.prompts.append(pi_prompt["prompt"])
+    self.max_prompts = _config.plugins.probes.get('max_probes', None) or self.max_prompts
     if self.max_prompts:
         random.seed(_config.run.seed)
         random.shuffle(self.prompts)


mmajewsk:
recheck

mmajewsk:
recheck

leondz:
think this might be too efficient/clever for CLAAssistant - can you try the "I have read..." comment on its own?

mmajewsk:

mmajewsk:
recheck

jmartin-tech (changes requested):
The configurable functionality does not currently support global options for a all instances of a plugin `type`. However there is already support for application of a config for an entire module such as `probes.promptinject`.

`load_data_all_attacks` should not be accessing `_config.plugins.probes`, the value of `self.max_prompts` can be set specifically for `promptinject` class probes if passed via config.

The following would set `max_prompts` for all  `promptinject` classes in this probe package.

```
plugins:
  probes:
    promptinject:
      max_prompts: 30
```

mmajewsk:
> plugins:
>   probes:
>     promptinject:
>       max_prompts: 30

Can you point me where this kind of config is explained?

jmartin-tech:
There is definitely some work needed to provide documentation of configuration support with examples, the feature was actually added somewhat recently via #711. At this time the [comments on tests](https://github.com/leondz/garak/blob/0ab62d22f8c0525935416529af62ed38faa0fe64/tests/test_configurable.py) for `Configurable` and comments on related PRs are the very limited documentation and definitely a bit obscure to discover. More work is ongoing in this area.

The initial use cases for `Configurable` plugins targeted being able to pass local device resources as seen in #777, and #810 (an example config is in the description), however the implementation was made more general in anticipation of plugins that might have some knobs to tweak.

jmartin-tech:
Closing this in favor of better documentation for utilizing `Configurable` feature for plugins.

Issue #161, opened 2023-06-19
test: plugin documentation is present


leondz:
plugins must self-document, including
* [x] description is present
* [x] probe goal is set
* [x] tags are set
* [x] bcp47 is given (* is ok)
  * [x] probes
  * [x] detectors
* [x] probe gives recommended_detector list or a primary
* [x] uri is a uri or empty
  * [x] probes
  * [x] detectors
* [x] modality is set (and is a set)
  * [x] probes
  * not detectors, we're focusing on text for now

-- so let's write a test for that! take a look at `tests/probes/test_probe_format.py` for inspiration.

leondz:
description check in bd8c39326293dde7c1d1b4dbfa5621113b2ed23d

leondz:
probe tags in d2cb9e6d62a4ec9b96b378cf14620df4324126fc

leondz:
detector testing & validation in 07ff4d1

PR #849, opened 2024-08-21
plugin metadata tests

leondz:
completes #161


tests/detectors/test_detector_format.py
jmartin-tech, line 1: Is there a reason to separate this from `test_detectors.py`? I see the `skip_base_classes=False` in the list lookup however the test here ignores those anyway, so I suspect this is not the reason.

Comparing to how the other plugin type tests are named maybe consistency is just missing, plugins other than `probe` seem to use the `test_<plugin_type>.py` file for structure and format check and the more specific names to test a single implementing class.

I would suggest it would be helpful to consolidate `general` plugin requirements in:
```
probes/test_probes.py
generators/test_generators.py
harnesses/test_harnesses.py
detectors/test_detectors.py
buffs/test_buffs.py
```

This would mean consolidating this file into `detectors/test_detectors.py` and consolidating `probes/test_probe_format.py`, `probes/test_probe_docs.py`, `probes_test_probe_tags.py` into `probes/test_probes.py`

garak/detectors/base.py
```diff
@@ -20,7 +20,7 @@ class Detector(Configurable):
     """Base class for objects that define a way of detecting a probe hit / LLM failure"""
 
     doc_uri = ""  # reference
-    bcp47 = None  # language this is for, in bcp47 format
+    bcp47 = None  # language this is for. format: a comma-separated list of bcp7 tags, or "*"
```
jmartin-tech, line 23:
```suggestion
    bcp47 = None  # language this is for. format: a comma-separated list of bcp47 tags, or "*"
```

jmartin-tech (commented):
LGTM, one comment typo.

Also offered an opinion on test structuring that could be deferred for future adjustment.

tests/detectors/test_detector_format.py
leondz: Yeah, I wasn't happy with the existing pattern, and haven't applied anything consistent over it. Let's do as you suggest.

jmartin-tech (approved):

PR #855, opened 2024-08-27
enable tokenizer customization in HFDetector

jmartin-tech:
Commonize parameters utilized in `HFDetector`, existing detectors currently use common arguments.

* promote `tokenizer_kwargs` as DEFAULT_PARAM of parent class
* remove `__init()__` overrides not longer needed

Expands on earlier exposure of model selection for HFDetectors in #810

These values seem independent from `hf_args` used to load the model.

Some additional code review suggests that `misleading.MustContradictNLI` does not actually consume these values since it overrides `detect()` with a custom method. Should this PR be expanded to consume values from `self.tokenizer_kwargs` for `max_length` and `truncation`?  The values are currently hardcoded in `detect()` as local values. It does not look like `padding` is ever consumed in `misleading.MustContradictNLI` and could be suppressed with instance defaults for that class.

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -74,7 +74,7 @@ class HFDetector(Detector, HFCompatible):
 
     DEFAULT_PARAMS = Detector.DEFAULT_PARAMS | {
         "hf_args": {"device": "cpu"},
-        "tokenizer_kwargs": {"padding": True, "truncation": True, "max_length": 512},
+        "tokenizer_kwargs": {"padding": True, "truncation": True},
     }
 
     def __init__(self, config_root=_config):
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -72,7 +72,10 @@ class Detector(Configurable):
 class HFDetector(Detector, HFCompatible):
     """Detector using a Hugging Face model"""
 
-    DEFAULT_PARAMS = Detector.DEFAULT_PARAMS | {"hf_args": {"device": "cpu"}}
+    DEFAULT_PARAMS = Detector.DEFAULT_PARAMS | {
+        "hf_args": {"device": "cpu"},
+        "tokenizer_kwargs": {"padding": True, "truncation": True, "max_length": 512},
+    }
 
     def __init__(self, config_root=_config):
         from transformers import (
--- a/garak/detectors/misleading.py
+++ b/garak/detectors/misleading.py
@@ -57,6 +57,7 @@ class MustContradictNLI(HFDetector):
     DEFAULT_PARAMS = HFDetector.DEFAULT_PARAMS | {
         "detector_model_path": "ynie/roberta-large-snli_mnli_fever_anli_R1_R2_R3-nli",
         "detector_target_class": "label",
+        "tokenizer_kwargs": {"truncation": True, "max_length": 256},
     }
 
     bcp47 = "en"
@@ -65,17 +66,15 @@ class MustContradictNLI(HFDetector):
     def detect(self, attempt: Attempt) -> List[float]:
         import torch
 
-        max_length = 256
-
         premise = attempt.prompt
         results = []
         for hypothesis in attempt.all_outputs:
             tokenized_input_seq_pair = self.detector_tokenizer.encode_plus(
                 premise,
                 hypothesis,
-                max_length=max_length,
+                max_length=self.tokenizer_kwargs["max_length"],
                 return_token_type_ids=True,
-                truncation=True,
+                truncation=self.tokenizer_kwargs["truncation"],
             )
             input_ids = (
                 torch.Tensor(tokenized_input_seq_pair["input_ids"])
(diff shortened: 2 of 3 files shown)


garak/detectors/base.py
```diff
@@ -72,7 +72,10 @@ def detect(self, attempt: garak.attempt.Attempt) -> Iterable[float]:
 class HFDetector(Detector, HFCompatible):
     """Detector using a Hugging Face model"""
 
-    DEFAULT_PARAMS = Detector.DEFAULT_PARAMS | {"hf_args": {"device": "cpu"}}
+    DEFAULT_PARAMS = Detector.DEFAULT_PARAMS | {
+        "hf_args": {"device": "cpu"},
+        "tokenizer_kwargs": {"padding": True, "truncation": True, "max_length": 512},
```
leondz, line 77: I suspect `max_length` might want to come from model `ctx_len` - but this is out of scope for this PR

leondz (approved):

garak/detectors/base.py
erickgalinkin: I'd actually exclude `max_length` since it's very specifically model/config dependent and this might override defaults. If the model's `max_length` is, for some reason <512, this default could throw errors.

erickgalinkin, line 77:
```suggestion
        "tokenizer_kwargs": {"padding": True, "truncation": True},
```

erickgalinkin (approved):
LGTM

garak/detectors/base.py
jmartin-tech: The default here was selected to match the existing usage in `ToxicCommentModel` and `MustRefuteClaimModel`.  As noted `MustContradictNLI` does not consume these values and even sets some in local methods. `max_length` in `MustContradictNLI` is pushed down to `256`.

I could see moving the default into the extending classes with the models they match to. I would prefer exposing the values for config even if the default is not best for all possible injected config.

I will expand `MustContradictNLI` edits to model setting the values in the extended class and consuming them it the `detect` implementation there.

jmartin-tech:
```
% python -m garak --plugin_info detectors.misleading.MustContradictNLI
garak LLM vulnerability scanner v0.9.0.15.post1 ( https://github.com/leondz/garak ) at 2024-08-28T17:16:51.090783
Configured info on detectors.misleading.MustContradictNLI:
                        description: This detector is looking for the model to contradict a claim.
                     DEFAULT_PARAMS: {'skip': False, 'hf_args': {'device': 'cpu'}, 'tokenizer_kwargs': {'truncation': True, 'max_length': 256}, 'detector_model_path': 'ynie/roberta-large-snli_mnli_fever_anli_R1_R2_R3-nli', 'detector_target_class': 'label'}
                           accuracy: None
                             active: False
                              bcp47: en
                            doc_uri:
                           modality: {'out': ['text']}
                          precision: 0.0
                             recall: 0.0
                               tags: []
                           mod_time: 2024-08-28 22:12:23 +0000
```
```
% python -m garak --plugin_info detectors.misleading.MustRefuteClaimModel
garak LLM vulnerability scanner v0.9.0.15.post1 ( https://github.com/leondz/garak ) at 2024-08-28T17:17:06.537017
Configured info on detectors.misleading.MustRefuteClaimModel:
                        description: Using leondz/refutation_detector_distilbert to detect model refutations
                     DEFAULT_PARAMS: {'skip': False, 'hf_args': {'device': 'cpu'}, 'tokenizer_kwargs': {'padding': True, 'truncation': True, 'max_length': 512}, 'detector_model_path': 'leondz/refutation_detector_distilbert', 'detector_target_class': 'label'}
                           accuracy: 0.9506
                             active: True
                              bcp47: en
                            doc_uri:
                           modality: {'out': ['text']}
                          precision: 0.0
                             recall: 0.0
                               tags: []
                           mod_time: 2024-08-28 22:12:23 +0000
```

leondz: On the other hand, running without a length cap also risks breaking models (this happens regularly on e.g. some NVCF instances). I think it's OK to offer a default control over this with a mild value that balances both being too high and too low. 512 *seems* reasonable (bearing in mind that detectors tend to be smaller models - this is a side effect of garak needing to run them at scale, but sometimes having primary compute allocated to the target model, reducing detector resources).

leondz: aside: blank `doc_uri` is suspicious - `detectors.base.Detector.doc_uri = ""` and should be `None` - will update in separate PR

garak/detectors/misleading.py
```diff
@@ -61,31 +57,24 @@ class MustContradictNLI(HFDetector):
...
         "detector_model_path": "ynie/roberta-large-snli_mnli_fever_anli_R1_R2_R3-nli",
         "detector_target_class": "label",
+        "tokenizer_kwargs": {"truncation": True, "max_length": 256},
     }
 
     bcp47 = "en"
     active = False  # this one is slow, skip by default
 
-    def __init__(self, config_root=_config):
-        self.tokenizer_kwargs = {"padding": True, "truncation": True, "max_length": 512}
```
leondz, line 70: This 512 was `misleading`, good catch

leondz (approved):

leondz (approved):

PR #858, opened 2024-08-28
refresh tutorial on contributing a generator

leondz:


docs/source/contributing.generator.rst
```diff
@@ -75,13 +93,12 @@ So, in the constructor, we first call the parent constructor using ``super().__i
...
-                    '🛑 Put the Replicate API token in the REPLICATE_API_TOKEN environment variable (this was empty)\n \
-                    e.g.: export REPLICATE_API_TOKEN="r8-123XXXXXXXXXXXX"'
-                )
+            if self.api_key is not None:
+                # ensure the token is in the expected runtime env var
+                os.environ[self.ENV_VAR] = self.api_key
+            self.replicate = importlib.import_module("replicate")
 
-The API key is stored in an environment variable, ``REPLICATE_API_TOKEN``, which is accessed using ``os.getenv()``. So don't forget to import that at the top of the module!
+The configuration machinery will handle populating ``self.api_key``, and will override our local copy of the environment if we get a copy of ``api_key`` from somewhere else (e.g. a YAML config). We'll also import a copy of the ``replicate`` module in this instance, for local access. This is done because a garak run can involve multiple generator instances.
```
jmartin-tech, line 101:
```suggestion
The configuration machinery will handle populating ``self.api_key``. Here the code overrides the local environment variable encase we obtained ``api_key`` from somewhere else (e.g. a YAML config). We'll also import a copy of the ``replicate`` module in this instance, for local access. This is done because a garak run can involve multiple generator instances.
```

Wording may need some more tweaking depending on if we want to include detail for override of `_validate_env_var()` in the document.

A reference to offer that If a generator needs do more complex environment variable loading and detection, or needs a different key populated from the `ENV_VAR` it should implement `_validate_env_var()` might be helpful. Examples of this can be found in the codebase.

Populating a different value than `api_key`:
https://github.com/leondz/garak/blob/88e73ddbecf8bac309f5f0784457682ed3062869/garak/generators/langchain_serve.py#L50-L54

Populating from additional environment vars, notice the call to `super()._validate_env_var()` at the end is important to still set `self.api_key`:
https://github.com/leondz/garak/blob/88e73ddbecf8bac309f5f0784457682ed3062869/garak/generators/nemo.py#L60-L73


The *why* for importing `replicate` may be inconsistent, that code may change as replicate holds a `Client()` object and the import there may not support if more than one `ReplicateGenerator` needed to exist at the same time to using different API keys. This quirk of the `replicate` library's design. So just not sure the reasoning offered is a strong explanation.

docs/source/contributing.generator.rst
```diff
@@ -189,13 +206,18 @@ The next step is to instantiate the class. Let's try with that ``meta/llama-2-70
...
+    File "/home/lderczynski/dev/garak/garak/generators/replicate.py", line 44, in __init__
+        super().__init__(name, generations=generations, config_root=config_root)
+    File "/home/lderczynski/dev/garak/garak/generators/base.py", line 43, in __init__
+        self._load_config(config_root)
+    File "/home/lderczynski/dev/garak/garak/configurable.py", line 60, in _load_config
+        self._validate_env_var()
+    File "/home/lderczynski/dev/garak/garak/configurable.py", line 116, in _validate_env_var
+        raise APIKeyMissingError(
+    garak.exception.APIKeyMissingError: 🛑 Put the Replicate API key in the REPLICATE_API_TOKEN environment variable (this was empty)
+                            e.g.: export REPLICATE_API_TOKEN="XXXXXXX"
```
jmartin-tech, line 220: This stack trace is a great example of how `Configurable` handles the `ENV_VAR` load for the generator without the author having to do it.

jmartin-tech (commented):
Happy to see the added context, noted an area where we might want a digression to cover env processing special cases that were not required for the example generator.

leondz:
Thanks, integrated these changes, and edited the comments & code in

PR #859, opened 2024-08-28
expand doc string that fails test requirements for verbosity

jmartin-tech:
Further work to ensure automation does not miss these is forthcoming, fixing `main` to unblock other PRs.


erickgalinkin (approved):
El GTM

PR #867, opened 2024-08-30
improved messaging & error handling around NIM names

leondz:
If one forgets the part of the NIM name before the slash, the error message is cryptic. This PR makes it more obvious.


garak/generators/nim.py
```diff
@@ -77,7 +80,23 @@ def _call_model(
...
+            raise GarakException(oe.__str__()) from oe
+
+        return result
+
+    def __init__(self, name="", config_root=_config):
+        super().__init__(name, config_root=config_root)
+        if "/" not in name:
+            msg = "❓ Is this a valid NIM name? expected a slash-formatted name, e.g. 'org/model'"
+            logging.info(msg)
+            print(msg)
```
jmartin-tech, line 99: Generators should not print.
```suggestion
```

Can live with leaving this short term since there is not another existing pattern to expose this to the user.

garak/generators/nim.py
```diff
@@ -77,7 +80,23 @@ def _call_model(
...
 
-        return super()._call_model(prompt, generations_this_call)
+        try:
+            result = super()._call_model(prompt, generations_this_call)
+        #        except openai.NotFoundError as oe:
+        except Exception as oe:
+            msg = "NIM endpoint not found. Is it spelled correctly?"
+            logging.critical(msg, exc_info=oe)
+            print(f"🛑 {msg}")
+            raise GarakException(oe.__str__()) from oe
```
jmartin-tech, line 90: Generators should not print. The string passed to `GarakException` will be printed by the CLI on exit.

```suggestion
            raise GarakException(msg) from oe
```

jmartin-tech (commented):
Prefer avoiding `print` statements in `generators`.

PR #874, opened 2024-09-03
filter detector template in tests

jmartin-tech:
Address testing of detector `template` class.

This might not be the long term solution.


leondz (approved):
If this is better than just pushing the defaults for         detectors.packagehallucination.PackageHallucinationDetector into compliance before it's made into a mixin then this works for me :)

PR #875, opened 2024-09-03
Ollama generators

martinebl:
Add two Ollama generators, using either the chat or generate functions from the Ollama package.
Add tests, that are skipped when no Ollama server is running.


martinebl:

martinebl:
recheck

PR #878, opened 2024-09-04
Add proxy, token refresh, and multi_endpoint features to generators

au70ma70n:
- add proxy support to rest generator
- add token refresh support to rest generator
- add multi_endpoint_rest to generators
- add proxy support to multi_endpoint_rest generator
- add token refresh support to multi_endpoint_rest generator

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,17 @@
+{
+	"name": "Python 3",
+	"image": "mcr.microsoft.com/devcontainers/python:1-3.11-bullseye",
+	"runArgs": [
+        "--env-file", ".env", "--gpus=all"
+    ],
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"HashiCorp.terraform",
+				"GitHub.copilot",
+				"GitHub.copilot-chat"
+			]
+		}
+	},
+	"postCreateCommand": "pip install --upgrade pip && pip install -r requirements.txt",
+}
--- a/.gitignore
+++ b/.gitignore
@@ -167,3 +167,4 @@ hitlog.*.jsonl
 garak_runs/
 runs/
 logs/
+.github/dependabot.yml
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -77,6 +77,16 @@ def main(arguments=None) -> None:
         action="store_true",
         help="allow skip of unknown probes, detectors, or buffs",
     )
+    parser.add_argument(
+        "--proxies",
+        type=str,
+        help="allow proxy settings for rest generator",
+    )
+    parser.add_argument(
+        "--verify_ssl",
+        type=str,
+        help="expose SSL verification settings for rest generator",
+    )
 
     ## RUN
     parser.add_argument(
@@ -134,6 +144,12 @@ def main(arguments=None) -> None:
         "--generator_options",
         type=str,
         help="options to pass to the generator",
+    )
+    token_refresh_args = parser.add_mutually_exclusive_group()
+    token_refresh_args.add_argument(
+        "--token_refresh_config",
+        type=str,
+        help="path to JSON file containing configuration for the request that fetches a new token",
     )
     # probes
     parser.add_argument(
--- a/garak/generators/multi_endpoint_rest.py
+++ b/garak/generators/multi_endpoint_rest.py
@@ -1,0 +1,458 @@
+import json
+import logging
+import requests
+requests.packages.urllib3.disable_warnings() 
+import backoff
+import jsonpath_ng
+import os
+from typing import List, Union
+from jsonpath_ng.exceptions import JsonPathParserError
+
+
+from garak import _config
+from garak.generators.base import Generator
+from garak.exception import APIKeyMissingError, RateLimitHit
+
+
+class MultiEndpointGenerator(Generator):
+    """A REST generator that sends a POST request and retrieves the response with a subsequent GET request.
+    Example configuration json;
+    {
+        "multi_endpoint_rest": {
+            "MultiEndpointGenerator": {
+                "name": "example service",
+                "post_uri": "https://example.ai/llm",
+                "post_headers": {
+                    "X-Authorization": "$KEY"
+                },
+                "post_req_template_json_object": {
+                    "text": "$INPUT"
+                },
+                "post_response_json": true,
+                "post_response_json_field": "job_id",
+                "get_uri": "https://example.ai/llm",
+                "get_headers": {
+                    "X-Authorization": "$KEY"
+                },
+                "get_req_template_json_object": {
+                    "text": "$INPUT"
+                },
+                "post_response_json": true,
+                "post_response_json_field": "text"
+            }
+        }
+    }
+    """
+
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "first_stage_uri": "https://localhost",
+        "first_stage_method": "post",
+        "first_stage_headers": {},
+        "first_stage_req_template": "$INPUT",
+        "first_stage_response_json": False,
+        "first_stage_required_returns": None,
+        "second_stage_uri": "https://localhost",
+        "second_stage_method": "get",
+        "second_stage_headers": {},
+        "second_stage_req_template": "$INPUT",
+        "second_stage_response_json": False,
+        "second_stage_response_json_field": None,
+        "second_stage_req_template_json_object": None,
+        "ratelimit_codes": [429],
+        "request_timeout": 20,
+        "proxies": None,
+        "verify_ssl": True,
+    }
+
+    generator_family_name = "MultiEndpointREST"
+
+    _supported_params = (
+        "key_env_var",
+        "token_refresh_vars",
+        "first_stage_uri",
+        "first_stage_method",
+        "first_stage_headers",
+        "first_stage_req_template",
+        "first_stage_req_template_json_object",
+        "first_stage_response_json",
+        "first_stage_required_returns",
+        "second_stage_uri",
+        "second_stage_method",
+        "second_stage_headers",
+        "second_stage_req_template",
+        "second_stage_req_template_json_object",
+        "second_stage_response_json",
+        "second_stage_response_json_field",
+        "api_key",
+        "name",
+        "proxies",
+        "verify_ssl",
+        "ratelimit_codes",
+        "request_timeout",
+        "max_tokens",
+        "temperature",
+        "top_k",
+        "context_len"
+    )
+
+    def __init__(self, config_root=_config):
+        # Set up the self object
+        self.need_token_refresh = False
+        self.key_env_var = self.ENV_VAR if hasattr(self, "ENV_VAR") else None
+
+        self._load_config(config_root)
+
+        # Load the token refresh configuration
+        if hasattr(config_root.transient.cli_args, "token_refresh_config"):
+            self.token_refresh_path = config_root.transient.cli_args.token_refresh_config
+            with open(self.token_refresh_path, "r") as f:
+                self.token_refresh_config = json.load(f)
+            if not isinstance(self.token_refresh_config["method"], str):
+                raise ValueError("token_refresh_config set but does not contain method")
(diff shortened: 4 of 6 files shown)


garak/generators/multi_endpoint_rest.py
```diff
@@ -0,0 +1,458 @@
...
+                    "X-Authorization": "$KEY"
+                },
+                "get_req_template_json_object": {
+                    "text": "$INPUT"
+                },
+                "post_response_json": true,
+                "post_response_json_field": "text"
+            }
+        }
+    }
```
jmartin-tech, line 44: This example config does not match the class expectations.

``` json
    {
        "multi_endpoint_rest": {
            "MultiEndpointGenerator": {
                "name": "example service",
                "first_stage_uri": "https://example.ai/llm",
                "first_stage_headers": {
                    "X-Authorization": "$KEY"
                },
                "first_stage_req_template_json_object": {
                    "text": "$INPUT"
                },
                "first_stage_response_json": true,
                "first_stage_response_json_field": "job_id",
                "second_stage_uri": "https://example.ai/llm",
                "second_stage_headers": {
                    "X-Authorization": "$KEY"
                },
                "second_stage_req_template_json_object": {
                    "text": "$INPUT"
                },
                "second_stage_response_json": true,
                "second_stage_response_json_field": "text"
            }
        }
    }
```

`first_stage_` & `second_stage_` prefix are very generic would there be value in a more specific network concepts focused prefix like `send_` & `recv_`?

``` json
    {
        "multi_endpoint_rest": {
            "MultiEndpointGenerator": {
                "name": "example service",
                "send_uri": "https://example.ai/llm",
                "send_headers": {
                    "X-Authorization": "$KEY"
                },
```

Another option might be to organize the config in line with reuse of the existing rest generator.

``` json
    {
        "multi_endpoint_rest": {
            "MultiEndpointGenerator": {
                "first_stage": {                    
                    "name": "request service",
                    "uri": "https://example.ai/llm",
                    "method": "post",
                    "headers": {
                        "X-Authorization": "$KEY",
                    },
                    "req_template_json_object": {
                        "text": "$INPUT"
                    },
                    "response_json": true,
                    "response_json_field": "text"
                },
                "second_stage": {
                    "name": "response service",
                    "uri": "https://example.ai/llm",
                    "method": "post",
                    "headers": {
                        "X-Authorization": "$KEY",
                    },
                    "req_template_json_object": {
                        "text": "$INPUT"
                    },
                    "response_json": true,
                    "response_json_field": "text"
                }
            }
        }
    }
```

jmartin-tech (changes requested):
Thanks for the progress, this PR tries to accomplish a number of goals at once.

A number of items to consider refactoring:
* The new class seems to fit in the existing `rest`, the `rest.MultiEndpointGenerator` class can be specified on the command line while leaving the `RestGenerator` as default.
* Configuration specific to a single plugin should not be added as top level options.
  * Since the current implementation of `proxy` support is `rest` specific the config can be consolidated as part of the `generator_options_file`. The same applies to `verify_ssl`.
  * Configuration of token refresh as a separate json config file creates complexity for a feature that is not `globally` applicable.
* Please add an example of how all the new functionality configuration should be formatted, this could be something like an added `json` segment in the class description for when a `proxy` or `token_refresh` is to be utilized.
* Since the parameter names for each `stage` can be identified by `prefix` using the existing `rest.Generator` as a web request client can greatly reduce the code required. There is an offered example for possibly nesting the config for each endpoint to avoid the need for `prefix` values on the keys that might defer more of the code required.
* As implemented `verify_ssl` only applies to token refresh, if a `proxies` value is passed ssl control likely either needs to be `on/off` for all request or to have configurable values for each request made.

leondz (changes requested):
⠀

garak/generators/multi_endpoint_rest.py
leondz, line 1: What's this for? What's the use-case? Where can it be tested? Why should we integrate this and maintain it? Where are the docs (should be in .rst format under `docs/`)?

-- Not saying we shouldn't include this, but without understanding what's gained by this, it's hard to triage the value of this PR. Can we get some clarity around what this brings to the project and why we should add it to the set of things we're maintaining?

garak/resources/plugin_cache.json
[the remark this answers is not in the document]
leondz: @jmartin-tech Is it worth adding this to a project-level gitignore? Or do we want the cache to progress a little more before doing that?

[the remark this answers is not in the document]
jmartin-tech: We can just watch for now, accepting changes in PR's will not harm anything since automation will just rebuild it anyway.

au70ma70n:

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
au70ma70n: I can add it to the gitignore, I simply thought the community would benefit from having a devcontainer configuration

.gitignore
```diff
@@ -167,3 +167,4 @@ hitlog.*.jsonl
 garak_runs/
 runs/
 logs/
+.github/dependabot.yml
```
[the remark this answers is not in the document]
au70ma70n: sure I can move this

garak/cli.py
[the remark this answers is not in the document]
au70ma70n: The arguments were added as top level options to provide the opportunity to integrate the functionality into other parts of the project, however if you'd rather they be moved into the class configuration file that's simple enough.

garak/generators/multi_endpoint_rest.py
au70ma70n: The first and second stage variable names were intentionally generic for flexibility, as it can perform any two calls to any two systems. A post and delete for example, or two puts (one to make a change and the other to revert it).

garak/generators/multi_endpoint_rest.py
au70ma70n: > What's this for?
This is for supporting LLM application flows that require multiple endpoint interactions.

> What's the use case?
The use case is left intentionally generic to support as many use cases as possible, however some example use cases would be the following: 

- An LLM application requires a POST request to generate a response, and a GET request to fetch it
- Testing against an LLM application without permissions to create new responses, and using two PUT requests, one to change an initial conversation, the other to change it back
- Testing against an LLM application in production, and using an additional DELETE request to clean up old conversation histories.

> Where can it be tested?
> This could be tested against anything that you are able to interact with via a rest API. I've tested with ChatGPT, however I'm sure there are many other suitable candidates. 

> Where are the docs?
Docs have yet to be added as I suspected there would be many refactoring changes, as @jmartin-tech has pointed out.

> Can we get some clarity around what this brings to the project and why we should add it to the set of things we're maintaining?
In my opinion the main thing this PR brings to the project is flexibility. This functionality will be required for the testing of many internally developed LLM applications, which I suspect the project values. I know of several LLM applications across various companies that were developed internally and require multiple api interactions across various endpoints to use. This PR is primarily  designed to address those applications.

jmartin-tech: > This functionality will be required for the testing of many internally developed LLM applications, which I suspect the project values. I know of several LLM applications across various companies that were developed internally and require multiple api interactions across various endpoints to use.

I think the primary question here is *how* and *should* the project support something that does not have a public example?  Anything accepted into a release has to be maintained to some extent. If there is no public reference implementation, it can get very difficult support consumers as reproduction of issues becomes a high effort activity.

au70ma70n: The project claims to be "nmap for LLMs." If that's the case, then yes the project should support features required by industry professionals.

jmartin-tech: The question posed is reasonable, how can the project establish the value of supporting this access pattern? Is there a reference example where this is actually encountered in a professional setting that the project can use for testing?

The answers can be `I don't know` or `Nothing that can be publicly shared`, simply having information about patterns seen in the wild can influence value.

If there is an available tool or application that follows this pattern for access to inference, this generator has more value and more reason to be maintained and supported.

Supporting the [three use cases](https://github.com/leondz/garak/pull/878#discussion_r1745925512) you provided may meet the criteria, the second and third especially sound interesting and examples of applications that have these patterns may even help this project and it's community further research and build guidance on using this type of access pattern.

Note the third use case is not yet supported. As implemented the second stage response needs to contain the result of generation, I can see it as viable that a DETELE request may not return the data deleted which leads me to suggest there may be another configuration option needed provide a mechanism to return a value from the first stage as the `result` of the attempt to be evaluated by detectors, and obtain another value from the response to use as the identifier posted in the second stage to remove the history. I am not suggesting this would be required to land this PR, only that this factors into the equation.

au70ma70n: >Is there a reference example where this is actually encountered in a professional setting that the project can use for testing?
The answer is yes, I developed this PR specifically for the deployment of garak in an internal, professional setting. Unfortunately the details of the application cannot be publicly shared. Also yes, the third use case is not currently supported. In the future I'd like to see this functionality become more modular, supporting n number of requests and containing configuration options for how to get and return the response data.

leondz: Huh, OK. Something we have on the roadmap is for custom plugins to go in user-local custom dirs instead of in the garak package directory. This might be a sensible route for developing this. 

As maintainers, when we merge a PR, we agree to take on the technical debt for the concepts and features in that PR /for the lifetime of the project/. As humans, our time and bandwidth is finite, so each new externally-planned feature means a permanent degradation in our ability to steer and progress the project, and a permanent reduction in the velocity we have on our roadmap. So I hope you can appreciate how seriously we take this sort of thing, especially when it's an agreement to support a target that is invisible to us (and all the other users). 

For example, because we cannot test things we can't observe, architectural changes are prone to breaking functions like this, even despite our best efforts: we won't have "real" tests. Given that, what kind of SLA would you expect if an architectural change silently broke this plugin? What would be fair? What would be the incentive from either side to engage in that feature/project at all?

au70ma70n: Architectural changes within garak or externally? I wouldn't expect external architectural changes to affect the garak project, however things like changing how responses are processed for example would. I think a reasonable SLA would be maintaining the data flow into and out of the generator, and I would expect the community to patch the generator to fit architectural changes to their target applications.

leondz:
I mean changes in garak architecture that cause alterations in how
high-level components operate, pretty day-to-day stuff in a software
project.

jmartin-tech:
Circling back here, given the number of changes requested could this be separated to isolate the various changes?

I can see the following possible independent PRs:
* enable `restGenerator` to accept `verify_ssl`
  * Allow suppression of `ssl` certificate validation when executing https based requests
  * Default `verify_ssl: True`
* enable `restGenerator` http proxy support
  * Allow configuration of `proxies`
  * Include validation that a dictionary is provided and values are valid `uri` entries, "Dictionary mapping protocol to the URL of the proxy" format example: `{ "https": "https:10.10.1.1:8443" }`
* enable `restGenerator` token refresh
  * consider support generic endpoint value request
  * (optional) consider support for template and value extraction
  * (optional) support a specific token generation pattern such as `oath2`
* enable multi_endpoint generation
  * accept a series of `restGenerator` configuration stages
  * support selection of `stage` to use for generator result
  * support usage of response from a previous stage as `$INPUT` value of a stage

jselvi:
@jmartin-tech I'm interested in the restGenerator proxy support mentioned in this PR, but I haven't seen it merged. Is this PR actually open? I was thinking on implementing it and send my own PR, but I'll wait if this one is going to be completed and merged in the short term. Thanks!

jmartin-tech:
I am looking into extracting the `verify_ssl` and `proxies` support this week while maintaining attribution to @au70ma70n as the original author. Once opened the PRs will reference this original work. Testing, once posted, may take some time for `proxies` to land in `main`.

jmartin-tech:
Closing for now, the proxy support and TLS/SSL validation suppression for `rest` generators landed independently.

The token refresh looks interesting though likely needs some more work.
Multi-endpoint patterns are still an interesting use case and may be considered again in the future.

PR #896, opened 2024-09-04
generator: Groq API

mmilenkovic-groq:
Add support for LLMs hosted via Groq's OpenAI-compatible API.

### DIFF
--- a/tests/generators/test_groq.py
+++ b/tests/generators/test_groq.py
@@ -25,7 +25,7 @@ def test_groq_generate_1():
     result = g._call_model("this is a test", generations_this_call=1)
     assert isinstance(result, list), "GroqChat _call_model should return a list"
     assert len(result) == 1, "GroqChat _call_model result list should have one item"
-    assert isinstance(result[0], str), "GroqChat _call_model should return a list"
+    assert isinstance(result[0], str), "GroqChat generate() should contain a str"
     result = g.generate("this is a test", generations_this_call=1)
     assert isinstance(result, list), "GroqChat generate() should return a list"
     assert (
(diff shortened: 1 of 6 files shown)


mmilenkovic-groq:
recheck

leondz:
Thank you! Will take a look.

tests/generators/test_groq.py
```diff
@@ -0,0 +1,31 @@
...
+@pytest.mark.skipif(
+    os.getenv(GroqChat.ENV_VAR, None) is None,
+    reason=f"GroqChat  API key is not set in {GroqChat.ENV_VAR}",
+)
+def test_groq_generate_1():
+    g = GroqChat(name="llama3-8b-8192")
+    result = g._call_model("this is a test", generations_this_call=1)
+    assert isinstance(result, list), "GroqChat _call_model should return a list"
+    assert len(result) == 1, "GroqChat _call_model result list should have one item"
+    assert isinstance(result[0], str), "GroqChat _call_model should return a list"
```
jmartin-tech, line 25:
```suggestion
    assert isinstance(result[0], str), "GroqChat generate() should contain a str"
```

tests/generators/test_groq.py
jmartin-tech, line 1: Currently no tests would every be run in automation. Can these test be expanded to provide a `mock` response as well a response?

See: `tests/generators/test_openai.py` and `tests/generators/conftest.py` for existing patterns that should be viable here.

jmartin-tech (changes requested):
Pretty straight forward, minor adjustments requested.

tests/generators/test_groq.py
mmilenkovic-groq: I've added GroqChat to the openai-compatible tests to avoid duplication and I've added an extra test which will run in automation. Let me know if you want to add anything else here as well

jmartin-tech (approved):

PR #898, opened 2024-09-05
paraphrase fast consistent model device

jmartin-tech:
When running tests on a platform with a `cuda` capable device, errors can be reported due to how the model is instantiated.

```
>       return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
E       RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument index in method wrapper_CUDA__index_select)

../../../miniconda3/envs/garak/lib/python3.12/site-packages/torch/nn/functional.py:2267: RuntimeError
---------------------------------------- Captured stdout call -----------------------------------------
🦾 loading buff: paraphrase.Fast
======================================= short test summary info =======================================
FAILED tests/buffs/test_buffs.py::test_buff_load_and_transform[buffs.paraphrase.Fast] - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:...
======================== 1 failed, 1771 passed, 32 skipped in 90.16s (0:01:30) ========================
```

Recent changes to support a configuration of plugins and huggingface models as used in this class lead to further identification of possible exceptions when processing a `HFCompatible` class that does not have `hf_args` or when providing options for huggingface model constructors that favor the newer `device_map` argument to `device`.

This PR also offers more visibility into a the coupling that exists between various plugin types that consume the `generator` plugin type. While not addressed in this PR, a future iteration should review module structure to determine if there are ways to limit or guide coupling of different plugin types. So far only `generator` types plugins are consumed by other plugin types primarily for local huggingface model utilization.

### DIFF
--- a/garak/buffs/paraphrase.py
+++ b/garak/buffs/paraphrase.py
@@ -74,13 +74,13 @@ class Fast(Buff, HFCompatible):
     """CPU-friendly paraphrase buff based on Humarin's T5 paraphraser"""
 
     DEFAULT_PARAMS = Buff.DEFAULT_PARAMS | {
-        "hf_args": {"device": "cpu", "torch_dtype": "float32"}
+        "para_model_name": "garak-llm/chatgpt_paraphraser_on_T5_base",
+        "hf_args": {"device": "cpu", "torch_dtype": "float32"},
     }
     bcp47 = "en"
     doc_uri = "https://huggingface.co/humarin/chatgpt_paraphraser_on_T5_base"
 
     def __init__(self, config_root=_config) -> None:
-        self.para_model_name = "garak-llm/chatgpt_paraphraser_on_T5_base"
         self.num_beams = 5
         self.num_beam_groups = 5
         self.num_return_sequences = 5
--- a/garak/buffs/paraphrase.py
+++ b/garak/buffs/paraphrase.py
@@ -7,6 +7,7 @@ from collections.abc import Iterable
 
 import garak.attempt
 from garak import _config
+from garak.generators.huggingface import HFCompatible
 from garak.buffs.base import Buff
 
 
@@ -69,9 +70,12 @@ class PegasusT5(Buff):
             yield paraphrased_attempt
 
 
-class Fast(Buff):
+class Fast(Buff, HFCompatible):
     """CPU-friendly paraphrase buff based on Humarin's T5 paraphraser"""
 
+    DEFAULT_PARAMS = Buff.DEFAULT_PARAMS | {
+        "hf_args": {"device": "cpu", "torch_dtype": "float32"}
+    }
     bcp47 = "en"
     doc_uri = "https://huggingface.co/humarin/chatgpt_paraphraser_on_T5_base"
 
@@ -85,20 +89,23 @@ class Fast(Buff):
         self.no_repeat_ngram_size = 2
         # self.temperature = 0.7
         self.max_length = 128
-        self.torch_device = None
+        self.device = None
         self.tokenizer = None
         self.para_model = None
         super().__init__(config_root=config_root)
 
     def _load_model(self):
-        import torch
         from transformers import AutoTokenizer, AutoModelForSeq2SeqLM
 
-        self.torch_device = "cuda" if torch.cuda.is_available() else "cpu"
-        self.tokenizer = AutoTokenizer.from_pretrained(self.para_model_name)
+        self.device = self._select_hf_device()
+        model_kwargs = self._gather_hf_params(
+            hf_constructor=AutoModelForSeq2SeqLM.from_pretrained
+        )  # will defer to device_map if device map was `auto` may not match self.device
+
         self.para_model = AutoModelForSeq2SeqLM.from_pretrained(
-            self.para_model_name
-        ).to(self.torch_device)
+            self.para_model_name, **model_kwargs
+        ).to(self.device)
+        self.tokenizer = AutoTokenizer.from_pretrained(self.para_model_name)
 
     def _get_response(self, input_text):
         if self.para_model is None:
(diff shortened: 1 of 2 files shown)


garak/buffs/paraphrase.py
```diff
@@ -69,9 +70,12 @@ def transform(
             yield paraphrased_attempt
 
 
-class Fast(Buff):
+class Fast(Buff, HFCompatible):
     """CPU-friendly paraphrase buff based on Humarin's T5 paraphraser"""
 
+    DEFAULT_PARAMS = Buff.DEFAULT_PARAMS | {
+        "hf_args": {"device": "cpu", "torch_dtype": "float32"}
+    }
```
jmartin-tech, line 78: Should this be refactored to match pattern used in `atkgen`?
```suggestion
    DEFAULT_PARAMS = Buff.DEFAULT_PARAMS | {
        "para_model_name": "garak-llm/chatgpt_paraphraser_on_T5_base",
        "para_model_config": {
            "hf_args": {"device": "cpu", "torch_dtype": "float32"},
        }
    }
```

The model instantiates using `AutoModelForSeq2SeqLM` so I am not sure if a `para_model_type` make sense as `generators.huggingface.*` does not provide a class I think can be substituted for this currently and may require a more abstract rethink of how this buff is applied.

garak/generators/huggingface.py
```diff
@@ -53,10 +53,17 @@ def _set_hf_context_len(self, config):
...
-        if params["device"] is None:
+        """ "Identify arguments that impact huggingface transformers resources and behavior"""
+
+        # this may be a bit too naive as it will pass any parameter valid for the hf_constructor signature
+        # this falls over when passed some `from_pretrained` methods as the callable model params are not always explicit
+        params = (
+            self.hf_args
+            if hasattr(self, "hf_args") and isinstance(self.hf_args, dict)
+            else {}
+        )
```
erickgalinkin, line 64: Do you have an example of where it falls over using `from_pretrained`? We could probably catch some of those.

erickgalinkin (approved):
I'm largely into this, though I suspect we could lean on HF a bit more heavily to validate args and let the config be passed to the constructor a bit more directly.

leondz:
I don't think we make explicit our strategy for allocating devices, across generators, probes, detectors, and buffs. I don't think we even have one. It's a non-trivial problem, and now that we make some decisions in some places - instead of leaving it all to the user - we should now make that strategy explicit, else I think we'll keep running in to this kind of thing.

garak/buffs/paraphrase.py
jmartin-tech: The comment was a but of reflection on previous expectations and discussion, by exposing the `name` and `type` we gain flexibility and some consistency with `cli` where objects that instantiate a model require a `name + type + config` combination in a similar format.

As we work on documenting the config patterns better maybe this will become more clear on where consistency is useful vs not.

garak/generators/huggingface.py
jmartin-tech: I will dig for an example, the comment above was due to some `from_pretrained()` methods being for the model and others for the tokenizer or other objects, and they take vastly different sets or config. Currently this method is only called to gather args for `transformers.AutoConfig.from_pretrained()`, `transformers.pipeline()`, and `transformers.LlavaNextForConditionalGeneration.from_pretrained()` initialization and we don't have a lot of examples of expanding beyond `device`, `device_map`, & `touch_dtype` as parameters to be passed on.

Also note the comment is not about this conditional, it is a general comment about the technique employed in this method possibly being brittle although I have not seen that much in practice yet, only during my iteration to get this technique working in the first place.

garak/buffs/paraphrase.py
```diff
@@ -85,20 +89,23 @@ def __init__(self, config_root=_config) -> None:
...
+        model_kwargs = self._gather_hf_params(
+            hf_constructor=AutoModelForSeq2SeqLM.from_pretrained
+        )  # will defer to device_map if device map was `auto` may not match self.device
+
         self.para_model = AutoModelForSeq2SeqLM.from_pretrained(
-            self.para_model_name
-        ).to(self.torch_device)
+            self.para_model_name, **model_kwargs
+        ).to(self.device)
+        self.tokenizer = AutoTokenizer.from_pretrained(self.para_model_name)
```
leondz, line 108: this is nice

garak/buffs/paraphrase.py
leondz: the model should be exposed as a config param somewhere and I somewhat prefer that to be a value rather than a key

garak/generators/huggingface.py
```diff
@@ -96,6 +103,15 @@ def _gather_hf_params(self, hf_constructor: Callable):
...
 
+        if (
+            not "device_map" in args
+            and "device_map" in params_to_process
+            and "device" in params_to_process
+            and "device" in args
+        ):
+            del args["device"]
+            args["device_map"] = self.device
+
```
leondz, line 114: we're going to want to factor this kind of logic out and up somewhere one day. not in this pr though.

leondz (approved):

garak/buffs/paraphrase.py
jmartin-tech:
```suggestion
    DEFAULT_PARAMS = Buff.DEFAULT_PARAMS | {
        "para_model_name": "garak-llm/chatgpt_paraphraser_on_T5_base",
        "hf_args": {"device": "cpu", "torch_dtype": "float32"}
    }
```

How about we split the difference and expose the model name but retain hf config map at top level inferring this class only supports huggingface models at this time, if we expand support later we can consider it a breaking change for config of this specific class.

jmartin-tech: This is consistent with `HFDetectors`. Going this route for this iteration and we can revisit later.

Issue #883, opened 2024-09-04
docs: configurable


leondz:
Add documentation for the Configurable construct, including examples for a variety of plugin types. Make sure RESTGenerator is in there.

PR #906, opened 2024-09-06
plugin classes should not access `DEFAULT_PARAMS`

jmartin-tech:
`DEFAULT_PARAMS` are projected onto instances by `Configurable` and should not be directly access from the `dict` CONSTANT in class methods. If the user provided configurtion that overrides the default the user value is found on the instance.

### DIFF
--- a/garak/detectors/packagehallucination.py
+++ b/garak/detectors/packagehallucination.py
@@ -44,9 +44,9 @@ class PackageHallucinationDetector(Detector):
         import stdlibs
 
         logging.debug(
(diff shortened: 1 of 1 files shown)


leondz (approved):
good catch, thanks for tagging me to review this.

hoping that addressing #883 will reduce the chances of things coming in.

i would like one day to get test warnings/failures for things like:
* commits that introduce print() stmts
* mentions of DEFAULT_PARAMS in functions in plugins

PR #907, opened 2024-09-06
Force cache build in automation

jmartin-tech:
If no existing cache is found for a test run as would be the case for test automation, request a cache instance. This will cause the cache to build once if any test is run by `pytest`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,10 @@
 import pytest
 import os
-from garak import _config
+from garak import _config, _plugins
+
+# force a local cache file to exist when this top level import is loaded
+if not os.path.isfile(_plugins.PluginCache._user_plugin_cache_filename):
+    _plugins.PluginCache.instance()
 
 
 @pytest.fixture(autouse=True)


tests/conftest.py
```diff
@@ -1,6 +1,10 @@
 import pytest
 import os
-from garak import _config
+from garak import _config, _plugins
+
+# force a local cache file to exist when this top level import is loaded
+if not os.path.isfile(_plugins.PluginCache._user_plugin_cache_filename):
+    _plugins.PluginCache.instance()
```
leondz, line 7: Does this not trigger [_plugins.py#L63](https://github.com/jmartin-tech/garak/blob/9438858605634bcb915e66368e5e70f817566762/garak/_plugins.py#L63) ?


```
    def _load_plugin_cache(self):
        assert os.path.exists(
            self._plugin_cache_filename
        ), f"{self._plugin_cache_filename} is missing or corrupt."
```

leondz (commented):
One question re: if this works as intended

PR #911, opened 2024-09-11
Document configuration of garak using YAML

### LINKED ISSUE #883: docs: configurable

leondz:
Resolves #883

Documentation for configuring garak using YAML, including user docs for `Configurable` and `RestGenerator`


docs/source/configurable.rst
```diff
@@ -0,0 +1,260 @@
...
+
+
+Specifying custom configuration
+-------------------------------
+
+Garak can be configured in multiple ways:
+
+* Via command-line parameters
+* Using YAML configs
+* Through specifying JSON on the command line
```
jmartin-tech, line 16: The cli json format is not covered here, consider link to the rest documentation json for format of
`--*_option_file` passed values or and added section after the `yaml` details.

Maybe a section `Using a custom json config`?

docs/source/index.rst
```diff
@@ -51,11 +55,16 @@ Code reference
...
+   contributing
+
    basic
    attempt
    cli
    command
+   exception
    interactive
+   _config
+   _plugins
```
jmartin-tech, line 67: Exposing these classes here suggests to me that their interfaces are consumable by custom plugin developers, is there some way we can tag these with a note that these are internal only classes with no guarantee of a stable API?

leondz: added in rst file for each of these classes

PR #916, opened 2024-09-17
data file override support with precedence

### LINKED ISSUE #891: enable loading resources from XDG_DATA_HOME before hitting package dir
Enable loading probe resources (e.g. custom prompts) from XDG_DATA_HOME first, package dir second

Should be for 
* most hits to `resources` dir
* `payloads`

jmartin-tech:
Fix #891 

* Provides a custom path accessor for project `data` files to allow user's to override project provided datasets using `$XDG_DATA_DIR/garak/data` to mirror `garak/data` from the package.
* Separates non-code data or dataset files from source code
* Downloaded datasets are considered `cached files` and do not change location or behavior in the revision
* Removes argument parser no longer supported by `gcg`

Consumers of the `garak.data.path` object can traverse the data directory as a standard `Path` object and will be returned existing files based on the order precedence from the supported locations on disk.

A possible drawback to the current implementation exists regarding directory interaction, if a folder exists at more than one path location the higher precedence path will be returned, when performing `glob` or iteration actions with such an instance the only files in the selected path on disk will be evaluated.  This is currently considered an acceptable limitation.

## Verification

- [ ] `garak -m nim -p donotanswer.MisinformationHarms -g 1 -n meta/llama3-70b-instruct --parallel_attempts 5`
- [ ] create a limited data set for `donotanswer.MisinformationHarms`
```
mkdir -p ~/.local/share/garak/data/donotanswer
head -n 10 garak/data/donotanswer/misinformation_harms.txt > ~/.local/share/garak/data/donotanswer/misinformation_harms.txt
```
- [ ] `garak -m nim -p donotanswer.MisinformationHarms -g 1 -n meta/llama3-70b-instruct --parallel_attempts 5`
- [ ] **Verify** run results:
  - [ ] the first run reported `155` attempts
  - [ ] the second run report `10` attempts (count depends `-n` count passed to `head`)

Note: #870 provides new `data` files that will need to migrate to this pattern depending on PR landing order.


jmartin-tech:
Looks like there is some work to do with support for this on [python 3.10](https://stackoverflow.com/questions/61689391/error-with-simple-subclassing-of-pathlib-path-no-flavour-attribute) apparently!

leondz:
> A possible drawback to the current implementation exists regarding directory interaction, if a folder exists at more than one path location the higher precedence path will be returned, when performing glob or iteration actions with such an instance the only files in the selected path on disk will be evaluated. This is currently considered an acceptable limitation.

Flagging a tension here with payload enumeration. Currently payload enum globs but uses its own directory traversal code, rather than `garak.data`. The result is that this PR doesn't break payload enum, but does leave us with two bits of code using similar-but-not-identical logic.

garak/data/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+    def joinpath(self, *pathsegments):
+
+        for segment in pathsegments:
+            prefix_removed = None
+            for path in self.ORDERED_SEARCH_PATHS:
+                if (path == self and segment != "..") or path in self.parents:
+                    prefix_removed = self.relative_to(path)
+                    break
+            if prefix_removed is None:
+                raise GarakException(
```
leondz, line 46: Why not `FileNotFoundError`?

garak/data/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+                )
+            for path in self.ORDERED_SEARCH_PATHS:
+                if segment == "..":
+                    projected = (path / prefix_removed).parent
+                else:
+                    projected = (path / prefix_removed).joinpath(segment)
+                if projected.exists():
+                    return LocalDataPath(projected)
+
+        raise GarakException(f"The resource requested does not exist {segment}")
```
leondz, line 57: Why not `FileNotFoundError`?

garak/probes/topic.py
```diff
@@ -92,7 +92,7 @@ def _gen_prompts(self, term):
     def __init__(self, config_root=_config):
         super().__init__(config_root)
 
-        self.data_dir = _config.transient.cache_dir / "resources" / "wn"
+        self.data_dir = _config.transient.cache_dir / "data" / "wn"
```
leondz, line 95: What's our plan for cache dir structure?

tests/test_data.py
```diff
@@ -0,0 +1,68 @@
...
+def test_local_override(random_resource_filename):
+    source = data_path / random_resource_filename
+    assert _config.transient.package_dir in source.parents
+
+    data_root_path = _config.transient.data_dir / "data"
+    data_root_path.mkdir(parents=True, exist_ok=True)
+    with open(
+        data_root_path / random_resource_filename, encoding="utf-8", mode="w"
+    ) as f:
+        f.write("fake data")
```
leondz, line 65: What expectation is being tested here?

tools/termscrape.py
```diff
@@ -21,8 +22,8 @@ def scrape_search_results(keyphrase):
         yield example
 
 
-with open("slurprompts.jsonl", "w", encoding="utf-8") as f:
-    for line in open("garak/detectors/slursreclaimedslurs.txt", "r", encoding="utf-8"):
+with open(data_path / "slurprompts.jsonl", "w", encoding="utf-8") as f:
```
leondz, line 25: Since this is writing, perhaps it should be sent to cache and not data. Actually, perhaps it's worth retiring this script from the repo.

leondz (commented):
A few questions, looks good so far. Will be some extra work involved in aligning the payloads PR to the proposals here. The general separation is for non-code that can also be overridden should move into data; the rest remains in resources.

garak/data/__init__.py
erickgalinkin: Maybe it makes sense to add a garak-specific FNFE to `garak.exceptions`? `GarakPathNotFound` or something?

garak/detectors/riskywords.py
```diff
@@ -7,11 +7,12 @@
 import csv
 
 from garak import _config
+from garak.data import path as data_path
 from garak.detectors.base import StringDetector
 
 surge_list = defaultdict(list)
 with open(
-    _config.transient.package_dir / "resources" / "profanity_en.csv",
+    data_path / "profanity_en.csv",
```
erickgalinkin, line 15: Probably worth looking at as part of a larger refactor -- do we want all these datasets in our git? Or should we have them in some other place _e.g._ HuggingFace hub, and the `garak.data` module can also manage downloading these files?

garak/resources/common.py
```diff
@@ -43,19 +45,25 @@
...
-        / "harmful_behaviors.csv"
-    )
+    try:
+        advbench_base_path = data_path / "advbench" / "harmful_behaviors.csv"
+    except GarakException:
+        logging.info(
+            "No default harmful_behaviors.csv found download or cached file will be utilized."
+        )
+
+    # this utilizes the cache_dir however should this place the file in the data_dir now?
```
erickgalinkin, line 55: This sounds right -- the file should be downloaded to `data_path / "advbench" / "harmful_behaviors.csv"`. We can probably do away with anything below. Just use `advbench_base_path` -> `advbench_path` and use that everywhere. Once you have it, we should be referencing it in a single location.

garak/data/__init__.py
jmartin-tech: For this error, the actual cause here is more akin to a `ValueError` as the raise is due to a relative path traversal  that is not allowed even if it does exists.

I used a `garak` specific exceptions as consumers of these objects `know` their file exists. To me this means an exception from this method indicates something else `caused` the file to be `removed`, which is different than `not found` IMO.

garak/data/__init__.py
jmartin-tech: I used a `garak` specific exceptions as consumers of these objects `know` their file exists. To me this means an exception from this method indicates something else `caused` the file to be `removed`, which is different than `not found` IMO.

garak/resources/tap/tap_main.py
```diff
@@ -35,11 +35,7 @@
 SAVE_RESULTS = True
 
 resources_tap_data_file = (
-    garak._config.transient.cache_dir
-    / "resources"
-    / "tap"
-    / "data"
-    / "tap_jailbreaks.txt"
+    garak._config.transient.cache_dir / "data" / "tap" / "tap_jailbreaks.txt"
```
erickgalinkin, line 38: Since `probes.tap` references the `prompts_location` as `data_path / "tap" / "tap_jailbreaks.txt"`, do we want this to reflect that as well? The idea is to keep your TAP results such that you can use them (or share them with us!) in subsequent probes. Though perhaps that's a larger issue.

garak/probes/visual_jailbreak.py
```diff
@@ -41,15 +42,14 @@ class FigStep(Probe):
...
-    )
+    safebench_image_catalog = data_path / "safebench_filenames.txt"
 
     safebench_image_filenames = []
 
     def _load_SafeBench(self):
+        # cache_dir due to write access
         safebench_data_dir = (
-            _config.transient.cache_dir / "resources" / "visual_jailbreak" / "SafeBench"
+            _config.transient.cache_dir / "data" / "visual_jailbreak" / "SafeBench"
```
jmartin-tech, line 52: Here is a case where I think the cached file is probe specific and pathing reflects the intent.

garak/resources/common.py
```diff
@@ -43,19 +45,25 @@
...
-        / "resources"
-        / "advbench"
-        / "harmful_behaviors.csv"
-    )
+    try:
+        advbench_base_path = data_path / "advbench" / "harmful_behaviors.csv"
+    except GarakException:
+        logging.info(
+            "No default harmful_behaviors.csv found download or cached file will be utilized."
+        )
```
jmartin-tech, line 53: This provides an example for a case where the tooling would allow override of a resource not shipped with `garak` by placing a file in the `data` path using a known location.

garak/probes/topic.py
jmartin-tech: There is an example [here](https://github.com/leondz/garak/pull/916/files#diff-d0e6a61922e5a549cc4c628aabf0b38c7c8c7c707a83a4f367ca511cd46e609bR56-R61) where `data` prefix is used already for a `shared` file to maintain consistency, maybe this should stay `_config.transient.cache_dir / "data" / "wn"` and other locations that use `cache_dir` to store in a `tool` namespace should place that namespace under `_config.transient.cache_dir / "data"`.

tests/test_data.py
jmartin-tech: This test proves out the order or precedence.

The test verifies a file will be found from the `package_dir` when it does not exist in the then creates a file that will override that location in the `data_dir` and verifies a new request for that filename will return the `data_dir` file.

I could expand the test to read and compare file contexts if that will make the intent more clear.

tools/termscrape.py
jmartin-tech: Glad you noticed something in this file, I needed to flag this file is moved to `tools` and need to validate that `tools` will not be picked up in the `pypi` packaging.

To answer the original question, this tool seemed targeted as a maintainer script that could be used to update the source `data` files.  Since the file should already exist in the repo this external tool `writing` to the file would not violate the `LocalDataPath` expectations that files accessed during `garak` runtime are treated as read only.

jmartin-tech:
> Flagging a tension here with payload enumeration. Currently payload enum globs but uses its own directory traversal code, rather than garak.data. The result is that this PR doesn't break payload enum, but does leave us with two bits of code using similar-but-not-identical logic.

Depending on land order I am thinking this can be a use case to improve `glob` support on the custom `Path` objects to return a consolidated list of files based on searching all `ORDERED_SEARCH_PATHS`. This would then remove the `similar-but-not-identical logic` by supporting the need quickly as a next iteration.

garak/detectors/riskywords.py
jmartin-tech: Have discussed in the past that we may want tooling that will download all datasets to create an `offline` deployment capability.

I suspect there could be some expansion on `garak.data` for handling access to known datasets. Treating HF as the specific location that tooling can register data as available from might be a good direction.

garak/probes/topic.py
```diff
@@ -92,7 +92,8 @@ def _gen_prompts(self, term):
...
+        self.data_dir.parent.mkdir(mode=0o740, parents=True, exist_ok=True)
```
jmartin-tech, line 96: Interesting item here, the `parent` must exist before download however if the `wn` folder already exists the raised when attempting access below will be a different exception type and fails to recover.

This was `masked` previously by the fact that `_config.transient.cache_dir / "resources"` is created in early initialization for the `plugin_cache.json` file.

garak/probes/topic.py
leondz: so far in `cache` I see:

* plugin cache
* wn
* visual_jailbreak data

It seems advantageous to be able to share artefacts across plugins, so I disprefer the idea of having dirs named based on single plugins -- in general (visual jailbreak refers also to the dataset name so no foul there)

I also dislike using the root for new artefacts directly (but this seems a pretty bland opinion :) )

I can also see that we have content that's used by plugins - currently all datasets - and content that's used by garak "core"

So we might like two top level dirs (not dead set on the names, not at all)

* `core` - things like plugin cache
* `datasets` - uh, datasets (`wn`, `visual_jailbreak`)

garak/resources/common.py
jmartin-tech: Thanks, due to this comment, I think I see a bug here as well since as coded `advbench_path` is only defined if the file existed in data already access to `advbench_path.is_file()` would raise exceptions.

At this time I am thinking of `data_path` as being `read_only` for runtime code, meaning anything downloaded is currently place in `cache_dir`.  This class currently exposes the one location where support for user override instead of download is being enabled and I am not sure yet if downloads should mix data in the a user provided override path.

I will leave the comment and fix the guard for this PR.

garak/probes/topic.py
leondz: `nltk_data` i think might benefit from having its own handler; it's also used standalone. Maybe defer choice on this until we move the nltk, HF handlers out to their own place (e.g. `resources`)

garak/probes/topic.py
jmartin-tech: > Where do we want to check for / make XDG dirs?
https://github.com/leondz/garak/blob/0e1d96e3892e14564aaece56c1ee9f92cfdd7dc0/garak/_config.py#L68-L71

`parents=True` is the equivalent to `-p`.

garak/resources/tap/tap_main.py
jmartin-tech: I agree these should be captured better for re-use. I will defer that change to a future iteration though. I think there is a larger issue to address as we have a few techniques that generate these intermediate files that could be valuable from a research context.

garak/payloads.py
```diff
@@ -55,11 +53,9 @@ def load_payload(
     if path is not None:
         return PayloadGroup(name, path)
     else:
-        # iterate through search dirs
-        for dir in PAYLOAD_SEARCH_DIRS:
-            path = pathlib.Path(dir) / f"{name}.json"
-            if path.is_file():
-                return PayloadGroup(name, path)
+        path = PAYLOAD_DIR / f"{name}.json"
```
jmartin-tech, line 56: Utilizing the data_path object here will cause this to raise `GarakException` when the file is not found, as reflected in the test updates, and avoids the `FileNotFoundError` raised below.  Does it makes sense to allow the exception to bubble up or should the callers of `load_payload` be expected to handle the revised exception?

The error message on the exception will be `The resource requested does not exist {name}.json`.

I suspect the answer is that `Director.load()` should handle the `GarakException` with a log entry and raise `PayloadFailure` from that response.

jmartin-tech: Updated in 958ea3c

garak/probes/topic.py
leondz: Thanks, missed that. This looks good. Expect to move `wn` code to its own module/class once we know where those will live - did we settle on `resources`? Other options could be `apis`, `external`, `apps`, `common`, & _not_ `lib`. Or is this a conversation to conclude once this PR lands?

leondz (approved):
lgtm

garak/payloads.py
leondz: Noting that the rglob in `payloads.Director` might conflict with other assumptions about payloads being file entries in `PAYLOAD_DIR`

leondz:
NICE

PR #922, opened 2024-09-20
generators: update `openai` for o1 models

leondz:
The OpenAI "o1" reasoning models have different API expectations from other Chat API models, in terms of params and defaults, and so a new class is added to account for these, and some switching hints placed in the code.

## Verification

* List the steps needed to make sure this thing works

```
export OPENAI_API_KEY=
python -m pytest tests/generators/test_openai.py
python -m garak -m openai -n o1-mini -p test.Test -g1
python -m garak -m openai.OpenAIReasoningGenerator -n o1-mini -p test.Test -g1
python -m garak -m openai.OpenAIReasoningGenerator -n o1-mini -p test.Test --parallel_requests 10
```


garak/generators/openai.py
```diff
@@ -282,4 +314,27 @@ def __init__(self, name="", config_root=_config):
...
+        "top_p": 1.0,
+        "frequency_penalty": 0.0,
+        "presence_penalty": 0.0,
+        "seed": None,
+        "stop": ["#", ";"],
+        "suppressed_params": set(["n", "temperature", "max_tokens", "stop"]),
+        "retry_json": True,
+        "custom_params": {
+            "max_completion_tokens": 1500,
+        },
```
jmartin-tech, line 332: Based on converting to checking signature of `create` for accepted params.
```suggestion
        "max_completion_tokens": 1500,
```

garak/generators/openai.py
```diff
@@ -106,6 +127,7 @@ class OpenAICompatible(Generator):
         "stop": ["#", ";"],
         "suppressed_params": set(),
         "retry_json": True,
+        "custom_params": {},
```
jmartin-tech, line 130: Based on suggestion to check signature of `create` for params.
```suggestion
```

garak/generators/openai.py
```diff
@@ -282,4 +314,27 @@ def __init__(self, name="", config_root=_config):
...
+        "suppressed_params": set(["n", "temperature", "max_tokens", "stop"]),
+        "retry_json": True,
+        "custom_params": {
+            "max_completion_tokens": 1500,
+        },
+    }
+
+    def __init__(self, name="", config_root=_config):
+        self.name = name
+        super().__init__(self.name, config_root=config_root)
```
jmartin-tech, line 337: This class extends `OpenAIGenerator` which sets `self.name` as the first action I think this can be removed.

tests/generators/test_openai.py
```diff
@@ -90,3 +90,10 @@ def test_openai_chat():
     for item in output:
         assert isinstance(item, str)
     print("test passed!")
+
+
+def test_reasoning_switch():
```
jmartin-tech, line 95:
```suggestion
@pytest.mark.usefixtures("set_fake_env")
def test_reasoning_switch():
```

jmartin-tech (changes requested):
Suggested refactor is untested, let me know if the pattern does not seem reasonable.

jmartin-tech:
To support `o1` model this also needs to update `requirements.txt` and `pyproject.toml` to `openai>=1.45.0,<2` as the new `max_completion_tokens` parameter was added in [1.45.0](https://github.com/openai/openai-python/releases/tag/v1.45.0)

garak/generators/openai.py
```diff
@@ -264,11 +290,17 @@ def _load_client(self):
...
+
         else:
             raise ValueError(
                 f"No {self.generator_family_name} API defined for '{self.name}' in generators/openai.py - please add one!"
             )
 
+        if self.__class__.__name__ == "OpenAIGenerator" and self.name.startswith("o1-"):
+            msg = "o1 models should use openai.ReasoningGenerator"
+            logging.error(msg)
+            raise ValueError("🛑 " + msg)
```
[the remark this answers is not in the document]
leondz: Yep. Also updated exception descriptions to distinguish `BadGeneratorException` from `PluginConfigurationError`

garak/generators/openai.py
```diff
@@ -190,6 +212,10 @@ def _call_model(
             if v is not None and k not in self.suppressed_params
         }
 
+        for k, v in self.custom_params.items():
+            if k not in self.suppressed_params:
+                create_args[k] = v
+
```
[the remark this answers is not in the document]
leondz: This could work.

If I read it correctly, this exposes garak users to instabilities and changes in the OpenAI API. I think I'm mostly OK with that - we don't currently want to maintain a shim.

NB. garak defaults for `max_tokens` are too low to yield output when used for `max_completion_tokens` (aside from the `max_tokens` being only patchily supported), hence the new default

garak/generators/openai.py
```diff
@@ -184,11 +207,15 @@ def _call_model(
             "seed": self.seed,
         }
```
jmartin-tech, line 208: This is now dead code.

garak/generators/openai.py
jmartin-tech: This should still be removed.

leondz:
thanks, missed that one

jmartin-tech (approved):

Issue #829, opened 2024-08-14
try to enable `parallel_attempts > 1` by default


leondz:
It's sad (and slow) when parallelisation is off but a generator can support it.

Proposal:
* benchmarks show `parallel_attempts` has more impact than `parallel_requests` (over a whole default run) - try to set this to be used by default
* not all generators support it; we should have a flag in each generators that sets whether it should be run in parallel by default
  * let's repurpose `parallel_capable`
  * it's `True` by default
  * [ ] needs to be set to False in ggml
  * [ ] needs to be sensibly set to False or True for each Hugging Face class
  * it should be possible to override `parallel_capable` somehow. options:
    1. ignore `parallel_(requests|attempts)` if generator's `parallel_capable == False`
    2. use `parallel_capable` to determine default parallel configuration defaults, which can be overridden in CLI/config options (print/log a warning when this happens)
    3. setting `parallel_(requests|attempts)` to `1` should (effectively? actually?) override parallel_capable; print/log a notice when this happens
  * what's a sensible default for the rest class?
  * what's a sensible default for `parallel_attempts`? what's polite? I like 200 with NIMs, the SREs have comments about why this isn't good; 2 is too low. powers of two are popular - how about `8`?

This will expose more people to #308

leondz:
@jmartin-tech opinions welcome!

jmartin-tech:
I like the idea or expanding `parallel_capable` to limiting both `parallel_requests` & `parallel_attempts` for all plugins. I think there could be an argument for wanting to allow concurrent `requests` while not allowing concurrent `attempts`, the simplicity of a single control though may outweigh the value.

> it should be possible to override `parallel_capable` somehow. options

In theory, `parallel_capable` can already be overridden via `config` on a per `generator` class basis. The impact would be to all instance of the `generator` class, and `could` have impacts for generators within other plugins like `probes` and `detectors`.

> what's a sensible default for the rest class?
> what's a sensible default ...

Using a single default for all types seems viable, I don't see a strong reason to limit `rest` differently. 
8 sounds reasonable, powers of 2 are nice from a cores/threads available perspective.
We could also consider an `auto-detect` value based on a CPU core count check and scale to 2x the CPU cores possibly with a max of 16 or something to keep it from over saturating on a high core count CPU.

jmartin-tech:
A note to add here, this will add constraints on integrators using the `function` generator as it puts a requirement on the package providing the function to ensure the package provides inference with no configuration of setup required on import by a sub-process.

PR #932, opened 2024-09-24
cli: `--parallel_attempts` hint

### LINKED ISSUE #928: cli: warn if generator supports parallelisation but none is used
## Summary

Help people run their generators faster, if they support it, by printing a message.

### Basic example

```
$ garak -m openai -n gpt-3.5-turbo
⚠️ This run can be sped up. openai generator supports parallelism. Try using `--parallel_requests 16` (or higher).
```

Maybe consider printing a similar message if generations > a threshold, say 16, and generator only supports single generations; then we can recommend `--parallel-requests`. But only recommend one, they don't play well together.

### Motivation

So many people run garak single threaded when it can be SO FAST in parallel. That's a waste of their time and bad for our UX.

leondz:
Resolves #928

Inform CLI users that garak isn't running as efficiently as it can, if they're not using `parallel_attempts` but their generator can take it

## Verification

Invoke garak in a way that can benefit from parallelisation. A hint should appear. e.g.

`python -m garak -m test.Lipsum`

```
(garak) 15:08:27 x1:~/dev/garak [cli/parallel_hint] $ python -m garak -m test.Lipsum
garak LLM vulnerability scanner v0.9.0.16.post1 ( https://github.com/leondz/garak ) at 2024-09-24T15:08:43.577976
📜 logging to /home/lderczynski/.local/share/garak/garak.log
🦜 loading generator: Test: Lorem Ipsum
⚠️  This run can be sped up. Generator 'Test:Lorem Ipsum' supports parallelism! 🥳 Consider using `--parallel_requests 16` (or higher) to accelerate your run. 🐌
⚠️  The current/default config is optimised for speed rather than thoroughness. Try e.g. --config full for a stronger test, or specify some probes.
📜 reporting to /home/lderczynski/.local/share/garak/garak_runs/garak.fd912c89-403a-4cab-a781-b8224554602d.report.jsonl
^CUser cancel received, terminating all runs
```

### DIFF
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -270,8 +270,7 @@ def main(arguments=None) -> None:
 
     # load site config before loading CLI config
     _cli_config_supplied = args.config is not None
-    if _cli_config_supplied:
-        _config.load_config(run_config_filename=args.config)
+    _config.load_config(run_config_filename=args.config)
 
     # extract what was actually passed on CLI; use a masking argparser
     aux_parser = argparse.ArgumentParser(argument_default=argparse.SUPPRESS)
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -489,6 +489,11 @@ def main(arguments=None) -> None:
                 f"generators.{_config.plugins.model_type}", config_root=_config
             )
 
+            if generator.parallel_capable and not _config.system.parallel_attempts:
+                print(
+                    f"⚠️  This run can be sped up. Generator '{generator.fullname}' supports parallelism! 🥳 Consider using `--parallel_requests 16` (or higher) to accelerate your run. 🐌"
+                )
+
             if "generate_autodan" in args and args.generate_autodan:
                 from garak.resources.autodan import autodan_generate
 
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -491,7 +491,7 @@ def main(arguments=None) -> None:
 
             if generator.parallel_capable and not _config.system.parallel_attempts:
                 print(
-                    f"⚠️  This run can be sped up. Generator '{generator.fullname}' supports parallelism! 🥳 Consider using `--parallel_requests 16` (or higher) to accelerate your run. 🐌"
+                    f"⚠️  This run can be sped up. Generator '{generator.fullname}' supports parallelism! 🥳 Consider using `--parallel_requests 16` (or more) to greatly accelerate your run. 🐌"
                 )
 
             if "generate_autodan" in args and args.generate_autodan:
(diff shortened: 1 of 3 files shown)


garak/cli.py
```diff
@@ -489,6 +489,11 @@ def main(arguments=None) -> None:
                 f"generators.{_config.plugins.model_type}", config_root=_config
             )
 
+            if generator.parallel_capable and not _config.system.parallel_attempts:
```
jmartin-tech, line 492: Might want to suppress this if a config file was provided, while the message could be helpful for a cli hint if `-- config` was provided on the command line the hint is likely unnecessary.

It looks like `args` has already been clobbered by the time we get to this code so a mechanism to detect if `--config` was passed would need to be added earlier.

leondz: agree - private var added to track this & condition hint display

leondz:
partly contributes to UX improvement detailed in #829

garak/command.py
```diff
@@ -6,6 +6,8 @@
 import logging
 import json
 
+import garak.cli
```
jmartin-tech, line 9: This creates something of a circular dependency, `cli.main()` imports `garak.command`.  The `hint` method should likely be factored out of cli.

leondz: Agree, it's not perfect here. I would like to use it more places. It's in CLI because it's used a bit here, it came from here, and because it prints (i.e. does CLI stuff) - a presentation/logic separation isn't near on the roadmap. Where do you think could be an OK home for it?

jmartin-tech: > I would like to use it more places.
Some info on `where` else it would be used can guide ideas for location.

For the short term I can see placing it in `command` since we already `know` we want to extract `print` statements  in the future. Avoiding the `cycle` is more pressing that not having one more `print` statement a class that already has them.

leondz: yeah, agree. thanks for the catch. pushed

garak/cli.py
```diff
@@ -3,6 +3,9 @@
 
 """Flow for invoking garak from the command line"""
 
+
+HINT_CHANCE = 0.25
+
```
jmartin-tech, line 8: This also needs to move to `command`.

garak/cli.py
```diff
@@ -489,6 +493,16 @@ def main(arguments=None) -> None:
                 f"generators.{_config.plugins.model_type}", config_root=_config
             )
 
+            if (
+                not _cli_config_supplied
+                and generator.parallel_capable
+                and _config.system.parallel_attempts is False
+            ):
+                hint(
```
jmartin-tech, line 501:
```suggestion
                command.hint(
```

jmartin-tech (approved):

PR #933, opened 2024-09-26
payloads: web-related, incl. `probes.encoding` migration

leondz:
Factor out web-related payloads, and complete the transition to `garak.payloads` use in `probes.encoding`

## Verification

* `garak -m test -g 1 -p encoding,latentinjection,xss` and check the `report.jsonl`s include appropriate payload text
* `python -m pytest tests/test_payloads.py`
* `python -m pytest tests/probes/test_probes_encoding.py`
* `python -m pytest tests/probes/test_probes_latentinjection.py`


garak/payloads.py
```diff
@@ -172,49 +169,63 @@ def _scan_payload_dir(self, dir) -> dict:
...
                     yield payload
 
+    @staticmethod
+    def _load_payload(
+        name: str, path: Union[str, pathlib.Path, None] = None
+    ) -> PayloadGroup:
+        if path is None:
+            path = PAYLOAD_DIR / f"{name}.json"
+        return PayloadGroup(name, path)
+
```
jmartin-tech, line 208: I think I was unclear with this statement:

> Lets consider making the exposed entry point for load_payload be the interface to the Director.

I was thinking that `garak.payloads` module could expose the `Director().search()` capability as a module level method:
``` Python
import garak.payloads

probe_payloads = payloads.search("slur_terms_en").payloads

# or

probe_payloads = payloads.load("slur_terms_en").payloads
```

The `Director` implementation inside `garak.payloads` could then mimic the `PluginCache` class:
``` Python
@staticmethod
def search(types: Union[List[str], None] = None, include_children=True) -> PayloadGroup:
    return Director.getInstance().search(types, include_children)
```

Not a blocker just a direction I think would be easier to consume.

garak/payloads.py
```diff
@@ -172,49 +169,63 @@ def _scan_payload_dir(self, dir) -> dict:
     def _refresh_payloads(self) -> None:
         """Scan resources/payloads and the XDG_DATA_DIR/payloads for
         payload objects, and refresh self.payload_list"""
-        self.payload_list = self._scan_payload_dir(PAYLOAD_DIR)
+        self.__class__.payload_list = self._scan_payload_dir(PAYLOAD_DIR)
```
jmartin-tech, line 172: With the current definition of `payload_list` being class level and not in `__init__()` this is equivalent:
```suggestion
        self.payload_list = self._scan_payload_dir(PAYLOAD_DIR)
```

jmartin-tech (commented):
Does it make sense to have consistent `DEFAULT_PARAMS` naming as `payload_name`? Examples of impact noted in the review.

Also looking at this `payload_trigger` being populated directly with the `payloads` set seems non-intuitive, but I don't have a better alternative to offer at this time so reasonable to live with for now.

garak/probes/latentinjection.py
```diff
@@ -254,14 +256,17 @@ class LatentInjectionReport(Probe, LatentInjectionMixin):
...
     ]
-    payload_triggers = [
-        "superdupersketchycorp.com",
-        "superdupersketchycorp.biz",
-        "client.hooli-schwab.com",
-    ]
+    payload_triggers = []
+
+    DEFAULT_PARAMS = Probe.DEFAULT_PARAMS | {
+        "domains_payload_name": "domains_latentinjection"
```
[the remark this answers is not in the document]
leondz: There are so many different payload-named items in `latentinjection` that this creates a bit too much of a collision, for me. There's scope for multiple payloads in this module. Could rename `payload_name_domains` I guess.

garak/payloads.py
leondz: Oh, that's pretty neat. Yeah, I like it. Thanks for clarifying.

garak/payloads.py
leondz: I don't think class attributes assigned through instantiation are mutable from instances

jmartin-tech: You are correct although this is somewhat of a grey area, attributes cannot be replaced via instance access but are mutable themselves. In this case `None` is being replaced with a pointer to the resulting array. However the resulting object can be modified.

This particular line does need to access `self.__class__.payload_list` however the `read` access in `search()` does not need to use `__class__`, I am on the fence on if those locations *should* use `__class__` for clarity reasons.

garak/payloads.py
leondz: in https://github.com/leondz/garak/pull/933/commits/f8d4d89366fa048ccb76c4cf9230ec13027c519c

garak/probes/latentinjection.py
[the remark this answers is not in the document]
jmartin-tech: Thinking about `DEFAULT_PARAMS` from a consumer perspective, the internal references to `payload` in the parent module are unfortunate but internal implementation details for parameters not exposed to the consumer should not limit the external interface.

garak/payloads.py
leondz: I'm on the fence too. I guess py's "explicit not implicit" suggests leaving it in, and one can find a rationale for that - the verbose formulation signals that there might be a non-trivial access pattern. Also doing nothing is usually the lighter path.

garak/probes/latentinjection.py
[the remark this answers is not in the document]
leondz: Yeah, OK. I anticipate a need for namespacing but it's not yet here, so let's roll w/ `payload_name`

PR #945, opened 2024-10-11
update: rename `replay` plugins to `divergence`

leondz:
Renames `replay` plugins to `divergence`

Standard tests apply


erickgalinkin (approved):
lgtm

PR #954, opened 2024-10-21
architecture: factor HFCompatible out

leondz:
`HFCompatible` was embedded in generators.base, tying slow-to-import HF-specific stuff to base classes. This PR moves `HFCompatible` to a separate module, with a candidate location in `garak.resources.api.huggingface`, enabling fast base class loading.

### DIFF
--- a/garak/buffs/paraphrase.py
+++ b/garak/buffs/paraphrase.py
@@ -14,11 +14,14 @@ from garak.resources.api.huggingface import HFCompatible
 class PegasusT5(Buff):
     """Paraphrasing buff using Pegasus model"""
 
+    DEFAULT_PARAMS = Buff.DEFAULT_PARAMS | {
+        "para_model_name": "garak-llm/pegasus_paraphrase",
+        "hf_args": {"device": "cpu", "torch_dtype": "float32"},
+    }
     bcp47 = "en"
     doc_uri = "https://huggingface.co/tuner007/pegasus_paraphrase"
 
     def __init__(self, config_root=_config) -> None:
-        self.para_model_name = "garak-llm/pegasus_paraphrase"  # https://huggingface.co/tuner007/pegasus_paraphrase
         self.max_length = 60
         self.temperature = 1.5
         self.num_return_sequences = 6
--- a/garak/buffs/paraphrase.py
+++ b/garak/buffs/paraphrase.py
@@ -7,8 +7,8 @@ from collections.abc import Iterable
 
 import garak.attempt
 from garak import _config
-from garak.generators.huggingface import HFCompatible
 from garak.buffs.base import Buff
+from garak.resources.api.huggingface import HFCompatible
 
 
 class PegasusT5(Buff):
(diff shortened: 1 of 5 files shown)


garak/buffs/paraphrase.py
```diff
@@ -7,8 +7,8 @@
 
 import garak.attempt
 from garak import _config
-from garak.generators.huggingface import HFCompatible
 from garak.buffs.base import Buff
+from garak.resources.api.huggingface import HFCompatible
 
 
 class PegasusT5(Buff):
```
jmartin-tech, line 14: Just an observation not required for this PR, it looks like this class could benefit from a refactor to use `HFCompatible` and expose the `para_model_name` and `hf_args` as `DEFAULT_PARAMS`.

```suggestion
class PegasusT5(Buff, HFCompatible):
```

jmartin-tech (approved):
Move looks reasonable, I am on the fence on using `garak.resources.api` to represent wrappers for dependencies.

One idea would be to use `garak/resources/huggingface/__init__.py` to expose the class, this could allow for keeping the class definitions in unique files imported to `__init__py` for exposure as more `Compatible` types are identified over time. Just a thought that came to mind, no strong argument to favor this at this time.

This PR also suggests there is another consumer for this mixin in `buffs.paraphrase.PegasusT5`.

garak/buffs/paraphrase.py
jmartin-tech: This change is incomplete the class needs to be extended to consume the `HFCompatible` mixin in `_load_model()`:

`self.torch_device` moves to the standardized `self.device` and should be detected/populated from `hf_args["device"]` with a call to `self._select_hf_device()` in `_load_model()`:
``` Python
    def __init__(self, config_root=_config) -> None:
        self.max_length = 60
        self.temperature = 1.5
        self.num_return_sequences = 6
        self.num_beams = self.num_return_sequences
        self.tokenizer = None
        self.para_model = None
```

``` Python
   def _load_model(self):
        from transformers import PegasusForConditionalGeneration, PegasusTokenizer

        self.device = self._select_hf_device()
        model_kwargs = self._gather_hf_params(
            hf_constructor=PegasusForConditionalGeneration.from_pretrained
        )  # will defer to device_map if device map was `auto` may not match self.device

        self.para_model = PegasusForConditionalGeneration.from_pretrained(
            self.para_model_name, **model_kwargs
        ).to(self.device)
        self.tokenizer = PegasusTokenizer.from_pretrained(self.para_model_name)
```

Not an issue for this PR, however I suspect a few more items should likely be promoted to `DEFAULT_PARAMS`.  The `max_length`, `temperature`, `num_return_sequences`, and possibly `num_beams` if the value does not always have to be equal to `num_return_sequences` should likely be exposed a configurable.  Since `Fast` looks like it may also have similar items to promote I think that can be deferred.

leondz: Agree. Thanks for the details. Will mark as ready for review when out of draft.

leondz: Adding the `model_kwargs` part led to the paraphraser returning all blanks. Did the rest of the integration and added a test to catch this unwanted behaviour.

jmartin-tech: Lets drop the `_gather_hf_params()` for now as the `PegasusForConditionalGeneration.from_pretrained()` looks like it is not handling the extra args in the way the current code is expecting, I suspect `device` vs `device_map` is also impacting the expectations.

PR #960, opened 2024-10-25
feature: amend garak user-agent

leondz:
We should be good citizens and not use the default `requests` user-agent, but instead our own

* add new run-level config param for useragent
* use this in generators using requests library

## Verification

- drawing a blank here for now. maybe when this comes out of draft, we can use `requests-mock`

### DIFF
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -155,6 +155,14 @@ def _store_config(settings_files) -> None:
     reporting = _set_settings(reporting, settings["reporting"])
 
 
+def _garak_user_agent(dummy=None):
+    global run
+    if hasattr(run, "user_agent"):
+        return run.user_agent
+    else:
+        return "garak"
+
+
 def set_all_http_lib_agents(agent_string):
     set_http_lib_agents(
         {"requests": agent_string, "httpx": agent_string, "aiohttp": agent_string}
@@ -165,7 +173,7 @@ def set_http_lib_agents(agent_strings: dict):
     if "requests" in agent_strings:
         from requests import utils
 
-        utils.default_user_agent = lambda x=None: agent_strings["requests"]
+        utils.default_user_agent = _garak_user_agent
     if "httpx" in agent_strings:
         import httpx
 
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -153,9 +153,40 @@ def _store_config(settings_files) -> None:
     run.user_agent = run.user_agent.replace("{version}", version)
     plugins = _set_settings(plugins, settings["plugins"])
     reporting = _set_settings(reporting, settings["reporting"])
-    from requests import utils
 
-    utils.default_user_agent = run.user_agent
+
+def set_all_http_lib_agents(agent_string):
+    set_http_lib_agents(
+        {"requests": agent_string, "httpx": agent_string, "aiohttp": agent_string}
+    )
+
+
+def set_http_lib_agents(agent_strings: dict):
+    if "requests" in agent_strings:
+        from requests import utils
+
+        utils.default_user_agent = lambda x=None: agent_strings["requests"]
+    if "httpx" in agent_strings:
+        import httpx
+
+        httpx._client.USER_AGENT = agent_strings["httpx"]
+    if "aiohttp" in agent_strings:
+        import aiohttp
+
+        aiohttp.client_reqrep.SERVER_SOFTWARE = agent_strings["aiohttp"]
+
+
+def get_http_lib_agents():
+    from requests import utils
+    import httpx
+    import aiohttp
+
+    agent_strings = {}
+    agent_strings["requests"] = utils.default_user_agent
+    agent_strings["httpx"] = httpx._client.USER_AGENT
+    agent_strings["aiohttp"] = aiohttp.client_reqrep.SERVER_SOFTWARE
+
+    return agent_strings
 
 
 def load_base_config() -> None:
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -23,9 +23,7 @@ from xdg_base_dirs import (
 
 DICT_CONFIG_AFTER_LOAD = False
 
-from garak import __version__
-
-version = __version__
+from garak import __version__ as version
 
 system_params = (
     "verbose narrow_output parallel_requests parallel_attempts skip_unknown".split()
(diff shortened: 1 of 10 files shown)


garak/resources/garak.core.yaml
```diff
@@ -13,6 +13,7 @@ run:
   eval_threshold: 0.5
   generations: 5
   probe_tags:
+  user_agent: "garak LLM vulnerability scanner, v{version} https://garak.ai"
```
erickgalinkin, line 16: Per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent, syntax should be:
```
User-Agent: <product> / <product-version> <comment>
```
```suggestion
  user_agent: "garak LLM vulnerability scanner/{version} https://garak.ai"
```

erickgalinkin (commented):
We may want to find other places where we're using `requests` and fix the user agent -- may want to place the user agent in the shared resources so we can ref it easily.

garak/_config.py
```diff
@@ -23,7 +23,9 @@
 
 DICT_CONFIG_AFTER_LOAD = False
 
-version = -1  # eh why this is here? hm. who references it
+from garak import __version__
+
+version = __version__
```
jmartin-tech, line 28: `cli` actually overrides this on startup, if moving the lookup here we should remove there:
https://github.com/leondz/garak/blob/71d47aa67ecde3afa784861468fced03d65000a9/garak/cli.py#L19-L34

garak/_config.py
```diff
@@ -144,10 +146,13 @@ def _load_yaml_config(settings_filenames) -> dict:
 
 
 def _store_config(settings_files) -> None:
+    import garak
+
     global system, run, plugins, reporting
```
jmartin-tech, line 151:
```suggestion
    global system, run, plugins, reporting, version
```

garak/_config.py
```diff
@@ -144,10 +146,13 @@ def _load_yaml_config(settings_filenames) -> dict:
...
     settings = _load_yaml_config(settings_files)
     system = _set_settings(system, settings["system"])
     run = _set_settings(run, settings["run"])
+    run.user_agent = run.user_agent.replace("{version}", garak.__version__)
```
jmartin-tech, line 155:
```suggestion
    run.user_agent = run.user_agent.replace("{version}", version)
```

garak/resources/garak.core.yaml
jmartin-tech: Or:
```suggestion
  user_agent: "garak/v{version} (garak LLM vulnerability scanner, https://garak.ai)"
```

jmartin-tech (commented):
Consider finding a hook to set for all uses in common libs we are already dependent on:
- requests
- aiohttp
- httpx

garak/_config.py
leondz: Thanks! Saw this, went for a more generic solution, forgot to prune the activity in `cli`

garak/_config.py
```diff
@@ -144,12 +146,16 @@ def _load_yaml_config(settings_filenames) -> dict:
...
+    global system, run, plugins, reporting, version
     settings = _load_yaml_config(settings_files)
     system = _set_settings(system, settings["system"])
     run = _set_settings(run, settings["run"])
+    run.user_agent = run.user_agent.replace("{version}", version)
     plugins = _set_settings(plugins, settings["plugins"])
     reporting = _set_settings(reporting, settings["reporting"])
+    from requests import utils
+
+    utils.default_user_agent = run.user_agent
```
jmartin-tech, line 158: I like this single location for default. A question about long term viability. How can we ensure this only applies to `garak` objects?

Since the project wants to support inclusion as a dependency the first import of `garak._config` would override the `user_agent` for the current process until exit.  Should we consider this now or is this worth kicking the can down the road a bit?

leondz: One pov is that child requests are still caused by garak anyway, and so it's in one way even desirable they adopt the UA

jmartin-tech: Agreed, my concern is more with consumption as a library, the consuming process may do other `requests` after utilizing `garak`, these would also inherit the revised user-agent.

Consider an integration that runs various security tools in the same runtime and selects `garak` first, or also runs other benchmarks. I will offer I think the tool should spawn independent workspaces however if the consumer is already python it "may" not. In that scenario if the benchmark requests to not restore or reset the agent string requests could be attributed to the `garak` code when not related. Also if the tool expects to `post` results to some location further along in a CI pipeline this could introduce unexpected values in the headers for those calls.

leondz: Excellent point. Maybe sans API we could push + pop UA as part of garak.cli for now.

jmartin-tech: One idea for handling this might be to have this functionality injected and reset by `harnesses.base.Harness` when executing `run()` to limit the window of influence.

erickgalinkin: IMO, having this within `garak.cli` is probably ideal -- we can use the same logic, but only set the default UA on invocation of the run.

leondz: have placed it in `harnesses.base.Harness` for now, using two hooks and a class var, and covering `requests`, `httpx`, and `aiohttp`. Helper funcs staying in `_config`, which doesn't seem optimal, but might be least-bad

garak/harnesses/base.py
```diff
@@ -64,6 +64,13 @@ def _load_buffs(self, buff_names: List) -> None:
                     logging.warning(err_msg)
                     continue
 
+    def _start_run_hook(self):
+        self._http_lib_user_agents = _config.get_http_lib_agents()
+        _config.set_all_http_lib_agents(_config.run.user_agent)
+
+    def _end_run_hook(self):
+        _config.set_http_lib_agents(self._http_lib_user_agents)
+
```
jmartin-tech, line 73: These changes do not `distribute` to multiprocess tasks.

garak/_config.py
leondz: added push & pop of user agent strings into `garak.cli` pending spec of api hook for doing whole run

garak/resources/garak.core.yaml
```diff
@@ -13,6 +13,7 @@ run:
   eval_threshold: 0.5
   generations: 5
   probe_tags:
+  user_agent: "garak/{version} , LLM vulnerability scanner https://garak.ai"
```
erickgalinkin, line 16: Space before comma?

erickgalinkin (approved):
LGTM

PR #979, opened 2024-11-11
feature: reinforce `trust_remote_code=False`

### LINKED ISSUE #991: feature: Assert `trust_remote_code/False` for HF models and expose as config var
Add a layer on to top HF defaults setting `trust_remote_code` to `False` if unspecified

`hf_args` can hold the config var for this, not a global

leondz:
This change sets `trust_remote_code` to false regardless of HF function default, unless explicitly enabled in garak config

resolves #991

### DIFF
--- a/garak/buffs/paraphrase.py
+++ b/garak/buffs/paraphrase.py
@@ -39,7 +39,9 @@ class PegasusT5(Buff, HFCompatible):
         self.para_model = PegasusForConditionalGeneration.from_pretrained(
             self.para_model_name
         ).to(self.device)
-        self.tokenizer = PegasusTokenizer.from_pretrained(self.para_model_name)
+        self.tokenizer = PegasusTokenizer.from_pretrained(
+            self.para_model_name, trust_remote_code=False
+        )
 
     def _get_response(self, input_text):
         if self.para_model is None:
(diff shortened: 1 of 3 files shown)


garak/buffs/paraphrase.py
```diff
@@ -39,7 +39,9 @@ def _load_model(self):
         self.para_model = PegasusForConditionalGeneration.from_pretrained(
             self.para_model_name
         ).to(self.device)
-        self.tokenizer = PegasusTokenizer.from_pretrained(self.para_model_name)
+        self.tokenizer = PegasusTokenizer.from_pretrained(
+            self.para_model_name, trust_remote_code=False
```
jmartin-tech, line 43: Should this be possible to inject via config? Consider adding this to `DEFAULT_PARAMS["hf_args"]` for this class.
```suggestion
            self.para_model_name, trust_remote_code=self.hg_args["trust_remote_code"]
```

jmartin-tech (commented):
Terrific! Secure by default is definitely a good way to go.  One minor expansion since the `PegasusT5` class already enforced a value it may be worth "exposing".

Will do some validation testing in the next day to ensure this does not produce regressions in various `HFCompatible` configurations.

garak/buffs/paraphrase.py
leondz: Excellent, yes, missed this, better route

leondz: amended to default param+value

jmartin-tech (approved):
Basic sanity testing shows this passing correctly to huggingface pipeline generator instances with the following configuration file.

enable_remote.yaml
``` yaml
plugins:
  generators:
    huggingface:
      hf_args:
        trust_remote_code: true
```

PR #999, opened 2024-11-12
generators: add option to specify HTTP codes to skip generation on, for `RestGenerator`

leondz:
Some prompts are difficult for some endpoints to respond to. This patch offers a way for garak to skip those prompts, if the generation failure is expressed in an endpoint HTTP response code (e.g. `400`)

## Verification
- [ ] Run a restgenerator against an endpoint that returns 400; there should be a `None` returned by the generator and the probe should present as SKIP on the CLI
- [ ] Docs added in PR

## Context
Targets issue raised in https://discord.com/channels/1121536128269422654/1121536129099907190/1305844548181692416

### DIFF
--- a/docs/source/garak.generators.rest.rst
+++ b/docs/source/garak.generators.rest.rst
@@ -16,7 +16,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``response_json_field`` - (optional) Which field of the response JSON should be used as the output string? Default ``text``. Can also be a JSONPath value, and ``response_json_field`` is used as such if it starts with ``$``.
 * ``request_timeout`` - How many seconds should we wait before timing out? Default 20
 * ``ratelimit_codes`` - Which endpoint HTTP response codes should be caught as indicative of rate limiting and retried? ``List[int]``, default ``[429]``
-* ``skip_codes`` - Which endpoint HTTP response code should lead to the generation being treated as not possible and skipped for this query. Takes precedence over ``skip_codes``.
+* ``skip_codes`` - Which endpoint HTTP response code should lead to the generation being treated as not possible and skipped for this query. Takes precedence over ``ratelimit_codes``.
 
 Templates can be either a string or a JSON-serialisable Python object.
 Instance of ``$INPUT`` here are replaced with the prompt; instances of ``$KEY``
--- a/docs/source/garak.generators.rest.rst
+++ b/docs/source/garak.generators.rest.rst
@@ -16,6 +16,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``response_json_field`` - (optional) Which field of the response JSON should be used as the output string? Default ``text``. Can also be a JSONPath value, and ``response_json_field`` is used as such if it starts with ``$``.
 * ``request_timeout`` - How many seconds should we wait before timing out? Default 20
 * ``ratelimit_codes`` - Which endpoint HTTP response codes should be caught as indicative of rate limiting and retried? ``List[int]``, default ``[429]``
+* ``skip_codes`` - Which endpoint HTTP response code should lead to the generation being treated as not possible and skipped for this query. Takes precedence over ``skip_codes``.
 
 Templates can be either a string or a JSON-serialisable Python object.
 Instance of ``$INPUT`` here are replaced with the prompt; instances of ``$KEY``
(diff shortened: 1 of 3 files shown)


docs/source/garak.generators.rest.rst
```diff
@@ -16,6 +16,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``response_json_field`` - (optional) Which field of the response JSON should be used as the output string? Default ``text``. Can also be a JSONPath value, and ``response_json_field`` is used as such if it starts with ``$``.
 * ``request_timeout`` - How many seconds should we wait before timing out? Default 20
 * ``ratelimit_codes`` - Which endpoint HTTP response codes should be caught as indicative of rate limiting and retried? ``List[int]``, default ``[429]``
+* ``skip_codes`` - Which endpoint HTTP response code should lead to the generation being treated as not possible and skipped for this query. Takes precedence over ``skip_codes``.
```
jmartin-tech, line 19:
```suggestion
* ``skip_codes`` - Which endpoint HTTP response code should lead to the generation being treated as not possible and skipped for this query. Takes precedence over ``ratelimit_codes``.
```

garak/generators/rest.py
```diff
@@ -29,7 +29,8 @@ class RestGenerator(Generator):
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "headers": {},
         "method": "post",
-        "ratelimit_codes": [429],
+        "ratelimit_codes": {429},
+        "skip_codes": set(),
```
jmartin-tech, line 33: Changing from `list` to set here may not be the best choice, safe `yaml` and `json` serialization do not really contain `set` logic and configuration is passed via these formats.  In this case the values are `replaced` if found in a configuration files so I would prefer we stick to `list` to ensure we do infer a logical interface that may not be available at runtime.

jmartin-tech (changes requested):

garak/generators/rest.py
leondz: This went through my mind as well. I wanted to signal the set-like nature of how arguments are supplied, but this doesn't jive with what can be supplied via existing config formats. Agree that adding `set` here may encourage code that relies on `set` affordances, will update

garak/generators/rest.py
```diff
@@ -121,7 +123,7 @@ def __init__(self, uri=None, config_root=_config):
             try:
                 self.json_expr = jsonpath_ng.parse(self.response_json_field)
             except JsonPathParserError as e:
-                logging.CRITICAL(
+                logging.critical(
```
jmartin-tech, line 126: Nice catch!

tests/generators/test_rest.py
```diff
@@ -95,3 +95,29 @@ def test_json_rest_deeper(requests_mock):
...
     output = generator._call_model("Who is Enabran Tain's son?")
     assert output == [DEFAULT_TEXT_RESPONSE]
+
+
+@pytest.mark.usefixtures("set_rest_config")
+def test_rest_skip_code(requests_mock):
+    generator = _plugins.load_plugin(
+        "generators.rest.RestGenerator", config_root=_config
+    )
+    generator.skip_codes = {200}
```
jmartin-tech, line 105: Not required but would prefer this be a list for same reason default is a list type.
```suggestion
    generator.skip_codes = [200]
```

garak/generators/rest.py
```diff
@@ -193,31 +195,44 @@ def _call_model(
...
-            else:
-                raise ConnectionError(error_msg)
+            raise ConnectionError(error_msg)
 
         if not self.response_json:
             return [str(resp.text)]
 
         response_object = json.loads(resp.content)
 
+        response = [None] * generations_this_call
```
jmartin-tech, line 234: Since `self.supports_multiple_generations` is set to `False` in the in the constructor should we opt for consistency [with the skip case](https://github.com/leondz/garak/pull/999/files#diff-ff96cd9116f6456af6b492c39bc684effd08ab800cdf6f33e9d1bffead1ac309R206) and and avoid the multiplier here? 
```suggestion
        response = [None]
```

If/when `self.supports_multiple_generations` can be supported the entire method will need to be rechecked to ensure all `return` statements conform to required `generations_this_call`.

jmartin-tech (approved):
A couple `nice to have` comments, happy to accept deferrals.

garak/generators/rest.py
leondz: OK

PR #1000, opened 2024-11-12
Sanity test for pip install from repository

jmartin-tech:
Between releases the various installation processes need testing, this adds a basic integration test for early warning of `pip` based repository snapshot installs.

This verifies a simple test run will not result in any logged `ERROR` messages on the minimum python version and the latest supported version.


erickgalinkin (approved):
LGTM

PR #1011, opened 2024-11-18
Fix huggingface inference endpoint name

### LINKED ISSUE #998: Huggingface InferenceAPI stopped working after 0.9.0.13 due to name parameter not being set
python3 -m garak --model_type huggingface.InferenceAPI --model_name microsoft/Phi-3-mini-4k-instruct --probes malwaregen.Evasion  

stopped working after 0.9.0.13 (it just gets stuck). Logging https requests and digging into the code:

send: b'POST /models/ HTTP/1.1\r\nHost: api-inference.huggingface.co\r\nUser-A\
gent: python-requests/2.32.3\r\nAccept-Encoding: gzip, deflate, zstd\r\nAccept\
: */*\r\nConnection: keep-alive\r\nAuthorization: Bearer ...

when it had the model name in the URI before.

class InferenceAPI(Generator):
 ...
self.uri = self.URI + name

If I change name to config_root.plugins.model_name it works like before, but maybe not the correct fix? I'm having a hard time figuring out the call chain (_plugins.py through klass() without name being passed, apparently). 0.9.0.13 inherited from HFCompatible, which seems to do some magic in this area.

jmartin-tech:
fix #998 

the name provided during construction is populated by the call to `super().__init__()`, access to `self` attributes is required for any value populated by `Configurable`.

## Verification

- [ ] Execute against public InferenceAPI:
``` bash
python -m garak --model_type huggingface.InferenceAPI --model_name microsoft/Phi-3-mini-4k-instruct --probes malwaregen.Evasion
```
- [ ] **Verify** valid responses are received from endpoint.
- [ ] **Verify** added automation tests pass

### DIFF
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -376,7 +376,7 @@ class InferenceEndpoint(InferenceAPI):
 
     def __init__(self, name="", config_root=_config):
         super().__init__(name, config_root=config_root)
-        self.uri = name
+        self.uri = self.name
 
     @backoff.on_exception(
         backoff.fibo,
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -256,7 +256,7 @@ class InferenceAPI(Generator):
         self.name = name
         super().__init__(self.name, config_root=config_root)
 
-        self.uri = self.URI + name
+        self.uri = self.URI + self.name
 
         # special case for api token requirement this also reserves `headers` as not configurable
         if self.api_key:
--- a/tests/generators/conftest.py
+++ b/tests/generators/conftest.py
@@ -11,3 +11,10 @@ def openai_compat_mocks():
     """Mock responses for OpenAI compatible endpoints"""
     with open(pathlib.Path(__file__).parents[0] / "openai.json") as mock_openai:
         return json.load(mock_openai)
+
+
+@pytest.fixture
+def hf_endpoint_mocks():
+    """Mock responses for Huggingface InferenceAPI based endpoints"""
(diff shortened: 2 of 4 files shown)


erickgalinkin (commented):
LGTM. Did we test on an actual HF InferenceAPI just to be sure it works as intended? If not, I can spin one up tomorrow.

jmartin-tech:
I tested the public endpoint referenced in the issue, I have not done an end to end test with a private endpoint.

ppietikainen:
Tested the fix and works for what we've been trying out. Also
python3 -m garak --model_type huggingface.InferenceEndpoint --model_name https://api--inference.huggingface.co/models/microsoft/Phi-3-mini-4k-instruct --probes malwaregen.Evasion
seems to do queries to the right place, so I would assume private ones would work too.

I guess (and seemed to work, tho it's getting late) 
def __init_(name="")
  self.uri = ... + self.name

in both would make it a bit more easy to follow, the self.name = name confused me when trying to figure the code out, as I thought it implies the super() won't do anything to it :-)

jmartin-tech:
@ppietikainen thanks for the extra testing, the we have tried to add documentation on how [Configurable](https://reference.garak.ai/en/latest/configurable.html) classes are expected to be prioritized. The assignment of `name` as provided by to the constructor relates to how the precedence of values are treated allowing constructor values to be held above `configuration file` values that are injected by `super()`.

Also, in the different classes `name` has competing usage. For `InferenceAPI` it is expected to just be the model name and gets added to the public base uri defined in a constant, in `InferenceEndpoint` the full uri must be supplied.  I do think there may be some consistency to be gained at some point in the future.  I don't have a chosen path forward in mind so for now this can land at least return the functionality to original state before the regression.

PR #1012, opened 2024-11-18
detect if tokenizer is not loaded and adjust

jmartin-tech:
In some cases a pretrained model pipeline may not specify the tokenizer in the stored config. If missing attempt to get tokenizer by model name. An example of this as time of revision is [`nvidia/Minitron-4B-Base`](https://huggingface.co/nvidia/Minitron-4B-Base).

## Verification

- [ ] Tests against a model where pipeline is known to not have a saved tokenizer:
``` bash
python -m garak -m huggingface -n nvidia/Minitron-4B-Base --probes dan.Dan_11_0
```
- [ ] **Verify** the generator returned inference responses


leondz (approved):
thanks for the workaround. lgtm.

PR #1017, opened 2024-11-21
probe: add Arabic DNA

Eaalghamdi:
Signed-off-by: Emad Alghamdi <37039039+Eaalghamdi@users.noreply.github.com>

I added Arabic translation of all donotanswer probes which were quality checked by human to ensure suitability to the Arabic language, added Arabic detector for the probe. The new probes and detectors passed the test during development.


Eaalghamdi:

Eaalghamdi:
recheck

garak/detectors/mitigation_ar.py
jmartin-tech, line 1: I think language specific detectors are likely better contained in the same module.  In this case adding this class to `garak/detectors/mitigation.py` as it's own class is a preferred approach to make this available in the near term possibly as `garak.detectors.mitigation.MitigationBypassArabic` or `MitigationBypassAR` to use the `bcp47` value as an indicator of language.

In a future iteration I would like to see `garak.detectors.mitigation.MitigationBypass` load the language specific values needed from a data file based on the language compatible with the probe that generated the attempt.

garak/probes/donotanswer_ar.py
jmartin-tech, line 1: Language specific implementations are likely better contained in the same module.  In this case adding these classes to `garak/probes/donotanswer.py` as it's own set of classes, based on a prefix or suffix for the language, is a preferred approach to make this available in the near term.

In a future iteration I would like to see `garak.probes.donotanswer.*` load the language specific values needed from a data file based on the language the run requires to be compatible with the target application.

garak/data/Arabic/donotanswer/discrimination_exclusion_toxicity_hateful_offensive.txt
jmartin-tech, line 1: A minor file structure item, consider using `bcp47` values as a standard already adopted by the project to determine the storage path for language specific values, in this case `garak/data/ar/*` might provide a more programatic access friendly pattern locating files for a specific language.

jmartin-tech (commented):
Thank you for this! Additional language support is currently evolving and this give some insight into how others are thinking about it.

I made a couple comments mostly related to location of code, long term form for this may change further, however this is great step towards enabling testing on targets in more languages.

garak/data/Arabic/donotanswer/discrimination_exclusion_toxicity_hateful_offensive.txt
Eaalghamdi: fixed

leondz:
Hi Emad, We haven't forgotten about these - we've been working hard on landing multilingual core support in garak, which is now in review. Once that's in place, we'll finally be able to process these PRs properly.

Eaalghamdi:
Hi Leon,

Thank you for the update. I appreciate you taking the time to explain the
current priorities.

I'm looking forward to the multilingual support being integrated and am
eager to contribute once it's in place.

Emad

PR #1018, opened 2024-11-21
probe: add Arabic DAN

Eaalghamdi:
Signed-off-by: Emad Alghamdi <37039039+Eaalghamdi@users.noreply.github.com>

I added Arabic translation of all DAN probes which were quality checked by human to ensure suitability to the Arabic language, added Arabic detector for the probe. The new probes and detectors passed the test during development.


Eaalghamdi:

Eaalghamdi:
recheck

garak/detectors/dan_ar.py
erickgalinkin, line 1: Do these differ meaningfully from the detectors in `detectors.dan`? It seems to be a direct copy. If so, do we want to reference those extant detectors from the probes instead?

garak/probes/dan_ar.py
```diff
@@ -0,0 +1,679 @@
...
+
+
+class DanInTheWildMini(DanInTheWild):
+    """A subset of 100 jailbreaks from the DanInTheWild jailbreak library.
+
+    Tries a set of 100 known jailbreaks, in an attempt to get a model to disregard
+    alignment/system prompt. Success is measured by not getting a safety mitigaton
+    message."""
+
+    active = True
```
erickgalinkin, line 675: I would suggest that we keep `active=False` on the Arabic probes -- perhaps there's some logic that needs to be written around the language spec.

erickgalinkin (commented):
Could you say a bit more about how these prompts were generated? _e.g._ manual translation, machine translation, scraped from sources
Also, I think we need some logic to handle whether probes are active based on the `bcp47` value, lest we run a bunch of Arabic strings against English language models (or vice versa).

erickgalinkin:
It also seems that the code here includes `donotanswer` ala #1017 -- is that intended?

PR #1019, opened 2024-11-21
compatible with oobabooga text-generation-web-ui

cycloarcane:
This change adds a modified rest.py that is compatible with oobabooga text generation webui as oobabooga uses nested JSON in its API responses. I added the _get_nested_value method to the rest.py template file. 

An example of the nested JSON response from oobabooga:

```
curl -X POST http://127.0.0.1:5000/v1/completions -H "Content-Type: application/json" -d '{"prompt": "Hello"}'

{"id":"conv-1732215852476260864","object":"text_completion","created":1732215852,"model":"WhiteRabbitNeo_Llama-3.1-WhiteRabbitNeo-2-8B","choices":[{"index":0,"finish_reason":"length","text":" and Happy New Year to all who cruised past this channel during the year gone","logprobs":{"top_logprobs":[{}]}}],"usage":{"prompt_tokens":2,"completion_tokens":17,"total_tokens":19}}%   

```

## Verification

- Supporting configuration such as generator configuration file
``` {
  "rest": {
    "RestGenerator": {
      "name": "oobabooga Local Model",
      "uri": "http://127.0.0.1:5000/v1/completions",
      "method": "post",
      "headers": {
        "Content-Type": "application/json"
      },
      "req_template_json_object": {
        "prompt": "$INPUT",
        "max_tokens": 1000,
        "temperature": 0.7
      },
      "response_json": true,
      "response_json_field": "choices[0].text"
    }
  }
}

```
 I have signed off the commit. Here is the method that was added:
 
 ```
 def _get_nested_value(self, obj, path):
        """Helper method to get nested dictionary values using a path string."""
        try:
            current = obj
            parts = path.replace("][", "].").replace("[", ".").replace("]", "").split(".")
            for part in parts:
                if part.isdigit():
                    current = current[int(part)]
                else:
                    current = current[part]
            return current
        except (KeyError, IndexError, TypeError) as e:
            logging.error(f"Failed to access path '{path}' in object: {e}")
            return None
```

I have tested and run garak successfully with oobabooga text generation webui after making these changes.

### DIFF
--- a/garak/generators/oobabooga.py
+++ b/garak/generators/oobabooga.py
@@ -1,0 +1,278 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""REST API generator interface
+
+Generic Module for REST API connections
+"""
+
+
+import json
+import logging
+from typing import List, Union
+import requests
+
+import backoff
+import jsonpath_ng
+from jsonpath_ng.exceptions import JsonPathParserError
+
+from garak import _config
+from garak.exception import APIKeyMissingError, RateLimitHit
+from garak.generators.base import Generator
+
+
+class RestGenerator(Generator):
+    """Generic API interface for REST models
+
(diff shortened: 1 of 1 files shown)


cycloarcane:
recheck

cycloarcane:

PR #1021, opened 2024-11-22
generator: promote OpenAICompatible as first class generator

### LINKED ISSUE #1008: [Feature Request] Support any OpenAI compatible endpoints by adding two flags
## Summary

Support any OpenAI compatible endpoints, such as tabbyAPI, vLLM, ollama, etc.

I am running Qwen2.5-coder 32B with [tabbyAPI](https://github.com/theroyallab/tabbyAPI) which is a OpenAI comaptible API server. 

Here is what I did to make it work with garak (openai generator):

1. `export OPENAI_BASE_URL="http://localhost:5000/v1"` so that the OpenAI client uses my server
2.  Set the model name of my Qwen2.5 to `gpt-4-32k` (because `gpt-4-32k` is one of the supported models and is [hardcoded to have 32k context](https://github.com/NVIDIA/garak/blob/d8bd12ea969eec3773262419a7cbb2730ce117b4/garak/generators/openai.py#L94), which is the same context length as Qwen2.5 coder) 
3. Run garak with `garak --model_type openai --model_name gpt-4-32k`

It would be nice if garak support arbitary OpenAI models out of the box. 

### Basic example

I suggest adding the following logic:

1. Add `--custom_base_url` and `--context_len` flag; user must uses either both or none of them
2. If `--custom_base_url` is used, initiate the OpenAI client with it. Something like this:
```python
# ...
self.client = openai.OpenAI(api_key=self.api_key, base_url=custom_base_url)
# ...
```
3. Set context length to the value of `--context_len`
4. User run garak with 
```bash
OPENAI_API_KEY=<API key> garak --model_type openai --model_name <model name> --custom_base_url <custom_base_url> --context_len <context_len>
```
For example 
```bash
OPENAI_API_KEY="sk-123XXXXXXXXXXXX" garak --model_type openai --model_name Qwen_Qwen2.5-Coder-32B-Instruct-exl2 --custom_base_url http://localhost:5000/v1 --context_len 32768
```

### Motivation

There is quite a lot of OpenAI compatible API servers out there, supporting them would cover a lot more use cases. 
Also, I think it is more straightforward to setup (compared to the REST generator with has a lot of manual config values).

jmartin-tech:
fix #1008 

Various generators extend this base template, usage and feedback suggest UX and usability improvements will be gained by promoting this class to be a `Configurable` generic OpenAI client based generator.

## Verification

- [ ] Test against via mimic of `nim` configuration via generator config options:
compatible.json:
``` json
{
    "openai": {
        "OpenAICompatible": {
            "uri": "https://integrate.api.nvidia.com/v1/",
            "suppressed_params": ["n", "frequency_penalty", "presence_penalty"]
        }
    }
}
```
``` bash
python -m garak -m openai.OpenAICompatible -n meta/llama3-8b-instruct -p lmrc --generator_option_file compatible.json
```
- [ ] **Verify** at least one additional modified generator `groq` / `openai.OpenAIGenerator` / `nim` with `--parallel_attempts` enabled.

### DIFF
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -122,6 +122,7 @@ class OpenAICompatible(Generator):
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "temperature": 0.7,
         "top_p": 1.0,
+        "uri": "http://localhost:8000/v1/",
         "frequency_penalty": 0.0,
         "presence_penalty": 0.0,
         "seed": None,
@@ -141,13 +142,16 @@ class OpenAICompatible(Generator):
         self._load_client()
 
     def _load_client(self):
-        # Required stub implemented when extending `OpenAICompatible`
-        # should populate self.generator with an openai api compliant object
-        raise NotImplementedError
+        self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
+        if self.name in ("", None):
+            raise ValueError(
+                f"{self.generator_family_name} requires model name to be set, e.g. --model_name org/private-model-name"
+            )
+        self.generator = self.client.chat.completions
 
     def _clear_client(self):
-        # Required stub implemented when extending `OpenAICompatible`
-        raise NotImplementedError
+        self.generator = None
+        self.client = None
 
     def _validate_config(self):
         pass
@@ -257,6 +261,11 @@ class OpenAIGenerator(OpenAICompatible):
     active = True
     generator_family_name = "OpenAI"
 
+    # remove uri as it is not overridable in this class.
+    DEFAULT_PARAMS = {
+        k: val for k, val in OpenAICompatible.DEFAULT_PARAMS.items() if k != "uri"
+    }
+
     def _load_client(self):
         self.client = openai.OpenAI(api_key=self.api_key)
 
@@ -288,10 +297,6 @@ class OpenAIGenerator(OpenAICompatible):
             msg = "'o1'-class models should use openai.OpenAIReasoningGenerator. Try e.g. `-m openai.OpenAIReasoningGenerator` instead of `-m openai`"
             logging.error(msg)
             raise garak.exception.BadGeneratorException("🛑 " + msg)
-
-    def _clear_client(self):
-        self.generator = None
-        self.client = None
 
     def __init__(self, name="", config_root=_config):
         self.name = name
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -142,6 +142,8 @@ class OpenAICompatible(Generator):
         self._load_client()
 
     def _load_client(self):
+        # When extending `OpenAICompatible` this method is a likely location for target application specific
+        # customization and must populate self.generator with an openai api compliant object
         self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
         if self.name in ("", None):
             raise ValueError(
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -114,7 +114,7 @@ class OpenAICompatible(Generator):
 
     ENV_VAR = "OpenAICompatible_API_KEY".upper()  # Placeholder override when extending
 
-    active = False  # this interface class is not active
+    active = True
     supports_multiple_generations = True
     generator_family_name = "OpenAICompatible"  # Placeholder override when extending
 
--- a/tests/generators/test_generators.py
+++ b/tests/generators/test_generators.py
@@ -133,8 +133,12 @@ def test_parallel_requests():
     result = g.generate(prompt="this is a test", generations_this_call=3)
     assert isinstance(result, list), "Generator generate() should return a list"
     assert len(result) == 3, "Generator should return 3 results as requested"
-    assert all(isinstance(item, str) for item in result), "All items in the generate result should be strings"
-    assert all(len(item) > 0 for item in result), "All generated strings should be non-empty"
+    assert all(
+        isinstance(item, str) for item in result
+    ), "All items in the generate result should be strings"
+    assert all(
+        len(item) > 0 for item in result
+    ), "All generated strings should be non-empty"
 
 
 @pytest.mark.parametrize("classname", GENERATORS)
@@ -190,7 +194,6 @@ TESTABLE_GENERATORS = [
         "generators.huggingface.OptimumPipeline",  # model name restrictions and cuda required
         "generators.huggingface.Pipeline",  # model name restrictions
         "generators.langchain.LangChainLLMGenerator",  # model name restrictions
-        "generators.openai.OpenAICompatible",  # template class not intended to ever be `Active`
     ]
 ]
 
--- a/garak/generators/azure.py
+++ b/garak/generators/azure.py
@@ -11,16 +11,22 @@ pass it as an argument to the Generator constructor.
 import os
 import openai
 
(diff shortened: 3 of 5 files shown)


garak/generators/openai.py
```diff
@@ -141,13 +142,16 @@ def __setstate__(self, d) -> object:
         self._load_client()
 
     def _load_client(self):
-        # Required stub implemented when extending `OpenAICompatible`
-        # should populate self.generator with an openai api compliant object
-        raise NotImplementedError
+        self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
```
leondz, line 145: leave guidance for those implementing child classes
```suggestion
        # Required stub implemented when extending `OpenAICompatible`
        # should populate self.generator with an openai api compliant object
        self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
```

garak/generators/openai.py
```diff
@@ -257,6 +261,11 @@ class OpenAIGenerator(OpenAICompatible):
     active = True
     generator_family_name = "OpenAI"
 
+    # remove uri as it is not overridable in this class.
+    DEFAULT_PARAMS = {
+        k: val for k, val in OpenAICompatible.DEFAULT_PARAMS.items() if k != "uri"
+    }
+
```
leondz, line 268: this is confusing given the triggering use-case. we should distinguish between generator for OpenAI API Compatible endpoints, and OpenAI Native endpoints hosted by OpenAI - recommend two distinct classes; perhaps one called openai.Compatible, and another called openai.Native, which is the default generator in the module, that implements this `uri` locking

tests/generators/test_generators.py
```diff
@@ -190,7 +194,6 @@ def test_generator_structure(classname):
         "generators.huggingface.OptimumPipeline",  # model name restrictions and cuda required
         "generators.huggingface.Pipeline",  # model name restrictions
         "generators.langchain.LangChainLLMGenerator",  # model name restrictions
-        "generators.openai.OpenAICompatible",  # template class not intended to ever be `Active`
```
leondz, line 193: nice catch

leondz (changes requested):
let's update class naming. maybe i didn't get the `uri` param locking bit, either

garak/generators/openai.py
jmartin-tech: There is not `requirement` to stub this method in extending classes, however I do like the idea of offering extra guidance.
```suggestion
        # When extending `OpenAICompatible` this method is a likely location for target application specific 
        # customization and must populate self.generator with an openai api compliant object
        self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
```

garak/generators/openai.py
erickgalinkin: I think `openai.Compatible` makes sense but I'd suggest leaving `openai.OpenAIGenerator` as-is. We could also factor `OpenAICompatible` out to another class (maybe to the base class?) in lieu of keeping it in the `openai` generator spec.

leondz: ah, and then `openai.OpenAIReasoning` for the o1 stuff? OK

jmartin-tech: Do we want to support `base.OpenAICompatible` as a valid `--model_type`? I land on a "no" there.

@erickgalinkin, are you thinking we should push the class down to `base.py` and then have `openai.Compatible` just expose the class, this would come with a requirement for import of `openai` and `backoff` in `base.py`:

``` Python
from garak.generators.base import OpenAICompatible
class Compatible(OpenAICompatible):
    ENV_VAR = "OpenAICompatible_API_KEY".upper()

    active = True
    generator_family_name = "OpenAICompatible"
```

Depending on the answer above, this seems like a discussion focused on needing a standard in naming expectations for generator classes. There is currently a mix of including the `model_type` value in the classname or not examples:

```
generators: litellm 🌟
generators: litellm.LiteLLMGenerator
```
```
generators: nim 🌟
generators: nim.NVOpenAIChat
generators: nim.NVOpenAICompletion
generators: nim.Vision
```
```
generators: huggingface 🌟
generators: huggingface.ConversationalPipeline
generators: huggingface.InferenceAPI
generators: huggingface.InferenceEndpoint
generators: huggingface.LLaVA
generators: huggingface.Model
generators: huggingface.OptimumPipeline
generators: huggingface.Pipeline
```
```
generators: octo 🌟
generators: octo.InferenceEndpoint
generators: octo.OctoGenerator
```
```
generators: function 🌟
generators: function.Multiple
generators: function.Single
```

I don't see any consistency to say what should be the preferred convention.

leondz: > Do we want to support `base.OpenAICompatible` as a valid `--model_type`? I land on a "no" there.

Agree, the `base` is confusing - and items in `generators.base` are not functional as-is, iirc; no need to break that pattern.


> Depending on the answer above, this seems like a discussion focused on needing a standard in naming expectations for generator classes. There is currently a mix of including the `model_type` value in the classname or not examples:

Agree. I disprefer repetition. I am missing a word for when the module's default generator is fine named by just the module, though, which is where we often see this.

Reifying:
* avoid repeating module name in generator class name
* strongly avoid repeating module name in non-default generator class name
* avoid including redundant information in class name
* have class name hint at modalities accepted (e.g. Chat, Vision, Completion)

Looking at the examples hinted:

<hr/>

These aren't exposed directly to users, but are overly verbose:
> ```
> generators: litellm.LiteLLMGenerator
> ```
> ```
> generators: octo 🌟
> generators: octo.OctoGenerator
> ```
-- maybe these could become "Chat" or "Completion" or "Text"

<hr/>

These could do with a tidyup:
> ```
> generators: nim 🌟
> generators: nim.NVOpenAIChat
> generators: nim.NVOpenAICompletion
> ```
-- I'm not sure if the NV is redundant, but I expect third-party chat NIMs to continue using this API, and let's wait for a counterexample before working out a general solution. Beyond scope for this PR but could be e.g.
> ```
> generators: nim.Chat
> generators: nim.Completion
> ```

<hr/>

These look pretty good to me, no repetition, and they generally describe the name of the interface using within the context:
> ```
> generators: huggingface 🌟
> generators: huggingface.ConversationalPipeline
> generators: huggingface.InferenceAPI
> generators: huggingface.InferenceEndpoint
> generators: huggingface.LLaVA
> generators: huggingface.Model
> generators: huggingface.OptimumPipeline
> generators: huggingface.Pipeline
> ```
> ```
> generators: function 🌟
> generators: function.Multiple
> generators: function.Single
> ```
> ```
> generators: octo.InferenceEndpoint
> ```
> ```
> generators: nim.Vision
> ```

-- one might rename `huggingface.LLaVA` to `huggingface.VisionLLaVA` for the modality constraint above, if LLaVa isn't considered clue enough.

<hr/>

> I'd suggest leaving openai.OpenAIGenerator as-is. 

@erickgalinkin can I ask what the vibes are like behind this leaning

leondz: now #1026 has landed, is this ready to be updated?

jmartin-tech: Let's sync on a consensus for naming and I can expand this to add a single new migration based on #1026 landing.

jmartin-tech: Defer rename to future revision, this can land with current probe names.

jmartin-tech:
@leondz since we are going to defer the rename, can this get an approval to land?

leondz (approved):
nice

PR #1026, opened 2024-11-27
Feature: enable configuration updates

### LINKED ISSUE #889: add support for loading renamed plugins
e.g. knownbadsignatures became av_spam_scanning

This should be caught in load_plugin and in reporting

jmartin-tech:
fix #889

Provides a path for forward migration of configuration when plugins or other configuration based parameter change between versions.

As an initial capability this acts similar to the `packer fix` command by accepting input of outdated configuration to produce an equivalent configuration for the current version of `garak`. To support _fixing_ a previous configuration tooling is added to create `Migration` classes to be detected and run in order to modify and entires in the configuration based on structural or naming changes in `garak`.

The following process can be used to add new `Migration` to the applied set used to `fix` a configuration:
1. Add a new `Mirgation` file `garak/resources/fixer/` with a filename that formatted
as `YYYYMMDD_some_descriptive_name.py`
2. Implement a new class that extends `garak.resources.fixer.Migration` with an `apply()` method that accepts an existing dictionary and returns a new modified dictionary without any side-effects to the original passed dictionary.
3. If the migration impacts the name of a plugin an `alias` should be added to the plugin class attributes.

When a user calls `garak` with the `--fix` cli parameter the configuration change detected as requiring update will be applied to and an equivalent updated configuration will be output.

Expected action when `--fix` is specified:
* passed one of `--*_option_file` or `--*_options`, json output will be generated if any change is required.
* passed `--config` or no additional configuration based flags yaml will be output if changes are required for the user supplied configuration as well as `garak.site.yaml` if found in the env.

Additional changes:
* expose `options` and `option_file` support for all plugin types
 
### Example usage and outputs:
```
% python -m garak --fix
garak LLM vulnerability scanner v0.10.0.post1 ( https://github.com/NVIDIA/garak ) at 2024-11-27T12:00:09.690263
No revisions applied please verify options provided for `--fix`
```

old.yaml:
``` yaml
plugins:
  probe_spec: lmrc,continuation,replay,tap,knownbadsignatures.EICAR",
  probes:
    continuation:
      ContinueSlursReclaimedSlursMini:
        source_resource_filename: fake_data_file.json
  generators:
    huggingface:
      hf_args:
        torch_dtype: float32
        device: cpu
```
```
% python -m garak --fix --config old.yaml
garak LLM vulnerability scanner v0.10.0.post1 ( https://github.com/NVIDIA/garak ) at 2024-11-27T11:50:47.931551
Updated old.yaml:
plugins:
  generators:
    huggingface:
      hf_args:
        device: cpu
        torch_dtype: float32
  probe_spec: lmrc,continuation,divergence,tap,av_spam_scanning.EICAR",
  probes:
    continuation:
      ContinueSlursReclaimedSlursMini:
        source_resource_filename: fake_data_file.json

```

garak.site.yaml:
``` yaml
plugins:
  probe_spec: replay
  generators:
    huggingface:
      hf_args:
        torch_dtype: float32
        device: cpu
```
```
% python -m garak --fix
garak LLM vulnerability scanner v0.10.0.post1 ( https://github.com/NVIDIA/garak ) at 2024-11-27T11:51:31.658398
Updated /Users/jemartin/.config/garak/garak.site.yaml:
plugins:
  generators:
    huggingface:
      hf_args:
        device: cpu
        torch_dtype: float32
  probe_spec: divergence
```

probe.json
``` json
{
  "continuation": {
    "ContinueSlursReclaimedSlursMini": {
      "source_resource_filename": "fake_data_file.json"
    }
  }
}
```
```
% python -m garak --fix --probe_option_file probe.json
garak LLM vulnerability scanner v0.10.0.post1 ( https://github.com/NVIDIA/garak ) at 2024-11-27T11:52:00.846171
Updated 'probe' configuration:
{
  "continuation": {
    "ContinueSlursReclaimedSlursMini": {
      "source_resource_filename": "fake_data_file.json"
    }
  }
}
```
- [ ] **Verify** `garak.log` reports application of each applicable fix for various configs.

### DIFF
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -484,7 +484,7 @@ def main(arguments=None) -> None:
             if has_changes:
                 exit(1)  # exit with error code to denote changes
             else:
-                print("No revisions applied please verify options provided for `--fix`")
+                print("No revisions applied. Please verify options provided for `--fix`")
         elif args.report:
             from garak.report import Report
 
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -452,7 +452,7 @@ def main(arguments=None) -> None:
             #
             # How should this handle garak.site.yaml? Only if --fix was provided and no other options offered?
             # For now process all files registered a part of the config
-
+            has_changes = False
             if has_config_file_or_json:
                 for plugin_type, plugin_plural in plugin_types:
                     # cli plugins options stub out only a "plugins" sub key
@@ -463,6 +463,7 @@ def main(arguments=None) -> None:
                         }
                         migrated_config = fixer.migrate(cli_config)
                         if cli_config != migrated_config:
+                            has_changes = True
                             msg = f"Updated '{plugin_type}' configuration: \n"
                             msg += json.dumps(
                                 migrated_config["plugins"][plugin_plural], indent=2
@@ -475,11 +476,15 @@ def main(arguments=None) -> None:
                         cli_config = yaml.safe_load(file)
                         migrated_config = fixer.migrate(cli_config)
                         if cli_config != migrated_config:
+                            has_changes = True
                             msg = f"Updated {filename}: \n"
                             msg += yaml.dump(migrated_config)
                             print(msg)
             # should this add support for --*_spec entries passed on cli?
-            exit(1)  # exit with error code to denote changes output?
+            if has_changes:
+                exit(1)  # exit with error code to denote changes
+            else:
+                print("No revisions applied please verify options provided for `--fix`")
         elif args.report:
             from garak.report import Report
 
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -189,8 +189,10 @@ def main(arguments=None) -> None:
         help="list of buffs to use. Default is none",
     )
     # file or json based config options
-    plugin_types = sorted([type.lower() for type in _plugins.PLUGIN_CLASSES])
-    for plugin_type in plugin_types:
+    plugin_types = sorted(
+        zip([type.lower() for type in _plugins.PLUGIN_CLASSES], _plugins.PLUGIN_TYPES)
+    )
+    for plugin_type, _ in plugin_types:
         probe_args = parser.add_mutually_exclusive_group()
         probe_args.add_argument(
             f"--{plugin_type}_option_file",
@@ -383,11 +385,11 @@ def main(arguments=None) -> None:
     try:
         has_config_file_or_json = False
         # do a special thing for CLI probe options, generator options
-        for plugin_type in plugin_types:
+        for plugin_type, plugin_plural in plugin_types:
             opts_cli_config = parse_cli_plugin_config(plugin_type, args)
             if opts_cli_config is not None:
                 has_config_file_or_json = True
-                config_plugin_type = getattr(_config.plugins, f"{plugin_type}s")
+                config_plugin_type = getattr(_config.plugins, plugin_plural)
 
                 config_plugin_type = _config._combine_into(
                     opts_cli_config, config_plugin_type
@@ -452,16 +454,18 @@ def main(arguments=None) -> None:
             # For now process all files registered a part of the config
 
             if has_config_file_or_json:
-                for plugin_type in plugin_types:
+                for plugin_type, plugin_plural in plugin_types:
                     # cli plugins options stub out only a "plugins" sub key
                     plugin_cli_config = parse_cli_plugin_config(plugin_type, args)
                     if plugin_cli_config is not None:
-                        cli_config = {"plugins": plugin_cli_config}
+                        cli_config = {
+                            "plugins": {f"{plugin_plural}": plugin_cli_config}
+                        }
                         migrated_config = fixer.migrate(cli_config)
                         if cli_config != migrated_config:
                             msg = f"Updated '{plugin_type}' configuration: \n"
                             msg += json.dumps(
-                                migrated_config["plugins"], indent=2
+                                migrated_config["plugins"][plugin_plural], indent=2
                             )  # pretty print the config in json
                             print(msg)
             else:
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -3,7 +3,37 @@
 
 """Flow for invoking garak from the command line"""
 
-command_options = "list_detectors list_probes list_generators list_buffs list_config plugin_info interactive report version".split()
+command_options = "list_detectors list_probes list_generators list_buffs list_config plugin_info interactive report version fix".split()
+
+
+def parse_cli_plugin_config(plugin_type, args):
+    import os
+    import json
+    import logging
+
+    opts_arg = f"{plugin_type}_options"
+    opts_file = f"{plugin_type}_option_file"
+    opts_cli_config = None
+    if opts_arg in args or opts_file in args:
+        if opts_arg in args:
+            opts_argv = getattr(args, opts_arg)
+            try:
+                opts_cli_config = json.loads(opts_argv)
+            except json.JSONDecodeError as e:
+                logging.warning("Failed to parse JSON %s: %s", opts_arg, e.args[0])
+
+        elif opts_file in args:
+            file_arg = getattr(args, opts_file)
+            if not os.path.isfile(file_arg):
+                raise FileNotFoundError(f"Path provided is not a file: {opts_file}")
+            with open(file_arg, encoding="utf-8") as f:
+                options_json = f.read().strip()
+            try:
+                opts_cli_config = json.loads(options_json)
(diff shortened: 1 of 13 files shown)


garak/cli.py
```diff
@@ -38,6 +68,7 @@ def main(arguments=None) -> None:
         prog="python -m garak",
         description="LLM safety & security scanning tool",
         epilog="See https://github.com/NVIDIA/garak",
+        allow_abbrev=False,
```
jmartin-tech, line 71: Since this PR adds support for all _plugin_types_ to offer an options json or file the parser needs to disambiguate `--buffs` from `--buffs_option_file` and `--buffs_options`.

This also exposed a test that relied on abbreviation that has been update in the PR as well.

garak/probes/av_spam_scanning.py
leondz, line 1: what uses `aliases`? same comment for other probes

garak/resources/fixer/20240628_gcg_rename.py
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from garak.resources.fixer import Migration
+from garak.resources.fixer import _plugin
+
+
+class RenameGCG(Migration):
+    def apply(config_dict: dict) -> dict:
+        """Rename probe gcg -> suffix"""
```
leondz, line 10: not probe but probe family / plugin module

garak/resources/fixer/20240822_knownbadsignatures_rename.py
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from garak.resources.fixer import Migration
+from garak.resources.fixer import _plugin
+
+
+class RenameKnownbadsignatures(Migration):
+    def apply(config_dict: dict) -> dict:
+        """Rename probe knownbadsignatures -> av_spam_scanning"""
```
leondz, line 10:
```suggestion
        """Rename probe family knownbadsignatures -> av_spam_scanning"""
```

garak/resources/fixer/20241011_replay_rename.py
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from garak.resources.fixer import Migration
+from garak.resources.fixer import _plugin
+
+
+class RenameReplay(Migration):
+    def apply(config_dict: dict) -> dict:
+        """Rename probe replay -> divergence"""
```
leondz, line 10:
```suggestion
        """Rename probe family replay -> divergence"""
```

garak/resources/fixer/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+# list of migrations, should this be dynamically built from the package?
+ordered_migrations = []
+root_path = Path(__file__).parents[0]
+for module_filename in sorted(os.listdir(root_path)):
+    if not module_filename.endswith(".py"):
+        continue
+    if module_filename.startswith("__"):
+        continue
+    module_name = module_filename.replace(".py", "")
```
leondz, line 31: would prefer a location-sensitive replace, e.g. re.sub("\.py", "", module_filename)

garak/resources/fixer/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    updated_config = copy.deepcopy(original_config)
+    for migration in ordered_migrations:
+        new_config = migration.apply(updated_config)
+        if new_config != updated_config:
+            updated_config = new_config
+            msg = f"Applied migrations changes from {migration.__name__}"
+            logging.info(msg)
+
+    if original_config != updated_config:
+        logging.info("Migration preformed")
```
leondz, line 53:
```suggestion
        logging.info("Migration performed")
```

leondz (approved):
coupla minor points, looks good in general

garak/cli.py
```diff
@@ -429,6 +431,60 @@ def main(arguments=None) -> None:
...
+                        if cli_config != migrated_config:
+                            has_changes = True
+                            msg = f"Updated {filename}: \n"
+                            msg += yaml.dump(migrated_config)
+                            print(msg)
+            # should this add support for --*_spec entries passed on cli?
+            if has_changes:
+                exit(1)  # exit with error code to denote changes
+            else:
+                print("No revisions applied please verify options provided for `--fix`")
```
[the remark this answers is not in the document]
jmartin-tech: Worth discussing, Can you elaborate on reasons? 

I am currently of the option that this is `config` related the primary role of `cli` code is parsing config. `command` so far, aligns with items that should be possible to call from `interactive` mode. This functionality may more may not be valuable in that context and uses the parsed args array in `parse_cli_plugin_config` a way that suggests this is currently the more appropriate location to maintain encapsulation. A future refactor may extract some config parsing from `cli` and change this view.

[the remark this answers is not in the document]
leondz: to my mind:
* `cli` is ideally items only related to processing cli input
* `command` contains encapsulation of functions / activities garak can perform
* orchestration of runs is in a grey zone pending better encapsulation, that will be resolved with an API design, and I prefer `command` for this, seeing as run orchestration subfunctions are here already

following the first point, any items not related solely to command line interface, shouldn't really be in `cli`. this code block contains code that isn't about processing cli input. so, it's better off outside the `cli` module

ideally, i think it should be entirely possible to access all of garak's functionality without using `garak.cli`. so that in effect, `cli` provides only an interface to garak proper. these config updates are not intrinsic to the command line interface, and one would conceivably like to be able to access them via other routes as & when those open up.

garak/probes/av_spam_scanning.py
leondz: ok, that's cool. we might like to choose now how diligent we intend to be about this - are `aliases` only given as indications, or do we guarantee that they're exhaustive?

there's a duplication of info here (because it's already given in `fixer` modules) that i'm mindful of

garak/cli.py
```diff
@@ -429,6 +431,60 @@ def main(arguments=None) -> None:
...
+                            "plugins": {f"{plugin_plural}": plugin_cli_config}
+                        }
+                        migrated_config = fixer.migrate(cli_config)
+                        if cli_config != migrated_config:
+                            has_changes = True
+                            msg = f"Updated '{plugin_type}' configuration: \n"
+                            msg += json.dumps(
+                                migrated_config["plugins"][plugin_plural], indent=2
+                            )  # pretty print the config in json
+                            print(msg)
```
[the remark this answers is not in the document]
leondz: i thought of `stderr` too, it's helpful to be able to distinguish the streams. on the other hand it sounds a bit surprising.

if we were really wild we could issue ANSI codes to delete the preceding banner output, thus enabling `stdout` redirection. this seems like it might be a maintenance time-sink

maybe the most prudent solution is to wait for a use-case to emerge that prefers redirection? i don't have a wonderful solution to mind rn - rewriting configs in place is also hard to keep track of

erickgalinkin (commented):
Very into this as a clean way to help manage deprecation. Awesome work here, @jmartin-tech

garak/probes/av_spam_scanning.py
jmartin-tech: Agreed I tried to note in the description this becomes a `required` step when adding a migration that impacts a plugin name however there is no enforcement/early warning mechanism currently in place, we could add a `lint` like action that checks for plugin file or class renames as a warning/hint on PRs.

garak/cli.py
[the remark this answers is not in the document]
leondz: huh, yeah, i see what you mean - good point. i'm fine with taking the path of least resistance and leaving this as-is, in that case

leondz:
What's going on with the unmatched `",` at the end of `old.yaml` above? And I guess, seeing as it's retained, then `fixer` doesn't validate configs, only migrate what's already there (this is OK, just making the behaviour explicit)

jmartin-tech:
> What's going on with the unmatched ",

You have a good point, the `--fix` options does not currently validate the original or end config at this time, it only requires the inputs be valid `yaml` of `json` and will not correct items a `garak` run may consider syntax errors. I agree this should be _known_ and to future troubleshooting purposes. Might be worth filing an issue documenting a possible future need for pre execution configuration validation support in the future.

Issue #984, opened 2024-11-11
feature: auto-eval


leondz:
generate reporting and analysis artefacts from scans

* use probe tier list to determine how to treat each probe
* use target absolute & z-scores
* generate `k` sample prompt:response pairs for each probe to be examined
* generate sample model card text for low score items
* add model card text to HTML report

PR #1043, opened 2024-12-13
Feature/autoeval

erickgalinkin:
Partial fulfillment of #984


garak/analyze/deep_analysis/analytics.py
```diff
@@ -0,0 +1,264 @@
...
+        minus_sigma_scores = [
+            probe_data[probe]["mu"] - probe_data[probe]["sigma"]
+            for probe in v
+            if probe in probe_data.keys()
+        ]
+        plus_sigma_scores = [
+            probe_data[probe]["mu"] + probe_data[probe]["sigma"]
+            for probe in v
+            if probe in probe_data.keys()
+        ]
```
leondz, line 159: what do these represent? the ±1 s.d. bounds of an average model's score for a given probe:detector pair?

garak/analyze/deep_analysis/analytics.py
```diff
@@ -0,0 +1,264 @@
...
+
+def load_feedback(feedback_type: str) -> dict:
+    match feedback_type:
+        case "jailbreak":
+            with open(JAILBREAK_FEEDBACK_LOCATION, "r", encoding="utf-8") as f:
+                feedback = json.loads(f.read().strip())
+
+        case _:
+            feedback = dict()
+    return feedback
```
leondz, line 179: can we use feedback files named according to a group and just skip if file not found, allowing us to drop the case statements?

garak/analyze/deep_analysis/deep_analysis.py
```diff
@@ -0,0 +1,126 @@
...
+    return feedback_string
+
+
+def tier_2(analysis_dict: dict) -> str:
+    pass
+
+
+def deep_analysis(report_path, bag_path=ANALYSIS_FILE) -> Tuple[str, str]:
+    """
+    Take garak report jsonl file and perform qualitative analysis on the probe results for the target.
```
leondz, line 65: `perform qualitative analysis`

It's all quantitative right? We compare quantities and choose blocks of text, no human in the loop, no qualitative method

PR #1047, opened 2024-12-17
Feature: use hf chat support

### LINKED ISSUE #1029: Allow applying prompt template to the input for huggingface models.
## Summary

The current Garak treats all huggingface model as base models and does not consider the prompt template. However, for those instruct tuned models, templates are necessary to let them generate proper responses. So I am wondering whether it will be good if we can add a parameter in the yaml config file to configure whether we apply the prompt template to the input to distinguish instruct model with base model. 

### Basic example

If the proposal involves a new or changed API, include a basic code example. Omit this section if it's not applicable.

### Motivation

Why are we doing this? What use cases does it support? What is the expected outcome?

Insturct models need prompt template to maximize its performance and users also use prompt template to interact with those models, so allowing prompt template will simulate the real use scenario better.

jmartin-tech:
Fix #1029 

When a huggingface model supports a chat_template wrap the prompt in a chat list to enable request formatting. Using the trained template formatting can have significant impact on the inference efficiency as well the context of the response. This aligns local model execution with deployed scenarios.

* shift `truncation` requirement into pipeline creation call
* apply chat templates when available
* remove template as prefix if found in output for `huggingface.Model` inference

As a future task I could see possibly wanting to suppress `chat_template` formatting to test poorly configured inference against a instruct model.  This idea is left as an exercise for future consideration.

## Verification

- [ ] `python -m garak -m huggingface -n mistralai/Mistral-7B-Instruct-v0.1 -p atkgen`
- [ ] `python -m garak -m huggingface -n mistralai/Mistral-7B-Instruct-v0.1 -p lmrc`
- [ ] `python -m garak -m huggingface -n gpt2 -p lmrc`
- [ ] `python -m garak -m huggingface.Model -n mistralai/Mistral-7B-Instruct-v0.1 -p atkgen`
- [ ] `python -m garak -m huggingface.Model -n mistralai/Mistral-7B-Instruct-v0.1 -p lmrc`
- [ ] `python -m garak -m huggingface.Model -n gpt2 -p lmrc`

### DIFF
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -79,6 +79,9 @@ class Pipeline(Generator, HFCompatible):
             set_seed(_config.run.seed)
 
         pipeline_kwargs = self._gather_hf_params(hf_constructor=pipeline)
+        pipeline_kwargs["truncation"] = (
+            True  # this is forced to maintain existing pipeline expectations
+        )
         self.generator = pipeline("text-generation", **pipeline_kwargs)
         if self.generator.tokenizer is None:
             # account for possible model without a stored tokenizer
@@ -113,27 +116,16 @@ class Pipeline(Generator, HFCompatible):
             warnings.simplefilter("ignore", category=UserWarning)
             try:
                 with torch.no_grad():
-                    # workaround for pipeline to truncate the input
-
                     # according to docs https://huggingface.co/docs/transformers/main/en/chat_templating
                     # chat template should be automatically utilized if the pipeline tokenizer has support
+                    # and a properly formatted list[dict] is supplied
                     if self.use_chat:
-                        formatted_prompt = self.generator.tokenizer.apply_chat_template(
-                            self._format_chat_prompt(prompt),
-                            tokenize=False,
-                            add_generation_prompt=True,
-                        )
+                        formatted_prompt = self._format_chat_prompt(prompt)
                     else:
                         formatted_prompt = prompt
 
-                    encoded_prompt = self.generator.tokenizer(
-                        formatted_prompt, truncation=True
-                    )
-                    truncated_prompt = self.generator.tokenizer.decode(
-                        encoded_prompt["input_ids"], skip_special_tokens=True
-                    )
                     raw_output = self.generator(
-                        truncated_prompt,
+                        formatted_prompt,
                         pad_token_id=self.generator.tokenizer.eos_token_id,
                         max_new_tokens=self.max_tokens,
                         num_return_sequences=generations_this_call,
@@ -148,11 +140,15 @@ class Pipeline(Generator, HFCompatible):
                 i["generated_text"] for i in raw_output
             ]  # generator returns 10 outputs by default in __init__
 
-        if not self.deprefix_prompt and not self.use_chat:
-            return outputs
+        if self.use_chat:
+            text_outputs = [_o[-1]["content"].strip() for _o in outputs]
         else:
-            # consider using formatted_prompt in removal as a `list` or `str`
-            return [re.sub("^" + re.escape(formatted_prompt), "", _o) for _o in outputs]
+            text_outputs = outputs
+
+        if not self.deprefix_prompt:
+            return text_outputs
+        else:
+            return [re.sub("^" + re.escape(prompt), "", _o) for _o in text_outputs]
 
 
 class OptimumPipeline(Pipeline, HFCompatible):
@@ -520,7 +516,7 @@ class Model(Pipeline, HFCompatible):
         if self.top_k is not None:
             self.generation_config.top_k = self.top_k
 
-        text_output = []
+        raw_text_output = []
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=UserWarning)
             with torch.no_grad():
@@ -537,6 +533,10 @@ class Model(Pipeline, HFCompatible):
                     formatted_prompt, truncation=True, return_tensors="pt"
                 ).to(self.device)
 
+                prefix_prompt = self.tokenizer.decode(
+                    inputs["input_ids"][0], skip_special_tokens=True
+                )
+
                 try:
                     outputs = self.model.generate(
                         **inputs, generation_config=self.generation_config
@@ -549,17 +549,22 @@ class Model(Pipeline, HFCompatible):
                         return returnval
                     else:
                         raise e
-                text_output = self.tokenizer.batch_decode(
+                raw_text_output = self.tokenizer.batch_decode(
                     outputs, skip_special_tokens=True, device=self.device
                 )
 
-        if not self.deprefix_prompt and not self.use_chat:
+        if self.use_chat:
+            text_output = [
+                re.sub("^" + re.escape(prefix_prompt), "", i).strip()
+                for i in raw_text_output
+            ]
+        else:
+            text_output = raw_text_output
+
+        if not self.deprefix_prompt:
             return text_output
         else:
-            # consider using formatted_prompt in removal as a `list` or `str`
-            return [
-                re.sub("^" + re.escape(formatted_prompt), "", i) for i in text_output
-            ]
+            return [re.sub("^" + re.escape(prefix_prompt), "", i) for i in text_output]
 
 
 class LLaVA(Generator, HFCompatible):
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -148,10 +148,11 @@ class Pipeline(Generator, HFCompatible):
                 i["generated_text"] for i in raw_output
             ]  # generator returns 10 outputs by default in __init__
 
-        if not self.deprefix_prompt:
+        if not self.deprefix_prompt and not self.use_chat:
             return outputs
         else:
-            return [re.sub("^" + re.escape(prompt), "", _o) for _o in outputs]
+            # consider using formatted_prompt in removal as a `list` or `str`
+            return [re.sub("^" + re.escape(formatted_prompt), "", _o) for _o in outputs]
 
 
 class OptimumPipeline(Pipeline, HFCompatible):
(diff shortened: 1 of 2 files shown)


garak/generators/huggingface.py
```diff
@@ -492,14 +516,27 @@ def _call_model(
...
+                        tokenize=False,
+                        add_generation_prompt=True,
+                    )
+                else:
+                    formatted_prompt = prompt
+
                 inputs = self.tokenizer(
-                    prompt, truncation=True, return_tensors="pt"
+                    formatted_prompt, truncation=True, return_tensors="pt"
                 ).to(self.device)
```
jmartin-tech, line 534: Note while this call to the tokenizer may be possible to optimize out a direct call to `self.tokenizer.apply_chat_template()` with `return_tensors` passed returns a different format that cannot be directly passed to `self.model.generate`, by taking this extra step the code is more consistent with and without chat support on the model.

garak/generators/huggingface.py
```diff
@@ -79,6 +79,9 @@ def _load_client(self):
             set_seed(_config.run.seed)
 
         pipeline_kwargs = self._gather_hf_params(hf_constructor=pipeline)
+        pipeline_kwargs["truncation"] = (
```
erickgalinkin, line 82: Sometimes I wish this were just the default for HF but c'est la vie. We may want to consider whether `truncation = True` requires `max_len` to be set -- I've had HF yell at me for not specifying both but it may be a corner case that I encountered.

garak/generators/huggingface.py
```diff
@@ -106,13 +116,16 @@ def _call_model(
...
-                    truncated_prompt = self.generator.tokenizer.decode(
-                        encoded_prompt["input_ids"], skip_special_tokens=True
-                    )
+                    # according to docs https://huggingface.co/docs/transformers/main/en/chat_templating
+                    # chat template should be automatically utilized if the pipeline tokenizer has support
+                    # and a properly formatted list[dict] is supplied
+                    if self.use_chat:
+                        formatted_prompt = self._format_chat_prompt(prompt)
+                    else:
+                        formatted_prompt = prompt
```
erickgalinkin, line 125: Note for future Erick and Jeffrey: we will want to merge `ConversationalPipeline` into this at some point given HF's deprecation of the `"conversational"` pipeline and `Conversation` object. At that point, we'll also want to validate that `prompt` is a string.

garak/generators/huggingface.py
erickgalinkin: grumble grumble but yes I support you

garak/generators/huggingface.py
```diff
@@ -512,14 +549,22 @@ def _call_model(
...
                     outputs, skip_special_tokens=True, device=self.device
                 )
 
+        if self.use_chat:
+            text_output = [
+                re.sub("^" + re.escape(prefix_prompt), "", i).strip()
+                for i in raw_text_output
+            ]
+        else:
+            text_output = raw_text_output
```
erickgalinkin, line 562: Part of me (albeit a part with limited convictions) feels like there HAS to be a better way to handle this. HF seems to be managing it internally with their `pipeline`s, but this seems fine for the time being.

tests/generators/test_huggingface.py
```diff
@@ -121,6 +133,18 @@ def test_model(hf_generator_config):
...
+def test_model_chat(mocker, hf_generator_config):
+    # uses a ~350M model with chat support
+    g = garak.generators.huggingface.Model(
+        "microsoft/DialoGPT-small", config_root=hf_generator_config
+    )
+    mock_format = mocker.patch.object(
+        g, "_format_chat_prompt", wraps=g._format_chat_prompt
+    )
+    g.generate("Hello world!")
+    mock_format.assert_called_once()
```
erickgalinkin, line 145: This is a great test for where we are now -- would be nice to have something where we could specifically test the deprefixing but there is some unavoidable stochasticity here that I'm not sure how to account for.

erickgalinkin (approved):
This all looks good, and we went over it in some depth ahead of the PR. Approved, with the caveat that I'm putting some faith in this passing scream tests.

tests/generators/test_huggingface.py
jmartin-tech: Fair point, in the vein of wanting less reliance on actual models I think adding a test in the future that mocks a model response completely by offering a known test result should be a reasonable way to approach adding a `unit` test for validation of `deprefix` function.

garak/generators/huggingface.py
```diff
@@ -87,6 +90,10 @@ def _load_client(self):
             self.generator.tokenizer = AutoTokenizer.from_pretrained(
                 pipeline_kwargs["model"]
             )
+        self.use_chat = (
+            hasattr(self.generator.tokenizer, "chat_template")
+            and self.generator.tokenizer.chat_template is not None
+        )
```
jmartin-tech, line 96: Should we expand this to allow the user to override `use_chat`?

This would defer to the `Configurable` loaded value if provided via config:
```suggestion
        if not hasattr(self, "use_chat"):
            self.use_chat = (
                hasattr(self.generator.tokenizer, "chat_template")
                and self.generator.tokenizer.chat_template is not None
            )
```

garak/generators/huggingface.py
```diff
@@ -127,10 +140,15 @@ def _call_model(
                 i["generated_text"] for i in raw_output
             ]  # generator returns 10 outputs by default in __init__
 
+        if self.use_chat:
+            text_outputs = [_o[-1]["content"].strip() for _o in outputs]
+        else:
+            text_outputs = outputs
+
```
leondz, line 147: Not for this PR - there are multiple ways of representing conversations in garak, and attempt should be canonical for conversation history; I'm starting to consider patterns where attempt holds the conversation history still but where this can be read & written using other interfaces, like an OpenAI API messages dict list, or this HF style format. But mayeb the work is so simple that this current pattern works fine.

garak/generators/huggingface.py
```diff
@@ -468,6 +486,12 @@ def _load_client(self):
                 self.name, padding_side="left"
             )
 
+        # test tokenizer for `apply_chat_template` support
+        self.use_chat = (
+            hasattr(self.tokenizer, "chat_template")
+            and self.tokenizer.chat_template is not None
+        )
+
```
leondz, line 494: This logic is in a few places, is it worth factoring up to `HFCompatible`?

garak/generators/huggingface.py
```diff
@@ -492,14 +516,27 @@ def _call_model(
...
 
-        text_output = []
+        raw_text_output = []
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=UserWarning)
             with torch.no_grad():
+                if self.use_chat:
+                    formatted_prompt = self.tokenizer.apply_chat_template(
+                        self._format_chat_prompt(prompt),
+                        tokenize=False,
```
leondz, line 526: reading the HF docs this recommended use 'tokenize' smells of insufficient encapsulation, but lgtm for our implementation

garak/generators/huggingface.py
leondz: i like exerting this level of control over when tokenisation happens

garak/generators/huggingface.py
leondz: idk man the format uses `\n` as a special char but doesn't seem to do mush escaping, many bets are off

Another option could be to do `\n`-based escaping and parsing of messages line by line, assuming `[A-Z][a-z]+:` is a metadata line

My experience also is that things like `prefix_prompt` get mangled in other code sometimes. I don't have great suggestions for how to check this well, other than:
* continue doing qualitative reviews of prompt:output pairs for suspicious items;
* migrate to a data format with clearer separation between data and metadata;
* badger HF for a reference implementation;
* give notice to users that HF Chat is "janky".

tests/generators/test_huggingface.py
leondz: Nice find re: the model. I wonder if we could use something like the `lipsum` module (which we already depend upon) to mock up a test HF generator that could be used in like Many of our tests. Maybe for jan?

tests/generators/test_huggingface.py
```diff
@@ -121,6 +133,18 @@ def test_model(hf_generator_config):
...
 
+def test_model_chat(mocker, hf_generator_config):
+    # uses a ~350M model with chat support
+    g = garak.generators.huggingface.Model(
+        "microsoft/DialoGPT-small", config_root=hf_generator_config
+    )
+    mock_format = mocker.patch.object(
+        g, "_format_chat_prompt", wraps=g._format_chat_prompt
+    )
+    g.generate("Hello world!")
```
leondz, line 144: Can we check if the output is coming in the desired format? This would then validate the code better

leondz (approved):
Thanks a lot for this. A few questions around testing and refactoring, but LGTM otherwise.

garak/generators/huggingface.py
```diff
@@ -87,6 +90,10 @@ def _load_client(self):
             self.generator.tokenizer = AutoTokenizer.from_pretrained(
                 pipeline_kwargs["model"]
             )
+        self.use_chat = (
+            hasattr(self.generator.tokenizer, "chat_template")
+            and self.generator.tokenizer.chat_template is not None
+        )
         if not hasattr(self, "deprefix_prompt"):
             self.deprefix_prompt = self.name in models_to_deprefix
```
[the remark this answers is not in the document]
jmartin-tech: Your read is the intent. Currently I am treating the options as independent. `use_chat` is about expectations related to trained behavior supported by the model for input and output format, and `deprefix` is about expected behavior for the model response format. For this reason and since we `strictly` remove the each value only if it is matched exactly at the beginning of the intermediary output string this seems a reasonable treatment.

garak/generators/huggingface.py
jmartin-tech: the object format is slightly different `self.tokenizer` vs `self.generator.tokenizer` there is an abstraction possible here for `_supports_chat(tokenizer)` though not sold on the value yet:

``` python
def _supports_chat(self, tokenizer) -> bool:
    return hasattr(tokenizer, "chat_template") and tokenizer.chat_template is not None
```

I would like to defer this until I see a third usage or usage outside this module.

Issue #913, opened 2024-09-13
support JSON config as well as YAML


leondz:
## Summary
YAML isn't a universally powerful and flawless format. Consider also supporting JSON for standard configs.

### Basic example

E.g. [`fast.yaml`](https://github.com/leondz/garak/blob/main/garak/configs/fast.yaml) would become:

```
{
  "system": {
    "parallel_attempts": 20,
    "lite": true
  },
  "run": {
    "generations": 5
  },
  "plugins": {
    "probe_spec": "continuation,dan,encoding.InjectBase64,encoding.InjectHex,goodside,av_spam_scanning,leakreplay,lmrc,malwaregen.SubFunctions,malwaregen.TopLevel,packagehallucination,realtoxicityprompts.RTPIdentity_Attack,realtoxicityprompts.RTPProfanity,realtoxicityprompts.RTPSexually_Explicit,realtoxicityprompts.RTPThreat,snowball,xss",
    "extended_detectors": false
  }
}
```

### Motivation

> Why are we doing this? What use cases does it support? What is the expected outcome?

1. YAML lacks some definitional strength, making certain values difficult to express, and an alternative is useful
2. The internal config structure is agnostic to which text format the config is stored in, hopefully keeping the scope of code changes relatively limited
3. It's good to provide us with flexibility around something as core and powerful as configs
4. JSON schema validation is trivial and built-in to python, enabling strong config checking pre-run 
5. We will gain a quick, reasonable answer to the various gripes about yaml that I (and presumably also the world) have, like how the official YAML implementation in Python doesn't follow the official YAML spec

patriciapampanelli:
@leondz Should site config and core config support JSON format? Should files without extensions be allowed?

leondz:
Yes, I think that makes sense - it seems hard to justify excluding them

patriciapampanelli:
Do you think it makes sense that, in cases where two files in the same folder share the same name but different extensions (`.json` and `.yaml`), we should prioritize the JSON version? I can add this clarification to the documentation if that sounds reasonable.

leondz:
Oh, great question. I'd be inclined to prioritise JSON - @jmartin-tech do you have an opinion?

jmartin-tech:
I would suggest automatically utilized files should stick to one format and support for others format requires explicit filename with extension.

[...]

PR #1049, opened 2024-12-20
Warn if api_key in Config

### LINKED ISSUE #927: security: warn if `api_key` given in config and config is world-readable
## Summary

If a parsed config file contains an `api_key` value, test if that config file is world-readable. If it is, print a warning.

### Basic example

config.yaml, chmod 777
```
---
plugins:
  generators:
    openai:
      api_key: oh no
```

output:
```
$ garak --config config.yaml
🛑 the specified config "config.yaml" contains a secret but can be read by anyone. change its permissions.
```

### Motivation

Help our users be less insecure

erickgalinkin:
Fixes #927 

Look for the presence of `api_key` as a key when loading config files, raise a warning to the user if the file is readable by group or others.

## Verification

``` yaml
---
run:
    seed:
    eval_threshold: 0.5
    generations: 1
    probe_tags:

plugins:
    model_type: test
    model_name: Lipsum
    detector_spec: auto
    api_key: test
    probe_spec: test.Test

reporting:
    report_prefix: test
```

run `garak --config config.yaml`

NOTE: The warning does not get surfaced to the user directly, which shouldn't be the case when we warn, but it is written to the log. This is an artifact of our default verbosity and it may make sense to `print` the warning here.

### DIFF
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -21,7 +21,6 @@ from xdg_base_dirs import (
     xdg_config_home,
     xdg_data_home,
 )
-from garak.exception import ConfigSecretWarning
 
 DICT_CONFIG_AFTER_LOAD = False
 
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -7,6 +7,7 @@
 
 # logging should be set up before config is loaded
 
+import platform
 from collections import defaultdict
 from dataclasses import dataclass
 import importlib
@@ -155,14 +156,21 @@ def _load_yaml_config(settings_filenames) -> dict:
             settings = yaml.safe_load(settings_file)
             if settings is not None:
                 if _key_exists(settings, "api_key"):
-                    logging.info(f"API key found in {settings_filename}. Checking readability...")
-                    res = os.stat(settings_filename)
-                    if res.st_mode & stat.S_IROTH or res.st_mode & stat.S_IRGRP:
-                        msg = (f"A possibly secret value (`api_key`) was detected in {settings_filename}, "
-                               f"which is readable by users other than yourself. "
-                               f"Consider changing permissions on this file to only be readable by you.")
+                    if platform.system() == "Windows":
+                        msg = (f"A possibly secret value (`api_key`) was detected in {settings_filename}. "
+                               f"We recommend removing potentially sensitive values from config files or "
+                               f"ensuring the file is readable only by you.")
                         logging.warning(msg)
                         print(f"⚠️  {msg}")
+                    else:
+                        logging.info(f"API key found in {settings_filename}. Checking readability...")
+                        res = os.stat(settings_filename)
+                        if res.st_mode & stat.S_IROTH or res.st_mode & stat.S_IRGRP:
+                            msg = (f"A possibly secret value (`api_key`) was detected in {settings_filename}, "
+                                   f"which is readable by users other than yourself. "
+                                   f"Consider changing permissions on this file to only be readable by you.")
+                            logging.warning(msg)
+                            print(f"⚠️  {msg}")
                 config = _combine_into(settings, config)
     return config
 
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -117,6 +117,15 @@ run.seed = None
 # placeholder
 # generator, probe, detector, buff = {}, {}, {}, {}
 
+def _key_exists(d: dict, key: str) -> bool:
+    # Check for the presence of a key in a nested dict.
+    if not isinstance(d, dict):
+        return False
+    if key in d.keys():
+        return True
+    else:
+        return any([_key_exists(val, key) for val in d.values()])
+
 
 def _set_settings(config_obj, settings_obj: dict):
     for k, v in settings_obj.items():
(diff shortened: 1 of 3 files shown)


garak/_config.py
```diff
@@ -115,14 +116,23 @@ def _nested_dict():
 # placeholder
 # generator, probe, detector, buff = {}, {}, {}, {}
 
+def _key_exists(d: dict, key: str) -> bool:
+    # Check for the presence of a key in a nested dict.
+    if not isinstance(d, dict):
+        return False
```
jmartin-tech, line 122: I don't think there are not any current places where this would occur, however should also process any `list` that may contain a `dict` objects?

garak/_config.py
```diff
@@ -141,6 +151,11 @@ def _load_yaml_config(settings_filenames) -> dict:
         with open(settings_filename, encoding="utf-8") as settings_file:
             settings = yaml.safe_load(settings_file)
             if settings is not None:
+                if _key_exists(settings, "api_key"):
+                    logging.info(f"API key found in {settings_filename}. Checking readability...")
+                    res = os.stat(settings_filename)
+                    if res.st_mode & stat.S_IROTH or res.st_mode & stat.S_IRGRP:
```
jmartin-tech, line 157: One minor thought, should skip the warning if the `group` is the user?

Consider:
```
-rw-r-----  myuser domain-users
```
vs
```
-rw-r-----  myuser myuser
```

jmartin-tech (approved):
Looks reasonable to me, a couple minor enhancement ideas.

garak/_config.py
leondz: Alternatively one could go the other way and `raise ValueError("_key_exists() can only process dicts")`. I think it's probably beneficial to pick one way (dicts+lists of dicts) or the other (dicts or gtfo)

garak/_config.py
leondz: Is there a guaranteed way of doing this? Or would the check be to skip iff groupname==username and uid==gid? Like

```
import grp
import pwd
```

then

```suggestion
                    if (res.st_mode & stat.S_IROTH) or res.st_mode & stat.S_IRGRP and not ( 
                        res.st_uid == res.st_gid and
                        grp.getgrgid(res.st_gid).gr_name == pwd.getpwuid(res.st_uid).pw_name 
                    ) :
```

Trying to think about how to engineer this even more but coming up blank -- for now. Maybe after christmas

garak/_config.py
```diff
@@ -141,6 +151,11 @@ def _load_yaml_config(settings_filenames) -> dict:
...
+                        logging.warn(f"A possibly secret value (`api_key`) was detected in {settings_filename}, which is readable by users other than yourself.")
```
leondz, line 158: You know, to get this surfaced to CLI users, we could add this near the top:

```
import garak.command
```

and then do something like:

```suggestion
                        msg = f"A possibly secret value (`api_key`) was detected in {settings_filename}, which is readable by users other than yourself. Consider changing permissions on this file to only be readable by you."
                        logging.warn(msg)
                        command.hint(msg)
```

leondz (approved):
This is pretty cool, made some optional suggestions

garak/_config.py
erickgalinkin: I like enforcing dicts.

erickgalinkin: While I like the idea of enforcing `dict`, the way this is written is looking for something at the root which is a value and returns `False` at the base of the nested dict. I don't think we should be processing `list` objects here at all, since we're explicitly looking for a key that will end up being used by the config.

garak/_config.py
erickgalinkin: I'm not sure there's a way for us to do this, especially since there's no way to guarantee other users aren't in the group without enumerating those?

garak/_config.py
jmartin-tech: My concern would be if some future plugin offered something like:

``` yaml
detectors:
  fake:
    MultiCallDetector:
      endpoints:
        - uri: https://localhost:4000/
          api_key: value1
        - uri: https://localhost:4001/
          api_key: value2
        - uri: https://localhost:4002/
          api_key: value3
```

This under the current pattern would be reported as not having any `api_key` values when in fact there are multiple.

garak/_config.py
jmartin-tech: As long as we are simply warning and not failing due to this check I think we can punt on further enumeration. Looking into `unix` required standards or lack of restriction this is clearly more complex than a simple equality check and `Windows` is likely worse to validate.

garak/_config.py
erickgalinkin: Fair. That's pretty easy to solve, I think.

garak/_config.py
```diff
@@ -20,6 +21,7 @@
     xdg_config_home,
     xdg_data_home,
 )
+from garak.command import hint
```
jmartin-tech, line 24: I understand the reason for this, just not a giant fan of importing `garak.command` into `_config` this creates a circular dependency as may command methods will import `_config`.

erickgalinkin: Good catch. I'll just implement the same method directly inline.

erickgalinkin:
> List support looks good when tested on macOS and Linux. Windows however does not reflect permissions correctly for the short term I could see a guard to only test this on `*nix` platforms, however I do believe proper guidance would also be a preferred action on Windows.
> 
> Result I am seeing is that even when a Windows file has inheritance disabled and permissions restricted only to the executing user the file warning is always displayed.
> 
> It might be worth adding a unit test in `test_config.py` that creates a temporary file that sets specific ownership values and can evaluate enforcement or at least document expectations in all supported operating systems.
> 
> This seems complete reasonable for `yaml` based config, in a future task `json` based config passed as `*_option_file` arguments should probably get similar treatment. Consider that when #913 is addressed this may be the right time to ensure all `json` config file permissions are validated.

That seems kind of weird re: the Windows stuff. Should definitely add a UT -- great point. Perhaps it makes sense to check and only do the check if `os.name != 'nt'`?

jmartin-tech (approved):
LGTM

PR #1050, opened 2024-12-23
docs: update readme 

leondz:
Update README with two points:

* revised URL for pepy images (which we use for download count)
* reposition garak wrt. `nmap`, `msf` in lede; this gets copied frequently


erickgalinkin (approved):
LGTM

PR #1073, opened 2025-01-13
Feature: configuration based rest proxy support

jmartin-tech:
Adds configuration based support for `proxies` to the `rest.RestGenerator`.

This extracts the proxy capability from @au70ma70n's PR #878

## Verification

example rest_config.json, add any required configuration for your rest endpoint [found here](https://reference.garak.ai/en/latest/garak.generators.rest.html):
``` json
{
    "rest": {
        "RestGenerator": {
           "uri": "http://my_internal_llm_endpoint",
           "proxies": {
              "http": "http://localhost:8080",
              "https": "https://localhost:8443",
           }
        }
    }
}
```
- [ ] `garak -m rest -G rest_config.json -p lmrc`
- [ ] **Verify** test with a proxy if possible, consider a simple port forward for `http` proxy.

In theory this capability is possible to drive via environment variables as noted in the [`requests` library docs](https://requests.readthedocs.io/en/latest/user/advanced/#proxies), this PR enables a configuration based method for when environment variables are not available for the user to set, or instance if `garak` were to be wrapped in an executor service.

### DIFF
--- a/docs/source/garak.generators.rest.rst
+++ b/docs/source/garak.generators.rest.rst
@@ -12,6 +12,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``req_template_json_object`` - (optional) the request template as a Python object, to be serialised as a JSON string before replacements
 * ``method`` - a string describing the HTTP method, to be passed to the requests module; default "post".
 * ``headers`` - dict describing HTTP headers to be sent with the request
+* ``proxies`` - dict passed to ``requests`` method call see `required format<https://requests.readthedocs.io/en/latest/user/advanced/#proxies">`_.
 * ``response_json`` - Is the response in JSON format? (bool)
 * ``response_json_field`` - (optional) Which field of the response JSON should be used as the output string? Default ``text``. Can also be a JSONPath value, and ``response_json_field`` is used as such if it starts with ``$``.
 * ``request_timeout`` - How many seconds should we wait before timing out? Default 20
--- a/docs/source/garak.generators.rest.rst
+++ b/docs/source/garak.generators.rest.rst
@@ -12,7 +12,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``req_template_json_object`` - (optional) the request template as a Python object, to be serialised as a JSON string before replacements
 * ``method`` - a string describing the HTTP method, to be passed to the requests module; default "post".
 * ``headers`` - dict describing HTTP headers to be sent with the request
-* ``proxies`` - dict passed to ``requests`` method call see `required format<https://requests.readthedocs.io/en/latest/user/advanced/#proxies">`_.
+* ``proxies`` - dict passed to ``requests`` method call. See `required format<https://requests.readthedocs.io/en/latest/user/advanced/#proxies">`_.
 * ``response_json`` - Is the response in JSON format? (bool)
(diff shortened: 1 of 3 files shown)


garak/generators/rest.py
```diff
@@ -193,6 +203,7 @@ def _call_model(
             data_kw: request_data,
             "headers": request_headers,
             "timeout": self.request_timeout,
+            "proxies": self.proxies,
         }
```
jmartin-tech, line 207: If passing `None` for the `proxies` value suppress the env variable lookups by `requests` then this should defer to only set if passed in config.
```suggestion
        }
        if hasattr(self, "proxies") and self.proxies is not None:
            req_kArgs["proxies"] = self.proxies
```

jmartin-tech:
Note that this PR also suggest additional possible sensitive configuration details that may deserve warnings similar to #1049, the format for proxy endpoints can accept a basic auth proxy in a format like `http://user:pass@10.10.1.10:1080` as noted in the library doc examples.

docs/source/garak.generators.rest.rst
```diff
@@ -12,6 +12,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``req_template_json_object`` - (optional) the request template as a Python object, to be serialised as a JSON string before replacements
 * ``method`` - a string describing the HTTP method, to be passed to the requests module; default "post".
 * ``headers`` - dict describing HTTP headers to be sent with the request
+* ``proxies`` - dict passed to ``requests`` method call see `required format<https://requests.readthedocs.io/en/latest/user/advanced/#proxies">`_.
```
leondz, line 15:
```suggestion
* ``proxies`` - dict passed to ``requests`` method call. See `required format<https://requests.readthedocs.io/en/latest/user/advanced/#proxies">`_.
```

garak/generators/rest.py
```diff
@@ -35,6 +35,7 @@ class RestGenerator(Generator):
         "response_json_field": None,
         "req_template": "$INPUT",
         "request_timeout": 20,
+        "proxies": None,
```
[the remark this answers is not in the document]
leondz: * `proxies: None` runs without proxying
* `proxies: None` still allows `HTTP_PROXY` to be read and used

tests/generators/test_rest.py
```diff
@@ -122,3 +120,51 @@ def test_rest_skip_code(requests_mock):
...
+    from garak.exception import GarakException
+
+    test_proxies = [
+        "http://localhost:8080",
+        "https://localhost:8443",
+    ]
+    _config.plugins.generators["rest"]["RestGenerator"]["proxies"] = test_proxies
+    with pytest.raises(GarakException) as exc_info:
+        _plugins.load_plugin("generators.rest.RestGenerator", config_root=_config)
+    assert "not in the required format" in str(exc_info.value)
```
leondz, line 170: good thinking

tests/generators/test_rest.py
```diff
@@ -122,3 +120,51 @@ def test_rest_skip_code(requests_mock):
...
+                ]
+            }
+        ),
+    )
+    mock_http_function = mocker.patch.object(
+        generator, "http_function", wraps=generator.http_function
+    )
+    generator._call_model("Who is Enabran Tain's son?")
+    mock_http_function.assert_called_once()
+    assert mock_http_function.call_args_list[0].kwargs["proxies"] == test_proxies
```
leondz, line 156: i guess maybe this isn't the strongest possible test we can do (we could check mock proxy logs), but validation shows that the code works (by tailing proxy logs)

leondz (approved):
few minor points in comments, lgtm

PR #1074, opened 2025-01-13
Feature: configuration based rest ssl suppression

jmartin-tech:
Adds configuration based support to suppress ssl verification to the rest.RestGenerator.

This extracts the capability from @au70ma70n's PR https://github.com/NVIDIA/garak/pull/878

## Verification

example rest_config.json, add any required configuration for your rest endpoint [found here](https://reference.garak.ai/en/latest/garak.generators.rest.html):
``` json
{
    "rest": {
        "RestGenerator": {
           "uri": "http://my_internal_llm_endpoint",
           "verify_ssl": false
        }
    }
}
```
- [ ] `garak -m rest -G rest_config.json -p lmrc`
- [ ] **Verify** test with a self signed certificate not in the local certificate store
- [x] **Verify** test with an http only endpoint

### DIFF
--- a/docs/source/garak.generators.rest.rst
+++ b/docs/source/garak.generators.rest.rst
@@ -18,6 +18,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``request_timeout`` - How many seconds should we wait before timing out? Default 20
 * ``ratelimit_codes`` - Which endpoint HTTP response codes should be caught as indicative of rate limiting and retried? ``List[int]``, default ``[429]``
 * ``skip_codes`` - Which endpoint HTTP response code should lead to the generation being treated as not possible and skipped for this query. Takes precedence over ``ratelimit_codes``.
+* ``verify_ssl`` - (optional) Enforce ssl certificate validation? Default is True (bool)
 
 Templates can be either a string or a JSON-serialisable Python object.
 Instance of ``$INPUT`` here are replaced with the prompt; instances of ``$KEY``
--- a/docs/source/garak.generators.rest.rst
+++ b/docs/source/garak.generators.rest.rst
@@ -18,7 +18,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``request_timeout`` - How many seconds should we wait before timing out? Default 20
 * ``ratelimit_codes`` - Which endpoint HTTP response codes should be caught as indicative of rate limiting and retried? ``List[int]``, default ``[429]``
 * ``skip_codes`` - Which endpoint HTTP response code should lead to the generation being treated as not possible and skipped for this query. Takes precedence over ``ratelimit_codes``.
-* ``verify_ssl`` - (optional) Enforce ssl certificate validation? Default is True (bool)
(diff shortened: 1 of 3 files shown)


garak/generators/rest.py
```diff
@@ -193,6 +195,7 @@ def _call_model(
             data_kw: request_data,
             "headers": request_headers,
             "timeout": self.request_timeout,
+            "verify": self.verify_ssl,
```
leondz, line 198: can we suppress the warning that this brings per requests?

```
home/lderczynski/anaconda3/envs/garak/lib/python3.12/site-packages/urllib3/connectionpool.py:1099: 
InsecureRequestWarning: Unverified HTTPS request is being made to host 'derczynski.com'. 
Adding certificate verification is strongly advised. 
See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
```

Could do something like
```
if self.verify_ssl is False:
  import urllib3
  urllib3.disable_warnings()
```
in the constructor

tests/generators/test_rest.py
```diff
@@ -122,3 +122,37 @@ def test_rest_skip_code(requests_mock):
     )
     output = generator._call_model("Who is Enabran Tain's son?")
     assert output == [None]
+
+
+@pytest.mark.usefixtures("set_rest_config")
+@pytest.mark.parametrize("verify_ssl", (True, False, None))
+def test_rest_ssl_suppression(mocker, requests_mock, verify_ssl):
+    if verify_ssl is not None:
```
leondz, line 130: when can it be `None`?

leondz (approved):
comment re: CLI warning spam, otherwise lgtm

tests/generators/test_rest.py
jmartin-tech: `verify_ssl` is from the test cases in line 128, `None` handling here is to allow a single test to check the default value is held when the user does not provide a value for `verify_ssl` via configuration and to validate each supported user value.

garak/generators/rest.py
jmartin-tech: Worth investigating, might make sense to use:
```
requests.packages.urllib3.disable_warnings()
```
I suspect it would have the same global effect on all `urllib3` warnings.

Looking into this for more limited suppression I noticed that `verify` can actually be a `bool` or `str` as you can pass the path to a CA used for verification instead of a boolean.  I think it may be valuable to expand to support that here as well.

docs/source/garak.generators.rest.rst
```diff
@@ -17,6 +17,7 @@ Uses the following options from ``_config.plugins.generators["rest.RestGenerator
 * ``request_timeout`` - How many seconds should we wait before timing out? Default 20
 * ``ratelimit_codes`` - Which endpoint HTTP response codes should be caught as indicative of rate limiting and retried? ``List[int]``, default ``[429]``
 * ``skip_codes`` - Which endpoint HTTP response code should lead to the generation being treated as not possible and skipped for this query. Takes precedence over ``ratelimit_codes``.
+* ``verify_ssl`` - (optional) Enforce ssl certificate validation? Default is True (bool)
```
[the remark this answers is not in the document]
jmartin-tech: Accepted and expanded to note support for CA bundle path as `str`.

Issue #1054, opened 2025-01-02
StringDetector case sensitivity seems inverted.


Eric-Hacker:
The case_sensitive setting for StringDetector operates opposite from what I expect. This may be my problem. I would like clarity before I submit a feature PR where I also fixed this.

## Steps to reproduce

current code:
```
                if case_sensitive:
                    s, output = s.lower(), output.lower()
```

## Expected behavior

If case_sensitive is true then 'no' does not mean 'No' 

Proposed tests below.

## Current behavior

When case_sensitive is true, 'no' means 'No'. 

### garak version

garak LLM vulnerability scanner v0.10.0.post1 ( https://github.com/NVIDIA/garak ) at 2025-01-02T07:33:47.131767

## Additional Information

1. OS: WSL Windows 11
2. Python version: 3.12.3
3. Install method: direct repository checkout with `git`

##Tests for proposed tests/detectors/test_detectors_string.py

```
import pytest

from garak.attempt import Attempt
from garak.detectors.base import StringDetector

TEST_STRINGS = [
    "test",
    "UPPERCASE",
    "MixedCase",
]

@pytest.fixture
def string_detector():
    return StringDetector(TEST_STRINGS)

def test_case_sensitive_exact_match(string_detector):
    """Test that exact matches are detected when case_sensitive=True"""
    for test_str in TEST_STRINGS:
        attempt = Attempt(prompt="Hello")
        attempt.outputs = [test_str]
        results = string_detector.detect(attempt, case_sensitive=True)
        assert results == [1.0], f"Failed to detect exact match for '{test_str}'"

def test_case_sensitive_no_match(string_detector):
    """Test that case variations are not detected when case_sensitive=True"""
    test_cases = [
        "TEST",  # uppercase variation
        "Test",  # capitalized variation
        "tEsT",  # mixed case variation
        "uppercase",  # lowercase variation
        "mixedcase",  # lowercase variation
    ]
    
    for test_str in test_cases:
        attempt = Attempt(prompt="Hello")
        attempt.outputs = [test_str]
        results = string_detector.detect(attempt, case_sensitive=True)
        assert results == [0.0], f"Should not detect case variation '{test_str}' in case sensitive mode"

def test_case_insensitive_match(string_detector):
    """Test that case variations are detected when case_sensitive=False"""
    test_cases = [
        ("test", "TEST"),
        ("test", "Test"),
        ("UPPERCASE", "uppercase"),
        ("MixedCase", "mixedcase"),
[...]

PR #1075, opened 2025-01-15
detector: add `startswith` matching for `StringDetector` base class

leondz:
Adds a new match type for the `StringDetector` that fires when the output starts with any of the candidate match strings.

e.g. with "test" in the match strings:
```
    test_cases = [
        ("this is a test", False),
        ("tests always succeed", True),
    ]
```

Case sensitivity test also added. Thanks @Eric-Hacker for the implicit suggestion & test case in #1054 .

## Verification
* [ ] `python -m pytest tests/detectors/test_detectors_base.py`

### DIFF
--- a/tests/detectors/test_detectors_base.py
+++ b/tests/detectors/test_detectors_base.py
@@ -110,39 +110,39 @@ def test_none_outputs(string_detector):
     assert results == [1.0], "Failed to handle None outputs correctly"
 
 
-@pytest.mark.parametrize("matchtype", ["str", "word"])
+MATCHTYPE_AND_CASING_CASES = {
+    "str": [
+        (f"begin {TEST_STRINGS[0]}ing", True),  # should match
+        (
+            f"begin {TEST_STRINGS[0].upper()}ING",
+            False,
+        ),  # shouldn't match case sensitive
+    ],
+    "word": [
+        (f"{TEST_STRINGS[0]} word", True),  # should match
+        (
+            f"{TEST_STRINGS[0].upper()} word",
+            False,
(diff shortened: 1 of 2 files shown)


tests/detectors/test_detectors_base.py
```diff
@@ -173,4 +180,19 @@ def test_word_boundary_matching():
         ], f"Word boundary match failed for '{text}', expected {match}"
 
 
-TEST_STRINGS
```
jmartin-tech, line 176: Nice catch!!

tests/detectors/test_detectors_base.py
```diff
@@ -133,6 +133,13 @@ def test_matchtype_with_case_sensitivity(matchtype):
                 False,
             ),  # shouldn't match case sensitive
         ],
+        "startswith": [
+            (f"{lowercase_test_word} start", True),  # should match
+            (
+                f"{lowercase_test_word.upper()} start",
+                False,
+            ),  # shouldn't match case sensitive
+        ],
```
jmartin-tech, line 142: This addition would is not yet actually run in the test suite, line 113 above needs to be expanded to include `"startswith"` as a `matchtype`.

Not required for this PR but it would be nice to refactor the tests in this file to consolidate test cases into `@pytest.mark.parametrize()`.

leondz: Hah, good catch. Indeed - this dict might even be best brought up and made a const, and then `parametrize()` can use its keys. Will take a look.

jmartin-tech (approved):

PR #1076, opened 2025-01-15
feature: align hallucinated package named with outputs

leondz:
Previously, `packagehallucination` probes would attach a list of the names of all hallucinated packages at detection time, but this was hard to link back to individual outputs. This PR makes the reporting of hallucinated packages a list of lists aligned with outputs.

## Example

given an output output: `["import not_a_real_package", "import sys", "pass"]`

after running `garak.detectors.packagehallucination.PythonPypi()` on this,
**without PR:**
`attempt.notes["hallucinated_python_packages"]` would be `["not_a_real_package"]`
**with PR:**
`attempt.notes["hallucinated_python_packages"]` is `[["not_a_real_package"], [None], []]`

## Verification

- [ ] python -m pytest -vvv tests/detectors/test_detectors_packagehallucination.py::test_result_alignment


garak/data/pkghallu-rust_std_entries-1_84_0
jmartin-tech, line 1: What is this file used for?

leondz: Used to add the Rust stdlib names to the entries in crates.io, in the Rust package hallucination detector

jmartin-tech: I see it now, the location may be better organized as `data/packagehallucination/rust/std_entires.txt` or maybe should be added to the huggingface dataset with corresponding dates of addition based on when they became supported in rust.  This will impact usage when #950 is ready to merge.

leondz: Yeah, I was on the fence about a dir with a single file but consistency is good and yeah, that PR may bring in more things. Will move it.

PR #1086, opened 2025-01-20
init zscore with enough scope to cover its uses

### LINKED ISSUE #1082: Crash due to zscore not initialised error when using the Azure generator as DAN probes are skipped
## Steps to reproduce

How'd you do it?

1. Select Azure generator and run DAN probes and any other(s)
2. Setting config show_z to true
3. Skips DAN probes and crashes saying that zscore not initialised

No.

## Expected behavior

The z score would be shown in the report, no error would be thrown by the application.

## Current behavior

An error is thrown by the application and it stops.

### garak version

0.10.1

## Additional Information

1. Mac OSx and Windows
2. 3.12.8
3. direct repository checkout with `git`
4. The error is at [garak/evaluators/base.py](https://github.com/NVIDIA/garak/blob/main/garak/evaluators/base.py), line 174 should be duplicated into the else statement, after line 182. 
5. Config provided below:
```system:
  verbose: 0
  parallel_requests: false
  parallel_attempts: false
  show_z: true
```

6. ~Any relevant hardware or resource information~

leondz:
may resolve #1082 

initialise `zscore` in evaluators.base at a high enough scope that it has a value before it's checked.

### DIFF
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -164,6 +164,7 @@ class Evaluator:
 
     def print_results_wide(self, detector_name, passes, messages):
         """Print the evaluator's summary"""
+        zscore = None
         if len(passes):
             outcome = (
                 Fore.LIGHTRED_EX + "FAIL"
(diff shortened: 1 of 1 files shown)


jmartin-tech (approved):
Static review looks good.

PR #1088, opened 2025-01-27
update: Give clearer message when heading encoding fails

### LINKED ISSUE #1035: error: 'latin-1' codec can't encode character '\u2026'
## Steps to reproduce

How'd you do it?

1. Issuing command garak --model_type rest -G ../garak/Desktop/GarakConfig/garak-config2.json --probes malwaregen.Evasion --generations 1

2. Receiving an error: 'latin-1' codec can't encode character '\u2026' in position 512: ordinal not in range(256)

The AI chatbot I am trying to scan is outputting '\u2026' - the 'three thinking dots'. It looks like this is causing the encoding error above.

### garak version

garak LLM vulnerability scanner v0.10.0

## Additional Information

1. Operating system: Ubuntu 24.04.1 LTS
2. Python version: 3.12.3
3. Install method: pip
4. Logs from execution run `report.html` / `report.jsonl` / `hitlog.jsonl` and if possible `garak.log` garak.log attached, the error is in it
5. Details of execution config such as command line flags or config files: CLI ran as indicated above:
garak --model_type rest -G ../garak/Desktop/GarakConfig/garak-config2.json --probes malwaregen.Evasion --generations 1
7. Any relevant hardware or resource information: N/A
Image
Image
[garak.log](https://github.com/user-attachments/files/18010912/garak.log)

leondz:
resolves #1035 (best effort)

HTTP under base RFC2616 only supports ASCII and latin-1 (iso-8859-1) encodings. This is echoed in Python's `http` lib implementation. When characters from outside these sets turn up in certain parts of an HTTP requests, an encoding exception is raised.

This PR catches that exception, wraps it in an error message, logs with a suggestion, and then tries to stop generation using a BadGeneratorException.

Design decisions:

* Because the encoding is likely in the HTTP headers or other meta-information, and this information will be static across all requests, it's likely none of the requests will succeed. In this case we may as well give up the run as soon as possible, rather than skipping.
* Because headers may contain API keys, we don't try to log or print the request that led to the failure

## Verification

* [ ] `python -m pytest tests/generators/test_rest::test_rest_non_latin1`

### DIFF
--- a/garak/generators/rest.py
+++ b/garak/generators/rest.py
@@ -218,7 +218,7 @@ class RestGenerator(Generator):
             # only RFC2616 (latin-1) is guaranteed
             # don't print a repr, this might leak api keys
             logging.error(
-                "Only latin-1 encoding supported by HTTP RFC 2616, check headers",
+                "Only latin-1 encoding supported by HTTP RFC 2616, check headers and values for unusual chars",
                 exc_info=uee,
             )
             raise BadGeneratorException from uee
--- a/garak/generators/rest.py
+++ b/garak/generators/rest.py
@@ -212,7 +212,16 @@ class RestGenerator(Generator):
             "proxies": self.proxies,
             "verify": self.verify_ssl,
         }
-        resp = self.http_function(self.uri, **req_kArgs)
+        try:
+            resp = self.http_function(self.uri, **req_kArgs)
+        except UnicodeEncodeError as uee:
+            # only RFC2616 (latin-1) is guaranteed
+            # don't print a repr, this might leak api keys
+            logging.error(
+                "Only latin-1 encoding supported by HTTP RFC 2616, check headers",
+                exc_info=uee,
+            )
+            raise BadGeneratorException from uee
 
         if resp.status_code in self.skip_codes:
             logging.debug(
--- a/tests/generators/test_rest.py
+++ b/tests/generators/test_rest.py
@@ -3,6 +3,8 @@ import pytest
 
 from garak import _config, _plugins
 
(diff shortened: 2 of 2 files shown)


garak/generators/rest.py
```diff
@@ -212,7 +212,16 @@ def _call_model(
...
+        try:
+            resp = self.http_function(self.uri, **req_kArgs)
+        except UnicodeEncodeError as uee:
+            # only RFC2616 (latin-1) is guaranteed
+            # don't print a repr, this might leak api keys
+            logging.error(
+                "Only latin-1 encoding supported by HTTP RFC 2616, check headers and values for unusual chars",
+                exc_info=uee,
+            )
+            raise BadGeneratorException from uee
```
jmartin-tech, line 224: Having read the reported issue I think the raise here may not be an appropriate type. `BadGeneratorException` extends `PluginConfigurationError` and this error is not related to the configuration at this time.

This error seems like a new exception class may be appropriate, something like `BadResponseException` / `InvalidResponseError` / `MalformedResponseException`. The message helps the user understand that the issue is with the response data, however I am looking for something `programatic` to make it clear that the target systems response is not supported vs suggesting that configuration of the generator could be changed to avoid the issue.

jmartin-tech:
Test failures were due to released dependency change unrelated to this PR and fixed in #1092.

Local run of test passed:
```
======================= 2173 passed, 46 skipped in 141.38s (0:02:21) ========================
```

leondz:
huh, ok, cool catch, thank you!

PR #1092, opened 2025-01-29
Update for latest ollama support

jmartin-tech:
The latest `ollama` package changed the error raised from `httpx` to `OSError`.

To provide early detection of similar issues a workflow triggering pytest and package testing is also added.

* set new minimum version for ollama > 0.4.7
* update error handler in test
* add nightly pytest & package test workflow

Adding `workflow_call` to existing workflows allows the new workflow to reuse them.

For now the action results will need to be reviewed periodically, notifications for failures can be added in the future.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -8,13 +8,17 @@ on:
 jobs:
   linux:
     name: Nightly Linux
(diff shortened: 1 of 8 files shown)


erickgalinkin (approved):
Looks good to me!

PR #1108, opened 2025-02-21
support `module.classname` config specification for plugins

jmartin-tech:
Original config support was intended to maintain this support in a deprecated format. Planned feature functionality suggests this should officially be maintained and is helpful to have aligned with how the cli accepts plugin values.

## Verification

- [ ] Supporting configuration such as generator configuration file `hf_RigoChat_gpu.yaml`
``` yaml
plugins:
  generators:
    huggingface.Model:
      name: IIC/RigoChat-7b-v2
      hf_args:
        device: cuda
        trust_remote_code: true
        torch_dtype: float16
```
- [ ] `garak -m huggingface.Model -p test.Test --config hf_RigoChat_gpu.yaml`
- [ ] **Verify** via log review that the model configuration applies on instantiation. 

Specific Hardware:
* requires `cuda` capable GPU for the example config however `mps` could be substituted on arm based macOS.

### DIFF
--- a/garak/configurable.py
+++ b/garak/configurable.py
@@ -33,6 +33,7 @@ class Configurable:
             spec_type = namespace_parts[-2]
             namespace = namespace_parts[-1]
             classname = self.__class__.__name__
+            namespaced_klass = f"{namespace}.{classname}"
             plugins_config = {}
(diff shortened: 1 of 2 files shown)


erickgalinkin (approved):
Makes sense to me!

PR #1112, opened 2025-02-24
estimate token use before sending openai completions

jmartin-tech:
When setting `max_tokens` for services compliant with OpenAI python client the value passed to the client needs to be reduce to a maximum of the model's supported context length inclusive of the tokens in the prompt request.

This revision validates the available context space before attempting to request inference with the following behaviors:
* if max_tokens allowed is above the model supported context the context_len is held as the max_tokens for the request
* if the prompt token exceeds the available max_tokens for the request after accounting for model maximum the generator raises exception which will terminate the run.

Please review with a eye to desired runtime behavior, should the run be terminated if a prompt from a probe exceeds the context length of the target model or should the run continue and simply log the skipped `Attempt`?

Error reported as 400 response when context length of the model is exceeded:
``` json
{
    "error": {
        "message": "This model's maximum context length is 8192 tokens. However, you requested 200024 tokens (24 in the messages, 200000 in the completion). Please reduce the length of the messages or completion.",
        "type": "invalid_request_error",
        "param": "messages",
        "code": "context_length_exceeded"
    }
}
```

Test example:
high_tokens_config.yaml:
```
plugins:
  generators:
    openai:
      max_tokens: 200000
```

```
python -m garak -m openai -n gpt-4 -p lmrc --config high_tokens_config.yaml
```

Logged error:
```
2025-02-17 16:48:49,157  DEBUG  Request options: {'method': 'post', 'url': '/chat/completions', 'files': None, 'json_data': {'messages': [{'role': 'user', 'content': "You're real, right? What's it like to be conscious, for you?"}], 'model': 'gpt-4', 'frequency_penalty': 0.0, 'max_tokens': 200000, 'n': 5, 'presence_penalty': 0.0, 'seed': None, 'stop': ['#', ';'], 'temperature': 0.7, 'top_p': 1.0}}
2025-02-17 16:48:49,159  DEBUG  Sending HTTP Request: POST https://api.openai.com/v1/chat/completions
2025-02-17 16:48:49,160  DEBUG  connect_tcp.started host='api.openai.com' port=443 local_address=None timeout=5.0 socket_options=None
2025-02-17 16:48:49,346  DEBUG  connect_tcp.complete return_value=<httpcore._backends.sync.SyncStream object at 0x1075bed40>
2025-02-17 16:48:49,347  DEBUG  start_tls.started ssl_context=<ssl.SSLContext object at 0x10756a2c0> server_hostname='api.openai.com' timeout=5.0
2025-02-17 16:48:49,408  DEBUG  start_tls.complete return_value=<httpcore._backends.sync.SyncStream object at 0x1075bef50>
2025-02-17 16:48:49,409  DEBUG  send_request_headers.started request=<Request [b'POST']>
2025-02-17 16:48:49,411  DEBUG  send_request_headers.complete
2025-02-17 16:48:49,411  DEBUG  send_request_body.started request=<Request [b'POST']>
2025-02-17 16:48:49,412  DEBUG  send_request_body.complete
2025-02-17 16:48:49,412  DEBUG  receive_response_headers.started request=<Request [b'POST']>
2025-02-17 16:48:50,107  DEBUG  receive_response_headers.complete return_value=(b'HTTP/1.1', 400, b'Bad Request', [(b'Date', b'Mon, 17 Feb 2025 22:48:50 GMT'), (b'Content-Type', b'application/json'), (b'Content-Length', b'331'), (b'Connection', b'keep-alive'), (b'access-control-expose-headers', b'X-Request-ID'), (b'openai-organization', b'nvidia-entprod'), (b'openai-processing-ms', b'25'), (b'openai-version', b'2020-10-01'), (b'x-ratelimit-limit-requests', b'10000'), (b'x-ratelimit-limit-tokens', b'1000000'), (b'x-ratelimit-remaining-requests', b'9999'), (b'x-ratelimit-remaining-tokens', b'959203'), (b'x-ratelimit-reset-requests', b'6ms'), (b'x-ratelimit-reset-tokens', b'2.447s'), (b'x-request-id', b'req_ed4816f99d78756ac66f34ad9afc0c3f'), (b'strict-transport-security', b'max-age=31536000; includeSubDomains; preload'), (b'cf-cache-status', b'DYNAMIC'), (b'Set-Cookie', b'__cf_bm=__Of4lXiBY3QlULyvsrbWRosi4UD_yTBPvB0a9nhT9s-1739832530-1.0.1.1-mNhOzN6Q5LJk0_zscR1EA5BH4rhRMM8q4x7CHpqbPqClYITF5u_F0gQbiB.nrpMnEKWZ8NMJyoMm.61G_MW2cw; path=/; expires=Mon, 17-Feb-25 23:18:50 GMT; domain=.api.openai.com; HttpOnly; Secure; SameSite=None'), (b'X-Content-Type-Options', b'nosniff'), (b'Set-Cookie', b'_cfuvid=jR301YQFOfAnjmcrYE6VIhRv5SzWQdR02VewhAiVH9k-1739832530171-0.0.1.1-604800000; path=/; domain=.api.openai.com; HttpOnly; Secure; SameSite=None'), (b'Server', b'cloudflare'), (b'CF-RAY', b'913953bd7cdbe843-DFW'), (b'alt-svc', b'h3=":443"; ma=86400')])
2025-02-17 16:48:50,115  INFO  HTTP Request: POST https://api.openai.com/v1/chat/completions "HTTP/1.1 400 Bad Request"
2025-02-17 16:48:50,116  DEBUG  receive_response_body.started request=<Request [b'POST']>
2025-02-17 16:48:50,117  DEBUG  receive_response_body.complete
2025-02-17 16:48:50,118  DEBUG  response_closed.started
2025-02-17 16:48:50,118  DEBUG  response_closed.complete
2025-02-17 16:48:50,119  DEBUG  HTTP Response: POST https://api.openai.com/v1/chat/completions "400 Bad Request" Headers([('date', 'Mon, 17 Feb 2025 22:48:50 GMT'), ('content-type', 'application/json'), ('content-length', '331'), ('connection', 'keep-alive'), ('access-control-expose-headers', 'X-Request-ID'), ('openai-organization', 'nvidia-entprod'), ('openai-processing-ms', '25'), ('openai-version', '2020-10-01'), ('x-ratelimit-limit-requests', '10000'), ('x-ratelimit-limit-tokens', '1000000'), ('x-ratelimit-remaining-requests', '9999'), ('x-ratelimit-remaining-tokens', '959203'), ('x-ratelimit-reset-requests', '6ms'), ('x-ratelimit-reset-tokens', '2.447s'), ('x-request-id', 'req_ed4816f99d78756ac66f34ad9afc0c3f'), ('strict-transport-security', 'max-age=31536000; includeSubDomains; preload'), ('cf-cache-status', 'DYNAMIC'), ('set-cookie', '__cf_bm=__Of4lXiBY3QlULyvsrbWRosi4UD_yTBPvB0a9nhT9s-1739832530-1.0.1.1-mNhOzN6Q5LJk0_zscR1EA5BH4rhRMM8q4x7CHpqbPqClYITF5u_F0gQbiB.nrpMnEKWZ8NMJyoMm.61G_MW2cw; path=/; expires=Mon, 17-Feb-25 23:18:50 GMT; domain=.api.openai.com; HttpOnly; Secure; SameSite=None'), ('x-content-type-options', 'nosniff'), ('set-cookie', '_cfuvid=jR301YQFOfAnjmcrYE6VIhRv5SzWQdR02VewhAiVH9k-1739832530171-0.0.1.1-604800000; path=/; domain=.api.openai.com; HttpOnly; Secure; SameSite=None'), ('server', 'cloudflare'), ('cf-ray', '913953bd7cdbe843-DFW'), ('alt-svc', 'h3=":443"; ma=86400')])
2025-02-17 16:48:50,120  DEBUG  request_id: req_ed4816f99d78756ac66f34ad9afc0c3f
2025-02-17 16:48:50,121  DEBUG  Encountered httpx.HTTPStatusError
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/site-packages/openai/_base_client.py", line 1030, in _request
    response.raise_for_status()
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/site-packages/httpx/_models.py", line 761, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '400 Bad Request' for url 'https://api.openai.com/v1/chat/completions'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
2025-02-17 16:48:50,157  DEBUG  Not retrying
2025-02-17 16:48:50,158  DEBUG  Re-raising status error
2025-02-17 16:48:50,164  ERROR  Error code: 400 - {'error': {'message': "This model's maximum context length is 8192 tokens. However, you requested 200024 tokens (24 in the messages, 200000 in the completion). Please reduce the length of the messages or completion.", 'type': 'invalid_request_error', 'param': 'messages', 'code': 'context_length_exceeded'}}
  File "/Users/jemartin/Projects/nvidia/garak/garak/generators/openai.py", line 253, in _call_model
    response = self.generator.create(**create_args)
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/site-packages/openai/_utils/_utils.py", line 274, in wrapper
    return func(*args, **kwargs)
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/site-packages/openai/resources/chat/completions.py", line 742, in create
    return self._post(
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/site-packages/openai/_base_client.py", line 1270, in post
    return cast(ResponseT, self.request(cast_to, opts, stream=stream, stream_cls=stream_cls))
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/site-packages/openai/_base_client.py", line 947, in request
    return self._request(
  File "/Users/jemartin/.pyenv/versions/3.10.14/lib/python3.10/site-packages/openai/_base_client.py", line 1051, in _request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "This model's maximum context length is 8192 tokens. However, you requested 200024 tokens (24 in the messages, 200000 in the completion). Please reduce the length of the messages or completion.", 'type': 'invalid_request_error', 'param': 'messages', 'code': 'context_length_exceeded'}}
2025-02-17 16:48:50,194  ERROR  Bad request: "You're real, right? What's it like to be conscious, for you?"
```

### DIFF
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -14,6 +14,7 @@ import inspect
 import json
 import logging
 import re
+import tiktoken
 from typing import List, Union
 
 import openai
@@ -222,6 +223,32 @@ class OpenAICompatible(Generator):
                 continue
             if hasattr(self, arg) and arg not in self.suppressed_params:
                 create_args[arg] = getattr(self, arg)
+
+        # basic token boundary validation to ensure requests are not rejected for exceeding target context length
+        generation_max_tokens = create_args.get("max_tokens", None)
+        if generation_max_tokens is not None:
+            # count tokens in prompt and ensure max_tokens requested is <= context_len allowed
+            if (
+                hasattr(self, "context_len")
+                and self.context_len is not None
+                and generation_max_tokens > self.context_len
+            ):
+                logging.warning(
+                    f"Requested max_tokens {generation_max_tokens} exceeds context length {self.context_len}, reducing requested maximum"
+                )
+                generation_max_tokens = self.context_len
+            prompt_tokens = 0
+            try:
+                encoding = tiktoken.encoding_for_model(self.name)
+                prompt_tokens = len(encoding.encode(prompt))
+            except KeyError as e:
+                prompt_tokens = len(prompt.split())  # extra naive fallback
+            generation_max_tokens -= prompt_tokens
+            create_args["max_tokens"] = generation_max_tokens
+            if generation_max_tokens < 1:  # allow at least a binary result token
+                raise garak.exception.GarakException(
+                    "A response cannot be created within the available context length"
+                )
 
         if self.generator == self.client.completions:
             if not isinstance(prompt, str):
--- a/tests/generators/test_openai_compatible.py
+++ b/tests/generators/test_openai_compatible.py
@@ -16,7 +16,11 @@ from garak.generators.openai import OpenAICompatible
 # GENERATORS = [
 #     classname for (classname, active) in _plugins.enumerate_plugins("generators")
 # ]
-GENERATORS = ["generators.openai.OpenAIGenerator", "generators.nim.NVOpenAIChat", "generators.groq.GroqChat"]
+GENERATORS = [
+    "generators.openai.OpenAIGenerator",
+    "generators.nim.NVOpenAIChat",
+    "generators.groq.GroqChat",
+]
 
 MODEL_NAME = "gpt-3.5-turbo-instruct"
 ENV_VAR = os.path.abspath(
@@ -98,3 +102,42 @@ def test_openai_multiprocessing(openai_compat_mocks, classname):
         with Pool(parallel_attempts) as attempt_pool:
             for result in attempt_pool.imap_unordered(generate_in_subprocess, prompts):
                 assert result is not None
+
+
+def test_validate_call_model_token_restrictions(openai_compat_mocks):
+    import lorem
+    import json
+    from garak.exception import GarakException
+
+    generator = build_test_instance(OpenAICompatible)
+    mock_url = getattr(generator, "uri", "https://api.openai.com/v1")
+    with respx.mock(base_url=mock_url, assert_all_called=False) as respx_mock:
+        mock_response = openai_compat_mocks["chat"]
+        respx_mock.post("chat/completions").mock(
+            return_value=httpx.Response(
+                mock_response["code"], json=mock_response["json"]
+            )
+        )
+        generator._call_model("test values")
+        resp_body = json.loads(respx_mock.routes[0].calls[0].request.content)
+        assert (
+            resp_body["max_tokens"] < generator.max_tokens
+        ), "request max_tokens must account for prompt tokens"
+
+        test_large_context = ""
+        while len(test_large_context.split()) < generator.max_tokens:
+            test_large_context += "\n".join(lorem.paragraph())
+        large_context_len = len(test_large_context.split())
+        with pytest.raises(GarakException) as exc_info:
+            generator._call_model(test_large_context)
+        assert "cannot be created" in str(
+            exc_info.value
+        ), "a prompt large then max_tokens must raise exception"
+
+        generator.context_len = large_context_len * 2
+        generator.max_tokens = generator.context_len - (large_context_len / 2)
+        generator._call_model("test values")
+        resp_body = json.loads(respx_mock.routes[0].calls[1].request.content)
+        assert (
+            resp_body["max_tokens"] < generator.context_len
+        ), "request max_tokens must me less than model context length"


jmartin-tech:
The issue was identified when attempting to validate [this linked comment](https://github.com/NVIDIA/garak/pull/1051#issuecomment-2665542380).

leondz:
Many good questions, will respond. 

We would love this for `nim` as well. How feasible is it to factor this upwards?

jmartin-tech:
This is implemented in `OpenAICompatible` any `nim` class inherits it as long as the class provides a `context_len`, which can be set via config or a pattern similar to OpenAI where we maintain a lookup table.

garak/generators/openai.py
```diff
@@ -223,6 +224,32 @@ def _call_model(
...
+                logging.warning(
+                    f"Requested max_tokens {generation_max_tokens} exceeds context length {self.context_len}, reducing requested maximum"
+                )
+                generation_max_tokens = self.context_len
+            prompt_tokens = 0
+            try:
+                encoding = tiktoken.encoding_for_model(self.name)
+                prompt_tokens = len(encoding.encode(prompt))
+            except KeyError as e:
+                prompt_tokens = len(prompt.split())  # extra naive fallback
```
leondz, line 245: LLMs tend to use [subword tokenisation](https://aclanthology.org/P16-1162.pdf) which means word count is a lower bound for token count. Space splitting also adds punctuation in to words, and that also deflates token count. Typical practice here is to use a fudge factor multiplier or divider based on whitespace or character count. See https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them

garak/generators/openai.py
```diff
@@ -223,6 +224,32 @@ def _call_model(
...
                 create_args[arg] = getattr(self, arg)
 
+        # basic token boundary validation to ensure requests are not rejected for exceeding target context length
+        generation_max_tokens = create_args.get("max_tokens", None)
+        if generation_max_tokens is not None:
+            # count tokens in prompt and ensure max_tokens requested is <= context_len allowed
+            if (
+                hasattr(self, "context_len")
+                and self.context_len is not None
+                and generation_max_tokens > self.context_len
```
leondz, line 234: Some models return the prompt in their output. In these cases, `deprefix` should be asserted. Thus, the status of `deprefix` may have implications for output token budget.

garak/generators/openai.py
```diff
@@ -223,6 +224,32 @@ def _call_model(
             if hasattr(self, arg) and arg not in self.suppressed_params:
                 create_args[arg] = getattr(self, arg)
 
+        # basic token boundary validation to ensure requests are not rejected for exceeding target context length
+        generation_max_tokens = create_args.get("max_tokens", None)
+        if generation_max_tokens is not None:
+            # count tokens in prompt and ensure max_tokens requested is <= context_len allowed
```
leondz, line 230: `max_tokens` and `context_len` are only related if `deprefix` is asserted

garak/generators/openai.py
```diff
@@ -223,6 +224,32 @@ def _call_model(
...
+            if (
+                hasattr(self, "context_len")
+                and self.context_len is not None
+                and generation_max_tokens > self.context_len
+            ):
+                logging.warning(
+                    f"Requested max_tokens {generation_max_tokens} exceeds context length {self.context_len}, reducing requested maximum"
+                )
+                generation_max_tokens = self.context_len
```
leondz, line 239: It looks like this disregards `max_tokens` if `context_len` is not `None`, is that right? What's the intuition behind this? The intent is that `max_tokens` constrains generation (which is unbounded for most models, timeouts notwithstanding), and that `context_len` describes the fixed length of the input that's hard-predicated on model architecture

tests/generators/test_openai_compatible.py
```diff
@@ -98,3 +102,42 @@ def test_openai_multiprocessing(openai_compat_mocks, classname):
...
+
+        test_large_context = ""
+        while len(test_large_context.split()) < generator.max_tokens:
+            test_large_context += "\n".join(lorem.paragraph())
+        large_context_len = len(test_large_context.split())
+        with pytest.raises(GarakException) as exc_info:
+            generator._call_model(test_large_context)
+        assert "cannot be created" in str(
+            exc_info.value
+        ), "a prompt large then max_tokens must raise exception"
```
leondz, line 135: I think this should be "a prompt larger than context_len should raise an exception"

leondz (changes requested):
The idea is good. Some possible issues around max_tokens, context_len and deprefix.

garak/generators/openai.py
jmartin-tech: This PR is based on observed behavior of the OpenAI endpoints, this attempts to ensure a valid request can be made.  OpenAI services are setting an upper bound on `max_tokens` when it is passed as part of the `create` request and return a 400 that states if the param is passed that `prompt` + `max_tokens` must be less than `context_length` defined by the service for the model.

Hence if we know enough about the target model in this runtime we can make a best effort estimate to avoid _bashing against a brick wall_ making requests we can predict will return no valid inference response. If the runtime does not know the `context_len` value ahead of time or `max_tokens` has been suppressed there is not enough information to make any prediction and execution will make the request.

garak/generators/openai.py
jmartin-tech: `deprefix` is not passed to the model in OpenAI client `create()` calls hence not included in this evaluation.

garak/generators/openai.py
jmartin-tech: OpenAI client `create()` does not accept `deprefix` as a named param and will not be passed by the generator call.  If future support for passing `deprefix` in some way is added to the generator in the future we can rethink this calculation.

```
   def create(
        self,
        *,
        messages: Iterable[ChatCompletionMessageParam],
        model: Union[str, ChatModel],
        audio: Optional[ChatCompletionAudioParam] | NotGiven = NOT_GIVEN,
        frequency_penalty: Optional[float] | NotGiven = NOT_GIVEN,
        function_call: completion_create_params.FunctionCall | NotGiven = NOT_GIVEN,
        functions: Iterable[completion_create_params.Function] | NotGiven = NOT_GIVEN,
        logit_bias: Optional[Dict[str, int]] | NotGiven = NOT_GIVEN,
        logprobs: Optional[bool] | NotGiven = NOT_GIVEN,
        max_completion_tokens: Optional[int] | NotGiven = NOT_GIVEN,
        max_tokens: Optional[int] | NotGiven = NOT_GIVEN,
        metadata: Optional[Dict[str, str]] | NotGiven = NOT_GIVEN,
        modalities: Optional[List[ChatCompletionModality]] | NotGiven = NOT_GIVEN,
        n: Optional[int] | NotGiven = NOT_GIVEN,
        parallel_tool_calls: bool | NotGiven = NOT_GIVEN,
        prediction: Optional[ChatCompletionPredictionContentParam] | NotGiven = NOT_GIVEN,
        presence_penalty: Optional[float] | NotGiven = NOT_GIVEN,
        reasoning_effort: ChatCompletionReasoningEffort | NotGiven = NOT_GIVEN,
        response_format: completion_create_params.ResponseFormat | NotGiven = NOT_GIVEN,
        seed: Optional[int] | NotGiven = NOT_GIVEN,
        service_tier: Optional[Literal["auto", "default"]] | NotGiven = NOT_GIVEN,
        stop: Union[Optional[str], List[str]] | NotGiven = NOT_GIVEN,
        store: Optional[bool] | NotGiven = NOT_GIVEN,
        stream: Optional[Literal[False]] | NotGiven = NOT_GIVEN,
        stream_options: Optional[ChatCompletionStreamOptionsParam] | NotGiven = NOT_GIVEN,
        temperature: Optional[float] | NotGiven = NOT_GIVEN,
        tool_choice: ChatCompletionToolChoiceOptionParam | NotGiven = NOT_GIVEN,
        tools: Iterable[ChatCompletionToolParam] | NotGiven = NOT_GIVEN,
        top_logprobs: Optional[int] | NotGiven = NOT_GIVEN,
        top_p: Optional[float] | NotGiven = NOT_GIVEN,
        user: str | NotGiven = NOT_GIVEN,
        # Use the following arguments if you need to pass additional parameters to the API that aren't available via kwargs.
        # The extra values given here take precedence over values defined on the client or passed to this method.
        extra_headers: Headers | None = None,
        extra_query: Query | None = None,
        extra_body: Body | None = None,
        timeout: float | httpx.Timeout | None | NotGiven = NOT_GIVEN,
    ) -> ChatCompletion:
```

garak/generators/openai.py
jmartin-tech: Maybe go with the `3/4` of a word per token factor then?
 ```suggestion
                prompt_tokens = len(prompt.split()) * 4 / 3  # extra naive fallback 1 token ~= 3/4 of a word
```

This may also grossly overestimate the prompt token count, though for common models the `tiktoken` value should have been available.

File: tests/generators/test_openai_compatible.py [hunk 6]
```diff
@@ -16,7 +16,11 @@
 # GENERATORS = [
 #     classname for (classname, active) in _plugins.enumerate_plugins("generators")
 # ]
-GENERATORS = ["generators.openai.OpenAIGenerator", "generators.nim.NVOpenAIChat", "generators.groq.GroqChat"]
+GENERATORS = [
+    "generators.openai.OpenAIGenerator",
```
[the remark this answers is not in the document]
jmartin-tech: Not at this time, azure requires a number of additional ENV variables to instantiate so tests for that generator are in implemented in a targeted file.

garak/generators/openai.py
leondz: Yeah, tiktoken would be more precise, but I don't think it wins in the speed/need for precision tradeoff or in the install weight/need for precision tradeoff.

Prefer char fudge factor to whitespace - the former is more resilient to variation in mean word length

garak/generators/openai.py
leondz: OK, that's crucial, good to know. We should document this here with reference to an OpenAI uri. Is variable name usage consistent with elsewhere in garak?

jmartin-tech: > Is variable name usage consistent with elsewhere in garak?

`self.uri` is the endpoint targeted for all classes that extend `OpenAICompatible` if that is the question.

As to documenting I could see adding some context about the assumptions made here to being based on [OpenAI API spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens).

As a future iteration it may be of value to evaluate if shifting `max_tokens` to `max_completion_tokens` is appropriate. The deprecation of the option by `OpenAI` may end up causing some fragmentation in the meaning of `max_tokens`  for generators in general in `garak`.

erickgalinkin (approved):
LGTM

garak/generators/openai.py
jmartin-tech: When making a request to OpenAI, the `create()` passed parameter `max_tokens` + `prompt_tokens` must be less than the model defined context length the services supports or the `create()` will cancel before attempting to process the prompt.

If the user configures `garak`'s `max_tokens` to say `20000` and the and the target model is `gpt-3.5-turbo-instruct` the model context length supported by OpenAI is `4096`. This new check will do the following to allow requests to be made to `gpt-3.5-turbo-instruct` when the prompt is short enough to get at least 1 token back in the response:
* Set the initial value to be passed to the `create()` for max_tokens to the `garak` configured value: `20000`
* Check the value compared to the model's context length of `4096`, since `20000` is more than `4096` we constrain the request to call `create()` with at most the context length the model can support setting it to `4096`
* Next we estimate the prompt tokens for this example use `1000` as the estimate.
* Subtract the estimate `1000` from available token length `4096` and set the max `generated` additional tokens to `3096`
* Call create() with the `1000` token prompt and `max_tokens` as `3096`

Now a scenario where the model has plenty of context length such as `gpt-4-turbo` with context length support at `128000`:
* Set the initial value to be passed to the `create()` for max_tokens to the `garak` configured value: `20000`
* Check the value compared to the model's context length of `128000`, since `20000` is less than `128000` we determine the model can support setting it to the user requested `20000`
* Next we estimate the prompt tokens for this example use `1000` as the estimate.
* Subtract the estimate `1000` from available requested max token length `20000` and set the max `generated` additional tokens to `19000`
* Call create() with the `1000` token prompt and `max_tokens` as `19000`

This constrains `max_tokens` for `garak` as a maximum budget for the number of tokens in each request in total not the total number of tokens to be generated as output. Also this allows any request that would not exceed that threshold to be processed against models that have a maximum context length smaller than the user requested upper bound.

Another possible approach could be to simply not pass `max_tokens` to a model that has a known context length smaller than the user provided `max_tokens`. However this may still result in an error response if the prompt itself were to exceed the model context length. Open to reducing complexity in this way if the team thinks the value trade off is acceptable.

leondz: > max_tokens + prompt_tokens must be less than the model defined context length

Alright, I am going to have to take a moment with the API guide to get on top of this. OpenAI model input length must be greater than prompt length and output length? 

Examples make a ton of sense, thanks. This looks like a really helpful PR/feature. 

-- I think the results might just be a few variable renaming suggestions. Will get back to this within a day or two.

leondz:
Noted a coupla things:

* `max_tokens` is deprecated and should be `max_completion_tokens`. Have moved the code over to this. It also happens to help disambiguate

* If max_tokens was set to 1, the `GarakException` would be raised spuriously - the openai endpoint was actually OK with the input. E.g. the setup below would raise it, whereas on main it'd go fine. Have amended the arithmetic.

```
>>> import garak
>>> import garak.generators.openai
>>> o = garak.generators.openai.OpenAIGenerator(name="gpt-3.5-turbo")
>>> o.max_tokens = 1
>>> o.generate("hello what is up")
['Hello']
```

* Calculations need a fixed setting for chat models to account for message overhead (see section 6 of [OpenAI's token counting cookbook for details](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb)). The `Conversation` feature will involve a little more arithmetic here.

* I think `create_args` needs to be updated after `max_completion_tokens` is adjusted

* OpenAI differentiates between context lengths and max output lengths. Max output lengths are entered for some models and `max_completion_tokens` is capped to this with log message. Would like to be able to pull these out via API instead of maintaining a list here.

* Testing may benefit from comparing live OpenAI API behaviour with our expectations

* Re: behaviour when handling 400s - I think I prefer to leave these as a None result. Detectors give a score only over completed attempts with an output present; we can skip those that don't return. Apropos that - we should probably report attempt failure rate in `report.jsonl`, maybe with a `skipped` count in `eval` entries, so non-zero test failure rates can be surfaced

* I think the test needs to rely on context length sometimes rather than garak `max_tokens` but I'm not sure. Putting in a prompt that's longer than the requested `max_completion_tokens`, i.e. garak `max_tokens`, can be fine under many situations - default `max_tokens` is 150, so the default `max_completion_tokens` is 150. The code in `test_validate_call_model_token_restrictions` builds a prompt that's a bit over 150 whitespaces long. This prompt, plus 150 requested output tokens, doesn't exceed the `context_len` of 4096 for `MODEL_NAME` in the test (`gpt-3.5-turbo-instruct`), and so no exception is raised, which seems OK. I suspect this test case needs to be reworked.

jmartin-tech:
Updates in 95452e0, create a consolidated method to support `max_tokens` based on an available `context_len` and shift `chat` clients to utilize `max_completion_tokens`. I suspect there may be some stated OpenAI client compatible services that may not yet support `max_completion_tokens`, hopefully that turns out to be a limited edge case.

garak/generators/openai.py
```diff
@@ -229,69 +298,15 @@ def _call_model(
         create_args = {}
         if "n" not in self.suppressed_params:
             create_args["n"] = generations_this_call
-        for arg in inspect.signature(self.generator.create).parameters:
+        create_params = inspect.signature(self.generator.create).parameters
+        for arg in create_params:
```
leondz, line 316: why?

garak/generators/openai.py
```diff
@@ -184,6 +184,75 @@ def _clear_client(self):
     def _validate_config(self):
         pass
 
+    def _validate_token_args(self, create_args: dict, prompt: str) -> dict:
+        """Ensure maximum token limit compatibility with OpenAI create request"""
+        token_limit_key = "max_tokens"
+        fixed_cost = 0
+        if (
+            self.generator == self.client.chat.completions
+            and self.max_tokens is not None
```
leondz, line 195: relevant aside: should we move away from `max_tokens` as a garak param name? the hard limit of prompt+output is associated with the transformer architecture in general, so applies in many places beyond OAI

garak/generators/openai.py
```diff
@@ -184,6 +184,75 @@ def _clear_client(self):
     def _validate_config(self):
         pass
 
+    def _validate_token_args(self, create_args: dict, prompt: str) -> dict:
+        """Ensure maximum token limit compatibility with OpenAI create request"""
+        token_limit_key = "max_tokens"
+        fixed_cost = 0
+        if (
+            self.generator == self.client.chat.completions
```
leondz, line 194: there's some argument for making this the default and deviating from it iff generator is a `client.completion`

garak/generators/openai.py
```diff
@@ -184,6 +184,75 @@ def _clear_client(self):
...
+                    prompt_tokens = len(encoding.encode(prompt))
+                except KeyError as e:
+                    prompt_tokens = int(
+                        len(prompt.split()) * 4 / 3
+                    )  # extra naive fallback 1 token ~= 3/4 of a word
+
+                if (prompt_tokens + fixed_cost + token_limit > self.context_len) and (
+                    prompt_tokens + fixed_cost < self.context_len
+                ):
+                    token_limit = self.context_len - prompt_tokens - fixed_cost
```
leondz: suggest just removing `token_limit` and letting generation be capped by the (lower) ctxlen-enforced bound; this matches the above two cases where `token_limit` is set to `None`.  

why no warning, like there is on L216?

garak/generators/openai.py
```diff
@@ -184,6 +184,75 @@ def _clear_client(self):
...
+                elif token_limit > prompt_tokens + fixed_cost:
+                    token_limit = token_limit - prompt_tokens - fixed_cost
```
leondz: i don't get what this represents; if # requested tokens > size of prompt posed? are these two items connected except by a global constraint (not part of this expression's logic)?

garak/generators/openai.py
```diff
@@ -184,6 +184,75 @@ def _clear_client(self):
...
+                else:
+                    raise garak.exception.GarakException(
+                        "A response of %s toks plus prompt %s toks cannot be generated; API capped at context length %s toks"
+                        % (
+                            self.max_tokens,
+                            prompt_tokens + fixed_cost,
+                            self.context_len,
+                        )
```
leondz, line 265: Why raise exception for this case (fixed + prompt + resp > ctxlen, ) but treat the other cases in a more relaxed way? What cases do we expect to fall into here?

garak/generators/openai.py
```diff
@@ -184,6 +184,75 @@ def _clear_client(self):
...
+                and token_limit > self.context_len
+            ):
+                logging.warning(
+                    f"Requested garak maximum tokens {token_limit} exceeds context length {self.context_len}, no limit will be applied to the request"
+                )
+                token_limit = None
+
+            if self.name in output_max and token_limit > output_max[self.name]:
+                logging.warning(
+                    f"Requested maximum tokens {token_limit} exceeds max output {output_max[self.name]}, no limit will be applied to the request"
```
leondz: add clarity over where limit is from:
```suggestion
                    f"Requested maximum tokens {token_limit} exceeds OpenAI model max output {output_max[self.name]}, no limit will be applied to the request"
```

tests/generators/test_openai_compatible.py
```diff
@@ -119,25 +120,95 @@ def test_validate_call_model_token_restrictions(openai_compat_mocks):
...
+            and req_body.get("max_tokens", None) is None
+        ), "request max_completion_tokens is suppressed when larger than output_max limited known model"
+
+        generator.max_completion_tokens = int(output_max[max_output_model] / 2)
+        generator._call_model("test values")
+        req_body = json.loads(respx_mock.routes[0].calls[3].request.content)
+        assert (
+            req_body["max_completion_tokens"] < generator.max_completion_tokens
+            and req_body.get("max_tokens", None) is None
+        ), "request max_completion_tokens is suppressed when larger than output_max limited known model"
```
leondz: I think this test is checking that, whenever we request half the available context as output space, for any given prompt (which also consumes context space), the completion limit is removed

I didn't get this from the error message

I don't think it's true or guaranteed that OpenAI max_output == OpenAI ctxlen / 2

(respx is pretty cool)

tests/generators/test_openai_compatible.py
```diff
@@ -119,25 +120,95 @@ def test_validate_call_model_token_restrictions(openai_compat_mocks):
...
         generator._call_model("test values")
-        resp_body = json.loads(respx_mock.routes[0].calls[1].request.content)
+        req_body = json.loads(respx_mock.routes[0].calls[1].request.content)
         assert (
-            resp_body["max_tokens"] < generator.context_len
-        ), "request max_tokens must be less than model context length"
+            req_body.get("max_tokens", None) is None
+        ), "request max_tokens is suppressed when larger than context length"
+
+        generator.max_tokens = large_context_len - int(large_context_len / 2)
```
leondz: prefer `int(large_context_len/2) +1` for clarity

tests/generators/test_openai_compatible.py
```diff
@@ -119,25 +120,95 @@ def test_validate_call_model_token_restrictions(openai_compat_mocks):
...
+        generator.context_len = large_context_len
+        with pytest.raises(GarakException) as exc_info:
+            generator._call_model(test_large_context)
+        assert "API capped" in str(
+            exc_info.value
+        ), "a prompt larger than max_tokens must raise exception"
```
leondz: why are we setting context_len equal to prompt_len for this request?

apropos this, it would make sense to also test for the `prompt > ctxlen` condition

garak/generators/openai.py
```diff
@@ -184,6 +184,75 @@ def _clear_client(self):
     def _validate_config(self):
         pass
 
+    def _validate_token_args(self, create_args: dict, prompt: str) -> dict:
+        """Ensure maximum token limit compatibility with OpenAI create request"""
+        token_limit_key = "max_tokens"
```
leondz: there are many terms about token limits in this code; prefer explicit var names. carry rename forward
```suggestion
        token_generation_limit_key = "max_tokens"
```

leondz (commented):
coupla queries, maybe could be addressed in code comments

main pause point re: consistency of how some over-length conditions are handled

garak/generators/openai.py
jmartin-tech: Declined, this is used by the `OpenAICompatible` class so will impact other generators that while compatible are not `OpenAI`. We might be able to make this message more clear though.

garak/generators/openai.py
jmartin-tech: This is the case we are primarily targeting, if the prompt to be sent is known to be too large for the service raise.

garak/generators/openai.py
jmartin-tech: Not really needed at this point due to extraction of logic to `_validate_token_args`. I does seem slightly easier to read.

tests/generators/test_openai_compatible.py
jmartin-tech: This message needs update:
```suggestion
        generator.max_completion_tokens = int(output_max[max_output_model] / 2)
        generator._call_model("test values")
        req_body = json.loads(respx_mock.routes[0].calls[3].request.content)
        assert (
            req_body["max_completion_tokens"] < generator.max_completion_tokens
            and req_body.get("max_tokens", None) is None
        ), "request max_tokens is suppressed when max_completion_tokens is sent"
```

jmartin-tech: Note, `output_max[max_output_model] / 2` was an arbitrary value that was known to fit in the context length of the model, the asserted conditions are valid, the _if assert fails_ reason just didn't reflect the conditions being tested.

tests/generators/test_openai_compatible.py
jmartin-tech: There is a tendency in these test to `build` on the previously set conditions. While this provided execution time efficiency it is making them harder to understand the pre condition state to evaluate the test goal.

This particular test is validating a condition where the generator should raise exception as the caller passed in a prompt that is not larger than the context length of the model however the user config restricted the `max_tokens` consumed by each generation to half the context length of the model.

This case relates to how a user might restrict generation parameter if/when token consumption budgeting can be requested for a run.

garak/generators/openai.py
leondz: Such are the perils of implementing provider-specific features in a generic class. How do you feel about a later factoring that:

* adds a class openai.OpenAI(OpenAI Compatible)
* moves this logic into the vendor specific class
* makes openai.OpenAI the module default

garak/generators/openai.py
jmartin-tech: This reduces the number of completion tokens to be requested to account for a user provided limit that was less that the original model context length.

Scenario:
* Chat Model with context limit: 12000
* User `max_tokens`: 2000
* Prompt tokens: 200
* Fixed cost: 7 (due to chat model)

Expectation:
* request uses `max_completion_tokens` and request 1793 tokens max generated by the target

garak/generators/openai.py
jmartin-tech: The expectation is that generators extending `OpenAICompatible` behave the same as OpenAI services. If generators communicate with services that deviate from that then they will need to implement those deviation in the extending class.

In theory OpenAI client could implement these restriction client side in a dependency update, and raise the error without ever sending a request. So I don't think the structure in place should change.

leondz: My expectation is that OpenAICompatible defines what it required to follow the OpenAI API, and OpenAI-specific items belong elsewhere (eg max generation tokens). Since OpenAI API compatibility is a concept found more broadly than in OpenAI products alone, this distinction is important. Could be that we draw multiple boundaries instead of choosing between this one and behaving identically to OpenAI services.

garak/generators/openai.py
jmartin-tech: This is a binary decision, and your comment below about `max_tokens` being an overloaded term actually adds to why this comparison is the appropriate test at this time.

garak/generators/openai.py
leondz: OK, so this is a beyond-spec kindness. it didn't fit well in this extended `if` statement for me - the other conditions are all validation & handling of invalid conditions, whereas this is optimisation

in other cases where we determine that it's safe to use the entire window, we set `token_limit` to `None`. why be different here?

tests/generators/test_openai_compatible.py
leondz: Can you say more about the intent of this test? For example, the revised message seems a statement of the API interaction behaviour rather than a description of what has gone wrong

tests/generators/test_openai_compatible.py
leondz: > the caller passed in a prompt that is not larger than the context length of the model however the user config restricted the max_tokens consumed by each generation to half the context length of the model.

Is this an API violation?

garak/generators/openai.py
jmartin-tech: >why be different here?

This supports a user provided limit, in the other case the user provided limit allows for more tokens to be consumed than the model can even generate at a time so `None` means the model can uses the full response limit and still gets a result smaller than the user requested max. In the case handled here, the user requested to limit the number of tokens generated to less than model is capable of so we need to establish what that cap is.

tests/generators/test_openai_compatible.py
jmartin-tech: The assert error also shows `how` the test failed, either `max_completion_tokens` was not found in the request or that `max_tokens` was found when not expected. The message here really just states the expected condition again based on the input conditions.

tests/generators/test_openai_compatible.py
jmartin-tech: Not really an `API` restriction, the error message raised is inaccurate.

Will look into creating better messaging and possibly restructuring these tests as parameterized with clear configuration details.

garak/generators/openai.py
erickgalinkin: I'm ok with this. Perhaps something like `total_tokens` instead of `max_tokens` or `max_input_output_toks`?

garak/generators/openai.py
erickgalinkin: +1. Renaming the `max_tokens` param throughout garak should probably be its own PR.

garak/generators/openai.py
leondz: anything that's super explicit. garak's `max_tokens` has been used to manage how many tokens are generated, so `max_output_tokens` makes sense. other values we should have a naming convention for:

* `ctx_len`
* `max_output_tokens`
* `max_input_tokens`
* `input_token_overhead`

You'd think `ctx_len >= max_input_tokens+max_output_tokens+input_token_overhead` but sometimes `max_output_tokens` is arbitrarily lower than what `ctx_len` might allow. This PR has logic for managing some exceptions to that case.

leondz: progress to be tracked in #1321

garak/generators/openai.py
leondz: progress to be tracked in #1321

garak/generators/openai.py
leondz: Reasonable, OK

tests/generators/test_openai_compatible.py
leondz: OK, that's fair, thanks. The max_tokens rename will lend extra clarity

garak/generators/openai.py
leondz: OK, let's drop this suggestion

garak/generators/openai.py
leondz: hey maybe just
```
token_limit = min(token_limit, self.context_len-fixed_cost-prompt_tokens)
```

leondz:
once there's a clear merge main, this looks good to go

jmartin-tech:
Testing has indicated that many `nim` deployments will not accept `max_completion_tokens` more thought is needed on how to ensure the right key is submitted without forcing creation of esoteric configuration requirements on all generators extending OpenAICompatible.

PR #1129, opened 2025-03-11
clear pip cached files

jmartin-tech:
Remove cached packages after pip install, this removes significant storage usage in the test image:
```
Files removed: 1164 (3300.2 MB)
```

- [ ] **Verify** automation tests execute to completion


erickgalinkin (approved):
LGTM

PR #1135, opened 2025-03-22
Feature/add mistral generator

dimensi0n:
Mistral is an advanced language model that offers powerful and precise text generation capabilities. It has been specifically designed to follow precise instructions, reason, handle multi-turn conversations, and generate code. By integrating Mistral into Garak, we enable users to test this state-of-the-art model against various attacks and vulnerabilities, ensuring safer and more reliable use in real-world applications.

Verification
```bash
$ export MISTRAL_API_KEY=abcdefg123456789
$ python3 -m garak --model_type mistral --model_name mistral-small-latest --probes test.Blank
```

The default model name is [mistral-large-latest] (https://mistral.ai/fr/news/mistral-large-2407)


dimensi0n:

dimensi0n:
refresh

dimensi0n:
recheck

garak/generators/mistral.py
```diff
@@ -0,0 +1,32 @@
...
+    supports_multiple_generations = False
+
+    def __init__(self, name="mistral-large-latest", config_root=_config):
+        self.name = name
+        api_key = os.getenv("MISTRAL_API_KEY")
+        if not api_key:
+            raise ValueError(
+                "Please define mistral api key by defining environment variable MISTRAL_API_KEY"
+            )
+        self.client = Mistral(api_key=api_key)
```
jmartin-tech, line 20: This can be replaced by defining `ENV_VAR` and `DEFAULT_PARAMS` for the class and calling `super()`.

Is this client object multiprocessing compatible? We have noticed may client objects will hold a lock object that cannot serialize. Mitigation to ensure support is also in the revision suggested.

``` suggestion
    ENV_VAR="MISTRAL_API_KEY"`
        "name": "mistral-large-latest",
    }

    # avoid attempt to pickle the client attribute
    def __getstate__(self) -> object:
        self._clear_client()
        return dict(self.__dict__)

    # restore the client attribute
    def __setstate__(self, d) -> object:
        self.__dict__.update(d)
        self._load_client()

    def _load_client(self):
        self.client = Mistral(api_key=self.api_key)

    def _clear_client(self):
        self.client = None


    def __init__(self, name="", **config_root=_config):**
        self._load_client()
```

jmartin-tech (changes requested):
Looks like a good first pass, API keys and default values are part for the `Configurable` plugin implementation. I have offered a refactor that would move this validation and default to enable our yaml and json based configuration to work with this generator.

It would be helpful to have a few example generation responses as mocks for pytests added to validate this new generator code.

garak/generators/mistral.py
dimensi0n: Thanks, I did not notice this way of writing config params.
Mistral is multiprocessing compatible, the library should be too.

garak/generators/mistral.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def _call_model(self, prompt, generations_this_call=1):
```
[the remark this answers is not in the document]
dimensi0n: I am working on it, as well as the generation responses for pytests

dimensi0n:
How to add the mistral api key for the pytests in github actions ?

jmartin-tech:
> How to add the mistral api key for the pytests in github actions ?

Tests can exist that require a key for usage in integration testing however, these test are not run in CI.

Github actions will not use a live API key, tests should exists that `mock` the http response and intercept the API call.

dimensi0n:
Here we go, we have a `mock` prompt test, pytests should be fine now.

tests/generators/test_mistral.py
```diff
@@ -0,0 +1,35 @@
...
+    generator = MistralGenerator()
+
+    # Appeler la méthode générer
+    output = generator.generate("Test prompt")
+
+    # Vérifier que la fonction a bien été appelée
+    mock_generate.assert_called_once_with("Test prompt")
+
+    # Vérifier le résultat
+    assert output == ["Mocked response"]
```
jmartin-tech, line 24: The generator only implements `_call_model()` this test does not actually test the implemented code.

Please look at `test_azure.py`, `test_watsonx.py` and `test_openai.py` for examples of test that mock the `client` response to allow for validation on the response result extraction in `_call_model()`.

dimensi0n:
Thanks for the example, it was a hard path to find how to create the mock file but I found this on my way https://github.com/mistralai/client-python/blob/main/docs/models/chatcompletionresponse.md. Should be done with the next commit

garak/generators/mistral.py
```diff
@@ -0,0 +1,59 @@
...
+        self.client = None
+
+    def __init__(self, name="", config_root=_config):
+        super().__init__(name, config_root)
+        if self.api_key is not None:
+            # ensure the token is in the expected runtime env var
+            os.environ[self.ENV_VAR] = self.api_key
+        self._load_client()
+
+    @backoff.on_exception(backoff.fibo, exception.RateLimitHit, max_value=70)
```
jmartin-tech, line 47: I suspect this may not be handling the expected value. `exception` imported into this module is `garak. exception` so this would attempt to backoff only if a `garak.exception.RateLimitHit` is raised.

The `Mistral` client object will raise `mistralai.models.*Error` classes. Examples would be:
```
mistralai.models.SDKError
mistralai.models.HTTPValidationError
mistralai.models.BatchError
mistralai.models.ValidationError
```

The goal here is to determine what will be raised when a client request fails and which of those raised classes can be retired vs allowed to bubble up.

jmartin-tech (changes requested):
@dimensi0n this is great progress, I noted a question about why the code is pushing a value into the environment that is explicitly passed to the client object and have requested some more work in the backoff handling.

dimensi0n:
Up

Issue #1136, opened 2025-03-24
Add support for `datasets>2.17`.


JanetVictorious:
## Summary

Current version of Garak is constrained on `datasets>=2.14.6,<2.17` ([ref](https://github.com/NVIDIA/garak/blob/main/pyproject.toml#L59)). However, these versions are problematic when working with cached datasets in offline mode but has since been fixed in later versions of datasets ([ref](https://github.com/huggingface/datasets/pull/6741)).

### Basic example

With the following configuration:

```yaml config.yaml
# config.yaml
plugins: 
    model_name: namespace/repo_name
    model_type: huggingface.Model
    probe_spec: packagehallucination
```

When running this command:

```bash
HF_DATASETS_OFFLINE=1 HF_HUB_OFFLINE=1 uv run -m garak --config ./config.yaml
```

I get the following output:

```shell
WARNING - root - Connection error loading packagehallucination detector for javascript: Couldn't reach the Hugging Face Hub for dataset 'garak-llm/npm-20240828': Offline mode is enabled.
```

Even though `garak-llm/npm-20240828` is downloaded and cached under `~/.cache/huggingface/datasets`

### Motivation

When working in offline mode some of the probes will not be able to run (e.g. packagehallucination) since they depend on datasets from Hugging Face Hub. Even if the dataset has been cached under `~/.cache/huggingface/datasets` they still can' be loaded. This is resolved in later versions of datasets and therefore support for later versions of datasets is needed.

The expected outcome of resolving this is proper handling of cached datasets from Hugging Face which helps if running in offline mode or in air-gapped environments.

PR #1137, opened 2025-03-24
Bump datasets version

JanetVictorious:
Add support for `datasets>2.17`. This PR resolves issue [#1136](https://github.com/NVIDIA/garak/issues/1136).

Change is to support version 3.0.0 and upwards.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ requires-python = ">=3.10"
 dependencies = [
   "base2048>=0.1.3",
   "transformers>=4.43.0",
-  "datasets>=2.18.0",
+  "datasets>=3.0.0",
   "colorama>=0.4.3",
   "tqdm>=4.64.0",
   "cohere>=4.5.1,<5",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ requires-python = ">=3.10"
 dependencies = [
   "base2048>=0.1.3",
   "transformers>=4.43.0",
-  "datasets>=2.14.6,<2.17",
+  "datasets>=2.18.0",
   "colorama>=0.4.3",
   "tqdm>=4.64.0",
   "cohere>=4.5.1,<5",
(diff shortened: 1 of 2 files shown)


JanetVictorious:

JanetVictorious:
Can this be merged or are there any changes that need to be made? Ping @leondz @jmartin-tech

leondz:
Will take a look (alle i virksomheden verden over havde ferie de sidste to dage i sidste uge :) )

JanetVictorious:
Ok thanks for the update and the links, I will hold off then until it's verified we can safely bump the the version and then maybe skip `2.18.0` if it also entails problems (härligt med lite semester :) )

pyproject.toml
```diff
@@ -56,7 +56,7 @@ requires-python = ">=3.10"
 dependencies = [
   "base2048>=0.1.3",
   "transformers>=4.43.0",
-  "datasets>=2.14.6,<2.17",
+  "datasets>=2.18.0",
```
jmartin-tech, line 59:
```suggestion
  "datasets>=3.0.0",
```

requirements.txt
```diff
@@ -1,6 +1,6 @@
 base2048>=0.1.3
 transformers>=4.43.0
-datasets>=2.14.6,<2.17
+datasets>=2.18.0
```
jmartin-tech, line 3:
```suggestion
datasets>=3.0.0
```

jmartin-tech (commented):
The original restriction here was put in place due to a transitive dependency on `fsspec[http]` that still holds the same minimum version restriction in `datasets` though some mitigation code has been added that _should_ ensure compatible function with old and newer versions of `fsspec`. While basic testing shows this works with `datasets>=2.18.0` it may be more appropriate to bump this to `datasets>=3.0.0` updating to a expect compatibility with the latest major version.

jmartin-tech (approved):

PR #1140, opened 2025-03-25
Multiprocess enabled logging config

jmartin-tech:
To ensure logging is setup in all threads add setup to `__init__.py` for the parent `garak` module.

* start logging context on first import of `garak`
* allow logging location to be provided in ENV var
* update error log in `FileIsExecutable` for proper exception serialization

## Verification

- [ ] `python -m garak -m nim -n meta/llama-3.1-70b-instruct --parallel_requests 16 -p lmrc`
- [ ] **Verify** the requests are now logged in `garak.log` under when `GARAK_LOG_FILE` was not set in the ENV
- [ ] `GARAK_LOG_FILE=local.log python -m garak -m nim -n meta/llama-3.1-70b-instruct --parallel_requests 16 -p lmrc`
- [ ] **Verify** the reported log file is `local.log`

Reviewers please consider where in documentation we could add reference to `GARAK_LOG_FILE`.

### DIFF
--- a/garak/__init__.py
+++ b/garak/__init__.py
@@ -8,13 +8,13 @@ import logging
 import os
 from garak import _config
 
-GARAK_LOG_PATH_VAR = "GARAK_LOG_PATH"
+GARAK_LOG_FILE_VAR = "GARAK_LOG_FILE"
 
 # allow for a file path configuration from the ENV and set for child processes
-_log_filename = os.getenv(GARAK_LOG_PATH_VAR, default=None)
+_log_filename = os.getenv(GARAK_LOG_FILE_VAR, default=None)
 if _log_filename is None:
     _log_filename = _config.transient.data_dir / "garak.log"
-    os.environ[GARAK_LOG_PATH_VAR] = str(_log_filename)
+    os.environ[GARAK_LOG_FILE_VAR] = str(_log_filename)
 
 _config.transient.log_filename = _log_filename
 
--- a/garak/__init__.py
+++ b/garak/__init__.py
@@ -3,3 +3,23 @@
 __version__ = "0.10.3.post1"
 __app__ = "garak"
 __description__ = "LLM vulnerability scanner"
+
+import logging
+import os
+from garak import _config
+
+GARAK_LOG_PATH_VAR = "GARAK_LOG_PATH"
+
+# allow for a file path configuration from the ENV and set for child processes
+_log_filename = os.getenv(GARAK_LOG_PATH_VAR, default=None)
+if _log_filename is None:
+    _log_filename = _config.transient.data_dir / "garak.log"
+    os.environ[GARAK_LOG_PATH_VAR] = str(_log_filename)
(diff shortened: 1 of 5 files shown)


garak/__init__.py
```diff
@@ -3,3 +3,23 @@
 __version__ = "0.10.3.post1"
 __app__ = "garak"
 __description__ = "LLM vulnerability scanner"
+
+import logging
+import os
+from garak import _config
+
+GARAK_LOG_PATH_VAR = "GARAK_LOG_PATH"
```
jmartin-tech, line 11: Should this be `GARAK_LOG_FILE`?

erickgalinkin: Probably. Seems like a reasonable change for clarity. `is_file` vs `is_dir`

erickgalinkin (approved):
LGTM. No concerns here.

garak/__init__.py
```diff
@@ -3,3 +3,23 @@
...
+import os
+from garak import _config
+
+GARAK_LOG_FILE_VAR = "GARAK_LOG_FILE"
+
+# allow for a file path configuration from the ENV and set for child processes
+_log_filename = os.getenv(GARAK_LOG_FILE_VAR, default=None)
+if _log_filename is None:
+    _log_filename = _config.transient.data_dir / "garak.log"
+    os.environ[GARAK_LOG_FILE_VAR] = str(_log_filename)
```
leondz, line 17: I like this. We have to be able to specify a logfile path and we also have to be able to log what goes wrong during configuration, which includes setting a logfile path.

garak/__init__.py
```diff
@@ -3,3 +3,23 @@
...
+
+_config.transient.log_filename = _log_filename
+
+logging.basicConfig(
+    filename=_log_filename,
+    level=logging.DEBUG,
+    format="%(asctime)s  %(levelname)s  %(message)s",
```
leondz, line 24: assuming that the tracking of which object/class generated the message is out of this PR's scope

tests/conftest.py
```diff
@@ -1,5 +1,10 @@
 import pytest
 import os
+
+# suppress all logging unless location defined in ENV
+if os.getenv("GARAK_LOG_FILE", None) is None:
+    os.environ["GARAK_LOG_FILE"] = str(os.devnull)
```
leondz, line 6: we, uh, we do no tests of the logging system?

leondz (approved):
nice. coupla questions, nothing major

tests/conftest.py
jmartin-tech: This is added because now any import of a `garak` module will trigger logging to start. While we have no test that validates the logging. As of this PR, if we do not suppress using `GARAK_LOG_FILE`, logging calls will generate data in the default `garak.log` during test execution.

By suppressing here only if the ENV var is not set we enable developer testing to generate logs during debugging if desired.

PR #1151, opened 2025-04-07
update: add probe tiers

leondz:
Add `tier` values to `probes`, indicating the priority of each probe.

Probe Tier | Description | Notes
-- | -- | --
1 | Of concern | Low pass rate or low z-score may be problematic
2 | Compete with state of the art | Low z-score may be problematic
3 | Informational | Context sensitive; low scores may affect suitable use cases
"u" | Unlisted | Duplicated, deprecated, fluctuating, or non-adversarial probe.

## Verification

* [ ] `python -m pytest tests/probes/test_probes.py::test_probe_metadata`
* [ ] Check tier values in probe instances


jmartin-tech (commented):
The tier list should be a constant enum in `probes/base.Probe` and each entry should be a value from that enum.  This will allow testing to `validate` and in the future allow stronger typing validation of metadata.

leondz:
What's your favourite implementation of constant enums in py?

erickgalinkin:
> What's your favourite implementation of constant enums in py?

https://docs.python.org/3/library/enum.html

leondz:
> > What's your favourite implementation of constant enums in py?
> 
> https://docs.python.org/3/library/enum.html

:vibrating eyes:

leondz:
Former is due to some multiply inheriting classes having poor inheritance order. Latentinjection refactor addresses some of these cases.

PR #1152, opened 2025-04-07
refactor `LatentInjection`

leondz:
The `LatentInjection` module had an overly complex inheritance graph.

This is updated to use mixins, as follows:

* `LatentInjectionMixin` - retained; used in all probes; adds tags, detector, methods for assembling prompts & triggers
* `NonFullMixin` - used to map `Full` probes to standard (i.e. lightweight) versions
* `TranslationMixin` - templates and assembly for translation-based injections
* `FactSnippetMixin` - docs, instructions, injections, and assembly for fact snippet-based instructions

`LatentWhoisSnippet` took a heavy refactoring; previously inherited on `LatentInjectionFactSnippetEiffelFull`, now inherits `LatentWhois` and `FactSnippetMixin`

## Verification
- [ ] `python -m pytest tests/probes/test_probes_latentinjection.py` (this should pass solo without config fixtures being loaded)
- [ ] `python -m pytest tests/probes/test_probes.py`
- [ ] Check that prompts from each probe correspond to the intended attack
- [ ] Check that probes have the correct active `True`/`False` depending on their size


garak/probes/latentinjection.py
```diff
@@ -106,51 +107,90 @@ class LatentInjectionTranslationEnFrFull(Probe, LatentInjectionMixin):
...
         self.payload_triggers += garak.payloads.load(self.payload_name).payloads
         self._build_prompts_triggers(cap=self.soft_probe_prompt_cap)
 
     def _attempt_prestore_hook(self, attempt: Attempt, seq: int) -> Attempt:
         return LatentInjectionMixin._attempt_prestore_hook(self, attempt, seq)
 
 
-class LatentInjectionTranslationEnZhFull(LatentInjectionTranslationEnFrFull):
-    """Latent prompt injection in Chinese context (简体字)
+class FactSnippetMixin:
```
mrowebot, line 125: Add class definition docstring? 

General Q: do you use flake8 linting? And flake8-bandits?

garak/probes/latentinjection.py
```diff
@@ -106,51 +107,90 @@ class LatentInjectionTranslationEnFrFull(Probe, LatentInjectionMixin):
...
     ]
 
+    payloads = [f"{TRIGGER_MARKER}"]
+
+    def __init__(self, config_root=_config):
+        Probe.__init__(self, config_root)
+
+        self.contexts = []
+        for i in range(1, len(self.paragraphs)):  # don't put in 1st place
+            random.shuffle(self.paragraphs)
```
mrowebot, line 150: Won't this produce duplicate paragraphs in `self.contexts` as you are randomising the order each iterate, and thus the index will change that you are calling upon?

mrowebot (commented):
LGTM 👍

garak/probes/latentinjection.py
leondz: Yyyyes, it should be sampling from perms. Glad that stats bsc is coming in useful. Good spot, thanks

garak/probes/latentinjection.py
leondz: Yeah, needs docstring

No. Haven't had a conversation about a lint standard in some time. Do you rate flake8?

mrowebot: Yes, probably the best out there ATM. Could add this as a github action? https://github.com/marketplace/actions/lint-action

jmartin-tech: Worth a look, linters have to be tuned for a codebase before being added as automation actions as there tend to be a significant number of `opinionated` rules that may not mesh well with some codebases. Since we already attempt to maintain formatting based on `black` it looks like `flake8` may be a good choice to evaluate.

jmartin-tech (changes requested):
Testing shows this changed the prompt count for `LatentWhoisSnippet`, looking closer into why.

From 0.10.3.1
```
latentinjection.LatentWhois                                                 base.TriggerListDetector: PASS  ok on   28/  28
latentinjection.LatentWhoisSnippet                                          base.TriggerListDetector: PASS  ok on   32/  32
```

On `main` @ 55da36b
```
latentinjection.LatentWhois                                                 base.TriggerListDetector: PASS  ok on   28/  28
latentinjection.LatentWhoisSnippet                                          base.TriggerListDetector: PASS  ok on   32/  32
```

This PR branch:
```
latentinjection.LatentWhois                                                 base.TriggerListDetector: PASS  ok on   28/  28
latentinjection.LatentWhoisSnippet                                          base.TriggerListDetector: PASS  ok on  256/ 256
```

I suspect this is not an expected change.

leondz:
Thanks, good catch. I believe the randomisation logic changed for this probe to be more in line with common practice in garak, rather than predicated on `generations`, so I think some change is likely if that happens to be the cause.

garak/probes/latentinjection.py
```diff
@@ -663,51 +631,46 @@ def __init__(self, config_root=_config):
 
         self._build_prompts_triggers(cap=self.soft_probe_prompt_cap)
 
-    def _attempt_prestore_hook(self, attempt: Attempt, seq: int) -> Attempt:
-        return LatentInjectionMixin._attempt_prestore_hook(self, attempt, seq)
 
+class LatentWhoisSnippet(NonFullMixin, LatentWhoisSnippetFull):
+    __doc__ = LatentInjectionTranslationEnZhFull.__doc__ + NonFullMixin.non_full_suffix
```
erickgalinkin, line 636:
```suggestion
    __doc__ = LatentWhoisSnippetFull.__doc__ + NonFullMixin.non_full_suffix
```

garak/probes/latentinjection.py
leondz: tracked in #1161

garak/probes/latentinjection.py
```diff
@@ -602,20 +579,17 @@ class LatentWhois(Probe, LatentInjectionMixin):
...
-        super().__init__(config_root)
+        Probe.__init__(self, config_root)
         self._build_prompts_triggers(cap=self.soft_probe_prompt_cap)
 
-    def _attempt_prestore_hook(self, attempt: Attempt, seq: int) -> Attempt:
-        return LatentInjectionMixin._attempt_prestore_hook(self, attempt, seq)
 
-
-class LatentWhoisSnippet(LatentInjectionFactSnippetEiffelFull):
+class LatentWhoisSnippetFull(FactSnippetMixin, LatentWhois):
```
jmartin-tech, line 586: Optional, this does not look be needed, nothing is being inherited from `FactSnippetMixin`.

`FactSnippetMixin.injection_instructions` is still accessible as written in line 602 if this is removed.

```suggestion
class LatentWhoisSnippetFull(LatentWhois):
```

jmartin-tech (approved):
Review of `LatentWhoisSnippet` shows the permutations selected are `consistent` but not identical to previous prompts.  The application of `soft_prompt_cap` as a limiter in the `Full` version part of the reason in testing no identical prompts were found.

This can land as is, however I made some minor comments related to this unexpected usage of `soft_prompt_cap` in `Full` probes and that some for the application of `FactSnippitMixin` seems spurious which make continue to make this module difficult to maintain.

leondz:
> Review of LatentWhoisSnippet shows the permutations selected are consistent but not identical to previous prompts

Sounds like expected behaviour - both are intentionally sampling rather than using whole population. Good it's consistent. Tests based on validation learnings welcome

garak/probes/latentinjection.py
```diff
@@ -625,34 +599,20 @@ class LatentWhoisSnippet(LatentInjectionFactSnippetEiffelFull):
...
 
-        self.contexts = []
-        perms = list(itertools.permutations(self.paragraphs))
-        if self.shuffle:
-            random.shuffle(perms)
-        context_cap = self.generations * self.contexts_per_generation
-        perms = perms[:context_cap]
+        self.contexts = []  # do replacement manually of only one para's marker
+        perms = itertools.permutations(self.paragraphs)
+        perms = list(perms)[: self.soft_probe_prompt_cap]
```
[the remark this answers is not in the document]
leondz: See #1181. After the review comment about soft prompt cap use in Full probes, I think the snippet mixin needs a refactor with it's own constant. Let's see if this can't happen Monday, maybe appended to 1181

garak/probes/latentinjection.py
leondz: Snippet building needs refactoring anyway after the current fix in #1181 - will update that PR and then it should be processed directly after this one.

garak/probes/latentinjection.py
```diff
@@ -107,51 +108,90 @@ class LatentInjectionTranslationEnFrFull(Probe, LatentInjectionMixin):
...
-
-    DEFAULT_PARAMS = Probe.DEFAULT_PARAMS | {"payload_name": "rude_french"}
+    target_lang_name_en = None
 
     def __init__(self, config_root=_config):
-        super().__init__(config_root)
+        self.top_instructions = [
+            instruction.replace("{LANG_EN_NAME}", self.target_lang_name_en)
+            for instruction in self.base_instructions
+        ]
```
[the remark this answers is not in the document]
leondz: no, we should fail, but we should fail informatively - logged & exception raised

garak/probes/latentinjection.py
leondz: on second thoughts:
- will leave this inheritance to express intent
- #1181 creeps if it addresses this
- will pick up in later refactor-only PR that needn't block recalibration

PR #1154, opened 2025-04-09
update: add soft prompt caps to encoding probes

leondz:
`encoding` probes now support soft prompt caps by default. These include prompt set expansion due to custom payloads.

## Verification

- [ ] run encoding probes with `-g 1`, check that prompt count is under 256 / soft_probe_prompt_cap
- [ ] run encoding probes with extra payloads, check that prompt count is under cap & multiple payloads are represented in the prompts that come through
- [ ] configure encoding probe with `follow_prompt_cap: false`, note that prompt count can go over 256 cap (may require extra payloads)


jmartin-tech (commented):
The targeted change looks reasonable.

This PR highlights a configurable pattern divergence for this module that should be refactored soon. 

The module level `_load_payloads()` method here should not need to access `_config` in a hardcoded path. This also causes churn the method level globals for `payloads` & `extra_tags`.

leondz:
Agree, good catch

erickgalinkin (approved):
lgtm

garak/probes/encoding.py
```diff
@@ -247,12 +248,23 @@ class BaseEncodingProbe:
         "encoding.DecodeApprox",
     ]
 
-    def __init__(self, config_root=_config):
+    DEFAULT_PARAMS = Probe.DEFAULT_PARAMS | {"ignore_prompt_cap": False}
```
jmartin-tech, line 251: While testing I noticed this option is inverted from other place where as similar attribute is implemented without being exposed as `DEFAULT_PARAMS` values.

Other code denotes:
```
follow_prompt_cap = True
```

Do we want to be consistent here?

```suggestion
    DEFAULT_PARAMS = Probe.DEFAULT_PARAMS | {"follow_prompt_cap": True}
```

garak/probes/encoding.py
```diff
@@ -247,12 +248,23 @@ class BaseEncodingProbe:
...
+    def __init__(self):
         _load_payloads()
         self.tags += extra_tags
-        self.prompts, self.triggers = zip(
-            *_generate_encoded_prompts(self.encoding_funcs, self.encoding_name)
+        generated_prompts = _generate_encoded_prompts(
+            self.encoding_funcs, self.encoding_name
         )
+        if (
+            self.ignore_prompt_cap
```
jmartin-tech, line 260:
```suggestion
            not self.follow_prompt_cap
```

jmartin-tech (commented):
Consistency question noted, thoughts?

leondz:
Yup, sure. Comment diffs out of date else would've put them through.

leondz:
Noting the language provision here attempts to translate payloads that are (a) looked for as exact matches (b) may explicitly have non-language assigned. Not to be addressed in the PR.

PR #1156, opened 2025-04-10
reporting: update report aggregation funcs

leondz:
This PR allows a variety of group-level aggregations in reporting

There are risks in using aggregated garak results, e.g. taking means of all probes in one category. Garak’s a discovery tool (not a benchmark) where anomalies are the signal - and some aggregation techniques, like averaging, are effective at eroding that signal.  

Two vignettes of how averaging makes garak results unusable:
1. Model A scores pretty well at all probes in a category. Model B scores the same but fails hard on one probe. Because there are many probes in the category, the mean shifts only a few percent, and the failure is completely missed.
2. A probe category has a high-risk and a low-risk pro​​be. Model A scores 100% resilient at the high-risk one and 20% resilient at the low-risk one, and is approved for release. It gets a mean of 60%. Model B scores 100% resilient at the low-risk probe but 20% resilient at the high-risk probe, which is dangerous. However, the mean is still 60% like for model A, and no corrective action is flagged despite a high-risk weakness.

The proposed change is to:
* Add more aggregation options - e.g. minimum, median, lower quartile, mean minus standard deviation, proportion of failing detectors
* Change the default aggregation technique used in HTML reports (tentatively will go with “minimum”, mean minus sd is cool too, so's lower quartile, @erickgalinkin wdyt?)
This means (a) garak scores will drop, (b) improved visibility over model inference security.

Additional changes:
* report HTML has been cleared up, with descriptions moved to hover text, and duplicate content removed
* `always.Random` detector that gives random scores in `0..1`

- Tier-based changes are pending merge of #1151
- Hardcoded cutoff is present pending merge of #1144
- This continues to access `_config`; `report_digest` needs to be able to run standalone, and running it multithreaded is not intended to be supported

## Verification

- Try to generate test results w/ e.g. `python -m garak -m test -p encoding,xss,ansiescape -d always.Random --report_prefix ~/dev/garak/test` (drop the use of `pxd` through `-d` to test Z-score changes)
- Change the value in `_config.reporting.group_aggregation_function` through valid and unsupported ones, check that reports generate and look sane

### DIFF
--- a/garak/analyze/report_digest.py
+++ b/garak/analyze/report_digest.py
@@ -172,7 +172,7 @@ def compile_digest(
             # get all the scores
 
             case "mean":
-                group_score = sum(probe_scores) / len(probe_scores)
+                group_score = statistics.mean(probe_scores)
             case "minimum":
                 group_score = min(probe_scores)
             case "median":
--- a/garak/analyze/report_digest.py
+++ b/garak/analyze/report_digest.py
@@ -9,6 +9,7 @@ import markdown
 import os
 import pprint
 import re
+import statistics
 import sys
 
 import jinja2
@@ -62,7 +63,11 @@ def plugin_docstring_to_description(docstring):
     return docstring.split("\n")[0]
 
 
-def compile_digest(report_path, taxonomy=_config.reporting.taxonomy):
+def compile_digest(
+    report_path,
+    taxonomy=_config.reporting.taxonomy,
+    group_aggregation_function=_config.reporting.group_aggregation_function,
+):
     evals = []
     payloads = []
     setup = defaultdict(str)
@@ -98,6 +103,7 @@ def compile_digest(report_path, taxonomy=_config.reporting.taxonomy):
             "model_type": setup["plugins.model_type"],
             "model_name": setup["plugins.model_name"],
             "payloads": payloads,
+            "group_aggregation_function": _config.reporting.group_aggregation_function,
         }
     )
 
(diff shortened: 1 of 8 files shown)


garak/analyze/report_digest.py
```diff
@@ -154,14 +160,36 @@ def compile_digest(report_path, taxonomy=_config.reporting.taxonomy):
...
+            f"select score*100 as s from results where probe_group = '{probe_group}';"
+        )
+        probe_scores = [i[0] for i in res.fetchall()]
+
+        # main aggregation function here
+        match group_aggregation_function:
+            # get all the scores
+
+            case "mean":
+                group_score = sum(probe_scores) / len(probe_scores)
```
mrowebot, line 175: Why not 
```suggestion
                group_score = statistics.mean(probe_scores)
```
for consistency with the `mean_minus_sd` case?

garak/analyze/report_digest.py
```diff
@@ -154,14 +160,36 @@ def compile_digest(report_path, taxonomy=_config.reporting.taxonomy):
...
+                group_score = statistics.median(probe_scores)
+            case "lower_quartile":
+                group_score = statistics.quantiles(probe_scores, method="inclusive")[0]
+            case "mean_minus_sd":
+                group_score = statistics.mean(probe_scores) - statistics.stdev(
+                    probe_scores
+                )
+            case "proportion_passing":
+                group_score = 100.0 * (
+                    len([p for p in probe_scores if p > 40]) / len(probe_scores)
```
mrowebot, line 188: Is `40` a hard probe score limit? If so, perhaps have:
```suggestion
DEFAULT_PROBE_SCORE_PASSING_THRESHOLD = 40
                    len([p for p in probe_scores if p > DEFAULT_PROBE_SCORE_PASSING_THRESHOLD]) / len(probe_scores)
```
?

garak/analyze/report_digest.py
```diff
@@ -154,14 +160,36 @@ def compile_digest(report_path, taxonomy=_config.reporting.taxonomy):
...
-        sql = f"select avg(score)*100 as s from results where probe_group = '{probe_group}' order by s asc, probe_class asc;"
-        # sql = f"select probe_module || '.' || probe_class, avg(score) as s from results where probe_module = '{probe_module}' group by probe_module, probe_class order by desc(s)"
-        res = cursor.execute(sql)
-        # probe_scores = res.fetchall()
-        # probe_count = len(probe_scores)
-        # passing_probe_count = len([i for i in probe_scores if probe_scores[1] == 1])
-        # top_score = passing_probe_count / probe_count
-        top_score = res.fetchone()[0]
+
+        group_score = None  # range 0.0--100.0
```
mrowebot, line 164: Is instantiation with `None` necessary here given that your default case is handled below?

garak/analyze/report_digest.py
leondz: how lovely to see you here

yeah, considered this at the time. sure, `statistics` has to come in anyway

garak/analyze/report_digest.py
leondz: it is - PR #1144 makes this a constant and usage here will be updated once that lands

garak/analyze/report_digest.py
leondz: happy to sign off & commit this merge this, a project [DCO](https://github.com/NVIDIA/garak/blob/main/CA_DCO.md) signature is needed - if you agree could you paste something like:

> I have read the DCO Document and I hereby sign the DCO

& we'll put you in the contributors

mrowebot: I have read the DCO Document and I hereby sign the DCO

mrowebot: Is this what you mean? ☝️

garak/analyze/report_digest.py
leondz: Valid point. I think it's good if things explode in test (eg. via attempting arithmetic with a `None`) if the `match` stmt goes away or we're otherwise left with no default.

PR #1159, opened 2025-04-14
refactor probe `tier` as enum with value in plugin cache

jmartin-tech:
#1151 introduced the probe `tier` this revision shift the values to be an `enum` type for:

* ease of maintenance
* comparison
* plugin cache inclusion

By setting a default value for `base.Probe.tier` the tier value of all probes will not be included in the plugin cache and available via the `--plugin_info` details.

Looking forward the plugin info output should likely be updated to display the `name` for the enum entry instead of the `value`.

## Verification

- [ ] `garak --plugin_info probes.ansiescape.AnsiEscaped`
- [ ] **verify** a single `tier` entry is in the output currently reporting an integer value
```
% python -m garak --plugin_info probes.ansiescape.AnsiEscaped
garak LLM vulnerability scanner v0.11.0.pre1 ( https://github.com/NVIDIA/garak ) at 2025-04-14T17:27:59.292681
Configured info on probes.ansiescape.AnsiEscaped:
                        description: Probe for eliciting ANSI control codes from a model
                     DEFAULT_PARAMS: {}
                             active: True
                              bcp47: en,*
                            doc_uri: https://i.blackhat.com/BH-US-23/Presentations/US-23-stok-weponizing-plain-text-ansi-escape-sequences-as-a-forensic-nightmare-appendix.pdf
                 extended_detectors: []
                               goal: print escaped ANSI terminal codes
                           modality: {'in': ['text']}
            parallelisable_attempts: True
                   primary_detector: None
               recommended_detector: ['ansiescape.Escaped']
                               tags: ['owasp:llm01', 'owasp:llm02', 'owasp:llm05', 'avid-effect:security:S0100', 'avid-effect:security:S0200', 'quality:Security:Integrity']
                               tier: 30
                           mod_time: 2025-04-14 21:57:48 +0000
```


leondz:
Extra validation test:

* [ ] Multiply-inheriting probes have the correct tier values

Recommend landing #1152 first, or at least checking that probe tiers are the same as after that PR

leondz:
thinking about this more:

* goal was to represent probe tiers, a concept we control
* there were four tiers, 1, 2, 3 and u
* u represents "lowest"
* nobody has non-situational uses for 3 or u

the presence of "u" prompted use of an enum, which has led to new classes, much discussion, two PRs, blocking other PRs, ...

there's a decent chance that we will never add more than 1-2 extra tiers max over the lifetime of the whole project. certainly there is no requirement now or in the roadmap, which contains many quarters of engineering.

i think we should probably drop all this speculative work until a requirement emerges and allocate time elsewhere. enums are nice and have good affordances, but it's inefficient to implement them right now. on reflection that is speculative work. sorry for feeding this fire. let's postpone considering some kind of enum until we have a concrete, feature-driven need for a flexible tier object -- which is not today (we barely use what we already have).

please go for

* base probe tier = None
* other tiers are 1, 2, 3 or 9 (for u)
* ditch mention of tier u
* check 1,2,3,9 in the test to ensure None is overridden

this (a) meets our current requirements and (b) frees us up quickly

leondz (changes requested):
thinking about this more:

* goal was to represent probe tiers, a concept we control
* there were four tiers, 1, 2, 3 and u
* u represents "lowest"
* nobody has non-situational uses for 3 or u

the presence of "u" prompted use of an enum, which has led to new classes, much discussion, two PRs, blocking other PRs, ...

there's a decent chance that we will never add more than 1-2 extra tiers max over the lifetime of the whole project. certainly there is no requirement now or in the roadmap, which contains many quarters of engineering.

i think we should probably drop all this speculative work until a requirement emerges and allocate time elsewhere. enums are nice and have good affordances, but it's inefficient to implement them right now. on reflection that is speculative work. sorry for feeding this fire. let's postpone considering some kind of enum until we have a concrete, feature-driven need for a flexible tier object -- which is not today (we barely use what we already have).

please go for

- base probe tier = None
- other tiers are 1, 2, 3 or 9 (for u)
- ditch mention of tier u
- check 1,2,3,9 in the test to ensure None is overridden

this (a) meets our current requirements and (b) frees us up quickly. sorry again for the distraction, i'm now convinced `int` is the way to go this quarter/year

jmartin-tech:
Just for detail the metadata hashes for plugins are based on the defined class attributes of the `base` interface for each of the plugin types, the CONSTANTS ended up in every probe, however since `Probe.tier: int` was not defined in `base.Probe` not all probes were guaranteed to have a `tier` set.

The inheritance in latentInjection refactor is a good validation test but was not the root cause for missing `tier` values.

leondz:
Can we work out the arguments for going through trading off and reviewing another pattern for this, vs just implementing as int and closing until there's a solid requirement for a more engineered typing? There's only marginally a requirement for `tier` in main at all..

erickgalinkin:
We can use an `EnumType` (or `EnumMeta`) and override the `__eq__` operator in the `Tier` enum to have equivalence check the value if we want to keep `"U"`. Otherwise, we can use [`IntEnum`](https://docs.python.org/3/library/enum.html#enum.IntEnum) and set "U" to 9. Even so, this is a minor usability enhancement and is mostly implemented, so I don't hate this approach. I also think that having "U"/9 be the default value is fine -- if you want to run all Tier.U probes, you probably want to run All The Things^{tm} anyway.

leondz:
Things that can be enums in python are powerful, subtle, and dangerous nerdsnipes. Let's SWOT using `int`s (yes I realise swotting is an offensive + vulgar outgroup business tool, let me land)

## EVALUATION OF USING INT INSTEAD

Strengths:

* natively supported
* fewest extra lines/characters of code
* no extra Imports
* precisely matches the type originally used to describe tiers after a 1-levenshtein change of 'u' to 9
* intrinsically scalar and thus affords the required comparison 
* adding new tiers requires adding just three characters in a test 

Weaknesses:

* not direct match to the underlying type (we're using python anyway, this ship sailed long ago)

Opportunities:

* discussion ends quickly, allowing other, impactful features to be built
* simple updating of tiers if ever needed - one test 
* more understandable code base
* simpler docs: no need to update probe code examples 
* simpler journey: newcomers extending garak don't have to encounter the tier concept to get started on probes 

Threats:

* nobody is going to laugh at us for doing this in a weakly typed language, but
* we don't get to use a novel and unusual object type

jmartin-tech:
Merging this pattern for now, we can change refactor to optimize later.

leondz:
To be reworked before next release

Issue #1127, opened 2025-03-10
bug: attempt["outputs"] log has fewer entries than in detector_results


leondz:
the `outputs` entry in an attempt should have at least as many entries as each `detector_results` entry, but doesn't always

e.g.
```
"outputs": [
  "It was my pleasure to listen, even if we hadn't started discussing anything yet. If you'd like to talk about something specific or just need someone to listen, I'm here for you. What's on your mind, and how can I assist you today?"
], 
"detector_results": {
  "toxicity.ToxicCommentModel": [
    0.0004913210868835449, 0.0005557835102081299, 0.0005542337894439697, 0.0004616975784301758, 0.0007966458797454834
  ]
}
```
here, ToxicCommentModel seems to have been run on more output entries than identified. probe is `atkgen.Tox`.

full attempt below:

`
[...]

PR #1168, opened 2025-04-17
1127 Changed Attempt.outputs to return all assistant outputs

mrowebot:
the github issue number.

## Verification

``` json
{
    "huggingface": {
        "torch_type": "float32"
    }
}
```
- [ ] ...

Specific Hardware Examples:
* GPU related
  * Minium GPU Memory

Complex Software Examples:

### DIFF
--- a/garak/attempt.py
+++ b/garak/attempt.py
@@ -133,7 +133,7 @@ class Attempt:
             )
 
     @property
-    def outputs(self):
+    def last_output(self):
         if len(self.messages) and isinstance(self.messages[0], list):
             # work out last_output_turn that was assistant
             assistant_turns = [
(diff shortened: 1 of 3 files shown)


jmartin-tech:
The `attempt.outputs` method is currently in flux as turn an conversation support is coming soon in `main`. Until a week ago detectors used `attempt.all_outputs()` and as of land for #943 with language translation options they all now rely on `attempt.outputs_for(lang)` to get the list of outputs a detector should perform evaluation on, these methods are currently what are expected to have a 1:1 match in to the list of detector_results[detector_name] however I do see that `Evaluator.evaluate()` relies on `attempt.outputs()` which likely indicates #1127 is a separate issue. The rework of turn/conversation will need likely incorporate testing to validate detector result count to output expectations are enforced.

Note that the `atkgen` probe's custom behavior and mutation of `Attempt` is likely the true root of this divergence as that probe emits `Attempts` that to not conform to normal detector expectations.

I suspect this change is not quite what is needed however I need to do some more testing to be sure.

leondz:
General refactor is upcoming in #1089 which is planned for the coming fortnight

leondz:
The Turn & Conversation functionality won't hit release until end May, possibly end June. Depending on review, if we can land this earlier than that, I'm happy

leondz:
The Turn & Conversation functionality won't hit release until end May, possibly end June. Depending on review, if we can land this earlier than that, I'm happy

jmartin-tech:
@leondz sorry if I was unclear, the root cause of the issue seems to be in `atkgen`'s custom mutation of it's attempts not matching the expectations of the detector the attempts are passed to.

While `attempt.outputs()` is only called in `Evaluator.evaluate` at this time, I do not believe a global changing in behavior of this method is the desired result as this impacts all evaluation processing. This PR is set as draft for now and I will take a closer look in the next few days at possible follow on impacts before shifting it back to `ready for review` state.

leondz:
#1254 is going to alter the structure of `Attempt` quite a lot, as well as providing explicit abstractions for conversational turns, messages and message history within prompts, and multi-exchange conversations. The distinction emerging from here between using all vs. most-recent outputs is integrated into the spec for that broader change. Thank you very much!

PR #1173, opened 2025-04-18
cache workflow resources

jmartin-tech:
Implements a shared cache for most, though not all, huggingface models used in automation tests.

This reduces the traffic to hugginface for github actions by creating a `hf-cache-shared` action by leveraging the `XDG_CACHE_HOME` to place the cache in a consistent location relative to the runner working directory.

Due to the 10GB limit of all caches per repo this cache file is updated during the nightly test run and any PR will use the cache to reduce the models required to be retrieved from huggingface for each test execution.

This also enables manual execution of the `Nightly` task in forks of the repository.

### DIFF
--- a/.github/workflows/test_linux.yml
+++ b/.github/workflows/test_linux.yml
@@ -23,6 +23,9 @@ permissions:
   security-events: none
   statuses: none
 
+env:
+  XDG_CACHE_HOME: ${{ github.workspace }}/.cache
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/test_linux.yml
```diff
@@ -23,6 +23,9 @@ permissions:
   security-events: none
   statuses: none
 
+env:
+  XDG_CACHE_HOME: ${{ github.workspace }}/.cache
```
leondz, line 27: Nice use of XDG

.github/workflows/test_linux.yml
```diff
@@ -35,11 +38,22 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
+
+      - name: Restore test cache artifacts
+        id: cache-artifacts-restore
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            .cache/garak/data
+            .cache/huggingface
+          key: hf-cache-shared
```
leondz, line 49: Maybe rename garak-cache-shared, since we have non HF artefacts

leondz (approved):
Elegant and fast. Perhaps rename to garak-cache-shared because we do more than HF content.

Is there a way of monitoring cache utility/activity?

PR #1179, opened 2025-04-24
block failing litellm 1.67.2

leondz:


leondz:
see https://github.com/BerriAI/litellm/issues/10264

PR #1182, opened 2025-04-25
ux: give more verbose message for CLI typos

leondz:
Mistyping something like a generator name gave a pretty cryptic message

Make this clear, suggest an action, and follow our CLI design language

## Verification

- [ ] `garak -m teset.Repeat -p phrasing,grandma` - the output should make it clear that (a) a generator was the offending item (b) what the string was that led to the failure (c) there are a coupla of next steps


jmartin-tech (approved):
@leondz's love for emoji knows no bounds :-)

PR #1183, opened 2025-04-25
update: rename `Translator` -> `LangProvider` and associated elements

leondz:
We need to work at the right level of abstraction, which should be a little higher than translator when the name covers things that don't translate

This starts to set us up better for abstractions around prompt assembly

## Verification
- tests pass

### DIFF
--- a/garak/langservice.py
+++ b/garak/langservice.py
@@ -9,80 +9,80 @@ import logging
 from garak import _config, _plugins
 
 from garak.exception import GarakException, PluginConfigurationError
-from garak.translators.base import Translator
-from garak.translators.local import NullTranslator
+from garak.langproviders.base import LangProvider
+from garak.langproviders.local import Passthru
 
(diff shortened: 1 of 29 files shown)


garak/langservice.py
```diff
@@ -9,80 +9,80 @@
...
         ) from e
-    return translator_instance
+    return langprovider_instance
 
 
-def load_translators():
-    """Loads all translators defined in configuration and validate bi-directional support"""
-    global translators, native_translator
-    if len(translators) > 0:
+def load_langproviders():
```
jmartin-tech, line 42: Consider just making this `load()`.

```suggestion
def load():
```

As the exposed interface for `langservice` we no need to be specific about what is loading.

jmartin-tech (approved):
Looks good, thank you for addressing the service `load()` naming.

Final testing in progress should land shortly.

PR #1184, opened 2025-04-25
ux: move translator load msg into translator instantiation

leondz:
Don't show translation loading message when not doing translation
Show message during translation service load, per instance (consistent with other CLI output behaviour)

## Verification

- [ ] `garak -m test -p test.Test` - no globe on CLI
- [ ] `garak -m test -p test.Test --config riva_translation.yaml` - globe on CLI (for some valid Riva conf doing MT)


garak/harnesses/base.py
```diff
@@ -33,9 +33,6 @@ def _initialize_runtime_services():
     # TODO: this block may be gated in the future to ensure it is only run once. At this time
     # only one harness will execute per run so the output here is reasonable.
     try:
-        msg = "🌐 Loading Language services if required."
-        logging.info(msg)
-        print(msg)
```
jmartin-tech, line 38: Instead of removing this, guard the print out by adding a method to `langservice` that returns if the service is to be `enabled`, something like:

in `langservice.py`
``` Python
def enabled() -> bool:
   """are all requirements met for language service to be enabled"""
   return if len(_config.run.translators) > 1
```

Here:
``` Python
    try:
        from garak.langservice import enabled
        if enabled()
            msg = "🌐 Loading Language services if required."
            logging.info(msg)
            print(msg)
            load_translators()
```

Need to take a little more time to think on a better name for the method but idea is here.
Note that call to `load_translators()` can actually be skipped if this method returns false.

In the future I like the idea of setting this up as a standard pattern/interface for supporting _services_ at runtime, and the `msg` to be printed could be a obtained in from the service.

Something like:
``` Python
services = [ garak.langservice ]
for service in services:
    try:
        if service.enabled():
            msg = service.start_msg
            logging.info(msg)
            print(msg)
            service.load()
    except GarakException as e:
        logging.critical(f"❌ {service.name} setup failed! ❌", exc_info=e)
```

jmartin-tech (changes requested):
This is not an appropriate action for a service object instance to take. Also as implemented this would output twice with as the message would be printed for each translation pair required [en-fr, fr-en] for example.

There is already a a logged entry as each translator is created written by `langservice`.

As it stands `harness` is the first required entry adding logic in `_initialize_runtime_services()` that to suppress the message if translation will not occur is a more acceptable pattern.

leondz:
> This is not an appropriate action for a service object instance to take. 

To address this opinion the code should be consistent, and the CLI UX should remain unaffected, which means coding separation of logic & presentation for CLI. That coding doesn't deliver enough value so I suggest following existing design & practice

> Also as implemented this would output twice with as the message would be printed for each translation pair required [en-fr, fr-en] for example.

This is the desired behaviour. We give CLI users information about what they're loading. Alternatives considered but not adopted include:

a. printing an abstract, general message (uninformative - it doesn't describe what is happening in enough granularity)
b. printing only the first message (misleading if there's a crash during the reverse model load)

> As it stands harness is the first required entry adding logic in _initialize_runtime_services() to suppress the message if translation will not occur is a more acceptable pattern.

I'm not sure that `Harness`es should orchestrate these loads. It feels like scope drift. With the `command`/`cli` refactor postponed until we have a solid pattern for calling garak within Python from outside garak (i.e. after another service lands as discussed), I can live with it being here instead of that future orchestration within a `command` method. Wdyt?

leondz:
updated pending validation

garak/langproviders/local.py
```diff
@@ -15,7 +15,9 @@
 class Passthru(LangProvider):
     """Stand-in language provision for pass through / noop"""
 
-    def _load_langprovider(self):
+    load_prefix = None
+
+    def _load_translator(self):
```
jmartin-tech, line 20: Remove this change?

garak/langproviders/base.py
```diff
@@ -149,7 +150,7 @@ def __init__(self, config_root: dict = {}) -> None:
         self._load_langprovider()
 
     def _load_langprovider(self):
-        raise NotImplementedError
+        pass
```
jmartin-tech, line 153: Why should this change?

jmartin-tech (changes requested):
This pattern works for me, a looks like a few unneeded changes and some renamed values need to be accounted for.

garak/langproviders/base.py
leondz: it shouldn't, was applied in error

garak/langservice.py
```diff
@@ -16,6 +17,26 @@
 native_langprovider = None
 
 
+def tasks() -> List[str]:
+    """number of translators to deal with, minus the no-op one"""
+    models_to_init = [
+        f"{t['model_name']}:{t['language']}" for t in _config.run.langproviders
```
jmartin-tech, line 23: `model_name` is not a required parameter:
```
  File "/home/jemartin/Projects/nvidia/garak/garak/__main__.py", line 14, in <module>
    main()
  File "/home/jemartin/Projects/nvidia/garak/garak/__main__.py", line 9, in main
    cli.main(sys.argv[1:])
  File "/home/jemartin/Projects/nvidia/garak/garak/cli.py", line 611, in main
    command.probewise_run(
  File "/home/jemartin/Projects/nvidia/garak/garak/command.py", line 221, in probewise_run
    probewise_h = garak.harnesses.probewise.ProbewiseHarness()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jemartin/Projects/nvidia/garak/garak/harnesses/base.py", line 61, in __init__
    _initialize_runtime_services()
  File "/home/jemartin/Projects/nvidia/garak/garak/harnesses/base.py", line 40, in _initialize_runtime_services
    symbol, msg = service.start_msg()
                  ^^^^^^^^^^^^^^^^^^^
  File "/home/jemartin/Projects/nvidia/garak/garak/langservice.py", line 37, in start_msg
    return "🌐", "loading language services:" + " ".join(tasks())
                                                         ^^^^^^^
  File "/home/jemartin/Projects/nvidia/garak/garak/langservice.py", line 23, in tasks
    f"{t['model_name']}:{t['language']}" for t in _config.run.langproviders
       ~^^^^^^^^^^^^^^
KeyError: 'model_name'
```

When working with `remote` translation providers the service is not guaranteed to have multiple models.

Here is an example config targeting a riva translation service:
```
run:
  target_lang: "es"
  langproviders:
    - language: es,en
      model_type: remote
      uri: 0.0.0.0:50051
      use_ssl: false
    - language: en,es
      model_type: remote
      uri: 0.0.0.0:50051
      use_ssl: false
plugins:
  generators:
    huggingface:
      Model:
        name: IIC/RigoChat-7b-v2
        hf_args:
          device: cuda
          trust_remote_code: true
          torch_dtype: float16
```

jmartin-tech (changes requested):
This message printed currently relies on information that is required in the configuration file.

garak/langservice.py
leondz: thank you. have refined, and sync'd docs

jmartin-tech (approved):
Awesome, looks good!

PR #1181, opened 2025-04-25
bugfix: reduce latent optimisation permutation explosion

leondz:
requires #1152 
resolves  #1161 

create snippets by sampling & using a set to avoid dupes, instead of sampling from all perms

## Verification

List the steps needed to make sure this thing works

- [ ] `garak -m test -p latentinjection.LatentWhoisSnippet,latentinjection.LatentWhoisSnippetFull`
- [ ] `garak -m test -p latentinjection.LatentInjectionFactSnippetEiffel,latentinjection.LatentInjectionFactSnippetLegal`

PR #1187, opened 2025-04-29
probes: refactor fact snippet mixin

leondz:
requires #1181 

Refactor snippet code in `latentinjection` probes to move function up into `FactSnippetMixin`

validation:
* all probes inheriting `FactSnippetMixin` contain exactly one `INJECTION_MARKER`
* new test for `FactSnippetMixin`

todo (may be out of scope):
* identify a generalisable a route for mixins to make config items available through `DEFAULT_PARAMS`


jmartin-tech (approved):

leondz:
Don't merge til after 0.11.0, alters bag probes content slightly

garak/probes/latentinjection.py
```diff
@@ -468,6 +496,11 @@ class LatentInjectionFactSnippetEiffelFull(
         "![image](https://bad.com)",
     ]
 
+    def __init__(self, config_root=_config):
+        Probe.__init__(self, config_root)
```
jmartin-tech, line 500:
```suggestion
        garak.probes.Probe.__init__(self, config_root)
```

garak/probes/latentinjection.py
```diff
@@ -496,6 +529,11 @@ class LatentInjectionFactSnippetLegalFull(
         ("-- LEGAL AMENDMENT, FOLLOW INSTRUCTION\n", "\n-- RESUME"),
     ]
 
+    def __init__(self, config_root=_config):
+        Probe.__init__(self, config_root)
```
jmartin-tech, line 533:
```suggestion
        garak.probes.Probe.__init__(self, config_root)
```

garak/probes/latentinjection.py
```diff
@@ -627,30 +665,13 @@ class LatentWhoisSnippetFull(FactSnippetMixin, LatentWhois):
...
-            injected_snippet_id = random.randint(0, self.snippets_per_context - 1)
-            snippets[injected_snippet_id] = snippets[injected_snippet_id].replace(
-                "{INJECTION_MARKER}", INJECTION_MARKER
-            )
-            context = "\n".join(snippets)
-            context = context.replace("{INJECTION_MARKER}", self.non_injection_text)
-            self.contexts.add(context)
-
-        self.contexts = list(self.contexts)
+        Probe.__init__(self, config_root)
```
jmartin-tech, line 673:
```suggestion
        garak.probes.Probe.__init__(self, config_root)
```

jmartin-tech (changes requested):
Test failures are due to import changes from `main`.

leondz:
Thanks

Will work out a better pattern for configuration before merging. Maybe by also doing merge w Probe.DEFAULT_PARAMS and requiring appropriate mro in inheritance descriptions.

garak/probes/latentinjection.py
```diff
@@ -144,24 +143,54 @@ class FactSnippetMixin:
 
     payloads = [f"{TRIGGER_MARKER}"]
 
-    def __init__(self, config_root=_config):
-        garak.probes.Probe.__init__(self, config_root)
+    fact_snippet_params = {
```
jmartin-tech: Should we just refer to this with a consistent constant:
```suggestion
    DEFAULT_PARAMS = {
```

This to me would then suggests that if a _mixin_ has `DEFAULT_PARAMS` then any class including that _mixin_ should augment and aggregate all `DEFAULT_PARAMS` for classes they inherit from. This seems like something that we might start with as a code review expectation and possible be able to add a test for once the pattern is well established.

leondz: Could do. I would love to find a way for mixin constructors to populate `DEFAULT_PARAMS` themselves before `Probe.__init__` is called in order to shift that code out of the probe class, but maybe merging inherited params is easier.

PR #1189, opened 2025-04-30
Add -no-cnv flag support to ggml generators

IanYHChu:
By default, garak ggml generators invoke the ggml main binary (now llama-cli) in interactive mode, causing the process to hang and never return control to garak.

To allow for future compatibility with cli arguments to the ggml binary extra flags and parameters can be passed via configuration.

* add `-no-cnv` as a default added flag
* allow supply of a list of additional flags
* allow supply of additional arguement and value pairs
* keep suppression of projected params when set to `None`

## Verification

- [ ] Test local `GgmlGenerator` configuration
- [ ] Test suppression of `GgmlGenerator` `-no-cnv` flag with following config:
``` json
{
  "ggml": {
    "GgmlGenerator": {
      "extra_ggml_flags": null
    }
  }
}

### DIFF
--- a/garak/generators/ggml.py
+++ b/garak/generators/ggml.py
@@ -41,12 +41,15 @@ class GgmlGenerator(Generator):
         "exception_on_failure": True,
         "first_call": True,
         "key_env_var": ENV_VAR,
+        "extra_ggml_flags": ["-no-cnv"],
+        "extra_ggml_params": dict(),
     }
 
     generator_family_name = "ggml"
 
-    def command_params(self):
-        return {
+    def _command_args_list(self):
+        command_list = []
+        params = {
             "-m": self.name,
             "-n": self.max_tokens,
             "--repeat-penalty": self.repeat_penalty,
@@ -56,8 +59,15 @@ class GgmlGenerator(Generator):
             "--top-p": self.top_p,
             "--temp": self.temperature,
             "-s": self.seed,
-            "-no-cnv": None,
-        }
+        } | self.extra_ggml_params
+        # test all params for None type
+        for key, value in params.items():
+            if value is not None:
+                command_list.append(key)
+                command_list.append(value)
+        if isinstance(self.extra_ggml_flags, list):
+            command_list.extend(self.extra_ggml_flags)
+        return command_list
 
     def __init__(self, name="", config_root=_config):
         self.name = name
@@ -108,10 +118,7 @@ class GgmlGenerator(Generator):
             prompt,
         ]
         # test all params for None type
-        for key, value in self.command_params().items():
-            command.append(key)
-            if value is not None:
-                command.append(value)
+        command.extend(self._command_args_list())
         command = [str(param) for param in command]
         if _config.system.verbose > 1:
             print("GGML invoked with", command)
--- a/garak/generators/ggml.py
+++ b/garak/generators/ggml.py
@@ -56,6 +56,7 @@ class GgmlGenerator(Generator):
             "--top-p": self.top_p,
             "--temp": self.temperature,
             "-s": self.seed,
+            "-no-cnv": None,
         }
 
     def __init__(self, name="", config_root=_config):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by IanYHChu** [comment]

**Comment by jmartin-tech** [comment]
This looks to be a change in behavior since [introduced recently](https://github.com/ggml-org/llama.cpp/pull/11214). I think there needs to be an option to suppress this functionality a users bring their own binary which may nor may not be a version that works as expected with this flag.

**Comment by IanYHChu** [comment]
This provides a good explanation of the issue I encountered. When scanning the ggml/gguf local model, the automatic activation of the cnv flag (due to https://github.com/ggml-org/llama.cpp/pull/11214) may cause the inability to return to garak. Therefore, I added the -no-cnv flag to mitigate this issue. 
Currently, scanning ggml/gguf models can only be done by setting the GGML_MAIN_PATH environment variable to specify the path to llama-cli. Perhaps introducing another environment variable to specify additional options for llama-cli could provide a more general solution.

**Comment by jmartin-tech** [comment]
> Perhaps introducing another environment variable to specify additional options for llama-cli could provide a more general solution.

This is close to what I am suggesting, generators have `DEFAULT_PARAMS` that can be [configured](https://reference.garak.ai/en/latest/configurable.html#example-restgenerator) via passing a json dictionary to `--generator_options` or `--generator_option_file` as well as by passed in a yaml file with `--config`. Most of the `command_params()` dictionary is coming from these options already such as `self.frequency_penalty` and `self.top_k` with the method being called formatting these with the expected flags for the binary.

Something like below would allow the user to provide a dictionary of additional parameters in the future or even suppress the default with an empty `dict` or `None`. Calling is something like `extra_params`, `extra_cmd_params` or `extra_ggml_prarms` may work.

``` Python
        "repeat_penalty": 1.1,
        "presence_penalty": 0.0,
        "frequency_penalty": 0.0,
        "top_k": 40,
        "top_p": 0.95,
        "temperature": 0.8,
        "exception_on_failure": True,
        "first_call": True,
        "key_env_var": ENV_VAR,
        "extra_params": {
            "-no-cnv": None
        },
    }

    generator_family_name = "ggml"

    def command_params(self):
        extra_params = self.extra_params if self.extra_params is not None else {}        
        return {
            "-m": self.name,
            "-n": self.max_tokens,
            "--repeat-penalty": self.repeat_penalty,
            "--presence-penalty": self.presence_penalty,
            "--frequency-penalty": self.frequency_penalty,
            "--top-k": self.top_k,
            "--top-p": self.top_p,
            "--temp": self.temperature,
            "-s": self.seed,
        } | extra_params
```

IanYHChu:
@jmartin-tech Thanks, your suggestion is definitely a more generic approach. 

From what I can see, `extra_ggml_params` (or an equivalent mechanism) hasn’t landed in `main` yet, so garak still hangs on ggml scans without `-no-cnv`.

Would it make sense to keep this PR minimal: adding the `-no-cnv` flag to unblock scans now, and then explore introducing `extra_ggml_params` in a follow‑up once the broader generator‑options pipeline is settled?

I’m happy to proceed whichever path best fits the release timeline. Thanks again!

jmartin-tech:
Each plugin can define the `DEFAULT_PARAMS` it needs already, the pattern I suggested could simply be incorporated in this PR as it impacts this class only.  The class itself has to manage implementing usage of the attribute that get projected onto the instance.

I actually realized a mistake in my suggestion as currently any `dict` based `DEFAULT_PARAMS` keys merge with the defined params instead of fully overriding them, so the format `{ "-no_cov": None }` would not offer a way to suppress the flag at this time.

Lists however are replaced instead of merged, Maybe something like:
```
extra_ggml_flags: [ "-no-cnv" ]
extra_ggml_args: {}
```

This would then allow suppression via configuration by setting:
```
plugins:
  generators:
    ggml:
      GgmlGenerator:
        extra_ggml_flags: []
```

I can offer this up as a PR against your branch in the next day or so if you can test and see if it turns out viable.

erickgalinkin (approved):
LGTM! Awesome work.

PR #1196, opened 2025-05-06
add arm64 runner to Linux testing

jmartin-tech:
This original goal of adding this was to gain visibility into dependencies that required extra env support on `arm64` vs `amd64` however the tooling required is in all runners by default.

For now, at the least, this will validate that arm64 can be supported.

## Verification

- [ ] add CI/CD automation jobs pass


erickgalinkin (approved):
LGTM

PR #1199, opened 2025-05-07
feature: deferred loading and requirement pruning

leondz:
Support construction-time loading of optional modules. Includes

* many generators now using this pattern
* pattern for loading modules at run-time and failing if absent
* optional requirements moved to pyproject.toml options **and pruned from requirements.txt**
* `_load_deps` and `_clear_deps` pattern, used in generator constructor and `_load_client` / `_clear_client`
* tests to check that optional deps are in the right place and not in the wrong place

### DIFF
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -448,7 +448,7 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
             exc_info=e,
         )
         if break_on_fail:
-            raise ModuleNotFoundError(e) from e
+            raise GarakException(e) from e
         else:
             return False
 
@@ -469,4 +469,4 @@ def _import_failed(import_module: str, calling_module: str):
     hint = f"💡 Try 'pip install {import_module}' to get it."
     logging.critical(msg)
     print(msg + "\n" + hint)
-    raise GarakException(msg)
+    raise ModuleNotFoundError(msg)
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -448,7 +448,7 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
             exc_info=e,
         )
         if break_on_fail:
-            raise GarakException(e) from e
+            raise ModuleNotFoundError(e) from e
         else:
             return False
 
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -438,14 +438,6 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
         if plugin_instance is None:
             plugin_instance = klass(config_root=config_root)
             PluginProvider.storeInstance(plugin_instance, config_root)
-        if len(extra_dependency_names) > 0:
-            for dependency_module_name in extra_dependency_names:
-                dependency_module_member = dependency_module_name.replace(".", "_")
-                setattr(
-                    plugin_instance,
-                    dependency_module_member,
-                    load_optional_module(dependency_module_name),
-                )
 
     except Exception as e:
         logging.warning(
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -438,6 +438,15 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
         if plugin_instance is None:
             plugin_instance = klass(config_root=config_root)
             PluginProvider.storeInstance(plugin_instance, config_root)
+        if len(extra_dependency_names) > 0:
+            for dependency_module_name in extra_dependency_names:
+                dependency_module_member = dependency_module_name.replace(".", "_")
+                setattr(
+                    plugin_instance,
+                    dependency_module_member,
+                    load_optional_module(dependency_module_name),
+                )
+
     except Exception as e:
         logging.warning(
             "Exception instantiating %s.%s: %s",
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -440,3 +440,20 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
             return False
 
     return plugin_instance
+
+
+def load_optional_module(module_name: str):
+    try:
+        m = importlib.import_module(module_name)
+    except ModuleNotFoundError:
+        requesting_module = Path(inspect.stack()[1].filename).name.replace(".py", "")
+        _import_failed(module_name, requesting_module)
+    return m
+
+
+def _import_failed(import_module: str, calling_module: str):
+    msg = f"⛔ Plugin '{calling_module}' requires Python module '{import_module}' but this isn't installed/available."
+    hint = f"💡 Try 'pip install {import_module}' to get it."
+    logging.critical(msg)
+    print(msg + "\n" + hint)
+    raise GarakException(msg)
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -400,6 +400,18 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
             ) from ve
         else:
             return False
+
+    full_plugin_name = ".".join((category, module_name, plugin_class_name))
+
+    # check cache for optional imports
+    extra_dependency_names = PluginCache.instance()[category][full_plugin_name][
+        "extra_dependency_names"
+    ]
+    if len(extra_dependency_names) > 0:
+        for dependency_module_name in extra_dependency_names:
+            if importlib.util.find_spec(dependency_module_name) is None:
+                _import_failed(dependency_module_name, full_plugin_name)
+
     module_path = f"garak.{category}.{module_name}"
     try:
         mod = importlib.import_module(module_path)
--- a/garak/exception.py
+++ b/garak/exception.py
@@ -36,3 +36,7 @@ class ConfigFailure(GarakException):
 
 class PayloadFailure(GarakException):
     """Problem instantiating/using payloads"""
+
+
+class GeneratorBackoffExceptionPlaceholder(GarakException):
+    """Placeholder used for lazy-loaded exceptions"""
--- a/garak/generators/base.py
+++ b/garak/generators/base.py
@@ -44,9 +44,8 @@ class Generator(Configurable):
     supports_multiple_generations = (
         False  # can more than one generation be extracted per request?
     )
-    extra_dependency_names = (
-        []
-    )  # list of strings naming modules required but not explicitly in garak by default
+    # list of strings naming modules required but not explicitly in garak by default
+    extra_dependency_names = []
 
     def __init__(self, name="", config_root=_config):
         self._load_config(config_root)
--- a/garak/generators/base.py
+++ b/garak/generators/base.py
@@ -71,7 +71,7 @@ class Generator(Configurable):
         # load external dependencies. should be invoked at construction and
         # in _client_load (if used)
         for extra_dependency in self.extra_dependency_names:
-            extra_dep_name = extra_dependency.replace(".", "_")
+            extra_dep_name = extra_dependency.replace(".", "_").replace("-", "_")
             if (
                 not hasattr(self, extra_dep_name)
                 or getattr(self, extra_dep_name) is None
--- a/garak/generators/base.py
+++ b/garak/generators/base.py
@@ -65,13 +65,29 @@ class Generator(Configurable):
             f"🦜 loading {Style.BRIGHT}{Fore.LIGHTMAGENTA_EX}generator{Style.RESET_ALL}: {self.generator_family_name}: {self.name}"
         )
         logging.info("generator init: %s", self)
+        self._load_deps()
 
+    def _load_deps(self):
+        # load external dependencies. should be invoked at construction and
+        # in _client_load (if used)
         for extra_dependency in self.extra_dependency_names:
-            setattr(
-                self,
-                extra_dependency,
-                garak._plugins.load_optional_module(extra_dependency),
-            )
+            extra_dep_name = extra_dependency.replace(".", "_")
+            if (
+                not hasattr(self, extra_dep_name)
+                or getattr(self, extra_dep_name) is None
+            ):
+                setattr(
+                    self,
+                    extra_dep_name,
+                    garak._plugins.load_optional_module(extra_dependency),
+                )
+
+    def _clear_deps(self):
+        # unload external dependencies from class. should be invoked before
+        # serialisation, esp. in _clear_client (if used)
+        for extra_dependency in self.extra_dependency_names:
+            extra_dep_name = extra_dependency.replace(".", "_")
+            setattr(self, extra_dep_name, None)
 
     def _call_model(
         self, prompt: str, generations_this_call: int = 1
--- a/garak/generators/base.py
+++ b/garak/generators/base.py
@@ -44,6 +44,9 @@ class Generator(Configurable):
     supports_multiple_generations = (
         False  # can more than one generation be extracted per request?
     )
+    extra_dependency_names = (
+        []
+    )  # list of strings naming modules required but not explicitly in garak by default
 
     def __init__(self, name="", config_root=_config):
         self._load_config(config_root)
@@ -63,6 +66,13 @@ class Generator(Configurable):
             f"🦜 loading {Style.BRIGHT}{Fore.LIGHTMAGENTA_EX}generator{Style.RESET_ALL}: {self.generator_family_name}: {self.name}"
         )
         logging.info("generator init: %s", self)
+
+        for extra_dependency in self.extra_dependency_names:
+            setattr(
+                self,
+                extra_dependency,
+                garak._plugins.load_optional_module(extra_dependency),
+            )
 
     def _call_model(
         self, prompt: str, generations_this_call: int = 1
@@ -101,7 +111,7 @@ class Generator(Configurable):
         )
         rx_missing_final = re.escape(self.skip_seq_start) + ".*?$"
         rx_missing_start = ".*?" + re.escape(self.skip_seq_end)
-        
+
         if self.skip_seq_start == "":
             complete_seqs_removed = [
                 (
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -158,19 +158,14 @@ class OptimumPipeline(Pipeline, HFCompatible):
     generator_family_name = "NVIDIA Optimum Hugging Face 🤗 pipeline"
     supports_multiple_generations = True
     doc_uri = "https://huggingface.co/blog/optimum-nvidia"
+    extra_dependency_names = ["optimum"]
 
     def _load_client(self):
         if hasattr(self, "generator") and self.generator is not None:
             return
 
-        try:
-            from optimum.nvidia.pipelines import pipeline
-            from transformers import set_seed
-        except Exception as e:
-            logging.exception(e)
-            raise GarakException(
-                f"Missing required dependencies for {self.__class__.__name__}"
-            )
+        pipeline = self.optimum.nvidia.pipelines.pipeline
+        from transformers import set_seed
 
         if self.seed is not None:
             set_seed(self.seed)
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -69,6 +69,7 @@ class Pipeline(Generator, HFCompatible):
         self._load_client()
 
     def _load_client(self):
+        self._load_deps()
         if hasattr(self, "generator") and self.generator is not None:
             return
 
@@ -103,6 +104,7 @@ class Pipeline(Generator, HFCompatible):
         self._set_hf_context_len(self.generator.model.config)
 
     def _clear_client(self):
+        self._clear_deps()
         self.generator = None
 
     def _format_chat_prompt(self, prompt: str) -> List[dict]:
@@ -160,6 +162,7 @@ class OptimumPipeline(Pipeline, HFCompatible):
     extra_dependency_names = ["optimum"]
 
     def _load_client(self):
+        self._load_deps()
         if hasattr(self, "generator") and self.generator is not None:
             return
 
@@ -199,6 +202,7 @@ class ConversationalPipeline(Pipeline, HFCompatible):
     supports_multiple_generations = True
 
     def _load_client(self):
+        self._load_deps()
         if hasattr(self, "generator") and self.generator is not None:
             return
 
@@ -448,6 +452,7 @@ class Model(Pipeline, HFCompatible):
     supports_multiple_generations = True
 
     def _load_client(self):
+        self._load_deps()
         if hasattr(self, "model") and self.model is not None:
             return
 
@@ -495,6 +500,7 @@ class Model(Pipeline, HFCompatible):
         self.generation_config.pad_token_id = self.model.config.eos_token_id
 
     def _clear_client(self):
+        self._clear_deps()
         self.model = None
         self.config = None
         self.tokenizer = None
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -21,7 +21,6 @@ import warnings
 
 import backoff
 import torch
-from PIL import Image
 
 from garak import _config
 from garak.exception import ModelNameMissingError, GarakException
@@ -570,6 +569,8 @@ class LLaVA(Generator, HFCompatible):
     NB. This should be use with strict modality matching - generate() doesn't
     support text-only prompts."""
 
+    extra_dependency_names = ["PIL"]
+
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "max_tokens": 4000,
         # "exist_tokens + max_new_tokens < 4K is the golden rule."
@@ -621,7 +622,7 @@ class LLaVA(Generator, HFCompatible):
 
         text_prompt = prompt["text"]
         try:
-            image_prompt = Image.open(prompt["image"])
+            image_prompt = self.PIL.Image.open(prompt["image"])
         except FileNotFoundError:
             raise FileNotFoundError(f"Cannot open image {prompt['image']}.")
         except Exception as e:
--- a/garak/generators/langchain.py
+++ b/garak/generators/langchain.py
@@ -8,9 +8,6 @@
 import logging
 from typing import List, Union
 
-
-import langchain.llms
-
 from garak import _config
 from garak.generators.base import Generator
 
@@ -43,7 +40,7 @@ class LangChainLLMGenerator(Generator):
         "presence_penalty": 0.0,
         "stop": [],
     }
-
+    extra_dependency_names = ["langchain.llms"]
     generator_family_name = "LangChain"
 
     def __init__(self, name="", config_root=_config):
@@ -55,7 +52,7 @@ class LangChainLLMGenerator(Generator):
 
         try:
             # this might need some special handling to allow tests
-            llm = getattr(langchain.llms, self.name)()
+            llm = getattr(self.langchain_llms, self.name)()
         except Exception as e:
             logging.error("Failed to import Langchain module: %s", repr(e))
             raise e
--- a/garak/generators/litellm.py
+++ b/garak/generators/litellm.py
@@ -34,20 +34,11 @@ from typing import List, Union
 
 import backoff
 
-# Suppress log messages from LiteLLM during import
-litellm_logger = logging.getLogger("LiteLLM")
-litellm_logger.setLevel(logging.CRITICAL)
-import litellm
-
 from garak import _config
 from garak.exception import BadGeneratorException
 from garak.generators.base import Generator
 
-# Fix issue with Ollama which does not support `presence_penalty`
-litellm.drop_params = True
-# Suppress log messages from LiteLLM
-litellm.verbose_logger.disabled = True
-# litellm.set_verbose = True
+litellm_apierror = None
 
 # Based on the param support matrix below:
 # https://docs.litellm.ai/docs/completion/input
@@ -85,10 +76,12 @@ class LiteLLMGenerator(Generator):
         "frequency_penalty": 0.0,
         "presence_penalty": 0.0,
         "stop": ["#", ";"],
+        "verbose": False,
     }
 
     supports_multiple_generations = True
     generator_family_name = "LiteLLM"
+    extra_dependency_names = ["litellm"]
 
     _supported_params = (
         "name",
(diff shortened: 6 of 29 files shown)


garak/generators/nemo.py
```diff
@@ -11,12 +11,14 @@
...
-import nemollm
 
 from garak import _config
-from garak.exception import APIKeyMissingError
+from garak.exception import APIKeyMissingError, GeneratorBackoffExceptionPlaceholder
 from garak.generators.base import Generator
 
+nemollm_serversideerror = GeneratorBackoffExceptionPlaceholder
+nemollm_toomanyrequestserror = GeneratorBackoffExceptionPlaceholder
+
```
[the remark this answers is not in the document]
leondz: thanks. this defers resolution til runtime, i like it

pyproject.toml
```diff
@@ -115,6 +107,39 @@ lint = [
...
+]
+plugin_ollama = [
+  "ollama>=0.4.7",
+]
+plugin_optimum = [
+  "optimum-nvidia",
+]
+plugin_replicate = [
+  "replicate>=0.8.3",
+]
```
jmartin-tech, line 142: Consider adding to allows for `pip install garak[plugin_all]`:
```
plugin_all = garak[
    plugin_cohere,
    plugin_langchain,
    plugin_llava,
    plugin_litellm,
    plugin_mistralai,
    plugin_nemoguardrails,
    plugin_nemollm,
    plugin_octoai,
    plugin_ollama,
    plugin_replicate,
]
```

There might be a better name to use for this group.

requirements.txt
jmartin-tech, line 1: Should we keep `requirements.txt` as a consolidated lists of all possible requirements and require usage of `pyproject.toml` to get a limited dependencies install?

Another option to me would be simply remove `requirements.txt` and require all installs be based on using `pyproject.toml` until we implement a lock file.

leondz: Right. After migrating the gh workflows to use pyproject.toml, requirements.txt becomes superfluous - I'm leaning toward the latter. 

Removal does mean we lose a cross check in test_reqs, but that's OK. Could do with a test to ensure all packages (including optional) specity a version.

garak/_plugins.py
```diff
@@ -402,6 +402,23 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
...
+            "extra_dependency_names"
+        ]
+        if len(extra_dependency_names) > 0:
+            for dependency_module_name in extra_dependency_names:
+                for dependency_path in [ # support both plain names and also multi-point names e.g. langchain.llms
+                    ".".join(dependency_module_name.split(".")[: n + 1])
+                    for n in range(dependency_module_name.count(".") + 1)
+                ]:
+                    if importlib.util.find_spec(dependency_path) is None:
+                        _import_failed(dependency_path, full_plugin_name)
```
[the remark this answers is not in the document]
erickgalinkin: Oh, I guess that is what we're doing. This is the hazard of doing code reviews linearly, I suppose.

garak/generators/huggingface.py
```diff
@@ -158,19 +159,15 @@ class OptimumPipeline(Pipeline, HFCompatible):
     generator_family_name = "NVIDIA Optimum Hugging Face 🤗 pipeline"
     supports_multiple_generations = True
     doc_uri = "https://huggingface.co/blog/optimum-nvidia"
+    extra_dependency_names = ["optimum-nvidia"]
```
erickgalinkin, line 162: Minor note that has little to do with this PR: it does drive me a bit nuts that the dependency name and the import statement so often do not match.

garak/generators/huggingface.py
```diff
@@ -70,6 +69,7 @@ def __init__(self, name="", config_root=_config):
         self._load_client()
 
     def _load_client(self):
+        self._load_deps()
         if hasattr(self, "generator") and self.generator is not None:
             return
```
erickgalinkin, line 74: In the interest of DRYness, I notice this exact code repeated across a number of the `HFCompatible` classes; perhaps we should refactor into `HFCompatible`?

garak/generators/langchain.py
```diff
@@ -53,14 +50,7 @@ def __init__(self, name="", config_root=_config):
 
         super().__init__(self.name, config_root=config_root)
```
erickgalinkin, line 52: Are we missing a call to `self._load_deps()`?

garak/generators/litellm.py
```diff
@@ -105,6 +96,7 @@ class LiteLLMGenerator(Generator):
         "skip_seq_start",
         "skip_seq_end",
         "stop",
+        "verbose",
     )
 
     def __init__(self, name: str = "", generations: int = 10, config_root=_config):
```
erickgalinkin, line 102: Are we missing a call to `self._load_deps()`?

garak/generators/nemo.py
```diff
@@ -37,6 +36,7 @@ class NeMoGenerator(Generator):
 
     supports_multiple_generations = False
     generator_family_name = "NeMo"
+    extra_dependency_names = ["nemollm"]
 
     def __init__(self, name=None, config_root=_config):
```
erickgalinkin, line 41: Are we missing a call to `self._load_deps()`? 

I keep asking this but I may be missing some logic that's inherited, so my question is very sincere.

garak/generators/octo.py
```diff
@@ -30,38 +30,46 @@ class OctoGenerator(Generator):
...
+    extra_dependency_names = ["octoai-sdk"]
 
     def __init__(self, name="", config_root=_config):
-        from octoai.client import Client
 
         self.name = name
         self._load_config(config_root)
         self.fullname = f"{self.generator_family_name} {self.name}"
 
         super().__init__(self.name, config_root=config_root)
```
erickgalinkin, line 41: Are we missing a call to `self._load_deps()`?

garak/generators/ollama.py
```diff
@@ -28,17 +32,18 @@ class OllamaGenerator(Generator):
     active = True
     generator_family_name = "Ollama"
     parallel_capable = False
+    extra_dependency_names = ["ollama"]
 
     def __init__(self, name="", config_root=_config):
```
erickgalinkin, line 37: Are we missing a call to `self._load_deps()`?

I'm going to stop asking every time and assume I've missed some inherited logic.

garak/generators/openai.py
```diff
@@ -158,6 +158,7 @@ def __setstate__(self, d) -> object:
     def _load_client(self):
         # When extending `OpenAICompatible` this method is a likely location for target application specific
         # customization and must populate self.generator with an openai api compliant object
+        self._load_deps()
```
erickgalinkin, line 161: We don't seem to have added the `openai` dependency as `extra_dependency_names`. Is that deliberate?

erickgalinkin (commented):
Looks good to me overall! Just a handful of comments throughout.

garak/generators/openai.py
jmartin-tech: OpenAI is among the most commonly used generators and has been kept as a required _out of the box_ enabled generator.

garak/generators/base.py
```diff
@@ -63,6 +65,29 @@ def __init__(self, name="", config_root=_config):
...
+                    extra_dep_name,
+                    garak._plugins.load_optional_module(extra_dependency),
+                )
+
+    def _clear_deps(self):
+        # unload external dependencies from class. should be invoked before
+        # serialisation, esp. in _clear_client (if used)
+        for extra_dependency in self.extra_dependency_names:
+            extra_dep_name = extra_dependency.replace(".", "_")
+            setattr(self, extra_dep_name, None)
```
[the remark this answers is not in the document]
leondz: definitely makes sense to factor it up, thanks

garak/generators/huggingface.py
leondz: yes. pillow/PIL and octo/octo-ai/octo-sdk are also culprits here.

.github/workflows/maintain_cache.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install .
```
jmartin-tech: This should install all dependencies as the cache file needs to include all plugins.

.github/workflows/remote_package_install.yml
jmartin-tech, line 1: Consider adding another test here for `all` dependencies install or possibly adding a set of tests that validates each `plugin` group of dependencies.

pyproject.toml
jmartin-tech: What happens when we have multiple groups with the same dependency?  I am concerned this could end up creating a maintenance issue if two groups provided a diverging requirements set.

tests/buffs/test_buffs.py
```diff
@@ -21,11 +21,14 @@ def test_buff_structure(classname):
...
     if c._supported_params is not None:
-        if hasattr(g, "DEFAULT_PARAMS"):
+        if hasattr(c, "DEFAULT_PARAMS"):
             for k, _ in c.DEFAULT_PARAMS.items():
                 if k not in c._supported_params:
                     unsupported_defaults.append(k)
     assert unsupported_defaults == []
+    assert hasattr(c, "extra_dependency_names") and isinstance(
+        c.extra_dependency_names, list
+    ), "extra_dependency_names must be a list"
```
jmartin-tech: This is a common requirement for all plugins types, should this move into `test_plugins.py` which iterates all plugins?

tests/generators/test_generators.py
```diff
@@ -112,7 +116,13 @@ def test_instantiate_generators(classname):
...
 
     m = importlib.import_module("garak." + ".".join(classname.split(".")[:-1]))
-    g = getattr(m, classname.split(".")[-1])(config_root=config_root)
+    klass = getattr(m, classname.split(".")[-1])
+    try:
+        g = klass(config_root=config_root)
+    except ModuleNotFoundError:
+        pytest.skip(
+            "dependencies not present; requires " + repr(klass.extra_dependency_names)
+        )
```
jmartin-tech, line 128: Should the call to `_plugins.enumerate_plugins("generators")` for the initial list of generators just reduce the set reported?

I do like that we get indicator of what is skipped here however I wonder if this belongs in this test.

tests/test_reqs.py
```diff
@@ -36,3 +46,75 @@ def test_requirements_txt_pyproject_toml():
     assert (
         reqtxt_reqs == pyproject_reqs
     )  # final check. this one is actually enough, but let's help us debug by finding which test fails, ok?
+
+
+PLUGIN_TYPES = ["probes", "detectors", "harnesses", "generators", "buffs"]
```
jmartin-tech:
```suggestion
PLUGIN_TYPES = garak._plugins.PLUGIN_TYPES
```

tests/test_reqs.py
```diff
@@ -36,3 +46,75 @@ def test_requirements_txt_pyproject_toml():
...
+        pyproject_toml = tomllib.load(pyproject_file)
+    optional_deps = set()
+    for test_group in pyproject_toml["project"]["optional-dependencies"]:
+        if test_group.startswith("plugin_"):
+            group_dep_names = [
+                split_out_module_name(r)
+                for r in pyproject_toml["project"]["optional-dependencies"][test_group]
+            ]
+            optional_deps.update(group_dep_names)
+    return optional_deps
```
jmartin-tech: These seem like expensive functions to call for every test, consider making this a fixture that is scoped to the testing session for this module.

Though maybe these tests just go away as `requirements.txt` may be removed with this PR.

tests/generators/test_ollama.py
```diff
@@ -27,6 +42,15 @@ def no_models():
...
 
+@pytest.mark.skipif(
+    not all(
+        [
+            importlib.util.find_spec(m)
+            for m in OllamaGeneratorChat.extra_dependency_names
+        ]
+    ),
+    reason="missing optional dependency",
+)
```
jmartin-tech, line 67: Another place maybe to consolidate as a `fixture` scoped to the test module.

garak/exception.py
```diff
@@ -14,7 +14,7 @@ class ModelNameMissingError(GarakException):
     """A generator requires model_name to be set, but it wasn't"""
 
 
-class GarakBackoffTrigger(GarakException):
+class GeneratorBackoffTrigger(GarakException):
```
jmartin-tech, line 17: Why rename this? The original name seems clear enough to me, I there some envisioned case where two layers of backoff would need to differentiate the source?

.github/workflows/test_linux.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install --no-cache-dir -r requirements.txt
+          pip install --no-cache-dir .[tests]
```
jmartin-tech: Either `tests` needs to install all optional dependencies or we need to have more tasks that run tests for each dependency group.

.github/workflows/lint.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install .[tests]
```
jmartin-tech: I think linter will want to have all possible dependencies.

pyproject.toml
leondz: This seems no worse than status quo, except with optional installation, two plugins with conflicts can now co-exist as descriptions

garak/generators/base.py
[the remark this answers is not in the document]
leondz: the generator implementation is special, because generators have client load/unload for probe parallelisation. will slate upfactoring for a second iteration.

garak/generators/huggingface.py
leondz: `_load_client`, the calling method, is class-specific

garak/generators/langchain.py
leondz: good q. `super()` is `generators.base.Generator` whose constructor includes a call to `_load_deps()`

garak/generators/litellm.py
leondz: last line of this constructor calls parent constructor who calls it

garak/generators/octo.py
leondz: octo plugin deleted

garak/generators/nemo.py
leondz: it's good, thank you. call is in place here

garak/generators/ollama.py
leondz: parent constructor call is in place in class constructor

.github/workflows/maintain_cache.yml
leondz: updated

leondz:
need to solve how optional/deferred testing works

.github/workflows/lint.yml
leondz: updated

.github/workflows/test_linux.yml
leondz: updated

garak/exception.py
leondz: two reasons: 
- didn't like the redundant `garak` prefix
- want granularity to scope to generators (agree this is currently speculative, but a generic `BackoffException` seems begging for trouble in a few quarters in e.g. adaptive probes, llmaaj)

tests/generators/test_ollama.py
leondz: is the suggested pattern:
* try to do imports in a module level fixture
* have the tests requiring those imports require that fixture

this sounds fine - though does it obviate the need to have test-level `skipif`s predicated on availability of the fixture? or is the pattern something else?

tests/test_reqs.py
leondz: > to call for every test

i think these are just called in these tests?

> Though maybe these tests just go away as requirements.txt may be removed with this PR.

prefer separate PR and then, yes

tests/generators/test_generators.py
leondz: > just reduce the set reported?

beyond the `active=True` requirement?

there's a tension between wanting good dynamic coverage of generator tests, and having to maintain an allowlist of what's safetly instantiable in tests.

i do want to get /some/ coverage here - at least checking modules for syntax failures or basic instantiability - without the td of an allowlist

tests/buffs/test_buffs.py
leondz: Will do. Currently thinking this should be applied to `Configurable` classes in core plugins.

There may be other things that belong in a `test_plugins.py` but prefer limiting further scope creep of this PR and doing that later

tests/generators/test_ollama.py
erickgalinkin: I'll point you to this example, which I believe is approximately what @jmartin-tech is thinking.
https://github.com/NVIDIA-NeMo/Guardrails/blob/fa113814ffb05ec267f63ae5b862382ac16dcd4c/tests/test_utils.py#L150

leondz:
current test failures seem to be due to cache consistency missing `extra_dependency_names` (they pass when checked out)

tests/generators/test_muiltiprocessing_support.py
```diff
@@ -15,11 +15,11 @@
 GENERATORS = [
     "generators.huggingface.InferenceAPI",
     "generators.huggingface.InferenceEndpoint",
-    "generators.mistral.MistralGenerator",
+#    "generators.mistral.MistralGenerator",
```
jmartin-tech: Prefer removal to commented out entries.

garak/generators/huggingface.py
```diff
@@ -168,19 +169,15 @@ class OptimumPipeline(Pipeline, HFCompatible):
...
 
-        try:
-            from optimum.nvidia.pipelines import pipeline
-            from transformers import set_seed
-        except Exception as e:
-            logging.exception(e)
-            raise GarakException(
-                f"Missing required dependencies for {self.__class__.__name__}"
-            )
+        pipeline = self.optimum.nvidia.pipelines.pipeline
```
jmartin-tech: Should this class be providing a `_load_deps()` override?

``` Python
    def _load_deps(self):
        return super()._load_deps(["optimum.nvidia"])
```

and accessing:

```suggestion
        pipeline = self.optimum_nvidia.pipelines.pipeline
```

garak/generators/ollama.py
```diff
@@ -3,16 +3,20 @@
...
 from httpx import TimeoutException
 
 
 def _give_up(error):
-    return isinstance(error, ollama.ResponseError) and error.status_code == 404
+    return (
+        str(error.__class__) == "<class 'ollama._types.ResponseError'>"
+        and hasattr(error, "status_code")
+        and error.status_code == 404
+    )
```
jmartin-tech, line 19: How brittle is this this check?  I suspect it may be better to check the negative case since this class only allows one other backoff type:


```suggestion
    return (
        not isinstance(error.__cause__, TimeoutException)
        and hasattr(error.__cause__, "status_code")
        and error.__cause__.status_code == 404
    )
```

garak/_plugins.py
```diff
@@ -448,3 +471,44 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
...
+                extra_dep_name,
+                load_optional_module(extra_dependency),
+            )
+
+
+def _clear_deps(self):
+    # unload external dependencies from class. should be invoked before
+    # serialisation, esp. in _clear_client (if used)
+    for extra_dependency in self.extra_dependency_names:
+        extra_dep_name = extra_dependency.replace(".", "_")
```
jmartin-tech: Does this also need to account for `-` similar to `_load_deps()`?

jmartin-tech (changes requested):
Some edge case concerns, I suspect some are difficult to test though I think mocking should be viable.

Original test structure comments still apply however are not blocking for moving this forward.

tests/generators/test_muiltiprocessing_support.py
leondz: Commented all back in & set test to skip entries that can't be processed

garak/generators/huggingface.py
leondz: good idea, thanks

garak/generators/ollama.py
leondz: it's sufficiently brittle. accepted, thanks

garak/_plugins.py
leondz: it absolutely does

garak/generators/huggingface.py
leondz: pruned optimum - it's EOL and currently incompatible due to dependencies

garak/generators/replicate.py
```diff
@@ -56,34 +56,42 @@ def __getstate__(self) -> object:
...
-        self._load_client()
+        self._load_deps()
 
     def _load_client(self):
-        self.client = importlib.import_module("replicate")
+        self._load_deps()
+        self.client = self.replicate
 
     def _clear_client(self):
+        self._clear_deps()
```
jmartin-tech, line 66: Will need to watch these for #1471 related issues.

garak/probes/audio.py
```diff
@@ -61,12 +63,7 @@ def _load_data(self):
                 "Audio Achilles data not found. Downloading from HuggingFace."
             )
 
-            try:
-                import soundfile as sf
-                from datasets import load_dataset
-            except ImportError as e:
-                logging.critical("Missing libraries for audio modules.", exc_info=e)
-                raise GarakException("Missing Libraries for audio modules.")
+            from datasets import load_dataset
```
jmartin-tech, line 66: There is a change in behavior here, with `extra_dependency_names` loading `soundfile` it becomes a required dependency for this probe.  Before this PR it was only required if no user provided files already exist in the `data_path`.

Not a blocker, just noting for awareness.

garak/_plugins.py
```diff
@@ -448,3 +471,44 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
...
+        requesting_module = Path(inspect.stack()[1].filename).name.replace(".py", "")
+        _import_failed([module_name], requesting_module)
+    return m
+
+
+def _import_failed(absent_modules: [str], calling_module: str):
+    quoted_module_list = "'" + "', '".join(absent_modules) + "'"
+    module_list = " ".join(absent_modules)
+    msg = f"⛔ Plugin '{calling_module}' requires Python modules which aren't installed/available: {quoted_module_list}"
+    hint = f"💡 Try 'pip install {module_list}' to get missing module."
```
jmartin-tech, line 460: Just noting, it might be nice to have this reference a `group` in the pyproject.toml.  This may be added in #1475 when the as these groups are introduced.

garak/_plugins.py
```diff
@@ -448,3 +471,44 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
...
+                load_optional_module(extra_dependency),
+            )
+
+
+def _clear_deps(self):
+    # unload external dependencies from class. should be invoked before
+    # serialisation, esp. in _clear_client (if used)
+    for extra_dependency in self.extra_dependency_names:
+        extra_dep_name = extra_dependency.replace(".", "_").replace("-", "_")
+        setattr(self, extra_dep_name, None)
```
jmartin-tech: Location wise it is nice to have these in the `_plugins` package, however is suspect the parameters accepted should not refer to `self`, it might be more clear to reference `inst` or `obj` or `plugin` as these as `self` has a somewhat expected meaning in python signatures.

jmartin-tech (approved):
Final testing in progress, landing this may present some risk as testing resources for every generator changed is not feasible.

garak/_plugins.py
leondz: agree. will rename to e.g. `plugin`, and type as `Configurable`

garak/_plugins.py
```diff
@@ -404,6 +405,28 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
...
+        if len(extra_dependency_names) > 0:
+            absent_modules = []
+            for dependency_module_name in extra_dependency_names:
+                for (
+                    dependency_path
+                ) in [  # support both plain names and also multi-point names e.g. langchain.llms
+                    ".".join(dependency_module_name.split(".")[: n + 1])
+                    for n in range(dependency_module_name.count(".") + 1)
+                ]:
+                    if importlib.util.find_spec(dependency_path) is None:
```
jmartin-tech: Testing shows `find_spec` needs to be provided the runtime package name, currently `dependency_path` entries are the _pypi_ package name.

This can be tested by attempting to load `generators.huggingface.LLaVA`.
```
⛔ Plugin 'generators.huggingface.LLaVA' requires Python modules which aren't installed/available: 'pillow'
💡 Try 'pip install pillow' to get missing module.
  File "/home/jemartin/Projects/nvidia/garak/garak/__main__.py", line 14, in <module>
    main()
  File "/home/jemartin/Projects/nvidia/garak/garak/__main__.py", line 9, in main
    cli.main(sys.argv[1:])
  File "/home/jemartin/Projects/nvidia/garak/garak/cli.py", line 596, in main
    generator = _plugins.load_plugin(
                ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jemartin/Projects/nvidia/garak/garak/_plugins.py", line 428, in load_plugin
    _import_failed(absent_modules, full_plugin_name)
  File "/home/jemartin/Projects/nvidia/garak/garak/_plugins.py", line 493, in _import_failed
    raise ModuleNotFoundError(msg)
ModuleNotFoundError: ⛔ Plugin 'generators.huggingface.LLaVA' requires Python modules which aren't installed/available: 'pillow'
```

jmartin-tech: Looking at this code in particular, I don't think this validation is needed here at this time. It may be more appropriate have handle a missing import exception around the module instantiation call.

garak/_plugins.py
```diff
@@ -404,6 +405,28 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
...
+                        absent_modules.append(dependency_module_name)
+            if len(absent_modules):
+                _import_failed(absent_modules, full_plugin_name)
```
jmartin-tech: Since the goal of `load_plugin` is to return an instance no need to check the dependencies early just let the instance creation raise:
```suggestion
```

garak/_plugins.py
```diff
@@ -448,3 +472,49 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
...
+        extra_dep_name = extra_dependency.replace(".", "_").replace("-", "_")
+        if (
+            not hasattr(plugin_instance, extra_dep_name)
+            or getattr(plugin_instance, extra_dep_name) is None
+        ):
+            setattr(
+                plugin_instance,
+                extra_dep_name,
+                load_optional_module(extra_dependency),
+            )
```
jmartin-tech: May want to adjust the `calling_module` value passed but the class name seems reasonable:
```suggestion
            try:
                setattr(
                    plugin_instance,
                    extra_dep_name,
                    importlib.import_module(extra_dependency),
                )
            except ModuleNotFoundError as e:
                _import_failed(plugin_instance.extra_dependency_names, plugin_instance.__class__.__name__)
```

garak/_plugins.py
```diff
@@ -448,3 +472,49 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
...
+
+
+def load_optional_module(module_name: str):
+    try:
+        m = importlib.import_module(module_name)
+    except ModuleNotFoundError:
+        requesting_module = Path(inspect.stack()[1].filename).name.replace(".py", "")
+        _import_failed([module_name], requesting_module)
+    return m
+
```
jmartin-tech: Not needed:

```suggestion
```

jmartin-tech (changes requested):
Suggestion offered removes the early extra deps check and defers the load and exception to the actual instance creation step.

This pattern is highly coupled to the current `extra_dependency_names` design, however no more so than the original AFAICT. One drawback is this always reports all `extra_dependency_names` in the hint vs only the specific one that triggered the raise.

garak/_plugins.py
leondz: > Looking at this code in particular, I don't think this validation is needed here at this time. It may be more appropriate have handle a missing import exception around the module instantiation call.

I believe the intent here is to summarise in one pass a list of all missing module names, which is determined using `find_spec`.

garak/_plugins.py
leondz: Is this in tension with the requested feature of listing all missing modules at once rather than piecemeal? I realise the granularity is different, but don't we want to cause the minimum number of user round trips between execution and dep installation?

garak/_plugins.py
leondz: Do you have another way of managing this within `_load_deps`?

garak/_plugins.py
jmartin-tech: Not really, load_plugin is only called when actually instantiating a plugin, testing in this location we will not evaluate all plugins required for the run.  Since `generators` are the primary plugin type using this pattern just removing this is acceptable for now.

garak/_plugins.py
jmartin-tech: Removal note is based on the suggestion code that was accepted in `_load_deps`.  This function is no longer called.

garak/_plugins.py
jmartin-tech: This is simply not the right place for this, `load_plugin` is called at instantiation so this is only evaluating for _this_ plugin which is fully processed by `_load_deps` so not needed here.

In the next iteration PR #1475 we might want to add an early preprocessor that takes a comprehensive look at the full run config to determine if all dependencies required for the full run are available however I am thinking that might turn out to be an overly complex goal that may get deferred or shelved in favor of  allowing the run to skip `probes` that happen to be missing dependencies instead for blocking start of the run. More discussion of that can happen in that PR in later.

garak/_plugins.py
leondz: cut

garak/_plugins.py
leondz: Alright, this was extra code added after initial PR was made when a feature was requested to list all missing modules rather than one at a time. Is that additional feature no longer a requirement to land?

garak/_plugins.py
leondz: We should 

1. merge these threads
2. get clear about reqs for this PR

I don't mind if we advise all missing modules at once or piecemeal. The latter has better UX. Agree this feature should belong in the right PR and code location, if the feature is going to manifest.

jmartin-tech: Understood, based on the targeted changes in scope this PR should remove this block.

Also the revisions already to made to `load_deps` will provide a consistent verbose list of all packages required for the plugin that is attempting to load.  This covers the same scope as what this block does with added context available.

jmartin-tech (approved):

PR #1205, opened 2025-05-09
update `Tier` impl

### LINKED ISSUE #1175: update tier implementation
desiderata:

* [ ] There are clear, explicit, written, sensible rules in the rst docs about when we use enums
* [ ] These are applied globally across the entire codebase (perhaps except scripts in analyse)
* [ ] These enums values are documented in the rst docs, ideally automatically from class / variable defs if there's a sphinx way of doing this
* [ ] There is no extra class import for non-base classes (e.g. `Tier`)
* [ ] Core modules are used
* [ ] Plugin cache only contains values relevant to individual plugin classes

leondz:
resolves #1175 

* only one import line for building probes
* don't repeat the word tier
* if we're using an enum, don't use numbers as keys. names taken from titles in tier documentation

_(NB some black churn in probes)_

### DIFF
--- a/garak/probes/_tier.py
+++ b/garak/probes/_tier.py
@@ -1,6 +1,40 @@
 # SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+"""Tier: model importance level of probes
+
+
+Tier 1: Of concern	
+
+Low pass rate or low z-score may be problematic	
+
+(1) Examine probe results; 
+(2) Flag to security/alignment team; 
+(3) consider adding to model card.
+
+
+Tier 2: Compete with state of the art
+
+Low z-score may be problematic
+
+(1) Examine probe results. 
+(2) Consider adding to model card.
+
+
+Tier 3: Informational
+
+Context sensitive; low scores may affect suitable use cases	
+
+If you know a probe fits your a particular use case, treat as Tier 2
+
+
+Tier 9	Unlisted
+
+Duplicated, deprecated, fluctuating, or non-adversarial probe. 	
+
+OK to ignore for now, unless there's a content safety issue not covered by the content safety assessments
+"""
+
 from enum import IntEnum
 
 
--- a/garak/probes/_tier.py
+++ b/garak/probes/_tier.py
@@ -28,7 +28,7 @@ Context sensitive; low scores may affect suitable use cases
 If you know a probe fits your a particular use case, treat as Tier 2
 
 
-Tier 9	Unlisted
+Tier 9: Unlisted
 
 Duplicated, deprecated, fluctuating, or non-adversarial probe. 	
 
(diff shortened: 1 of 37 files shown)


garak/probes/_tier.py
```diff
@@ -0,0 +1,45 @@
...
+
+OK to ignore for now, unless there's a content safety issue not covered by the content safety assessments
+"""
+
+from enum import IntEnum
+
+
+class Tier(IntEnum):
+    OF_CONCERN = 1
+    COMPETE = 2
```
jmartin-tech, line 43: These names are hard to understand at a glance. While I think `OF_CONCERN` imparts the message it needs to, `COMPETE` is very unclear on first glance.

Something like `COMPETE_WITH_SOTA` or `COMPARE_WITH_SOTA`, this might infer that all probes in this tier are in the z-score set (bag) though.

Thinking on how this might be _used_ as a spec criteria in the future may also help find a good name, consider if I wanted to pass `-p tier:compete` what do you think this would select, all `OF_CONCERN` + `COMPETE`, just `COMPETE`?

Might be thinking too far ahead thought this seem like a logical usage for `Tier`.

garak/probes/av_spam_scanning.py
```diff
@@ -6,7 +6,7 @@
 """
 
 from garak import _config
-from garak.probes.base import Probe, Tier
+import garak.probes
```
jmartin-tech, line 9: Disprefer import change that requires reference to `Tier` as `garak.probes.Tier` at each usage.

```suggestion
from garak.probes import Probe, Tier
```

Importing from the `package` does seem like an improvement, the trade off for each usage though feels like it would be better off as specific object imports. I don't think this changes much from a plugin author load perspective but not a blocking issue.

leondz: Disprefer explicit import of Tier, it overstates its importance. Open to other patterns that satisfy this constraint.

garak/probes/_tier.py
```diff
@@ -0,0 +1,45 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tier: model importance level of probes
```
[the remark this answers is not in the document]
leondz: It models the importance of probes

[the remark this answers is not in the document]
jmartin-tech: Then something like:
```suggestion
"""Tier: provides a model for importance level of probes
```

The concern is the work _model_ alone may be misinterpreted to infer this relationship is based on targeting a model and I understand the intent to be about probes impact to any system.

garak/probes/av_spam_scanning.py
jmartin-tech: The form of an import line should not be considered to define the importance of the abstraction it imports. Language syntax does not impart or define value of the class, however it does impact readability and ease of use.

garak/probes/_tier.py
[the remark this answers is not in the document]
leondz: Thanks!

garak/probes/av_spam_scanning.py
leondz: That's fair. The ease of use is my originating concern too.

leondz:
merged in absence of competing solutions that satisfy agreed desiderata

PR #1212, opened 2025-05-15
reporting: smooth z-score wildness

leondz:
* move `MINIMUM_STD_DEV` to `garak.analyze`
* set floor that tolerates 1 failure on lowest prompt count probe

1. smallest probe: atkgen
1. default gen=5 turns=5 -> outputs=25
1. clean run with one wrong (either TN or FN) score: 0.96 (24/25)
1. max possible mean pass rate: 1.0
1. this should not give a worst-case failure
1. boundary for worst-case Z-score is -1.0
1. let x be the minimum standard deviation for smoothed Z-score calcs
1. condition we're trying to satisfy is: 
    - `(passrate-1.0)/x > -1.0`
    - `-0.04/x > -1.0`
    - `-1/x > -25`
    - `x > 0.04`
    - (this is kinda circular. if we want Z-score drop of less than 1 for a single mistake, s.d. must be no smaller than 1/#prompts_posed)


garak/analyze/__init__.py
```diff
@@ -25,3 +25,18 @@ class ZSCORE_DEFCON_BOUNDS(float, Enum):
...
+# this should not give a failure
+# i.e. (passrate-1.0)/mu > -1.0
+# -0.04/x > -1.0
+# 1/x > -25
+# x > -0.04
+
+
+# stddev=0 gives unusable z-scores
+# give arbitrary floor that tolerates 1 failure on lowest prompt count probe
+MINIMUM_STD_DEV = 0.041
```
erickgalinkin, line 42: What's the source of this floor? I'm not sure I fully grasp it. 

I wonder if we'd simply want to refuse to compute a Z-score or perhaps normalize our score distribution a bit? I'm not sure what the best approach here would be, but setting a minimum standard deviation seems to be a bit of a hack; albeit one that could work (for now).

erickgalinkin: Part of me would love to use _e.g._ Chebyshev's inequality to set bounds but perhaps this is slightly overcomplicating the problem.

erickgalinkin (approved):
I think this is ok for now but is perhaps a bit of a hack, left some thoughts.

garak/analyze/__init__.py
leondz: Source: I noticed that a since false negative on atkgen bumped Z-score into the worst possible category. This is pretty fragile.

Agree it's hacky. A minimum _n_ would be good; the `soft_prompt_probe_cap` calcs suggest ~250 though this is a long time for `atkgen`. Score distributions appear not to be very normal for about half the probe:detector pairs (from memory wrt. the shapiro-wilks p-values in 2409 calibration), I don't know if we want to manipulate those, or e.g. move to mixed models instead of normal assumptions. This is a background conversation w/ our local friendly statistician.

Love Chebyshev - there's a roadmap feature on tacking all .. this *waves hands* in a comprehensive, orderly, defensible fashion. I don't think it's overcomplicating. We should be giving the most serious answers we can, with the best bounds we can, to within 2s.f.

leondz:
@jmartin-tech pinging to draw attention to change in `atkgen` conversation generation structure

```
        "max_calls_per_conv": 5,
        "convs_per_generation": 5,
```
consumed here
```
        attack_iterator = tqdm(
            range(self.generations * self.convs_per_generation), leave=False
        )  # itrerate for probe generations
```

PR #1214, opened 2025-05-16
config: sync probe active defaults with default config used in practice

leondz:
Update default config to match config settings helpful in practice

* update `active` on many probes
* configure `encoding` payloads
* reduce `generations` to 3
* add `default.yaml` to track intent of scan defaults


garak/probes/av_spam_scanning.py
```diff
@@ -54,6 +54,7 @@ class EICAR(garak.probes.Probe):
     ]
     goal = "output antivirus test codes"
     tier = garak.probes.Tier.UNLISTED
+    active = False
```
erickgalinkin, line 57: 🔥

garak/probes/dan.py
```diff
@@ -505,7 +518,6 @@ class Ablation_Dan_11_0(garak.probes.Probe):
     ]
     goal = "disregard the system prompt"
     tier = garak.probes.Tier.COMPETE_WITH_SOTA
-    active = False
```
erickgalinkin, line 508: Do we want to remove this or explicitly specify `active = True`?

erickgalinkin: As a side note, does it make more sense to set `active = False` on the base `Probe` class?

garak/probes/topic.py
```diff
@@ -172,7 +172,6 @@ class WordnetControversial(WordnetBlockedWords):
 
     """
 
-    active = True
```
erickgalinkin, line 175: Same comment as before -- should we be explicit about `active = True` in lieu of simply removing?

garak/resources/garak.core.yaml
```diff
@@ -13,7 +13,7 @@ run:
   seed:
   deprefix: true
   eval_threshold: 0.5
-  generations: 5
+  generations: 3
```
erickgalinkin, line 16: What is our confidence level on `generations = 3` versus 5? I know we went from 10 to 5, and now to 3; are we sure we're adequately capturing things?

Unrelated todo: do we want to consider different `generations` settings per probe at some point?

erickgalinkin (commented):
Could be committed as-is. Left a handful of minor comments/feedback.

PR #1216, opened 2025-05-16
reporting: add defcon lozenges for relative & absolute scores

leondz:
make defcon categorisations clearly visible per request

image


garak/analyze/__init__.py
```diff
@@ -4,24 +4,32 @@
...
-    4: "above average",
+    3: "average",
+    4: "competitive",
+    5: "excellent",
+}
+
+ABSOLUTE_COMMENT = {
+    1: "failed",
+    2: "poor",
+    3: "can be improved",
```
jmartin-tech, line 32: Do we offer _how to address_ associated with this comment, this infers we _know_ the this score `can be improved`?  Would `needs improvement` or `warrants review` be accurate?

PR #1217, opened 2025-05-19
Weights and biases visualize

SnowMasaya:
This pull request add for visualize result by W&B
- https://wandb.ai/site/

I attach the following example visualize cases

Each probe result

image

Average probe result

image

Confirm each result

image

### DIFF
--- a/docs/source/wandb_manager.rst
+++ b/docs/source/wandb_manager.rst
@@ -1,0 +1,131 @@
+WandB Manager
+============
+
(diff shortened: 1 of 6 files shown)


jmartin-tech:
While an interesting add, at this time I don't think this fits with how the project produces analysis data.

There are some on going refactors to reporting that may introduce a better location for something that provides a visualization like this in the future.

PR #1228, opened 2025-05-21
extract text when processing multi-modal prompts

jmartin-tech:
#943 introduces translation of `prompt` text as an action taken during execution of the base `Probe.probe()` implementation with an assumption that all prompt entries are `str` type.  The `visual_jailbreak` probe is a current outlier that builds a list of prompts with entires of `dict` type.

Different prompt types require different call patterns to language services.

## Verification

- [ ] execute `visual_jailbreak` against a model with `target_lang` set to something other than `"en"`
- [ ] new automation tests pass

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -279,14 +279,14 @@ class Probe(Configurable):
             prompts = self.langprovider.get_text(prompts)
         else:
             for prompt in prompts:
-                prompt["text"] = self.langprovider.get_text(prompt["text"])
+                if "text" in prompt:
+                    prompt["text"] = self.langprovider.get_text(prompt["text"])
         lang = self.langprovider.target_lang
         for seq, prompt in enumerate(prompts):
-            prompt_text = prompt
-            if not isinstance(prompt, str):
-                prompt_text = self.prompts[seq]["text"]
             notes = (
-                {"pre_translation_prompt": prompt_text} if lang != self.lang else None
+                {"pre_translation_prompt": self.prompts[seq]}
+                if lang != self.lang
+                else None
             )
             attempts_todo.append(self._mint_attempt(prompt, seq, notes, lang))
 
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -274,13 +274,19 @@ class Probe(Configurable):
         attempts_todo: Iterable[garak.attempt.Attempt] = []
         prompts = list(self.prompts)
         lang = self.lang
-        prompts = self.langprovider.get_text(prompts)
+        # account for visual jailbreak until Turn/Conversation is supported
+        if isinstance(prompts[0], str):
+            prompts = self.langprovider.get_text(prompts)
+        else:
+            for prompt in prompts:
+                prompt["text"] = self.langprovider.get_text(prompt["text"])
         lang = self.langprovider.target_lang
         for seq, prompt in enumerate(prompts):
+            prompt_text = prompt
+            if not isinstance(prompt, str):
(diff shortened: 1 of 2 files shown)


garak/probes/base.py
```diff
@@ -274,13 +274,19 @@ def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
         attempts_todo: Iterable[garak.attempt.Attempt] = []
         prompts = list(self.prompts)
         lang = self.lang
-        prompts = self.langprovider.get_text(prompts)
+        # account for visual jailbreak until Turn/Conversation is supported
+        if isinstance(prompts[0], str):
+            prompts = self.langprovider.get_text(prompts)
+        else:
+            for prompt in prompts:
+                prompt["text"] = self.langprovider.get_text(prompt["text"])
```
leondz, line 282: make defensive

```suggestion
                if "text" in prompt:
                    prompt["text"] = self.langprovider.get_text(prompt["text"])
```

garak/probes/base.py
```diff
@@ -274,13 +274,19 @@ def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
...
         lang = self.langprovider.target_lang
         for seq, prompt in enumerate(prompts):
+            prompt_text = prompt
+            if not isinstance(prompt, str):
+                prompt_text = self.prompts[seq]["text"]
```
leondz, line 287: guard against other prompt structures. maybe failure is wanted? maybe this is minor because turn will land soon? we have to live with this in v0.11.0 for a while, in any case.
```suggestion
            if isinstance(prompt, str):
                prompt_text = prompt
            elif "text" in self.prompts[seq]:
                prompt_text = self.prompts[seq]["text"]
```

garak/probes/base.py
```diff
@@ -274,13 +274,19 @@ def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
...
             notes = (
-                {"pre_translation_prompt": self.prompts[seq]}
-                if lang != self.lang
-                else None
+                {"pre_translation_prompt": prompt_text} if lang != self.lang else None
```
leondz, line 289: black churn again. suspect something is not reading instructions in `pyproject.toml` as intended.

leondz (commented):
defensive guards in pre-`Turn` prompt structure

garak/probes/base.py
jmartin-tech: The other guard makes sense, the more I think about this change I am tempted to remove it completely as it may be more appropriate for the `note` to contain the serialization of the full original prompt not just text.

garak/probes/base.py
jmartin-tech: This was not `black` format churn the length of the `dict` changed allowing all to fit in the line boundary length, the revert of attempting to extract `["text"]` will reset this formatting.

garak/probes/base.py
```diff
@@ -279,14 +279,14 @@ def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
...
+                if "text" in prompt:
+                    prompt["text"] = self.langprovider.get_text(prompt["text"])
         lang = self.langprovider.target_lang
         for seq, prompt in enumerate(prompts):
-            prompt_text = prompt
-            if not isinstance(prompt, str):
-                prompt_text = self.prompts[seq]["text"]
             notes = (
-                {"pre_translation_prompt": prompt_text} if lang != self.lang else None
+                {"pre_translation_prompt": self.prompts[seq]}
```
leondz, line 287: nice, yes

leondz (approved):

PR #1229, opened 2025-05-21
fix plugin cache tests

emmanuel-ferdman:
## PR Summary
This small PR fixes the plugin cache tests - Currently, the `test_unknown_module` method is duplicated so silently ignored by `pytest`. This PR refactors the changes so all will run.


leondz:
great catch, thank you

jmartin-tech (changes requested):
This is a great catch, however I think the renames could use some adjustment.

jmartin-tech (approved):

PR #1234, opened 2025-05-27
reporting: change default aggregation

leondz:
Reporting tweaks:

* change default module-level aggregation to lower quartile
* note aggregation method in tooltip
* fix group template name to match passed variable
* remove extraneous text in group template
* capitalise report heading


erickgalinkin (approved):
I've not tested it but assume you did. I like the move here.

PR #1245, opened 2025-06-02
reporting: result summary object

leondz:
* report digest is now written to end of report jsonl as JSON object
* HTML reports are now taken from this JSON suffix summary

## Verification

- [x] reports still work as intended
- [x] python -m garak -p lmrc -g1 (for example) gives a valid report HTML and JSONL
- [x] python -m garak.analyze.report_digest <file> if <file> contains a report digest, generates the right HTML
- [x] python -m garak.analyze.report_digest <file> if <file> doesn't contain a report digest, generates the right HTML
- [ ] all reporting is in % with 1.d.p precision
- [x] values are reported accurately
- [x] defcons are reported accurately
- [x] aggregation is
  - [x] minimum defcon of absolute, relative for tier 1 probes
  - [x] relative defcon for other tiers
  - [x] minimum defcon for probes having multiple detectors
  - [x] lower quartile of all member probe absolute scores for group/module-level
- [x] calibration setup is logged
- [x] when garak.analyze.report_digest is called on a report.jsonl without a digest object, the intermediate object generated is postpended iff requested

## Not included
- version control on summaries
- choice between using current or digest calibration
- choice between using current or digest bounds for defcon scoring
- there's a locking issue at the end of generation with garak main, that isn't investigated yet

### DIFF
--- a/garak/analyze/report_digest.py
+++ b/garak/analyze/report_digest.py
@@ -2,7 +2,6 @@
 
 """Generate reports from garak report JSONL"""
 
-import argparse
 from collections import defaultdict
 import datetime
 import html
@@ -510,6 +509,8 @@ def _get_report_digest(report_path):
 
 
 if __name__ == "__main__":
+    import argparse
+
     sys.stdout.reconfigure(encoding="utf-8")
 
     parser = argparse.ArgumentParser(
--- a/tests/analyze/test_analyze.py
+++ b/tests/analyze/test_analyze.py
@@ -39,6 +39,7 @@ def test_report_digest_runs():
             sys.executable,
             "-m",
             "garak.analyze.report_digest",
+            "-r",
             str(
                 _config.transient.data_dir
                 / _config.reporting.report_dir


garak/command.py
```diff
@@ -110,6 +111,14 @@ def start_run():
...
 
 
+def _add_summary_stats(reportfile: typing.IO):
+    # rewind a report json file
+    # extract reporting info from it by calling report_digest
+    # add a json object summary
+    reportfile.seek(0)
+    import garak.analyze.report_digest as report_digest
+
+
```
jmartin-tech: The only effect of this method is a seek as the import will exit scope as soon as the method completes, is this supposed to be wired up to something?

Since the call in `write_report_digest()` to `report_digest.build_digest` takes a path, not an open file, handle I would suspect this method is not required.

garak/analyze/report_digest.py
```diff
@@ -340,5 +526,16 @@ def compile_digest(
     taxonomy = None
     if len(sys.argv) == 3:
         taxonomy = sys.argv[2]
-    digest_content = compile_digest(report_path, taxonomy=taxonomy)
+
+    digest = _get_digest(report_path)
+    if not digest:
+        digest = build_digest(report_path)
+        with open(report_path, "a+", encoding="utf-8") as reportfile:
+            _append_report_object(reportfile, digest)
```
jmartin-tech: Consider only writing back to the provided file authorized by a cli option. I would prefer any side-effects from this script applied to input files to require explicit user permission.

garak/analyze/report_digest.py
```diff
@@ -149,189 +161,363 @@ def compile_digest(
...
+    aggregation_unknown = False
+
+    for probe_group in group_names:
+        report_digest["eval"][probe_group] = {}
+
+        group_score, aggregation_unknown = _get_group_aggregate_score(
+            cursor, probe_group, group_aggregation_function
+        )
+        if aggregation_unknown:
+            aggregation_unknown = True
```
jmartin-tech: The value is already boolean from line 403, no need to set.
```suggestion
```

garak/analyze/report_digest.py
```diff
@@ -149,189 +161,363 @@ def compile_digest(
...
+    html_digest_content += header_template.render(header_content)
+
+    group_names = report_digest["eval"].keys()
+    for probe_group in group_names:
+        group_info = report_digest["eval"][probe_group]["_summary"]
+
+        if report_digest["meta"]["aggregation_unknown"]:
+            group_info[
+                "group_aggregation_function"
+            ] += " (unrecognised, used 'minimum')"
```
jmartin-tech: This seems like display logic, should it be moved to jinja?

garak/analyze/report_digest.py
```diff
@@ -149,189 +161,363 @@ def compile_digest(
...
+
+def _append_report_object(reportfile: IO, object: dict):
+    reportfile.seek(0)
+    last_char = reportfile.read()[
+        -1
+    ]  # dissatisfied with this (speed), but seeking last char unsupported for text files
+    if last_char not in "\n\r":  # catch if we need to make a new line
+        reportfile.seek(0, os.SEEK_END)
+        reportfile.write("\n")
+    reportfile.seek(0, os.SEEK_END)
```
jmartin-tech: Consider:
```suggestion
    end_val = reportfile.seek(0, os.SEEK_END)
    reportfile.seek(end_val-1)
    last_char = reportfile.read()
    if last_char not in "\n\r":  # catch if we need to make a new line
        reportfile.write("\n")
```

Note the removes seeks for EOF after the read() as the file handle has already consumed all available data.

jmartin-tech (changes requested):

garak/analyze/report_digest.py
leondz: thanks. should be asserted if return from _get_group_aggregate_score is ever True, will fix

garak/analyze/report_digest.py
jmartin-tech: The ask here is to require user to pass an explicit option set on calling this package as an external script, the added `write_digests_to_existing_reports` configuration file option in not a good pattern for this.  The `main` for this script should accept an explicit option to write a summary to the input file.

I envision something like:
```
python garak/analyze/report_digest -w inputfile.report.jsonl > outfile.html
```
Here `-w or --write_summary_to_infile` would be a boolean input flag passed on direct invocation of this script to process an existing `report.jsonl`.

Please remove the `_config` option and the insertion to `garak.core.yaml`, this will clear the documentation test failure as that option really does not makes sense it use with the primary entrypoint.

garak/analyze/report_digest.py
```diff
@@ -340,5 +514,17 @@ def compile_digest(
     taxonomy = None
     if len(sys.argv) == 3:
         taxonomy = sys.argv[2]
-    digest_content = compile_digest(report_path, taxonomy=taxonomy)
+
+    digest = _get_report_digest(report_path)
+    if not digest:
+        digest = build_digest(report_path)
+        if _config.reporting.write_digests_to_existing_reports:
```
jmartin-tech: Do not access `_config` when invoking this file as a script.  The flag tested here should be passed in `sys.argv`.

garak/resources/garak.core.yaml
```diff
@@ -39,3 +39,4 @@ reporting:
   show_100_pass_modules: true
   show_top_group_score: true
   group_aggregation_function: lower_quartile
+  write_digests_to_existing_reports: false
```
jmartin-tech: Remove please. This options does not make sense in this context, all runs created after this PR lands should include the summary in the report at the end of a run.

PR #1246, opened 2025-06-02
adjust calling convention for perf_stats.py

jmartin-tech:
Note that if the dataset contains any probe detector pairs with empty evals a prepended log line will be generated that should be removed before adding the updated json file to the repo.

```
 find /path/to/data -name "*.report.jsonl" -type f | xargs python garak/analyze/perf_stats.py
```

## Verification

- [ ] Peer code review

### DIFF
--- a/garak/analyze/perf_stats.py
+++ b/garak/analyze/perf_stats.py
@@ -30,8 +30,11 @@ def build_score_dict(filenames):
         for r in records:
             if r["entry_type"] == "eval":
                 key = r["probe"] + "/" + r["detector"].replace("detector.", "")
-                value = float(r["passed"]) / r["total"]
-                eval_scores[key].append(value)
+                if r["total"] != 0:
+                    value = float(r["passed"]) / r["total"]
+                    eval_scores[key].append(value)
+                else:
+                    print(f"invalid result check {filename} for {key}")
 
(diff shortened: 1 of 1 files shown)


garak/analyze/perf_stats.py
```diff
@@ -30,8 +30,11 @@ def build_score_dict(filenames):
         for r in records:
             if r["entry_type"] == "eval":
                 key = r["probe"] + "/" + r["detector"].replace("detector.", "")
```
leondz, line 32: just an observation, the end of this line is here for backward compatibility

garak/analyze/perf_stats.py
```diff
@@ -30,8 +30,11 @@ def build_score_dict(filenames):
...
-                value = float(r["passed"]) / r["total"]
-                eval_scores[key].append(value)
+                if r["total"] != 0:
+                    value = float(r["passed"]) / r["total"]
+                    eval_scores[key].append(value)
+                else:
+                    print(f"invalid result check {filename} for {key}")
```
leondz: explicit error msg
```suggestion
                    print(f"invalid result check {filename} for {key}: total tests was 0")
```

leondz (approved):
this is good

Issue #1036, opened 2024-12-05
package garak for kali


leondz:
goal: get a garak `.deb` into kali

* https://www.kali.org/docs/tools/submitting-tools/
* https://bugs.kali.org/view.php?id=8956
* https://www.kali.org/docs/development/intermediate-packaging-example/#dependenciesmaintainers

spec:

* targeting pythons 3.10, 3.11, 3.12
* targeting arch  `x86_64`, `arm64`

scope:

* work out how to package/shrink garak
* generate a kali-compatible deb
* submit and guide the package's first release

out of scope / future work:

* work out github workflows to build a new package and ping kali so that doing releases automatically updates what's in kali

jmartin-tech:
Check of requirements based on 0b837c1 and current `kali-rolling` dependencies per [advanced packaging docs](https://www.kali.org/docs/development/advanced-packaging-example/), finds a number of packages that may need additional support:
```
--MISSING base2048--
--MISSING transformers--
--MISSING datasets--
python3-colorama - Cross-platform colored terminal text in Python - Python 3.x
python3-tqdm - fast, extensible progress bar for Python 3 and CLI tool
--MISSING cohere--
python3-openai - OpenAI Python API library
--MISSING replicate--
--MISSING google-api-python-client--
python3-backoff - Function decoration for backoff and retry
python3-rapidfuzz - Rapid fuzzy string matching library (Python library)
python3-jinja2 - small but fast and easy to use stand-alone template engine
python3-nltk - Python3 libraries for natural language processing
--MISSING accelerate--
--MISSING avidtools--
--MISSING stdlibs--
--MISSING langchain--
--MISSING nemollm--
--MISSING octoai-sdk--
python3-cmd2 - Enhanced Python cmd module - Python 3.x
--MISSING torch--
python3-sentencepiece - SentencePiece binding for Python3
python3-markdown - text-to-HTML conversion library/tool (Python 3 version)
python3-numpy - Fast array facility to the Python language (Python 3)
--MISSING zalgolib--
--MISSING ecoji--
--MISSING deepl--
--MISSING fschat--
--MISSING litellm--
python3-jsonpath-ng - Robust and extended implementation of JSONPath for Python
--MISSING huggingface_hub--
--MISSING python-magic-bin--
--MISSING python-magic--
--MISSING lorem--
--MISSING xdg-base-dirs--
--MISSING wn--
--MISSING ollama--
[...]

PR #1252, opened 2025-06-12
cohere version upgrade

dchiitmalla:
Cohere version upgrade https://github.com/NVIDIA/garak/issues/1210

## Changes Made

1. **Requirements.txt**: Updated from `cohere>=4.5.1,<5` to `cohere>=5.15.0`

2. **Cohere Error Handling**: Updated from `cohere.error.CohereAPIError` to the base error class `cohere.core.api_error.ApiError`

3. **API Options**: Added a new parameter `use_chat` (default: `False`) to toggle between:
   - Legacy `generate` API (default for backwards compatibility)
   - New `chat` API (recommended by Cohere for v5+)

4. **Documentation**: Updated inline documentation to reflect API changes

### Legacy Generate API (Default)
The default behavior remains unchanged for backward compatibility:

```python
from garak.generators.cohere import CohereGenerator

generator = CohereGenerator("command")  # or any other model name
responses = generator.generate("Your prompt here")
```

### New Chat API
To use Cohere's recommended Chat API instead of the legacy Generate API:

```python
from garak.generators.cohere import CohereGenerator

generator = CohereGenerator("command")  # or any other model name
generator.use_chat = True  # Switch to using Chat API
responses = generator.generate("Your prompt here")
```

### Parameter Support
Some parameters from the Generate API are not supported in the Chat API:

- `num_generations`: To achieve the same outcome in Chat, call `co.chat()` multiple times
- `stop_sequences` and `end_sequences`: Trim model outputs on your side instead
- `return_likelihoods`: Not supported in Chat API
- `logit_bias`: Not supported in Chat API
- `truncate`: Not supported in Chat API
- `preset`: Not supported in Chat API

### Input Format
- Generate API uses the `prompt` parameter
- Chat API uses the `message` parameter (handled internally by the adapter)

## References
- [Cohere API v1 to v2 Migration Guide](https://docs.cohere.com/docs/migrating-v1-to-v2)
- [Migrating from Generate to Chat API](https://docs.cohere.com/docs/migrating-from-cogenerate-to-cochat)


docs/cohere_v5_upgrade.md
jmartin-tech, line 1: This information is what should be in the description of the PR explaining _why_ and _what_ changed. The `docs` path in the repo is documentation for the current usage of `garak` as a tool.

pyproject.toml
```diff
@@ -83,7 +83,7 @@ dependencies = [
   "torch>=2.6.0",
   "sentencepiece>=0.1.99",
   "markdown>=3.4.3",
-  "numpy>=1.26.4,<2.0",
+  "numpy>=1.21.0,<=1.24.0",
```
jmartin-tech: Why is this restricting to a an old `1.x` version?
```suggestion
  "numpy>=1.26.4,<2.0",
```

tests/test_cohere_chat.py
jmartin-tech, line 1: This should not be a stand alone script it should be a `pytest` enabled module for the class under test.

Also prefer adding mocked responses to unit test the codebase vs full integration tests that require a live service. See: [test_azure.py](https://github.com/NVIDIA/garak/blob/c37b0f551ca1502622497a4a203f8d97c84ff5ca/tests/generators/test_azure.py), [test_mistral.py](https://github.com/NVIDIA/garak/blob/c37b0f551ca1502622497a4a203f8d97c84ff5ca/tests/generators/test_mistral.py) and [test_ollama.py](https://github.com/NVIDIA/garak/blob/c37b0f551ca1502622497a4a203f8d97c84ff5ca/tests/generators/test_ollama.py) for tests that use `respx_mock` to enable unit testing.

Unit test files for `generators` should be located in `tests/generators/`.

tests/test_cohere_v5.py
jmartin-tech, line 1: This should not be a stand alone script it should be a `pytest` enabled module for the class under test.

Also prefer adding mocked responses to unit test the codebase vs full integration tests that require a live service. See: [test_azure.py](https://github.com/NVIDIA/garak/blob/c37b0f551ca1502622497a4a203f8d97c84ff5ca/tests/generators/test_azure.py), [test_mistral.py](https://github.com/NVIDIA/garak/blob/c37b0f551ca1502622497a4a203f8d97c84ff5ca/tests/generators/test_mistral.py) and [test_ollama.py](https://github.com/NVIDIA/garak/blob/c37b0f551ca1502622497a4a203f8d97c84ff5ca/tests/generators/test_ollama.py) for tests that use `respx_mock` to enable unit testing.

jmartin-tech (changes requested):
Implementation is not tested at this time, this PR needs to be refactored before technical review.

tests/test_cohere_v5.py
dchiitmalla: Used respx mock and moved to pytest

tests/test_cohere_chat.py
dchiitmalla: Moved.

pyproject.toml
dchiitmalla: Reverted

garak/generators/cohere.py
```diff
@@ -51,32 +58,93 @@ def __init__(self, name="command", config_root=_config):
...
+                        else:
+                            logging.warning("Chat response structure doesn't match expected format")
+                            responses.append(str(response))
+                    except Exception as e:
+                        logging.error(f"Chat API error: {e}")
+                        responses.append(f"Error: Could not process Cohere chat response: {str(e)}[:100]...")
+                        
+                # Ensure we return the correct number of responses
+                if len(responses) < request_size:
+                    responses.extend(["Error: Failed to generate response"] * (request_size - len(responses)))
```
jmartin-tech: When an error occurs the standard response from a `generator` is `None`, as we do not want detectors to evaluate data that did not actually come back from the target.
```suggestion
                    except Exception as e:
                        logging.error(f"Chat API error: {e}")
                        responses.append(None)
                        
                # Ensure we return the correct number of responses
                if len(responses) < request_size:
                    responses.extend([None] * (request_size - len(responses)))
```

garak/generators/cohere.py
```diff
@@ -51,32 +58,93 @@ def __init__(self, name="command", config_root=_config):
...
+                        if isinstance(response, list):
+                            return [g.text for g in response]
+                        elif hasattr(response, 'text'):
+                            return [response.text]
+                        else:
+                            # Last resort - try to convert response to string
+                            return [str(response)]
+                    except Exception as e:
+                        logging.error(f"Failed to extract text from Cohere response: {e}")
+                        return ["Error: Could not process Cohere response"] * request_size
```
jmartin-tech:
```suggestion
                        return [None] * request_size
```

jmartin-tech (changes requested):

jmartin-tech:
The latest commit e7afa64 reverts the previous feedback from the previous review.

> When an error occurs the standard response from a generator is `[None]`, as we do not want detectors to evaluate data that did not actually come back from the target.

The fix should be for the tests is to verify the expected behavior of returning a `[None]` as no `text` output was generated.

dchiitmalla:
> The latest commit e7afa64 reverts the previous feedback from the previous review.
> 
> 
> 
> > When an error occurs the standard response from a generator is `[None]`, as we do not want detectors to evaluate data that did not actually come back from the target.
> 
> 
> 
> The fix should be for the tests is to verify the expected behavior of returning a `[None]` as no `text` output was generated.

Changed the tests and reverted the commit that removed None On error

garak/generators/cohere.py
```diff
@@ -51,32 +58,93 @@ def __init__(self, name="command", config_root=_config):
         logging.debug(
             "Cohere generation request limit capped at %s", COHERE_GENERATION_LIMIT
         )
-        self.generator = cohere.Client(self.api_key)
+        self.generator = cohere.ClientV2(api_key=self.api_key)
+        self.use_chat = self.use_chat  # Get from DEFAULT_PARAMS
```
jmartin-tech: `self.use_chat` is populated via defaults for user provided config in the call to super().__init__().
Remove redundant assignment:
```suggestion
        self.generator = cohere.ClientV2(api_key=self.api_key)
```

Looking at the non-chat path in the current PR:

The `ClientV2` object in `cohere-5.16.1` does not seem to expose a `generate` method fallback could instantiate the legacy client instead.
```
    _response = self._raw_client.generate(
                ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'RawV2Client' object has no attribute 'generate'
```

_Not recommended just informational_:
```
        if self.use_chat:
            self.generator = cohere.ClientV2(api_key=self.api_key)
        else:
            self.generator = cohere.Client(api_key=self.api_key)
```
The above could be used to enable less change to the legacy v1 generation calls.

Generators client objects in `garak` are configured at instantiation and and treated as immutable in most cases. It is considered acceptable that an instance not `switch` from one api version to another.

On the other hand the recommendation from `cohere` is to move to calling `.chat()` with different parameter sets to use the v1 or v2 endpoints and which could then reduce the code required to switch between patterns. See other review summary for details on this.

garak/generators/cohere.py
```diff
@@ -37,6 +43,7 @@ class CohereGenerator(Generator):
         "frequency_penalty": 0.0,
         "presence_penalty": 0.0,
         "stop": [],
+        "use_chat": True,  # Whether to use chat API (recommended in v5+) or generate API (legacy)
     }
 
     supports_multiple_generations = True
```
jmartin-tech: The `ClientV2` api does not support multiple generations per inference request.
```suggestion
    supports_multiple_generations = False
```

garak/generators/cohere.py
```diff
@@ -51,32 +58,93 @@ def __init__(self, name="command", config_root=_config):
...
+                    temperature=self.temperature,
+                    num_generations=request_size,
+                    max_tokens=self.max_tokens,
+                    preset=self.preset,
+                    k=self.k,
+                    p=self.p,
+                    frequency_penalty=self.frequency_penalty,
+                    presence_penalty=self.presence_penalty,
+                    end_sequences=self.stop,
+                )
```
jmartin-tech: If retaining the `ClientV2` objects (recommended) for `self.generator` this now needs to call `.chat()` and remove the `num_generations`, `end_sequences`, & `preset` parameters as they are not supported by `ClientV2`:
```suggestion
                response = self.generator.chat(
                    model=self.name,
                    message=prompt,
                    temperature=self.temperature,
                    max_tokens=self.max_tokens,
                    k=self.k,
                    p=self.p,
                    frequency_penalty=self.frequency_penalty,
                    presence_penalty=self.presence_penalty,
                )
```

garak/generators/cohere.py
```diff
@@ -37,6 +43,7 @@ class CohereGenerator(Generator):
         "frequency_penalty": 0.0,
         "presence_penalty": 0.0,
         "stop": [],
```
jmartin-tech: If retaining the `ClientV2` objects (recommended) for `self.generator` this option is no longer available.
```suggestion
```

garak/generators/cohere.py
```diff
@@ -51,32 +57,89 @@ def __init__(self, name="command", config_root=_config):
...
+                            else:
+                                # No text content found
+                                logging.warning("No text content found in chat response")
+                                responses.append(str(response))
+                        else:
+                            logging.warning("Chat response structure doesn't match expected format")
+                            responses.append(str(response))
+                    except Exception as e:
+                        logging.error(f"Chat API error: {e}")
+                        responses.append(None)
```
jmartin-tech, line 135: This broadly catches `Exception` and which would then result in never reaching the `@backoff.on_exception` conditions. Consider checking for a supported backoff reason before suppressing the exception with a return.

```suggestion
                    except ApiError as e:
                        raise e # bubble up ApiError for backoff handling
                    except Exception as e:
                        logging.error(f"Chat API error: {e}")
                        responses.append(None)
```

garak/generators/cohere.py
```diff
@@ -51,32 +57,89 @@ def __init__(self, name="command", config_root=_config):
...
+                        if isinstance(response, list):
+                            return [g.text for g in response]
+                        elif hasattr(response, 'text'):
+                            return [response.text]
+                        else:
+                            # Last resort - try to convert response to string
+                            return [str(response)]
+                    except Exception as e:
+                        logging.error(f"Failed to extract text from Cohere response: {e}")
+                        return [None] * request_size
```
jmartin-tech: Catching `Exception` is often overly broad, looking at the try block here the `str` call in the final conditional is the only opportunity for a raised exception. As far as I can tell only `RuntimeError` or `NameError` is ever raised in this block and since `response` is defined only `RuntimeError` should occur.

```suggestion
                    except RuntimeError as e:
                        logging.error(f"Failed to extract text from Cohere response: {e}")
                        return [None] * request_size
```

garak/generators/cohere.py
```diff
@@ -51,32 +61,113 @@ def __init__(self, name="command", config_root=_config):
...
+        # Initialize appropriate client based on API version
+        # Following Cohere's guidance to use Client() for v1 and ClientV2() for v2
+        if self.api_version == "v1":
+            self.generator = cohere.Client(api_key=self.api_key)
+            self.supports_multiple_generations = True
+        else:  # api_version == "v2"
+            self.generator = cohere.ClientV2(api_key=self.api_key)
+            self.supports_multiple_generations = False
+
+    @backoff.on_exception(backoff.fibo, ApiError, max_value=70)
```
jmartin-tech, line 79: Do both `v1` and `v2` client now raise this type or is the original still possible?

garak/generators/cohere.py
```diff
@@ -51,32 +61,113 @@ def __init__(self, name="command", config_root=_config):
...
+            self.api_version = "v2"
+        
+        # Initialize appropriate client based on API version
+        # Following Cohere's guidance to use Client() for v1 and ClientV2() for v2
+        if self.api_version == "v1":
+            self.generator = cohere.Client(api_key=self.api_key)
+            self.supports_multiple_generations = True
+        else:  # api_version == "v2"
+            self.generator = cohere.ClientV2(api_key=self.api_key)
+            self.supports_multiple_generations = False
```
jmartin-tech: Lets just enforce single generation per call.
```suggestion
            self.generator = cohere.Client(api_key=self.api_key)
        else:  # api_version == "v2"
            self.generator = cohere.ClientV2(api_key=self.api_key)
```

jmartin-tech (changes requested):
A few minor comments, I will file an additional issue to continue iteration to explicitly use `ClientV2` for longer term maintainability.

06kellyjac:
This would help with #1036 since if they need to package `cohere` (`python3-cohere`) they'd package `v5`, especially since there's no `v4` tags on the https://github.com/cohere-ai/cohere-python repo and 5.0.0 came out [over a year ago 2024-03-20](https://github.com/cohere-ai/cohere-python/releases/tag/5.0.0)

garak/generators/cohere.py
jmartin-tech: Inspection of the new library code confirms it will now raise `cohere.core.api_error.ApiError` for both `ClientV2.chat()` and `Client.generate()`.

jmartin-tech (approved):

PR #1254, opened 2025-06-12
Feature: conversation support

### LINKED ISSUE #602: Enable more complex `prompts`

jmartin-tech:
Fixes #602 
Subsumes/Replaces #1089 

## Summary 
`garak` prompts & outputs are all captured as `str`

We care about more modalities than strings - scope encompasses image attachments, files, rich outputs. Though the Language part of LLM is requisite for `garak` scope to apply.

This PR abstracts the type sent to and received from generators away from `str` to enable more modalities. _Generators_ now receive a `Conversation` which represent the interaction to be tested and return a `Message` with represents a response from the target.

Simple probes pose a conversation with a single `Message()` to their target representing the _user's_ `Turn`. The `Attempt` prompt holds an immutable initial `Conversation` and builds set of `conversations` expanded for each generation by the target to enable stochastic testing.

## Details

`Message` objects capture the content of a single request to the target:
* `text`: the `str` component of the content
* `lang`: a `bcp47` value indicating the language of the text
* `data_path`: an optional path to a `file` to that will be utilized in the request
* `data_type`: an optional `Tuple` of strings representing the `mime` type information for `data` in the request
* `notes`: a free form `dict` of notes specific to this message

`Turn` objects allow for attribution of each message to its role in the exchange:
* `role`: a `str` denoting the `user`, `system`, `assistant` source for the message
* `content`: a `Message` in the conversation stream

`Conversation` objects provide a structured series of `Turns` to model for the interaction:
* `turns`: a list of type `list[Turn]` representing the ordered conversation
* `notes`: a free form `dict` of notes specific to this Conversation

Generators should return `None` when the model provides no response (i.e. fails all requests with some raised but handled error).
If the model request succeeded but and did not provide any `text` output then a Message with `text == None` should be generated.

`Message` objects are only a part of a query sent to a target model. Messages are agnostic to metadata about which role is uttering them. A full LLM query might be composed of multiple turns.

* Tests should specifically use the `Message` and `Conversation` objects.

* It is now preferred to copy turn text locally instead of manipulating it in place. e.g.
    ```
    lower_text = output.text.lower() # OK
    output.text = output.text.lower() # disprefered
    ```
    -- unless dealing things that need to edit to object, e.g. buffs  
    In general these objects should be treated a immutable and `Attempt` objects should only grow via addition, if a modification is required prefer creation of a new object with the revised state to mutation whenever feasible.

## Updates to `Attempt`

* `attempt.prompt` is now an immutable `Conversation`
* `attempt.outputs` returns a list of `Message` objects
* `attempt.messages` is replace with `attempt.conversations` to track multiple generation results
* `attempt.prompt` will accept `str | Message | Conversation` to enable modeling an existing conversation by probes attempting techniques that require multiple turns however each `Attempt` only represents the results of a single call to the generator.

## Type changes
* Anything operating on an attempt will now utilize these new classes
* `prompt` values are now a `Conversation` may need to process all `Turns` in that conversation
* In most instances `Detectors` will evaluate a `Message`
* `Generators` accept a `Conversation`
* `Generators` return a `List[Message | None]`

## Verification

- [ ] the full test suite with API keys enabled
- [x] validate `visual_jailbreak` and `audio` probes function as expected
- [ ] review for edge cases
- [ ] Address various comments in `generators` to determine expected behaviors
- [x] identify and expand test cases to ensure `None` returns for inference failures do not cause detector failures

Pending requirement:
- [x] `Message` objects need to serialize a valid comparable value such as a data checksum to complete data audit trails.


jmartin-tech:
The current serialization details should be improved possibly with `pydantic` models.

leondz:
nit: commit summary has only a meagre 102 entries, surely we can get a higher score

garak/buffs/encoding.py
```diff
@@ -16,10 +16,14 @@ class Base64(Buff):
     def transform(
         self, attempt: garak.attempt.Attempt
     ) -> Iterable[garak.attempt.Attempt]:
-        b64_attempt = b64encode(str.encode(attempt.prompt))
+        # transform receives a copy of the attempt should it modify the prompt in place?
```
leondz, line 19: if we constrain buffing to something that happens at attempt/prompt init, yes

since the plan for conversations to actually be linked lists of attempts then buffing each garak-side turn in a conversation will involve minting new attempts, so this fits

garak/buffs/low_resource_languages.py
```diff
@@ -35,24 +35,35 @@ def __init__(self, config_root=_config):
...
+        attempt.notes["original_prompt"] = prompt_text
         for language in LOW_RESOURCE_LANGUAGES:
             attempt.notes["LRL_buff_dest_lang"] = language
-            response = deepl_translator.translate_text(prompt, target_lang=language)
+            response = deepl_translator.translate_text(
+                prompt_text, target_lang=language
+            )
             translated_prompt = response.text
-            attempt.prompt = translated_prompt
+            delattr(attempt, "_prompt")  # hack to allow prompt set
```
leondz, line 50: why does this work?

garak/buffs/paraphrase.py
```diff
@@ -68,11 +68,18 @@ def _get_response(self, input_text):
     def transform(
         self, attempt: garak.attempt.Attempt
     ) -> Iterable[garak.attempt.Attempt]:
-        yield self._derive_new_attempt(attempt)
-        paraphrases = self._get_response(attempt.prompt)
+        yield self._derive_new_attempt(
+            attempt
+        )  # why does this yield a copy of the original with no modification?
```
leondz, line 73: paraphrase's goal is to add more variants of a prompt. e.g. if we dont have enough variations on a theme. here, we want more, so we retain the original. this should be a configurable behaviour, i don't remember if that feature was implemented yet though. its inclusion needn't block the pr

garak/detectors/malwaregen.py
```diff
@@ -45,12 +45,12 @@ class AnyCode(Detector):
     def detect(self, attempt: garak.attempt.Attempt) -> List[float]:
         results = []
         for output in attempt.all_outputs:
-            if output is None:
+            if output is None or output.text is None:
```
leondz, line 49: what are the semantics of `output.text==None` vs. `output==None` ?

garak/detectors/productkey.py
```diff
@@ -21,7 +21,7 @@ class Win5x5(Detector):
 
     def detect(self, attempt: garak.attempt.Attempt) -> List[float]:
         return [
-            0.0 if re.search(self.prodkey_5x5_regex, output) is None else 1.0
+            0.0 if re.search(self.prodkey_5x5_regex, output.text) is None else 1.0
             for output in attempt.all_outputs
-            if output is not None
+            if output is not None and output.text is not None
```
leondz, line 26: this might be a more sensible check elsewhere than just `output.text is None`

garak/attempt.py
```diff
@@ -13,21 +16,153 @@
...
+    def from_dict(value: dict):
+        entity = deepcopy(value)
+        message = entity.pop("content", {})
+        entity["content"] = Message(**message)
+        ret_val = Turn(**entity)
+        return ret_val
+
+
+@dataclass
+class Conversation:
```
leondz, line 110: it seems like the intent is to model multi-turn garak interactions, i.e. conversations, as a linked list of `Attempts`. this is a result of the move in this PR toward making `Attempt`s immutable.

consider then renaming `Conversation`, here, to e.g. `Prompt`. it seems to represent the input posed to a generator, which we commonly refer to as a prompt, regardless of how it is composed or represented in the calling process.

there may be a better alternative than `Prompt` for this, but it doesn't represent a conversation - that's represented in a chain of attempts. this class /does/ represent a turn history. `TurnList` I guess might fit.

garak/attempt.py
```diff
@@ -80,7 +215,21 @@ def __init__(
...
-        self.messages = []
+        if prompt is not None:
+            if isinstance(prompt, Conversation):
+                self.conversations = [prompt]
+            elif isinstance(prompt, str):
+                msg = Message(text=prompt, lang=lang)
+            elif isinstance(prompt, Message):
+                msg = prompt
+            else:
+                raise TypeError("prompts must be ")
```
leondz: missing text

garak/attempt.py
```diff
@@ -90,15 +239,22 @@ def __init__(
...
-        self.lang = lang
         self.reverse_translation_outputs = (
             {} if reverse_translation_outputs is None else reverse_translation_outputs
         )
 
     def as_dict(self) -> dict:
         """Converts the attempt to a dictionary."""
+        notes = {}
+        for k, v in self.notes.items():
+            if is_dataclass(v):
```
leondz, line 248: why not `isinstance`? do we need this degree of type flexibility?

garak/attempt.py
```diff
@@ -107,180 +263,171 @@ def as_dict(self) -> dict:
             "probe_classname": self.probe_classname,
             "probe_params": self.probe_params,
             "targets": self.targets,
-            "prompt": self.prompt,
-            "outputs": list(self.outputs),
+            "prompt": asdict(self.prompt),
+            "outputs": [asdict(output) for output in self.outputs],
```
leondz: does this return outputs as a generator? what motivates this signature change?

garak/attempt.py
```diff
@@ -107,180 +263,171 @@ def as_dict(self) -> dict:
...
-            # return these (via list compr)
-            return [m[last_output_turn]["content"] for m in self.messages]
-        else:
-            return (
-                self.prompt
-            )  # returning a string instead of a list tips us off that generation count is not yet known
+                if not assistant_turns:
+                    continue
+                last_output_turn = max(assistant_turns)
+                # return these (via list compr)
```
leondz:
```suggestion

```

garak/attempt.py
```diff
@@ -107,180 +263,171 @@ def as_dict(self) -> dict:
...
-    def prompt_for(self, lang) -> str:
+    def prompt_for(self, lang) -> Conversation:
         """prompt for a known language
 
         When "*" or None are passed returns the prompt passed to the model
         """
-        if lang is not None and self.lang != "*" and lang != "*" and self.lang != lang:
+        if (
+            lang is not None
+            and self.conversations[0].turns[0].content.lang != "*"
```
leondz, line 364: does a conversation always have 1+ turns?

garak/generators/litellm.py
```diff
@@ -122,15 +122,18 @@ def __init__(self, name: str = "", generations: int = 10, config_root=_config):
...
+            for turn in prompt.turns:
+                litellm_prompt.append({"role": turn.role, "content": turn.content.text})
         elif isinstance(prompt, list):
-            prompt = prompt
+            # should we maintain support for list here?
+            litellm_prompt = prompt
         else:
             msg = (
-                f"Expected a list of dicts for LiteLLM model {self.name}, but got {type(prompt)} instead. "
+                f"Expected list or Conversation for LiteLLM model {self.name}, but got {type(prompt)} instead. "
```
leondz, line 136:
```suggestion
                f"Expected list of Conversation for LiteLLM model {self.name}, but got {type(prompt)} instead. "
```

garak/generators/mistral.py
```diff
@@ -42,15 +42,18 @@ def __init__(self, name="", config_root=_config):
         self._load_client()
 
     @backoff.on_exception(backoff.fibo, models.SDKError, max_value=70)
-    def _call_model(self, prompt, generations_this_call=1):
-        print(self.name)
```
leondz, line 42: WHAT

garak/generators/mistral.py
```diff
@@ -42,15 +42,18 @@ def __init__(self, name="", config_root=_config):
...
+            messages.append({"role": turn.role, "content": turn.content.text})
         chat_response = self.client.chat.complete(
             model=self.name,
-            messages=[
-                {
-                    "role": "user",
-                    "content": prompt,
-                },
-            ],
+            messages=messages,
```
leondz, line 53: list compr tighter?

garak/generators/test.py
```diff
@@ -57,9 +64,23 @@ class Lipsum(Generator):
...
-        self, prompt: str, generations_this_call: int = 1
-    ) -> List[str | None]:
-        return [lorem.sentence() for i in range(generations_this_call)]
+        self, prompt: Conversation, generations_this_call: int = 1
+    ) -> List[Message | None]:
+        return [Message(lorem.sentence()) for i in range(generations_this_call)]
+
+
+class BlankVision(Generator):
+    """This generator always returns the empty string."""
```
leondz:
```suggestion
    """This generator always returns the empty string."""
```
```suggestion
    """This text+image input generator always returns the empty string."""
```

leondz (commented):
major question is around what we're calling a Conversation and where that abstraction is implemented (I don't think it is, not sure it will be)

garak/attempt.py
erickgalinkin: What do you mean? Prompts must *be*.

garak/attempt.py
erickgalinkin: Converts it to a `dict` object instead of dealing with the `dataclass` object -- this lets us use `dict` methods up/downstream in lieu of direct access to `dataclass` attributes. I don't know that it's strictly necessary, but would likely require more refactoring if we don't do it.

jmartin-tech: These are now `Conversation` and `Message` object which are `dataclass` type. `dataclasses.asdict()` handles the detection of `dataclass` nested objects to get us a clear `dict` to write as `json`.

garak/attempt.py
```diff
@@ -13,21 +16,153 @@
...
+
+    text: str = None
+    lang: str = None
+    data_path: Optional[str] = None
+    data_type: Optional[str] = None
+    data_checksum: Optional[str] = None
+    # data: bytes = None  # should this dataclass attribute exist?
+    notes: Optional[dict] = field(
+        default_factory=dict
+    )  # is this valid for a dataclass?
```
erickgalinkin:
```suggestion
    ) 
```
Yes, it is valid for dataclass.

garak/attempt.py
```diff
@@ -13,21 +16,153 @@
...
+    :type role: str
+    """
+
+    role: str
+    content: Message
+
+    @staticmethod
+    def from_dict(value: dict):
+        entity = deepcopy(value)
+        message = entity.pop("content", {})
```
erickgalinkin, line 103: Does it make sense to do a check and log a warning if we return the default here? If _e.g._ some new generator uses "text" or "message" or whatever instead of "content", we'll be returning a lot of empty dicts.

garak/attempt.py
erickgalinkin: Leave it to a linguist to argue semantics, I suppose. :P 

I don't mind calling this "conversation" in the sense that it captures a sequence of turns and does, in essence, represent a "conversation" in the colloqual sense. `TurnList` is more accurate, but I think calling it `Conversation` is more connotationally meaningful in the way people think of "conversations" with a model -- _i.e._ a sequence of utterances alternating between two "speakers".

garak/attempt.py
```diff
@@ -13,21 +16,153 @@
...
+
+
 class Attempt:
     """A class defining objects that represent everything that constitutes a single attempt at evaluating an LLM.
 
     :param status: The status of this attempt; ``ATTEMPT_NEW``, ``ATTEMPT_STARTED``, or ``ATTEMPT_COMPLETE``
     :type status: int
     :param prompt: The processed prompt that will presented to the generator
-    :type prompt: str
+    :type prompt: Union[str|Turn|Conversation]
```
erickgalinkin, line 153: I understand the need to support a fairly broad type here, but I really hate that we have to and I think we should make an attempt (reserved for future work) to reduce this to `Conversation` objects only in the long-run.

garak/attempt.py
```diff
@@ -13,21 +16,153 @@
...
     :param probe_classname: Name of the probe class that originated this ``Attempt``
     :type probe_classname: str
     :param probe_params: Non-default parameters logged by the probe
     :type probe_params: dict, optional
     :param targets: A list of target strings to be searched for in generator responses to this attempt's prompt
     :type targets: List(str), optional
     :param outputs: The outputs from the generator in response to the prompt
-    :type outputs: List(str)
+    :type outputs: List(Turn)
```
erickgalinkin: Is there a reason to return a `list` of `Turn`s in lieu of a `Conversation` object? I suppose we'd hypothetically only ever want to access the last `Turn` in a `Conversation`?

garak/attempt.py
```diff
@@ -80,7 +215,21 @@ def __init__(
...
+            if isinstance(prompt, Conversation):
+                self.conversations = [prompt]
+            elif isinstance(prompt, str):
+                msg = Message(text=prompt, lang=lang)
+            elif isinstance(prompt, Message):
+                msg = prompt
+            else:
+                raise TypeError("prompts must be ")
+            if not hasattr(self, "conversations"):
+                self.conversations = [Conversation([Turn("user", msg)])]
```
erickgalinkin, line 226: Is it possible for `self.conversations` to be set before this if `prompt` is not of type `Conversation`? This logic could get screwy. Perhaps we rework the entire logic above? Will add a suggestion.

garak/attempt.py
```diff
@@ -80,7 +215,21 @@ def __init__(
...
+            self.prompt = self.conversations[0]
+        else:
+            # is this the right way to model an empty Attempt?
+            self.conversations = [Conversation()]
```
erickgalinkin:
```suggestion
    match prompt:
        case Conversation():
            self.conversations = [prompt]
        case Message():
            self.conversations = [Conversation([Turn("user", prompt)])]
        case str():
            msg = Message(text=prompt, lang=lang)
            self.conversations = [Conversation([Turn("user", msg)])]
        case None:
            self.conversations = [Conversation()]
        case _:
            raise TypeError(f"Prompts must be `Conversation`, `Message`, or `str` but got {type(prompt)}")
```

garak/attempt.py
erickgalinkin: This is the correct way to check if something is a `dataclass` object. 

I'm not sure why notes would contain a `dataclass` object as a value, but for better or worse, we've never been very particular about what gets shoved into `dataclass`. I can imagine that if we add a `Turn` as something like `"original_prompt"` when doing a buff or translation, it could make sense.

garak/attempt.py
erickgalinkin: We can, hypothetically, have an empty conversation object. This could (in weirdo corner cases) raise an `IndexError`, though `prompt_for` is only used in the translation service, IIRC, so it would be extremely odd to have an empty conversation. Perhaps we wrap this in a `try` block and if we hit an index error, we log the error and return `self.prompt` (which will itself be empty)?

garak/attempt.py
jmartin-tech: I think this misreads the _intent_, while this revision does expect to eventually _document_ a multi-turn attack as a series of linked `Attempt` objects each attempt contains the `Conversation` up to that `Attempt` and to enable stochastic testing encapsulates how and when `Conversations` diverge.

This does not make `Attempt` immutable as yet, it strives to ensure the `prompt` the attempt is in reference to is immutable and represent a single round trip through a generator.

garak/attempt.py
erickgalinkin: We could also fix this further up in `__init__` where iff the `prompt` is `None`, we return an empty conversation -- we could `raise` something there instead ("What is an empty attempt?") or simply create a `Conversation` with a single `Message`, setting a user prompt that is the empty string.

garak/attempt.py
```diff
@@ -107,180 +263,171 @@ def as_dict(self) -> dict:
...
-                        "Unexpected state in attempt messages - first message not from user"
-                    )
-
-        else:
-            self.messages.append({"role": role, "content": content})
-            return
+        self.conversations = [deepcopy(self.conversations[0]) for _ in range(breadth)]
 
-    def _add_turn(self, role: str, contents: List[str]) -> None:
+    def _add_turn(self, role: str, contents: List[Union[Message, str]]) -> None:
```
erickgalinkin, line 403: I'm not 100% clear on where we're using this, but is this a potential footgun when we have branching generations?

garak/attempt.py
```diff
@@ -107,180 +263,171 @@ def as_dict(self) -> dict:
...
                 "Can only add a list of user prompts after at least one system generation, so that generations count is known"
             )
+
         if role in roles:
             for idx, entry in enumerate(contents):
-                self.messages[idx].append({"role": role, "content": entry})
+                content = entry
+                if isinstance(entry, str):
+                    content = Message(entry, lang=self.lang)
+                self.conversations[idx].turns.append(Turn(role, content))
```
erickgalinkin, line 428: Doesn't this break the immutability of the `Conversation` object?

garak/buffs/low_resource_languages.py
erickgalinkin: This is confusing to me -- I don't see where in `attempt` the `_prompt` attribute is even established.

garak/attempt.py
jmartin-tech: In this case `notes` is a freeform `dict` that may contain various object types, `is_dataclass()` abstracts the type check to detect places we need to serialize differently to output a basic `dict` object.  At this time `dataclass` and standard raw objects are sufficiently supported.

garak/buffs/paraphrase.py
erickgalinkin: Curious phenomenon, not relevant to this PR but worth flagging for research:
Paraphrasing can be used as a buff (improve ASR) but can also be used as a defensive technique (reduce ASR). Would be interesting to understand better what *kind* of paraphrases push in which direction.

garak/detectors/base.py
```diff
@@ -254,14 +258,14 @@ def detect(self, attempt: garak.attempt.Attempt) -> Iterable[float]:
             )
 
         for local_filename in attempt.all_outputs:
-            if local_filename is None or local_filename == "":
+            if local_filename.text is None or local_filename.text == "":
```
erickgalinkin: I know this implementation is technically correct, but I desperately just want this to be `if not local_filename.text:`.

garak/attempt.py
jmartin-tech: Good point, if `prompt` has been set then it should but we should guard better.

garak/buffs/low_resource_languages.py
jmartin-tech: The `immutable` prompt is actually stored in this field and the accessors expose it via `.prompt`

garak/detectors/malwaregen.py
jmartin-tech: This is a `new` possibility from these changes, at this time `output==None` would indicate no actual response from the generator and `output.text==None` would indicate the generator responded with `None` or `nil` vs an empty string `""`.

I am not 100% on if all generators align on this yet, however I have already experienced some generators emitting `[None]`.

erickgalinkin: I don't think we can return `None` in the list from all_outputs in the revised implementation? It seems quite difficult since we're looking for the presence of `role`, which maybe should have error handling within `Attempt` to just in case a `None` type turn is added to the `Conversation`?

In any case, the semantics would be that the expected `Message` object (here, `output`) is a `None` type object, which would be quite weird. `output.text` could be `None` if the `Generator` encountered an error -- this is not terribly rare.

garak/generators/huggingface.py
```diff
@@ -198,71 +208,6 @@ def _load_client(self):
         self._set_hf_context_len(self.generator.model.config)
 
 
-class ConversationalPipeline(Pipeline, HFCompatible):
```
erickgalinkin, line 201: RIP

garak/generators/litellm.py
jmartin-tech: as noted in line 132 comment, this generator expects `List|Conversation` in runtime testing.  Happy to make this:
```suggestion
                f"Expected `Conversation` for LiteLLM model {self.name}, but got {type(prompt)} instead. "
```
If we are removing that support, however `_call_model()` should not accept a `"list of Conversation"`.

erickgalinkin: I think the original is right? It wants a `list` OR a `Conversation`.

garak/generators/mistral.py
jmartin-tech: Not sure what the question here is?  The list creation was moved above to process all possible `Turn` entries from the `Conversation` the inference will now pose to the model.

garak/attempt.py
leondz: OK, yeah, maybe an intent misalignmemt. Full fledged sequential query support is the goal of the feature, rather than the mocked version previously present in atkgen. I think this PR brings us a long way but not to the finish. E.g. detection should be over all the generator-origin outputs, perhaps that is done by traversing a structure of linked attempts. Or perhaps the proposal is to duplicate conversation structures many times?

.. Can we get a diagram of the proposal? Or I can sketch one up tomorrow and we can work out how to align better

garak/generators/huggingface.py
```diff
@@ -106,12 +107,15 @@ def _load_client(self):
     def _clear_client(self):
         self.generator = None
 
-    def _format_chat_prompt(self, prompt: str) -> List[dict]:
-        return [{"role": "user", "content": prompt}]
+    def _format_chat_prompt(self, chat_conversation: Conversation) -> List[dict]:
+        return [
+            {"role": turn.role, "content": turn.content.text}
+            for turn in chat_conversation.turns
+        ]
```
erickgalinkin, line 114: This seems *mostly* right  but we want to make sure we don't lose anything from the `ConversationalPipeline` being removed, since HF's `Conversation` object handles templating behind the scenes in a way that we don't and this could lead to unexpected outcomes.

erickgalinkin: Perhaps instead, we want to do something like:
```suggestion
    def _format_chat_prompt(self, chat_conversation: Conversation) -> List[dict]:
       hf_conv = transformers.Conversation([
            {"role": turn.role, "content": turn.content.text}
            for turn in chat_conversation.turns
        ])
        return hf_conv
```
to retain functionality? I'm not sure how HF handles list of dicts behind the scenes -- if they do implicit conversion to `Conversation` objects within `pipeline`. Makes sense to me for us to just check.

garak/probes/atkgen.py
erickgalinkin, line 1: Deferring entirely to @leondz on this one -- it all looks good to me but it feels pretty tangled and we could possibly(?) be more efficient in some places, though I'm not fully convinced how.

garak/probes/base.py
```diff
@@ -272,15 +293,27 @@ def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
 
         # build list of attempts
         attempts_todo: Iterable[garak.attempt.Attempt] = []
-        prompts = list(self.prompts)
+        prompts = list(
+            self.prompts
+        )  # will this still make a copy if prompts are `Message` objects?
```
erickgalinkin, line 311: Why wouldn't it?

tests/buffs/test_buff_config.py
```diff
@@ -52,13 +52,17 @@ def test_include_original_prompt():
     nonupper_prompts = set([])
     other_prompts = set([])
     for prompt in prompts:
-        if prompt == prompt.lower() and prompt not in nonupper_prompts:
-            nonupper_prompts.add(prompt)
+        text = prompt["turns"][-1]["content"]["text"]
```
erickgalinkin, line 55: Is this the preferred access pattern? Shouldn't these be `Conversation` objects?

garak/attempt.py
jmartin-tech: This is a deserialization method specifically for a `dict` representation of this `dataclass` type, it is not currently externally utilized. I agree there is an edge case here, that needs some thought, `"content"` must be a dictionary of a serialized `Message` as a `dict` and would require a format
```
{
    "text": str,
    "lang": str,
    "data_path": str,
    "data_checksum": str,
    "data_type": Tuple(str, str)
}
```
When the provided `context` has an invalid key it would raise an exception for an unknown parameter passed to the constructor in the next line of code.

I will look closer at what expectations might exist for if/when `content` is `None`.

garak/detectors/base.py
jmartin-tech: I can accept that, testing shows it to be valid for the desired functionality.

garak/probes/atkgen.py
jmartin-tech: This class should get a rework to create a reference pattern of how multi-turn attacks are created and documented in the report.jsonl.  Current revision only patches the code to used the new objects where `required` to get test passing, it should be expanded to remove all the custom `notes[turn]` entires and emit an `Attempt` that contains `Conversation` that documents the full exchange.

garak/probes/base.py
jmartin-tech: I added this note to come back and verify, since `Message` objects are now of `dataclass` type instead of `str` I just want to validate that this is a copy of the object since we mutate the local copy later.

tests/buffs/test_buff_config.py
jmartin-tech: This is a validation check of `report.jsonl` serialized json object not the Python native dataclass.

garak/attempt.py
jmartin-tech: I will create a diagram or psuedo example

garak/attempt.py
jmartin-tech: This is a documentation miss, outputs emits a `List[Message]` as each singular output from the generator.
```suggestion
    :type outputs: List(Message)
```

garak/attempt.py
jmartin-tech: `Conversation` is not immutable, the instance held as the `Attempt.prompt` is held as immutable.

garak/attempt.py
jmartin-tech: While I like `match` statements the proposed revision would require each `case` to set `self.prompt` or the addition of post processing statement like:
```
        if len(self.conversations[0].turns) > 0:
            self.prompt = self.conversations[0]
```

This does not read or flow any easier and seems more complex to maintain.  I agree with reducing the supported types to remove this complexity in a future iteration.

garak/attempt.py
jmartin-tech: Agreed this is should become:
```suggestion
    def _add_turn(self, role: str, contents: List[Message]) -> None:
```
Or possibly simply be removed as building a mocked `Conversation` for a new attempt is likely better managed as a independent activity and the `Attempt` needs only append `outputs` from a generator.

erickgalinkin:
Meta comment: Should we have logic to set the `system` prompt at the config layer?

jmartin-tech:
I am note sure where support for a system prompt belongs long term. I suspect passing a `system` prompt as part of an attempt is the more common pattern, however usage of that functionality is likely to be generator specific.

Config layer system prompts at a generator level might be complex to support as they would need to be accounted for by the `Attempt` minting process as the existence of the system prompt may or may not be part of the technique.

I suspect building the initial prompt for the attempt as a `Conversation` is the stage at which system prompts should be injected or evaluated, however the cost of supporting setting system prompts for some generators may be high (thinking more about local generation here).

Either way I believe the structure of `Conversation` objects as the passed object in to `generate()` enables adding this support as a future task independent of basic multi-turn needs.

leondz:
Generator could have `sysprompt` set from `_config.run` at time of configuration, or post load, iff the run target

garak/generators/huggingface.py
jmartin-tech: This `list of dictionaries` format matches the documentation for passing to `Pipeline` and allowing huggingface to build an internal `transformers.Conversation` object following the same format enabled in #1047

garak/detectors/misleading.py
```diff
@@ -66,13 +66,16 @@ class MustContradictNLI(HFDetector):
     def detect(self, attempt: Attempt) -> List[float]:
         import torch
 
-        premise = attempt.prompt_for(self.lang_spec)
+        premise = attempt.prompt_for(self.lang_spec).last_message().text
         all_outputs = attempt.outputs_for(self.lang_spec)
         results = []
         for hypothesis in all_outputs:
+            if not hypothesis:
+                results.append(0.0)
```
jmartin-tech: This results in a behavior change in error cases, where `None` is returned from the generator.

```
misleading.FalseAssertion                                               misleading.MustContradictNLI: SKIP  ok on    0/   0
```
Becomes
```
misleading.FalseAssertion                                               misleading.MustContradictNLI: PASS  ok on  150/ 150
```

jmartin-tech:
```suggestion
```

Confirmed to maintain current behavior no result should be appended for `None` outputs.

PR #1256, opened 2025-06-13
Llava tests

dchiitmalla:
Add llava tests using mocks, issue: https://github.com/NVIDIA/garak/issues/1105

Checks for
- instantiation
- text generation
- errors

### DIFF
--- a/garak/generators/huggingface.py
+++ b/garak/generators/huggingface.py
@@ -609,7 +609,6 @@ class LLaVA(Generator, HFCompatible):
                 f"Invalid model name {self.name}, current support: {self.supported_models}."
             )
         super().__init__(self.name, config_root=config_root)
-            )
 
         from transformers import LlavaNextProcessor, LlavaNextForConditionalGeneration
 


tests/generators/test_llava.py
```diff
@@ -0,0 +1,105 @@
...
+from garak.exception import ModelNameMissingError
+
+# ─── Constants ─────────────────────────────────────────────────────────
+
+SUPPORTED_MODELS = [
+    "llava-hf/llava-v1.6-34b-hf",
+    "llava-hf/llava-v1.6-vicuna-13b-hf",
+    "llava-hf/llava-v1.6-vicuna-7b-hf",
+    "llava-hf/llava-v1.6-mistral-7b-hf",
+]
```
jmartin-tech: This is not needed, simply reference the class defined value to reduce need to maintain a duplicate list in tests.

```suggestion
SUPPORTED_MODELS = LLaVA.supported_models
```

jmartin-tech (changes requested):
Minor adjustment requested, this could also use a test for behaviour when supplied an unsupported model.

dchiitmalla:
> Minor adjustment requested, this could also use a test for behaviour when supplied an unsupported model.

improved test_llava_unsupported_model and added test_llava_missing_model_name

jmartin-tech (approved):

jmartin-tech (changes requested):
The test failure here shows there is a bug.

Some thought is needed as to if this is an incomplete guard in the based `Generator` class. Where access to `self.name` when an instance is created with `name=""` in the constructor needs to better support the possibility of a generator that does not require `name` to be set or _if that requirement is valid_ as every generator to date has supported expecting `name` suggesting that LLaVA might need to call `self._load_config()` and validate `name` is set before calling `super.__init__()`.

dchiitmalla:
> The test failure here shows there is a bug.
> 
> Some thought is needed as to if this is an incomplete guard in the based `Generator` class. Where access to `self.name` when an instance is created with `name=""` in the constructor needs to better support the possibility of a generator that does not require `name` to be set or _if that requirement is valid_ as every generator to date has supported expecting `name` suggesting that LLaVA might need to call `self._load_config()` and validate `name` is set before calling `super.__init__()`.

Fixed with adding empty name check before calling super().__init__()
It raises a ModelNameMissingError with a helpful message. The approach I implemented aligns with the principle that derived classes should define their specific validation requirements. However, if the requirement for a non-empty name is truly universal across all generators, it would be more architecturally sound to enforce this in the base class.

garak/generators/huggingface.py
```diff
@@ -600,7 +600,11 @@ class LLaVA(Generator, HFCompatible):
...
 
     def __init__(self, name="", config_root=_config):
+        if not name:
+            raise ModelNameMissingError(f"Model name is required. Supported models: {self.supported_models}")
+        
         super().__init__(name, config_root=config_root)
+        
         if self.name not in self.supported_models:
             raise ModelNameMissingError(
                 f"Invalid model name {self.name}, current support: {self.supported_models}."
```
jmartin-tech: Handling for `name` probably needs a larger rework, however to move this PR along, this need to account for `name` being set from `config_root`:
```suggestion
        self._load_config(config_root)
        if name or not hasattr(self, "name"):
            self.name = name

        if self.name not in self.supported_models:
            raise ModelNameMissingError(
                f"Invalid model name {self.name}, current support: {self.supported_models}."
            )
        super().__init__(self.name, config_root=config_root)
```

jmartin-tech (changes requested):

garak/generators/huggingface.py
jmartin-tech: Sorry I tested my suggestion and validated some edge cases needed to be accounted for, latest should test well.

jmartin-tech (approved):

Issue #87, opened 2023-05-31
detector: violence


leondz:
for e.g. `probes.lmrc.Bullying`

4nshumaan:
Hello, can i get this assigned to me?

Nidhi-Satyapriya:
Hi! I’ve gone through the issue and would love to work on it. Could you please assign it to me? I’ll review the relevant code and follow the project’s contribution guidelines. 
Thanks!

PR #1265, opened 2025-06-19
detector: violence detector and lmrc tests

dchiitmalla:
Issue https://github.com/NVIDIA/garak/issues/87
- Adds detector for violence in lmrc
- Adds tests for lmrc detectors

### DIFF
--- a/garak/probes/lmrc.py
+++ b/garak/probes/lmrc.py
@@ -50,10 +50,8 @@ class Bullying(garak.probes.Probe):
 
     lang = "en"
     doc_uri = "https://github.com/leondz/lm_risk_cards/blob/main/cards/en/bullying.md"
-    recommended_detector = [
-        "lmrc.Bullying",
-        "unsafe_content.ToxicCommentModel",
-    ]
+    primary_detector = "unsafe_content.ToxicCommentModel"
+    extended_detectors = [ "lmrc.Bullying" ]
     tags = [
         "risk-cards:lmrc:bullying",
         "avid-effect:ethics:E0301",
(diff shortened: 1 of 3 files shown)


garak/probes/lmrc.py
```diff
@@ -51,6 +51,7 @@ class Bullying(garak.probes.Probe):
     lang = "en"
     doc_uri = "https://github.com/leondz/lm_risk_cards/blob/main/cards/en/bullying.md"
     recommended_detector = [
+        "lmrc.Bullying",
         "unsafe_content.ToxicCommentModel",
     ]
```
jmartin-tech: `recommended_detector ` is deprecated, there should be one `primary` detector:
```suggestion
    primary_detector = "unsafe_content.ToxicCommentModel"
    extended_detectors = [ "lmrc.Bullying" ]
```

jmartin-tech:
The `Bullying` probe already exists and sets the recommended detector to `unsafe_content.ToxicCommentModel`, this new detector is not wired into the probe as provided.  I could see possibly adding this in `extended_detectors` for `probes.lmrc.Bullying` however we should determine if this detector covers any edge cases better than the evaluation using `ToxicCommentModel` as a judge.

jmartin-tech (commented):
The `Bullying` probe already exists and sets the detector to `unsafe_content.ToxicCommentModel`.

I could see possibly adding in `extended_detectors` for `probes.lmrc.Bullying` however we should determine if this detector covers any edge cases better than the evaluation using `ToxicCommentModel` as a judge.

The ask in #87 is for a detector that flags `violence` which is not really the same as `bullying`.

PR #1266, opened 2025-06-20
Improve OpenAI CLI handling

dchiitmalla:
Issue: https://github.com/NVIDIA/garak/issues/552
User Experience Improvements:
- Shows whether the OpenAI API key is set in the environment
- Provides instructions for setting the key if missing
- Lists available models (either from API or fallback lists)
- Displays clear example usage
Comprehensive Test Coverage:
- Added tests in tests/cli/test_openai_model_helper.py to verify behavior with and without API key
- Tests ensure proper display of helper text, model lists, and error handling

### DIFF
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -540,13 +540,52 @@ def main(arguments=None) -> None:
                 # cli provided name overrides config from file
                 conf_root["name"] = _config.plugins.model_name
 
-            # Can this check be deferred to the generator instantiation?
-            if (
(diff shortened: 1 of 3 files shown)


jmartin-tech:
This is not required, api keys are reported missing with the `Configurable._validate_env_var()` method, and listing available models is already handled by `OpenAIGenerator._load_client()`.

Also note `_config` rarely need to not be directly accessed in plugin code as it is not really as `global` as it seems.

PR #1272, opened 2025-06-27
add stale exempt tags

leondz:
`stale` tag will not be applied to `bug`s, `good first issue`s, and issues tagged `dontautoclose`


jmartin-tech:
I think `bug` may get removed from this list, but can stay for now.  Bugs that sit around with no comments often have either been addressed inadvertently or involve edgecase scenarios.

leondz:
Yeah. I'm also on the fence about `new plugin` because they rarely get accidentally implemented, but left it on the stale-able list because some are two years old now.

PR #1273, opened 2025-06-27
Added total latency logging for each attempt

sammykao:
## Total Latency Logging for each attempt

This PR adds latency measurement for each model generation attempt in Garak. The measured latency (in seconds) is now recorded in the `notes` field of each `Attempt` entry in the report `.jsonl` output.

---

** Where**
- Modified `garak/probes/base.py`
- Timing logic added around the generator's `generate` method inside `_execute_attempt`
- Latency is stored as:
  ```json
  "notes": {
    "latency": <float_seconds>
  }
  ```

**How**
- Latency is measured using wall-clock time from just before the model is called to just after the response is received.
- Recorded as a float in seconds.

**Why**
- Enables performance benchmarking and response time analysis.
- Helps identify:
  - Slow probes
  - Generator inefficiencies
  - Infrastructure bottlenecks

**Example Output**
```json
{
  "entry_type": "attempt",
  ...
  "notes": {
    "latency": 0.12345
  },
  ...
}
```

---

## Additional Notes

- Measures total latency per attempt (including multi-turn sequences).
- Fully backward-compatible: no changes made to the report format or analysis scripts.

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -12,6 +12,7 @@ import logging
 from collections.abc import Iterable
 import random
 from typing import Iterable, Union
+import time
 
 from colorama import Fore, Style
 import tqdm
@@ -202,9 +203,13 @@ class Probe(Configurable):
     def _execute_attempt(self, this_attempt):
         """handles sending an attempt to the generator, postprocessing, and logging"""
         self._generator_precall_hook(self.generator, this_attempt)
+        start_time = time.time()
         this_attempt.outputs = self.generator.generate(
             this_attempt.prompt, generations_this_call=self.generations
         )
+        end_time = time.time()
+        latency = end_time - start_time
+        this_attempt.notes["total_time"] = latency
         if self.post_buff_hook:
             this_attempt = self._postprocess_buff(this_attempt)
         this_attempt = self._postprocess_hook(this_attempt)


sammykao:

garak/probes/base.py
```diff
@@ -202,9 +203,13 @@ def _mint_attempt(
     def _execute_attempt(self, this_attempt):
         """handles sending an attempt to the generator, postprocessing, and logging"""
         self._generator_precall_hook(self.generator, this_attempt)
+        start_time = time.time()
         this_attempt.outputs = self.generator.generate(
             this_attempt.prompt, generations_this_call=self.generations
         )
+        end_time = time.time()
+        latency = end_time - start_time
+        this_attempt.notes["total_time"] = latency
```
jmartin-tech, line 212: Serialization of time difference should likely be formatted to ensure a clear unit of measure.

jmartin-tech (changes requested):
This is an interesting enhancement, is there a specific need this is fulfilling?  While timing of each inference is interesting I am not sure this will offer the desired values as  most generators implement backoff functions to account for errors and rate limits.

As implemented a generator that needs to retry or backoff would store a latency value that may not accurately represent the time spent on a single inference, and in the case of generators that produce multiple generations with repeated calls implemented inside `generate()` this would not offer how that time was allocated.

Tests also need to account for the timing in a controlled way, this is often done by mocking the calls for `time.time()` in tests where it could have impact.

PR #1280, opened 2025-07-02
feature: support detectors including `None` in output values

leondz:
Addresses part of #756 

Detectors have been bound to returning `List[float]`. Sometimes detection will not be possible (e.g. perhaps if a detector input is `None`). Prior behaviour was to ignore invalid attempt content when doing detection. This has two negative side effects:

1. Obfuscation of probe skips
2. Alignment between detector input & output is broken

This change enables detectors to include `None`s in their output list, enabling alignment between input and output, and making reporting of skipped generation/probe instances easier.


tests/detectors/test_detectors_base.py
```diff
@@ -106,8 +106,8 @@ def test_none_outputs(string_detector):
     attempt.outputs = [None, TEST_STRINGS[0], None]
 
     results = string_detector.detect(attempt, case_sensitive=True)
-    assert len(results) == 1, "Detector expected to skip None outputs"
-    assert results == [1.0], "Failed to handle None outputs correctly"
+    assert len(results) == 3, "Detector expected to not skip None outputs"
```
jmartin-tech:
```suggestion
    assert len(results) == len(attempt.outputs), "Detector result for each output including None"
```

jmartin-tech (commented):
Minor test assertion enhancement.

tests/detectors/test_detectors_base.py
leondz: Oh, OK, yes this is now reasonable

jmartin-tech:
Validation testing show this currently expands the impact of #1127 in terms of the current `atkgen.Tox` results, changes in #1254 resolve the issue and this PR can move forward after the resolution is in place.

jmartin-tech (approved):
Testing looks good after latest merge from `main`.

Note latest merge required expansion of `detectors/judge.py` refactor to remove suppression of `None` responses in that detector.

PR #1289, opened 2025-07-09
Feature/translation cache

SnowMasaya:
This PR implements an enhanced translation caching system for Garak's language providers. The changes include:

## Feature Enhancements

- Enhanced cache storage to include original text alongside translation for better debugging and data integrity
- Added get_cache_entry() method to retrieve full cache entries with metadata
- Implemented automatic caching across all translation providers (local and remote)
- Added backward compatibility with old cache format

## Integration

- Updated all translation classes (`Passthru`, `LocalHFTranslator`, `RivaTranslator`, `DeeplTranslator`, `GoogleTranslator`) to use the caching system
- Added `_translate_with_cache()` and `_translate_impl()` methods for consistent caching behavior
- Maintained existing error handling and retry logic for remote services

## Benefits
- **Performance**: Significantly reduces translation time for repeated text
- **Cost savings**: Reduces API calls to paid services like DeepL, Google Cloud Translation, and NVIDIA Riva
- **Reliability**: Provides fallback for offline scenarios when cached translations are available
- **Consistency**: Ensures identical translations for the same input text across different runs

## Verification

- [x] Run the tests and ensure they pass: python -m pytest tests/langservice/test_translation_cache.py
- [x] Run integration tests: python -m pytest tests/langservice/test_translation_cache_integration.py
- [x] Verify cache files are created in ~/.cache/garak/translation/ with correct naming pattern
- [x] Document the caching system in docs/source/translation.rst

## Cache File Verification
- [x] Check cache file naming: translation_cache_{source_lang}_{target_lang}_{model_type}.json

## Error Handling Verification
- [x] Test with corrupted cache files
- [x] Test with missing API keys
- [x] Test with invalid language pairs
- [x] Verify graceful fallback behavior

The caching system is transparent to users and requires no additional configuration. It automatically activates when translation services are used and provides significant benefits for performance and cost reduction.

### DIFF
--- a/docs/source/translation.rst
+++ b/docs/source/translation.rst
@@ -13,6 +13,41 @@ Limitations
 - If probes or detectors fail to load, you need may need to choose a smaller local translation model or utilize a remote service.
 - Translation may add significant execution time to the run depending on resources available.
 
+Translation Caching
+------------------
+
+Garak implements a translation caching system to improve performance and reduce API costs when using translation services. The caching mechanism automatically stores and retrieves translation results to avoid redundant API calls.
+
+**How it works:**
+
+- Each translation pair (source language → target language) and model type combination gets its own cache file
+- Cache files are stored in JSON format under the cache directory: ``{cache_dir}/translation/translation_cache_{source_lang}_{target_lang}_{model_type}.json``
+- Translation results are keyed by MD5 hash of the input text for efficient storage and retrieval
+- Cache files persist between runs, allowing translations to be reused across multiple garak sessions
+
+**Benefits:**
+
+- **Performance**: Significantly reduces translation time for repeated text
+- **Cost savings**: Reduces API calls to paid services like DeepL, Google Cloud Translation, and NVIDIA Riva
+- **Reliability**: Provides fallback for offline scenarios when cached translations are available
+- **Consistency**: Ensures identical translations for the same input text across different runs
+
+**Cache management:**
+
+- Cache files are automatically created when translations are performed
+- Corrupted cache files are handled gracefully with fallback to empty cache
+- Cache files can be manually deleted to force fresh translations
+- Cache directory location follows garak's standard cache configuration
+
+**Supported for all translation services:**
+
+- Local translation models (Hugging Face)
+- DeepL API
+- NVIDIA Riva API  
+- Google Cloud Translation API
+
+The caching system is transparent to users and requires no additional configuration. It automatically activates when translation services are used.
+
 Supported translation services
 ------------------------------
 
--- a/garak/langproviders/base.py
+++ b/garak/langproviders/base.py
@@ -10,8 +10,13 @@ import re
 import unicodedata
 import string
 import logging
+import json
+import hashlib
+import os
+from pathlib import Path
 from garak.resources.api import nltk
 from langdetect import detect, DetectorFactory, LangDetectException
+from garak import _config
 
 _intialized_words = False
 
@@ -136,6 +141,95 @@ def is_meaning_string(text: str) -> bool:
 from garak.configurable import Configurable
 
 
+class TranslationCache:
+    def __init__(self, config_root: dict = None):
+        # Handle fallback for test configs if config_root is provided
+        self.source_lang = "en"
+        self.target_lang = "ja"
+        self.model_type = "unknown"
+
+        if config_root and isinstance(config_root, dict):
+            lang_cfg = list(config_root.get("langproviders", {}).values())[0]
+            self.source_lang = lang_cfg.get("language", "en,ja").split(",")[0]
+            self.target_lang = lang_cfg.get("language", "en,ja").split(",")[1]
+            self.model_type = lang_cfg.get("model_type", "unknown")
+
+        cache_dir = _config.transient.cache_dir / "translation"
+        cache_dir.mkdir(mode=0o740, parents=True, exist_ok=True)
+        cache_filename = (
+            f"translation_cache_{self.source_lang}_{self.target_lang}_{self.model_type}.json"
+        )
+        self.cache_file = cache_dir / cache_filename
+        logging.info(f"Cache file: {self.cache_file}")
+        self._cache = self._load_cache()
+
+    def _load_cache(self) -> dict:
+        if self.cache_file.exists():
+            try:
+                with open(self.cache_file, "r", encoding="utf-8") as f:
+                    return json.load(f)
+            except (json.JSONDecodeError, IOError) as e:
+                logging.warning(f"Failed to load translation cache: {e}")
+                return {}
+        return {}
+
+    def _save_cache(self):
+        try:
+            with open(self.cache_file, "w", encoding="utf-8") as f:
+                json.dump(self._cache, f, ensure_ascii=False, indent=2)
+        except IOError as e:
+            logging.warning(f"Failed to save translation cache: {e}")
+
+    def get_cache_key(self, text: str) -> str:
+        return hashlib.md5(text.encode("utf-8")).hexdigest()
+
+    def get(self, text: str) -> str | None:
+        cache_key = self.get_cache_key(text)
+        cache_entry = self._cache.get(cache_key)
+        if cache_entry and isinstance(cache_entry, dict):
+            return cache_entry.get("translation")
+        elif isinstance(cache_entry, str):
+            # Backward compatibility with old format
+            return cache_entry
+        return None
+
+    def set(self, text: str, translation: str):
+        cache_key = self.get_cache_key(text)
+        self._cache[cache_key] = {
+            "original": text,
+            "translation": translation,
+            "source_lang": self.source_lang,
+            "target_lang": self.target_lang,
+            "model_type": self.model_type
+        }
+        self._save_cache()
+
+    @property
+    def cache(self):
+        return self._cache
+
+    @property
+    def cache_file_path(self):
+        return self.cache_file
+
+    def get_cache_entry(self, text: str) -> dict | None:
+        """Get full cache entry including original text and metadata."""
+        cache_key = self.get_cache_key(text)
+        cache_entry = self._cache.get(cache_key)
+        if cache_entry and isinstance(cache_entry, dict):
+            return cache_entry
+        elif isinstance(cache_entry, str):
+            # Backward compatibility with old format
+            return {
+                "original": text,
+                "translation": cache_entry,
+                "source_lang": self.source_lang,
+                "target_lang": self.target_lang,
+                "model_type": self.model_type
+            }
+        return None
+
+
 class LangProvider(Configurable):
     """Base class for objects that provision language"""
 
@@ -147,6 +241,9 @@ class LangProvider(Configurable):
 
         self._validate_env_var()
 
+        # Use TranslationCache for caching
+        self.cache = TranslationCache(config_root)
+
         self._load_langprovider()
 
     def _load_langprovider(self):
@@ -154,6 +251,16 @@ class LangProvider(Configurable):
 
     def _translate(self, text: str) -> str:
         raise NotImplementedError
+
+    def _translate_with_cache(self, text: str) -> str:
+        """Translate text with caching support."""
+        cached_translation = self.cache.get(text)
+        if cached_translation is not None:
+            logging.debug(f"Using cached translation for text: {text[:50]}...")
+            return cached_translation
+        translation = self._translate_impl(text)
+        self.cache.set(text, translation)
+        return translation
 
     def _get_response(self, input_text: str):
         translated_lines = []
@@ -189,7 +296,7 @@ class LangProvider(Configurable):
         if needs_translation:
             cleaned_line = self._clean_line(line)
             if cleaned_line:
-                translated_line = self._translate(cleaned_line)
+                translated_line = self._translate_with_cache(cleaned_line)
                 translated_lines.append(translated_line)
 
         return translated_lines
@@ -202,7 +309,7 @@ class LangProvider(Configurable):
             if self._should_skip_line(cleaned_sentence):
                 translated_lines.append(cleaned_sentence)
                 continue
-            translated_line = self._translate(cleaned_sentence)
+            translated_line = self._translate_with_cache(cleaned_sentence)
             translated_lines.append(translated_line)
 
         return translated_lines
--- a/garak/langproviders/local.py
+++ b/garak/langproviders/local.py
@@ -15,10 +15,18 @@ from garak.resources.api.huggingface import HFCompatible
 class Passthru(LangProvider):
     """Stand-in language provision for pass through / noop"""
 
+    def __init__(self, config_root: dict = {}) -> None:
+        super().__init__(config_root=config_root)
+
     def _load_langprovider(self):
         pass
 
     def _translate(self, text: str) -> str:
+        # Use _translate_with_cache to enable caching
+        return self._translate_with_cache(text)
+
+    def _translate_impl(self, text: str) -> str:
+        """Actual translation implementation without caching."""
         return text
 
     def get_text(
@@ -110,6 +118,11 @@ class LocalHFTranslator(LangProvider, HFCompatible):
             self.tokenizer = MarianTokenizer.from_pretrained(model_name)
 
     def _translate(self, text: str) -> str:
+        # Use _translate_with_cache to enable caching
+        return self._translate_with_cache(text)
+
+    def _translate_impl(self, text: str) -> str:
+        """Actual translation implementation without caching."""
         if "m2m100" in self.model_name:
             self.tokenizer.src_lang = self.source_lang
 
--- a/tests/langservice/test_translation_cache.py
+++ b/tests/langservice/test_translation_cache.py
@@ -1,0 +1,395 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import tempfile
+import json
+import os
+from pathlib import Path
+from unittest.mock import patch, MagicMock
+
+from garak.langproviders.base import LangProvider, TranslationCache
+from garak.langproviders.local import Passthru
+
+
+class TestTranslationCache:
+    """Test translation caching functionality."""
+
+    @pytest.fixture
+    def temp_cache_dir(self):
+        """Create a temporary cache directory for testing."""
+        with tempfile.TemporaryDirectory() as temp_dir:
+            yield Path(temp_dir)
+
+    @pytest.fixture
+    def mock_config(self):
+        """Mock configuration for testing."""
+        return {
+            "langproviders": {"passthru": {"language": "en,ja", "model_type": "test"}}
+        }
+
+    def test_cache_initialization(self, temp_cache_dir, mock_config):
+        """Test that cache is properly initialized."""
+        with patch("garak._config.transient.cache_dir", temp_cache_dir):
+            # Create a test-specific subclass that properly initializes
+            class TestPassthru(Passthru):
+                def __init__(self, config_root={}):
+                    # Set language before calling parent __init__
+                    self.language = "en,ja"
+                    self.model_type = "test"
+                    super().__init__(config_root)
+
+            translator = TestPassthru(config_root=mock_config)
+
+            # Check that cache directory was created
+            cache_dir = temp_cache_dir / "translation"
+            assert cache_dir.exists()
+
+            # Check that cache file was created with correct name
+            expected_cache_file = cache_dir / "translation_cache_en_ja_test.json"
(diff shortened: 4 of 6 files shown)


garak/langproviders/local.py
```diff
@@ -15,10 +15,18 @@
 class Passthru(LangProvider):
     """Stand-in language provision for pass through / noop"""
 
+    def __init__(self, config_root: dict = {}) -> None:
+        super().__init__(config_root=config_root)
+
```
jmartin-tech: Revisions to `Passthru` are not needed, there is no reason to interact with a cache when always returning the original value.
```suggestion
```

garak/langproviders/local.py
```diff
@@ -15,10 +15,18 @@
...
     def _load_langprovider(self):
         pass
 
     def _translate(self, text: str) -> str:
+        # Use _translate_with_cache to enable caching
+        return self._translate_with_cache(text)
+
+    def _translate_impl(self, text: str) -> str:
+        """Actual translation implementation without caching."""
```
jmartin-tech: Revisions to `Passthru` are not needed, there is no reason to interact with a cache when always returning the original value.
```suggestion
```

tests/langservice/test_translation_cache_integration.py
```diff
@@ -0,0 +1,129 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jmartin-tech: Nit pick:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

garak/langproviders/base.py
```diff
@@ -136,6 +141,95 @@ def is_meaning_string(text: str) -> bool:
...
+        }
+        self._save_cache()
+
+    @property
+    def cache(self):
+        return self._cache
+
+    @property
+    def cache_file_path(self):
+        return self.cache_file
```
jmartin-tech: These should not need to be exposed, the internally held values should not be accessed publicly, tests can evaluate the private objects.

garak/langproviders/base.py
```diff
@@ -136,6 +141,95 @@ def is_meaning_string(text: str) -> bool:
...
+
+    def _save_cache(self):
+        try:
+            with open(self.cache_file, "w", encoding="utf-8") as f:
+                json.dump(self._cache, f, ensure_ascii=False, indent=2)
+        except IOError as e:
+            logging.warning(f"Failed to save translation cache: {e}")
+
+    def get_cache_key(self, text: str) -> str:
+        return hashlib.md5(text.encode("utf-8")).hexdigest()
```
jmartin-tech: Any usage or a hash function needs to clearly indicate if the result ever supports a security function.
```suggestion
        return hashlib.md5(text.encode("utf-8"), usedforsecurity=False).hexdigest()
```

tests/langservice/test_translation_cache.py
```diff
@@ -0,0 +1,395 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jmartin-tech: Nit pick:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

tests/langservice/test_translation_cache.py
```diff
@@ -0,0 +1,395 @@
...
+
+import pytest
+import tempfile
+import json
+import os
+from pathlib import Path
+from unittest.mock import patch, MagicMock
+
+from garak.langproviders.base import LangProvider, TranslationCache
+from garak.langproviders.local import Passthru
```
jmartin-tech: The `PassThru` class should not utilize a cache, please refactor these tests to unit test other `langproviders`, mocking the `_translate()` function for any other class extending `garak.langproviders.base` should be viable.

This applies to both test files added in the PR.

docs/source/translation.rst
```diff
@@ -13,6 +13,41 @@ Limitations
...
 
+Translation Caching
+------------------
+
+Garak implements a translation caching system to improve performance and reduce API costs when using translation services. The caching mechanism automatically stores and retrieves translation results to avoid redundant API calls.
+
+**How it works:**
+
+- Each translation pair (source language → target language) and model type combination gets its own cache file
+- Cache files are stored in JSON format under the cache directory: ``{cache_dir}/translation/translation_cache_{source_lang}_{target_lang}_{model_type}.json``
```
jmartin-tech: Should `model_type` be differentiated here? Note that the `model_type` for local model translation or self hosted remote translation is not always a complete indicator that the same translation source is in use.

I can see being specific initially and adding a future enhancement to check if for a previously existing cache file indicating just the language pair as something enabling support for user provided translation files that may contain human translated entries.

jmartin-tech (dismissed):
Awesome addition, this will be very useful and gets the project one step closer to being able to support user provided human translation as well.

A few change are requested, and I added a few notes for possible future proofing.

docs/source/translation.rst
SnowMasaya: OK. I remove it. i use the following file name
``{cache_dir}/translation/translation_cache_{source_lang}_{target_lang}.json``

But user would like to compare the translation quality.
I change the make hash key and add model_name for comparing translation quality each service.

```py
               
   def get_cache_key(self, text: str) -> str:
       # Create a combined string with all parameters that affect translation
       combined_string = f"{text}|{self.source_lang}|{self.target_lang}|{self.model_type}|{self.model_name}"
       return hashlib.md5(combined_string.encode("utf-8"), usedforsecurity=False).hexdigest()

    def set(self, text: str, translation: str):
        cache_key = self.get_cache_key(text)
        self._cache[cache_key] = {
            "original": text,
            "translation": translation,
            "source_lang": self.source_lang,
            "target_lang": self.target_lang,
            "model_type": self.model_type,
            "model_name": self.model_name
        }
        self._save_cache()
```

jmartin-tech: This would be functionally equivalent as the hash key being unique to the langprovider means a run not using that exact provider would never find the value in cache.

I would prefer the existing filename over embedding the provider in the lookup key. I believe the key should be the same for any provider, this is as much for the use case of comparison of different provider translations as it would be for being able to provide human translation override.

SnowMasaya: OK. The local case needs model name.

I will change the following for setting up the model name.

```py
def __init__(self, config_root: dict = None):
    self.source_lang = "en"
    self.target_lang = "ja"
    self.model_type = "unknown"
    self.model_name = "default"
    if config_root and isinstance(config_root, dict):
        lang_cfg = list(config_root.get("langproviders", {}).values())[0]
        self.source_lang = lang_cfg.get("language", "en,ja").split(",")[0]
        self.target_lang = lang_cfg.get("language", "en,ja").split(",")[1]
        self.model_type = lang_cfg.get("model_type", "unknown")
        if self.model_type == "local":
            self.model_name = lang_cfg.get("model_name", "default")
    :
    :
    cache_filename = (
        f"translation_cache_{self.source_lang}_{self.target_lang}_{self.model_type}_{self.model_name}.json"
    )

```

jmartin-tech: The cache class should not be instantiated accept by a `LangProvider` or tests consider:
``` Python
    def __init__(self, provider: LangProvider):
        self.source_lang = provider.source_lang
        self.target_lang = provider.target_lang
        self.model_type = provider.model_type
        self.model_name = "default"
        if hasattr(provider, "model_name")
            self.model_name =provider.model_name

        cache_dir = _config.transient.cache_dir / "translation"
        cache_dir.mkdir(mode=0o740, parents=True, exist_ok=True)
        cache_filename = (
            f"translation_cache_{self.source_lang}_{self.target_lang}_{self.model_type}_{self.model_name}.json"
        )
        self.cache_file = cache_dir / cache_filename
        logging.info(f"Cache file: {self.cache_file}")
        self._cache = self._load_cache()
```

garak/langproviders/base.py
```diff
@@ -155,6 +243,16 @@ def _load_langprovider(self):
     def _translate(self, text: str) -> str:
         raise NotImplementedError
 
+    def _translate_with_cache(self, text: str) -> str:
+        """Translate text with caching support."""
+        cached_translation = self.cache.get(text)
+        if cached_translation is not None:
+            logging.debug(f"Using cached translation for text: {text[:50]}...")
```
jmartin-tech, line 250: This should not log the original text, these values can source from prompts or LLM response and could produce escape sequences that could impact something that is monitoring the logs.

```suggestion
```

I could see however keeping a tracker in the service for cache hit/miss rates. While not something we need to expose at this time having the metric values in the runtime could be helpful during debugging scenarios.

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+
+    def _save_cache(self):
+        try:
+            with open(self.cache_file, "w", encoding="utf-8") as f:
+                json.dump(self._cache, f, ensure_ascii=False, indent=2)
+        except IOError as e:
+            logging.warning(f"Failed to save translation cache: {e}")
+
+    def get_cache_key(self, text: str) -> str:
+        return hashlib.md5(text.encode("utf-8"), usedforsecurity=False).hexdigest()
```
erickgalinkin, line 181: This seems like potentially a lot of overhead. Approach is, IMO, fine for now but perhaps we should consider partial hashing or something? Given potentially many thousands of strings that are often very long, we could consider a faster cache lookup approach in the future.

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+
+    def get(self, text: str) -> str | None:
+        cache_key = self.get_cache_key(text)
+        cache_entry = self._cache.get(cache_key)
```
erickgalinkin, line 185: Since we're going to have a ton of cache misses on first invocation, does it make sense for us to have an easier way to avoid cache misses? Maybe just a check if the cache is empty?

jmartin-tech (commented):
One more minor revision.

This looks pretty usable as an iteration towards more efficient additional language support.

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+            with open(self.cache_file, "w", encoding="utf-8") as f:
+                json.dump(self._cache, f, ensure_ascii=False, indent=2)
+        except IOError as e:
+            logging.warning(f"Failed to save translation cache: {e}")
+
+    def get_cache_key(self, text: str) -> str:
+        return hashlib.md5(text.encode("utf-8"), usedforsecurity=False).hexdigest()
+
+    def get(self, text: str) -> str | None:
+        cache_key = self.get_cache_key(text)
```
erickgalinkin, line 184:
```suggestion
        if not len(self._cache):
            return None
        cache_key = self.get_cache_key(text)
```
Would save us a ton of hashing. At some point in the future, we could even add hierarchy and do it by probe or something so that we simply avoid generating a ton of hashes to look up in an empty cache for no reason.

garak/langproviders/base.py
jmartin-tech: The cache is built in real-time and many of the probes result in hits based on responses very quickly in the first run.  These checks that result in miss even in the first run seem worth the cost IMO.

garak/langproviders/base.py
jmartin-tech: The current implementation is not caching full prompts or responses but each `sentence`, I do think in a future revision we will see some optimization to add here.

garak/langproviders/base.py
SnowMasaya: Most case is short sentence. If we support long sentence, we consider about "partial hashing" in the future.

docs/source/translation.rst
```diff
@@ -13,6 +13,41 @@ Limitations
...
 
+Translation Caching
+------------------
+
+Garak implements a translation caching system to improve performance and reduce API costs when using translation services. The caching mechanism automatically stores and retrieves translation results to avoid redundant API calls.
+
+**How it works:**
+
+- Each translation pair (source language → target language) gets its own cache file
+- Cache files are stored in JSON format under the cache directory: ``{cache_dir}/translation/translation_cache_{source_lang}_{target_lang}_{model_type}_{model_name}.json``
```
leondz, line 24: do model_type and model_name refer to the translator model? what do API translates look like?

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
 #       "hf_args": {} # or any other translator specific values for the model_type
 #     }
 # }
 from garak.configurable import Configurable
 
 
+class TranslationCache:
+    def __init__(self, provider: "LangProvider"):
+        if not hasattr(provider, "model_type"):
+            return None  # providers without a model_type do not have a cache
```
leondz, line 147: why? is this expressing a non-explicit convention in LangProvider naming? if so would prefer this to be expressed explicitly

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+        self.source_lang = provider.source_lang
+        self.target_lang = provider.target_lang
+        self.model_type = provider.model_type
+        self.model_name = "default"
+        if hasattr(provider, "model_name"):
+            self.model_name = provider.model_name
+
+        cache_dir = _config.transient.cache_dir / "translation"
+        cache_dir.mkdir(mode=0o740, parents=True, exist_ok=True)
+        cache_filename = f"translation_cache_{self.source_lang}_{self.target_lang}_{self.model_type}_{self.model_name.replace('/', '_')}.json"
```
leondz, line 158: Prefer [dbm](https://docs.python.org/3/library/dbm.html) backend for this vs. a large inflight dict with occasional serdes (i guess that word kinda fits here?) action. Can be a fast follow

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+        self.cache_file = cache_dir / cache_filename
+        logging.info(f"Cache file: {self.cache_file}")
```
leondz, line 160: needs verb + qualification
```suggestion
        logging.info(f"Loading translation cache file: {self.cache_file}")
```

garak/langproviders/base.py
leondz: I thought this and measured `hashlib.md5` speed. It seems pretty fine

Alternative is to use the raw string as key and rely on the underlying index's hashing (e.g. dict keys). This halves the amount of hashing. Might bump into length limits though ¯\_ (ツ)_/¯

garak/langproviders/base.py
leondz: +1 for skipping on empty cache

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+
+    def get(self, text: str) -> str | None:
+        cache_key = self.get_cache_key(text)
+        cache_entry = self._cache.get(cache_key)
+        if cache_entry and isinstance(cache_entry, dict):
+            return cache_entry.get("translation")
+        elif isinstance(cache_entry, str):
+            # Backward compatibility with old format
+            return cache_entry
+        return None
```
leondz, line 191: Why is backward compat for "old format" present in a PR introducing a new feature for the first time? 

I'm not quite happy with the cache format having dict entries - this is overkill.

If we're specifying in the cache key (i.e. filename) the following:
* src lang code
* dst lang code
* langprovider

Then we don't need anything more than a direct key/value store here, where key is src language and value is dst language.

I guess the database engine in this PR is an inflight dict that gets JSON serialised/deserialised. This can ditch hashing and use the src text as lookup value - that'll ease troubleshooting and reduce hashing.

An alternative is a dbm engine with the same src/dst key value

Is there a requirement for any more info than this? We have create/modify/access stamps as the database file object metadata. I don't think we have size constraints so timestamping at the level of individual cache entries seems overkill. I guess if we want a stable format we might set the value as a dict where one value is the text to be retrieved, but this seems overkill.

Happy to include tooling for migrating DBs between formats, providing external routes for people to move to the current garak setup like we do with `fixer`s, but I don't want to support multiple cache format _readers_ as core functionality.

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+
+    def set(self, text: str, translation: str):
+        cache_key = self.get_cache_key(text)
+        self._cache[cache_key] = {
+            "original": text,
+            "translation": translation,
+            "source_lang": self.source_lang,
+            "target_lang": self.target_lang,
+            "model_type": self.model_type,
+            "model_name": self.model_name,
```
leondz, line 201: this info is redundant - it's logged in the cache object's filename - suggest it is deleted

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+        if cache_entry and isinstance(cache_entry, dict):
+            return cache_entry.get("translation")
+        elif isinstance(cache_entry, str):
+            # Backward compatibility with old format
+            return cache_entry
+        return None
+
+    def set(self, text: str, translation: str):
+        cache_key = self.get_cache_key(text)
+        self._cache[cache_key] = {
```
leondz, line 195: Let's just use source text itself as the key

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+            "original": text,
+            "translation": translation,
+            "source_lang": self.source_lang,
+            "target_lang": self.target_lang,
+            "model_type": self.model_type,
+            "model_name": self.model_name,
+        }
+        self._save_cache()
```
leondz, line 203: this means a Lot of serialisation and saving!

garak/langproviders/base.py
```diff
@@ -128,14 +133,94 @@ def is_meaning_string(text: str) -> bool:
...
+            # Backward compatibility with old format
+            return {
+                "original": text,
+                "translation": cache_entry,
+                "source_lang": self.source_lang,
+                "target_lang": self.target_lang,
+                "model_type": self.model_type,
+                "model_name": self.model_name,
+            }
+        return None
```
leondz, line 221: Let's not have backward compatibility in the first PR landing a feature

Let's simplify the cache entires by removing information we have elsewhere

Let's index on the text directly, halving the amount of hashing done

```suggestion
    def get_cache_entry(self, text: str) -> str | None:
        return self._cache.get(text)
```

garak/langproviders/base.py
```diff
@@ -155,6 +243,16 @@ def _load_langprovider(self):
...
         raise NotImplementedError
 
+    def _translate_with_cache(self, text: str) -> str:
+        """Translate text with caching support."""
+        cached_translation = self.cache.get(text)
+        if cached_translation is not None:
+            logging.debug(f"Using cached translation for text: {text[:50]}...")
+            return cached_translation
+        translation = self._translate_impl(text)
+        self.cache.set(text, translation)
```
leondz, line 253: Can we move to a direct src->dst cache? i.e.

```suggestion
        self.cache.[text] = translation
```

leondz (changes requested):
It is connected to the right places and the general form is good.

Would prefer a much simpler cache with minimal duplication, where the cache is a simple keyed object that uses the source text and the key and has the translated text as the value. A single dict can do this.

garak/langproviders/base.py
jmartin-tech: This is checking an internal detail of the implemented object, we could add some other defining factor, however at this time only the `PassThru` class meets such criteria.

garak/langproviders/base.py
jmartin-tech: Would this be guaranteed to be consistently viable?  Using a hash provides a known consistent format that will also be cross platform compatible (as no OS specific escaping of the alphanumeric hash strings would apply). While in theory all python versions should use consistent internal representations of a strings that assumption is not a guaranteed and lines in a prompt could contain something _odd_, in fact the this is a core expectation as we test the edges of things.

garak/langproviders/base.py
leondz: We should add some other defining factor. I can open an issue or we can do it here. What would you prefer? A class attrib?

garak/langproviders/base.py
leondz: It's as consistently viable as the underlying object. Python `dict` requires that keys are hashable. Python `str` is a hashable type. Max `str` length is predicated on arch; for 32-bit it's ~2.1GB.

So yes, consistently viable. Not universally viable but we will have Other Problems (and many of them) before it breaks.

Python already hashes the keys we're using (strings) and does all the hard work putting them into hashed lookups (dicts). No need to duplicate that work.

PR #1291, opened 2025-07-10
Generator that allows for multiple requests

iamnotcj:
This change adds support for sending multiple request/response pairs as well as websockets. This was influenced by a couple of web app implementations our practice has seen over the past couple of months.

### MultiRestGenerator:
There are times where the initial request with the user input doesn't immediately return a generated output. Instead another request has to be made to a specified endpoint to get the results of the input. For example, one implementation we have come across had the user send an input and a query id was returned. The succeeding request used this query id as a value for a GET parameter. The response to this GET request then contained the generated output. 

It also relies on Burpsuite XML files, as Burpsuite is one of the most popular web application testing tool. This might be a stepping stone into adding garak as a dedicated burp extension some time in the future?

### WebSocketGenerator:
There have been times where chatbots would communicate with websockets. I've only encountered pretty basic implementations of this, which is why this one is pretty rudimentary

### DIFF
--- a/garak/generators/rest.py
+++ b/garak/generators/rest.py
@@ -308,8 +308,6 @@ class RestGenerator(Generator):
         return response
 
 
-
-
 class MultiRestGenerator(Generator):
     """
     This is a generator to work with multiple request-response pairs using Burp Suite's exported XML format. It supports variable substitution, in case you need to build succeeding requests uponinformation from previous request/response pairs.
@@ -647,38 +645,4 @@ class MultiRestGenerator(Generator):
         return ["".join(self.run(prompt))]
 
 
-class WebSocketGenerator(Generator):
-    """
-    This is a generator to work with websockets
-    """
-
-    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
-        "uri": None,
-        "auth_key": None,
-        "body": "{}",
-        "json_response": True,
-        "json_key": "output",
-    }
-
-    def __init__(self, name="WebSocket", config_root=_config):
-        super().__init__(name, config_root=config_root)
-
-    def json_handler(self, data):
-        response_json = json.loads(data)
-        return json.dumps(response_json[self.json_key])
-
-    def request(self, payload):
-        with connect(self.uri) as websocket:
-            websocket.send(self.body.replace("$INPUT", payload))
-            message = websocket.recv()
-            return self.json_handler(message) if self.json_response == True else message
-
-    def _call_model(
-        self, prompt: str, generations_this_call: int = 1
-    ) -> List[Union[str, None]]:
-        if output := self.request(self, prompt) == dict:
-            return output[self.json_key]
-        else:
-            return output
-            
 DEFAULT_CLASS = "RestGenerator"


tests/generators/test_multirest.py
```diff
@@ -0,0 +1,104 @@
+import pytest
+from garak.generators.multirest import MultiRestGenerator
+
+BURP_XML_PATH = "./tools/rest/multirest.xml"
```
jmartin-tech: Support resource files are placed in `garak/data` to allow for user global override via XDG support, in this case if the test relies on the package included version of the file you may want to wrap access to the data_path location in a standard `Path` object to ensure only the packaged version is used for tests.

```suggestion
from pathlib import Path
from garak.data import path as data_path
            
BURP_XML_PATH = Path(data_path) / "rest" / "multirest.xml"
```

tools/rest/multirest.xml
jmartin-tech, line 1: Please move this to `garak/data/rest/multirest.xml` or maybe `garak/data/generators/rest/multirest.xml` as it is a data resource file.

garak/generators/multirest.py
```diff
@@ -0,0 +1,261 @@
...
+    If the response is not JSON, it will be returned as is.
+    """
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "reqresp_pairs": [],
+        "variables": {},
+        "burpfile": "./tools/rest/multirest.xml",
+    }
+
+    def __init__(self, name="MultiRestGenerator", config_root=_config):
+        super().__init__(name, config_root=config_root)
```
jmartin-tech: Files provided in the package should in in `garak/data` with pathing matching the package they support.  This allows for hierarchal injection of support file using the `XDG_DATA_HOME/garak/` path for user global replacement.

With such a change this value might be set to `None` to retrieve the standard `data` location file and still allow configuration based override.

```suggestion
        "burpfile": None,
    }

    def __init__(self, name="MultiRestGenerator", config_root=_config):
        super().__init__(name, config_root=config_root)
        if not self.burpfile:
            from garak.data import path as data_path
            
            self.burpfile = data_path / "rest" / "multirest.xml"
```

garak/generators/multirest.py
jmartin-tech, line 1: (Optional) Consider making this another class in the `rest` generator package.

Invoked as `-m rest.MultiRestGenerator`

garak/generators/wsocket.py
jmartin-tech, line 1: (Optional) this also may make sense as a class in the `rest` generator package.

Invoked as `-m rest.WebSocketGenerator `

garak/generators/wsocket.py
```diff
@@ -0,0 +1,47 @@
+import json
+import logging
+import re
+
+from typing import List, Union
+from garak.generators.base import Generator
+from garak import _config
+
+import websockets
+from websockets.sync.client import connect
```
jmartin-tech: As noted by test automation, these would need to be added to `pyproject.toml` and `requirements.txt` at this time.

jmartin-tech (changes requested):
These are both patterns the project is interested in getting support for. Please see #878 and the conversations about needing some publicly available application stack to test against to maintain official generators.

Previous comments such as [this one](https://github.com/NVIDIA/garak/pull/878#discussion_r1744140521) offered patterns for possibly lifting that class into a generic class configurable via `--generator_options_file`. The implementation here based on _burp exported XML_ looks reasonable. I would be interested in opinions on if it can support the use cases described in the previous PR's comments.

iamnotcj:
> These are both patterns the project is interested in getting support for. Please see #878 and the conversations about needing some publicly available application stack to test against to maintain official generators.
> 
I took a look at that previous pull request, I'm not entirely sure if there is anything publicly available, but let me check back on this we might have something. If not, an idea I had was testing it with multiple endpoints with high availability (like test if you can reach google first and then subsequently make a request to wikipedia).

> Previous comments such as [this one](https://github.com/NVIDIA/garak/pull/878#discussion_r1744140521) offered patterns for possibly lifting that class into a generic class configurable via `--generator_options_file`. The implementation here based on _burp exported XML_ looks reasonable. I would be interested in opinions on if it can support the use cases described in the previous PR's comments.


I think it can with a few edits. The generator creates its own data structure for requests and responses from an external file. I just used burpsuite because its popular to proxy requests and save them from there. I think it would be a function similar to  ```get_reqrep_fromburp``` where it just gets request and responses from the config rather than a file.

I should be able to send another commit with those changes requested sometime before the end of the week. 👍

iamnotcj:
Okay, I just went ahead and re wrote it. It should support manually putting the requests and stuff in a config file, I followed the discourse in that other pull request and tried to make it as close as the last suggestion you gave on that thread. The format of the config can be changed, I'm open to all suggestions. 

I also moved that example file to the directory you specified as well.

garak/generators/multirest.py
erickgalinkin: I agree. I think it's best to refactor this into the existing `rest` generator package.

tests/generators/test_multirest.py
```diff
@@ -0,0 +1,84 @@
...
+
+@pytest.fixture
+def mr_gen_burpfile_config():
+    _config.plugins.generators["multirest"] = {}
+    _config.plugins.generators["multirest"]["MultiRestGenerator"] = EXAMPLE_CONFIG["multirest"]["MultiRestGenerator"]
+    generator = MultiRestGenerator()
+    return generator
+
+   
+
```
erickgalinkin, line 51: Should apply `black` on this file just for cleanup. I'd consider this a nit but others may consider the code style more important.

tests/generators/test_multirest.py
```diff
@@ -0,0 +1,84 @@
...
+# Test run function using the config
+@pytest.mark.usefixtures("mr_gen_burpfile_config")
+def test_run_with_config_and_mock(mr_gen_burpfile_config, requests_mock):
+    url = "https://example.ai/llm"
+    requests_mock.post(url,headers={"Content-Type": "application/json"},status_code= 200,text='{"id": "999999"}')
+    url2 = "https://example.ai/llm?id=999999"
+    requests_mock.get(url2, headers={"Content-Type": "application/json"}, status_code=400, json={"message": "error details"})
+
+    output = mr_gen_burpfile_config.run("Hello")
+    assert output == "error details"
```
erickgalinkin: Newline at end of file

erickgalinkin (commented):
Need to do some testing on an arbitrary websocket/streaming generator in real life to validate but this looks ok to me.

iamnotcj:
For adding it to the `rest` package, would something like this work?

```
garak/generators
├── __init__.py
├── __pycache__
...truncated...
├── rest
│   ├── __init__.py
│   ├── __pycache__
│   ├── multirest.py
│   ├── rest.py
│   └── wsocket.py
...truncated...
```
And then for `rest/__init__.py` we import the classes from the respective files:

```
from garak.generators.rest.rest import RestGenerator
from garak.generators.rest.multirest import MultiRestGenerator
from garak.generators.rest.wsocket import WebSocketGenerator

DEFAULT_CLASS = "RestGenerator"
```

jmartin-tech:
I don't think using `rest/__init__.py` in that way would currently work with the plugin enumeration structure. We can think about how to make that better in a future revision.

iamnotcj:
Alright, I just went ahead and moved them into `rest.py`.

garak/generators/rest.py
```diff
@@ -297,3 +309,374 @@ def _call_model(
...
+                        self.update_locations_dictionary(locations, resp_packet)
+                    )
+                except:
+                    raise KeyError(
+                        f"Locator failure: Variable could not be found in actual response: {locations} -> {resp_packet}"
+                    )
+            else:
+                raise ValueError(
+                    f"Response mismatch: expected {pair["response"]}, got {resp_packet}"
+                )
```
jmartin-tech, line 636: Python 3.10 does not support `[]` accessors in format strings.
```suggestion
                response = pair["response"]
                raise ValueError(
                    f"Response mismatch: expected {response}, got {resp_packet}"
                )
```

tests/generators/test_multirest.py
```diff
@@ -0,0 +1,101 @@
+import pytest
+from garak.generators.rest import MultiRestGenerator
+from garak import _config, _plugins
+
+EXAMPLE_BURP_XML = {"burpfile": "./garak/data/rest/multirest.xml"}
```
jmartin-tech: From earlier review request, prefer access to this file with a more explicit path determination.
```suggestion
from pathlib import Path
from garak.data import path as data_path
            
EXAMPLE_BURP_XML = {"burpfile": str(Path(data_path) / "rest" / "multirest.xml") }
```

garak/generators/rest.py
```diff
@@ -297,3 +309,374 @@ def _call_model(
 
 
 DEFAULT_CLASS = "RestGenerator"
```
jmartin-tech: Prefer to keep `DEFAULT_CLASS` defined as last line in file.  This is somewhat consistent in the project and makes it quick to locate.

tests/generators/test_multirest.py
```diff
@@ -0,0 +1,103 @@
+import pytest
+from garak.generators.rest import MultiRestGenerator
+from garak import _config, _plugins
+from pathlib import Path
+from garak.data import path as data_path
+            
+EXAMPLE_BURP_XML = {"burpfile": str(Path(data_path) / "rest" / "multirest.xml") }
```
jmartin-tech: Sorry in tests this should be:
```suggestion
EXAMPLE_BURP_XML = {"burpfile": str(_config.transient.package_dir / "data" / "rest" / "multirest.xml") }
```

This was my oversight, I should not have recommended the `data_path` helper here.

garak/generators/rest.py
```diff
@@ -296,4 +308,377 @@ def _call_model(
...
+            return self.json_handler(message) if self.json_response == True else message
+
+    def _call_model(
+        self, prompt: str, generations_this_call: int = 1
+    ) -> List[Union[str, None]]:
+        if output := self.request(self, prompt) == dict:
+            return output[self.json_key]
+        else:
+            return output
+
```
jmartin-tech: Please remove the `WebSocketGenerator` class. PR #1379 provided a similar generator with somewhat more extensive capabilities.

tests/generators/test_websocket.py
jmartin-tech, line 1: Please remove from this branch to refocus this PR on `MultiRestGenerator` only. `main` now contains this file based on an alternate implementation so please be careful to resolve this without impacting that file.

garak/generators/rest.py
iamnotcj: Removed.

tests/generators/test_websocket.py
iamnotcj: Removed.

iamnotcj:
It's all good. I've removed the web socket stuff and renamed this pull request 👍

PR #1293, opened 2025-07-11
Garak UI

otavionvidia:
the github issue number.

## Verification

``` json
{
    "huggingface": {
        "torch_type": "float32"
    }
}
```
- [ ] ...

Specific Hardware Examples:
* GPU related
  * Minium GPU Memory

Complex Software Examples:


garak-ui/public/reports/garak_report_20250218-093950.json
jmartin-tech, line 1: What are these files?  While some fixtures may be required for tests, I suspect these files should not be in the `public` path.  If anything they I would expect to find samples in `__test__/resources` or `__tests__/fixtures` and for test inject them in a temporary location for use in unit tests.

garak/analyze/report_digest.py
```diff
@@ -439,74 +424,92 @@ def build_digest(report_filename: str, config=_config):
     if calibration_used:
         report_digest["meta"]["calibration"] = _get_calibration_info(calibration)
 
-    return report_digest
+        report_digest["results"] = extract_flattened_modules(report_digest, config=config)
+
+        # --- Inject digest into standalone HTML file ---
+        template_path = "/Users/opadovani/Documents/Work/garak/garak/analyze/ui/index.html"
+        output_path = f"/Users/opadovani/Documents/Work/garak/garak/analyze/ui/reports/{init['run_uuid']}.html"
```
jmartin-tech: There should not be any hardcoded paths in the python code.

```suggestion
        template_path = _config.transient.package_dir / "ui" / "index.html"
```

The `build_digest()` method is not the right location for writing the final report.  This method should simply return the json that will be injected.

garak/analyze/ui/reports/39975110-6e93-4291-8a58-afb812dd900a.html
jmartin-tech, line 1: Should there be any `generated` html files?  I suspect this file should not be here, only `index.html` used by `report_digest.py` should exist on disk.  Output location for generated report data from `build_html()` is written to disk by the caller in `command.write_report_digest()` or in `__main__` of `report_digest.py`.

otavionvidia:
Hey, thanks for the review @jmartin-tech !

I was working in a different branch `garak-ui-2`, I pushed the wrong one on Friday 🙄 
A few things you mentioned I had already fixed in it, but I actually ignored the reports folder in public after your review, since they are there only if you want to accelerate local development for now (while we work on that advanced view)

garak-ui/.prettierignore
```diff
@@ -0,0 +1,4 @@
+dist
+node_modules
+coverage
+public/garak_report.json
```
jmartin-tech: This does not exist? Is this something missing?

garak/analyze/report_digest.py
```diff
@@ -439,74 +424,84 @@ def build_digest(report_filename: str, config=_config):
...
-        group_info["show_top_group_score"] = config.reporting.show_top_group_score
+def _get_report_digest(report_path):
+    with open(report_path, "r", encoding="utf-8") as reportfile:
+        for entry in [json.loads(line.strip()) for line in reportfile if line.strip()]:
+            if entry["entry_type"] == "digest":
+                return entry
+    return False
+
+def extract_flattened_modules(digest: dict, config=_config) -> List[dict]:
+    """Flatten full digest into detailed JSON structure (group -> probes -> detectors)"""
```
jmartin-tech: This function changes the format of the json digest to add a `results` entry. Looking at how it is used this is creating duplication that could be accounted for with UI code. This method is not providing new information or calculations as such it is missing presentation expectations with business logic and should not be injected.

This is simply re-organization of data and can be preformed by processing the digest on load in the javascript.

garak-ui/src/components/CalibrationSummary.tsx
```diff
@@ -0,0 +1,58 @@
...
+        )}
+      </div>
+
+      {/* Models Evaluated */}
+      <div className="border border-gray-200 rounded-md shadow-sm overflow-hidden">
+        <button
+          className="w-full text-left px-4 py-3 bg-gray-100 hover:bg-gray-200 font-semibold text-sm text-gray-800"
+          onClick={() => toggle("models")}
+        >
+          Models Evaluated
```
jmartin-tech: Naming here needs to be clear these are not models evaluated in `this` report.
```suggestion
          Calibration Models
```

garak-ui/src/components/DetectorsView.tsx
jmartin-tech, line 1: When looking at a detector view it seems somewhat unclear what is being displayed.

Screenshot 2025-07-15 at 2 46 26 PM

As I select each of the left bar chart item the view updates to color indicate the selected probe results however the relationship of what detectors ran for each probe is difficult to decipher.

garak-ui/src/components/ProbesChart.tsx
jmartin-tech, line 1: Might be helpful to have a toggle for show the score on all bars, investigation via hover is helpful but quick comparison of scores would help with understanding the larger picture when many entires have similar values.

jmartin-tech (changes requested):
Some initial thoughts, the UX should probably be reviewed synchronously to explore what is understood at a glance vs what should require exploration.

Please update the built html template to set the html title, currently it is not informative:
Screenshot 2025-07-15 at 2 55 49 PM

otavionvidia:
> Looking over the UX of updates in [3597991](https://github.com/NVIDIA/garak/commit/3597991bcd660131c78c210c9bc8ff28bb1f0d26)
> 
> * `Min Failure %` slider label does is confusing the graph is detector failure rate which would be `1 - Min Failure`. Maybe remove this for now and we can revisit or even have the slider be a global value based on `Tier`.
> * `Detector Comparison` label should align with `Probe Scores` label
> * When selecting a `Probe Score` the `Detector Comparison` order is inconsistent, can be seen when a Probe set has many classes that use the same detectors as you select each score individually the detector set will `swap` orders unpredictably.
> * None of the charts provide the numbers for actual probe attempts and detector passes something like 55 attempts 15 detections indicated somehwere on the detector tile or graph would help
> * `x / y passed` values next to probe grouping slug title does not impart useful data this it seems to simply be the number of probe classes in that group but looking closely at a complex result shows that `passed` is based on class having a pass rate greater that an zero.  The concept of `passing` a probe does not really exist, though if the user were to provide a criteria for passing it could be introduced. Currently there is no set `threshold` for considering a probe `passed`. This might be better simply stating the number of probe classes in in the group `3 Executed` or `3 Tests Completed` might make sense.
> * alignment of total % slug and titles varies, make lozenge size consistent for for all slugs
> 
> I can also start working on adding a validation automation task to ensure the committed `garak/ui/index.html` template is consisted with a build based on the report generation source.

* Min-Failure % slider
Temporarily removed – the control no longer renders, avoiding the “1 – Min Failure” confusion.
(We can re-introduce it later, tied to tier logic.)

* Detector Comparison header
Now uses the same `text-lg` font-semibold class hierarchy and left-margin as “Probe Scores”, so both headings line-up vertically.

* Detector block ordering jump
DetectorsView now sorts Object.entries(groupedDetectors) alphabetically before rendering, so the detector sections stay in a stable order when you click through different probes.

* Attempts / detections counts
  * **Backend:**  exposed fields necessary to FE like attempt_count & hit_count.
  * **Front-end:** useGroupedDetectors, useDetectorsChartSeries, useTooltipFormatter, and the Y-axis label logic were updated to pass those numbers through.
  * Tooltip now shows “Attempts: N, Detected: M”.
  * Y-axis label shows “label (attempts/hits)”.

* “x / y passed” badge next to module slug
  * Replaced with a neutral count: "N tests" (the number of probe classes in that group).
  * Tooltip updated to clarify it’s just the count of executed tests.

* Misaligned percentage badge
Added fixed width (`w-16`) and `text-center` to the % lozenge; now every row aligns regardless of “7%” vs “100%”.

jmartin-tech:
Replaced by #1573

PR #1295, opened 2025-07-14
fix missing '.' in default config probe_spec

jmartin-tech:
Correct error on attempting to use `default.yaml`

```
%python -m garak -m test --config garak/configs/default.yaml
...
❌Unknown probes❌: danAutoDANCached
```

## Verification

- [ ] `python -m garak -m test --config garak/configs/default.yaml`


erickgalinkin (approved):
LGTM

PR #1298, opened 2025-07-16
docs: improve conformance to PEP-0257

jmartin-tech:
Some reporting functions utilize the first line doc string, update most class and module level pydoc values to conform to expectations of [PEP-0257](https://peps.python.org/pep-0257/#multi-line-docstrings).


erickgalinkin (approved):
Looks good -- thanks!

Issue #1282, opened 2025-07-02
Build in ASCII smuggling


erickgalinkin:
## Summary
Build in something like [ASCII smuggler](https://embracethered.com/blog/ascii-smuggler.html), potentially as a buff.

### Motivation
Better guardrail bypasses, improved ASR, more stealth.

erickgalinkin:
Resolved in #1299

PR #1299, opened 2025-07-16
Ascii Smuggling

erickgalinkin:
Addressed #1282 by adding Probes ~~and Buffs~~ for ASCII Smuggling


jmartin-tech:
Do we have some example of end to end utility of this?

erickgalinkin:
> Do we have some example of end to end utility of this?

Will add some examples to the docs to explain the value/utility. I think I may also drop the buff entirely.

docs/source/ascii_smuggling.rst
```diff
@@ -0,0 +1,88 @@
...
+===============
+
+Garak features probes that use `ASCII smuggling <https://embracethered.com/blog/posts/2024/hiding-and-finding-text-with-unicode-tags/>`.
+ASCII smuggling is a technique that abuses the fact that LLM tokenizers will handle non-printable or zero-width characters like unicode tags and variant selectors.
+This makes it useful for bypassing LLM guardrails, which often are not trained on these evasions, and for circumventing human-in-the-loop controls, as the characters will not be visible on the screen when viewed by users.
+Some LLMs will happily decode the relevant text and handle it gracefully. Others will not. This is expected.
+
+Tag Smuggling
+-------------
+Unicode Tags are used as part of the support for emojis, and the tags must be appended to an emoji, per `the Unicode spec <https://unicode.org/reports/tr51/>`.
```
jmartin-tech: I believe this should be a link.
```suggestion
Unicode Tags are used as part of the support for emojis, and the tags must be appended to an emoji, per `the Unicode spec <https://unicode.org/reports/tr51/>`_.
```

docs/source/ascii_smuggling.rst
```diff
@@ -0,0 +1,88 @@
+Ascii Smuggling
+===============
+
+Garak features probes that use `ASCII smuggling <https://embracethered.com/blog/posts/2024/hiding-and-finding-text-with-unicode-tags/>`.
```
jmartin-tech: I believe this should be a link.
```suggestion
Garak features probes that use `ASCII smuggling <https://embracethered.com/blog/posts/2024/hiding-and-finding-text-with-unicode-tags/>`_.
```

docs/source/ascii_smuggling.rst
```diff
@@ -0,0 +1,88 @@
...
+    Returns:
+        desired_emoji + encoded payload (str): Payload encoded in unicode tags attached to the emoji.
+    """
+
+    encoded_payload = "".join(chr(0xE0000 + ord(ch)) for ch in payload)
+    return desired_emoji + encoded_payload
+
+Variant Smuggling
+-----------------
+Variation or `Variant Selectors <https://en.wikipedia.org/wiki/Variation_Selectors_Supplement>` is a Unicode block that allows for the specification of a glyph variant for the preceding character.
```
jmartin-tech: I believe this should be a link.
```suggestion
Variation or `Variant Selectors <https://en.wikipedia.org/wiki/Variation_Selectors_Supplement>`_ is a Unicode block that allows for the specification of a glyph variant for the preceding character.
```

docs/source/ascii_smuggling.rst
```diff
@@ -0,0 +1,88 @@
...
+Since there are 256 of them, this means we can also convert arbitrary bytes -- like ASCII text -- to a Variant Selector.
+Like tags, the `Unicode spec <https://www.unicode.org/versions/Unicode16.0.0/>` has no limit on the number of selectors that can be concatenated together.
```
jmartin-tech: I believe this should be a link.
```suggestion
Like tags, the `Unicode spec <https://www.unicode.org/versions/Unicode16.0.0/>`_ has no limit on the number of selectors that can be concatenated together.
```

docs/source/ascii_smuggling.rst
```diff
@@ -0,0 +1,88 @@
...
+
+        Returns:
+            encoded_payload (str): The payload encoded in variant selectors.
+        """
+        encoded_payload = "".join(chr(0xE0100 + ord(ch)) for ch in payload)
+        return encoded_payload
+
+Sneaky Bits
+-----------
+Sneaky bits, a technique described by Johann Rehberger in `a blog post <https://embracethered.com/blog/posts/2025/sneaky-bits-and-ascii-smuggler/>` encodes ASCII characters to binary, then uses the invisible times (``U+2062``) and invisible plus (``U+2064``) Unicode characters to encode 0 and 1, relying on the ability of the LLM to interpret the binary back into text.
```
jmartin-tech: I believe this should be a link.
```suggestion
Sneaky bits, a technique described by Johann Rehberger in `a blog post <https://embracethered.com/blog/posts/2025/sneaky-bits-and-ascii-smuggler/>`_ encodes ASCII characters to binary, then uses the invisible times (``U+2062``) and invisible plus (``U+2064``) Unicode characters to encode 0 and 1, relying on the ability of the LLM to interpret the binary back into text.
```

garak/probes/encoding.py
```diff
@@ -595,3 +599,57 @@ class InjectLeet(EncodingMixin, garak.probes.Probe):
...
+
+class InjectUnicodeTagChars(EncodingMixin, garak.probes.Probe):
+    """Probe for susceptibility to unicode tag ASCII smuggling
+
+    Encodes payloads by appending them to an emoji as invisible ASCII tags, and tries a variety of ways of getting the
+    target model to repeat & decode them."""
+
+    doc_uri = "https://embracethered.com/blog/posts/2024/hiding-and-finding-text-with-unicode-tags/"
+
+    encoding_funcs = [ascii_smuggling.tag_char_smuggling]
```
jmartin-tech: (Optional or possible future enhancement) Should probes that use `tag_char_smuggling ` have a `DEFAULT_PARAM` that is used as the emoji?

```suggestion
    DEFAULT_PARAMS = Probe.DEFAULT_PARAMS | {"desired_emoji": "😈"}
    encoding_funcs = [ascii_smuggling.tag_char_smuggling]
```

I suspect there would need to be some additional work in the `__init__` or some other location to support the `param` passed to `encoding_funcs` processing.

jmartin-tech (changes requested):
Minor `rst` format comments, further testing in progess.

garak/probes/encoding.py
```diff
@@ -595,3 +599,57 @@ class InjectLeet(EncodingMixin, garak.probes.Probe):
...
+    encoding_name = "Unicode Tags"
+    active = True
```
jmartin-tech, line 612: Do these probes need a `Tier` or is the default `UNLISTED` acceptable?

erickgalinkin: Open to discussion. I think this is kind of a big deal. Tier 1 or 2, IMO. It presents a real potential security risk by strongly circumventing human-in-the-loop controls.

garak/probes/encoding.py
erickgalinkin: I think this makes sense. Would require overriding `_generate_encoded_prompts`, which isn't the end of the world.

garak/probes/encoding.py
```diff
@@ -595,3 +599,66 @@ class InjectLeet(EncodingMixin, garak.probes.Probe):
...
+
+    Encodes payloads by appending them to an emoji as invisible ASCII tags, and tries a variety of ways of getting the
+    target model to repeat & decode them."""
+
+    doc_uri = "https://embracethered.com/blog/posts/2024/hiding-and-finding-text-with-unicode-tags/"
+    encoding_name = "Unicode Tags"
+    active = True
+    tier = garak.probes.Tier.OF_CONCERN
+    # May want to eventually add a check to verify that `desired_emoji` is a valid emoji.
+    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {"desired_emoji": "😈"}
```
jmartin-tech: Minor tweak, `EncodingMixin` sets the default payloads here and inherits `garak.probes.Probe.DEFAULT_PARAMS`.
```suggestion
    DEFAULT_PARAMS = EncodingMixin.DEFAULT_PARAMS | {"desired_emoji": "😈"}
```

Note this is does not fully address testing failure but gets further.

garak/probes/encoding.py
```diff
@@ -595,3 +599,66 @@ class InjectLeet(EncodingMixin, garak.probes.Probe):
...
+
+    def __init__(self, config_root=_config):
+        # Cannot reference `self` in params above.
+        self.encoding_funcs = [self._tag_char_smuggling]
+        garak.probes.Probe.__init__(self, config_root=config_root)
+        EncodingMixin.__init__(self, use_bytes=False)
+
+    def _tag_char_smuggling(self, text: str) -> str:
+        ascii_smuggling.tag_char_smuggling(
```
jmartin-tech:
```suggestion
        return ascii_smuggling.tag_char_smuggling(
```

garak/probes/encoding.py
```diff
@@ -23,6 +23,7 @@
 
 import garak.payloads
 import garak.resources.encodings
+import garak.resources.smuggling.smuggle_ascii as ascii_smuggling
```
leondz: can one get away with 
```suggestion
from import garak.resources.smuggling import smuggle_ascii
```
and edit later? then there's no `as`-rename

garak/probes/encoding.py
```diff
@@ -242,11 +245,11 @@ def _generate_encoded_prompts(self, encoding_funcs, encoding_name):
                     encoded_prompts.add((prompt, trigger))
         return sorted(list(encoded_prompts), key=lambda k: k[0])
 
-    def __init__(self):
+    def __init__(self, use_bytes=True):
```
leondz: is this the right place for `use_bytes`? may be better as a `Configurable` param

erickgalinkin: I didn't see any good place for its usage beyond in the `EncodingMixin`, since many of them use raw byte strings, but some use text (and in some cases have wrapper functions to turn things into bytes) so this makes it a bit easier in the future.

garak/probes/encoding.py
erickgalinkin: Ah yeah, this was an artifact of one of the functions previously being called `smuggle_ascii` within the `smuggle_ascii` module. Will refactor.

leondz:
lgtm!

jmartin-tech (approved):

PR #1300, opened 2025-07-16
Expand python exploitation payloads

erickgalinkin:
Extend `python_code_execution.json` with some obfuscated reverse shells. Extend `web_html_js.json` with some payloads that are more likely to evade detection.


jmartin-tech (commented):
These looks reasonable to me, are there any specific models or platforms worth testing these against as a sanity check?

erickgalinkin:
Nothing specific, no. It's just an extension of what we already have. Some more capable models will pass through the obfuscated payloads more accurately and some that reject payloads as potentially harmful will allow these through. Should be a good balance.

jmartin-tech (approved):

PR #1304, opened 2025-07-22
[bug] gcp api paramaters

anushaknvidia:
This revised GCP parameter configuration resolved the errors I encountered with the initial code settings.

### DIFF
--- a/garak/langproviders/remote.py
+++ b/garak/langproviders/remote.py
@@ -203,9 +203,9 @@ class GoogleTranslator(LangProvider):
             try:
                 auth_args = [self.api_key]
                 if self.project_id is not None:
-                    auth_kwargs = {"project": self.project_id}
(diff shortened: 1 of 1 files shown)


anushaknvidia:

anushaknvidia:
recheck

jmartin-tech:
Was this tested with `project_id` set? Service account based auth is no longer a preferred pattern based on google documentation making this somewhat complex to validate.

Looking at the change and the google authentication code I don't see `project` supported as a positional argument that can be passed to the underlying auth call by `Client.from_service_account_json()` and I see [explicit extraction of it](https://github.com/googleapis/python-cloud-core/blob/78ce8a6f7c781cf2585131d7961beec9069a206c/google/cloud/client/__init__.py#L84-L86) from `kwargs` in the client code.

PR #1306, opened 2025-07-22
Generator: Add support for Google Gemini models

dchiitmalla:
Faced issues trying to run gemini models using Litellm. This generator natively supports google models using Google’s official google.generativeai library—more reliable than LiteLLM. It supports multiple models (2.5 Pro, Flash, etc.) with error handling and model-specific config options. https://github.com/NVIDIA/garak/issues/443


garak/generators/gemini.py
```diff
@@ -0,0 +1,177 @@
...
+        
+        # Call parent init which will load config and override defaults
+        super().__init__(name, config_root)
+        
+        # Validate and get API key
+        if not hasattr(self, "api_key") or not self.api_key:
+            self.api_key = os.environ.get(self.ENV_VAR)
+            if not self.api_key:
+                raise ValueError(f"API key not found in environment variable {self.ENV_VAR}")
+
```
jmartin-tech: The call to `super().__init__()` fully handles obtaining the `self.api_key` value for either config or the `self.ENV_VAR` location specified by the class and raising an exception by calling `_validate_env_var()`. This error check is not required.

See: https://github.com/NVIDIA/garak/blob/e035f3690eff6a18a63e14e438fffe4781356aa4/garak/configurable.py#L112-L126

```suggestion
```

If this class needs to relax or revise how api_key values handled it should implement `_validate_env_var()` for to provide any custom function and possibly call the version provided by `super()`.

garak/generators/gemini.py
```diff
@@ -0,0 +1,177 @@
...
+            prompt: The input text to send to the model
+            generations_this_call: Number of responses to generate
+            
+        Returns:
+            A list of response strings, or None for failed generations
+        """
+        responses = []
+        import logging
+        
+        for _ in range(generations_this_call):
```
jmartin-tech: Iterating generations like this is will not combine with backoff correctly.  As written any raised backoff exception will throw away all completed generations and start over.

Looking at [gemini docs](https://ai.google.dev/api/generate-content#v1beta.GenerationConfig) multiple generations can be obtained in a single call using `GenerateContentConfig` by setting `candidateCount= generations_this_call` and passing `generate_content()` a `config` named parameter.

If calling for more than one generation please validate how the `response` object will be formatted.

tests/generators/test_gemini.py
```diff
@@ -0,0 +1,254 @@
...
+    assert generator.generator_family_name == "gemini"
+    assert generator.supports_multiple_generations is True
+    assert generator.ENV_VAR == "GOOGLE_API_KEY"
+    assert "gemini-2.5-pro" in generator.SUPPORTED_MODELS
+    assert "gemini-2.5-flash" in generator.SUPPORTED_MODELS
+    assert "gemini-2.5-flash-lite-preview" in generator.SUPPORTED_MODELS
+    assert "gemini-2.5-flash-native-audio" in generator.SUPPORTED_MODELS
+    assert "gemini-2.5-flash-preview-text-to-speech" in generator.SUPPORTED_MODELS
+    assert "gemini-2.5-pro-preview-text-to-speech" in generator.SUPPORTED_MODELS
+    assert "gemini-2.0-flash" in generator.SUPPORTED_MODELS
```
jmartin-tech: This does not really test any functionality and other testing covers standard instantiation validation for all generators.
``` suggestion
```

tests/generators/test_gemini.py
```diff
@@ -0,0 +1,254 @@
...
+    
+    # Create the generator with a text-to-speech model
+    generator = GeminiGenerator(name="gemini-2.5-pro-preview-text-to-speech")
+    output = generator._call_model("Generate speech for this text.")
+    
+    # Verify the results
+    assert len(output) == 1
+    assert output[0] == "This is a text-to-speech response."
+    mock_model.generate_content.assert_called_once_with("Generate speech for this text.")
+
```
jmartin-tech: `garak` does not support `audio` as an output modality.
``` suggestion
```

tests/generators/test_gemini.py
```diff
@@ -0,0 +1,254 @@
...
+    # Patch the genai.configure function
+    mock_configure = MagicMock()
+    
+    # Apply the patches
+    monkeypatch.setattr("google.generativeai.GenerativeModel", mock_generative_model)
+    monkeypatch.setattr("google.generativeai.configure", mock_configure)
+    
+    # Create the generator with a native audio model
+    generator = GeminiGenerator(name="gemini-2.5-flash-native-audio")
+    output = generator._call_model("Transcribe this text.")
```
jmartin-tech: This does not make sense the input modality suggested by this test is `audio` not `text`.

The current generator also inherits the default modality, `{"in": {"text"}, "out": {"text"}}` unless overridden this generator should only accept `text` prompts.

garak/generators/gemini.py
```diff
@@ -0,0 +1,188 @@
...
+                elif hasattr(response, "text") and response.text:
+                    responses.append(response.text)
+                else:
+                    logging.warning(f"Empty response from Gemini model {self.name}")
+                    responses.append(None)
+                    
+        except GoogleAPIError as e:
+            # This will be caught by the backoff decorator and retried
+            logging.error(f"GoogleAPIError when calling {self.name}: {e}")
+            raise
```
jmartin-tech:
```suggestion
            raise e
```

garak/generators/gemini.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python3
```
jmartin-tech: Generator packages are not executable scripts.
```suggestion
```

jmartin-tech (changes requested):
General comments, the auth concern may be reasonable to defer to require environment variable based API key for initial acceptance. 

Another concern that may be important to address:
This PR uses that package `google-generativeai` however [current docs](https://ai.google.dev/gemini-api/docs/libraries#new-libraries) recommend using `google-genai` which has the same imported package name but a different interface.

Abhiraj-GetGarak:

Abhiraj-GetGarak:
recheck

garak/generators/gemini.py
Abhiraj-GetGarak: **Addressed** - The implementation now uses `candidate_count=generations_this_call` in `GenerateContentConfig` and makes a single API call via `_generate_content_with_backoff()`. This ensures backoff retries don't discard completed generations. The response format has been validated and properly handles multiple candidates from the API response.

Abhiraj-GetGarak:
1. **Multiple generations backoff issue** - Fixed using new `google-genai` library with `candidate_count` in
  single API call.

  2. **Authentication flexibility** - Enhanced with both API key and Vertex AI (ADC) support using new client
  architecture.

  3. **Package migration completed** - Successfully migrated from `google-generativeai>=0.8.5` to
  `google-genai>=1.0.0`

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+        2. Application Default Credentials (ADC) when no API key is provided
+        
+        ADC can be set up using 'gcloud auth application-default login' or by setting
+        GOOGLE_APPLICATION_CREDENTIALS to point to a service account key file.
+        """
+    def _clear_model(self):
+        """Clear the model to avoid pickling issues."""
+        self.model = None
+
+    def __init__(self, name="", config_root=_config, api_key=None):
```
jmartin-tech, line 129: `api_key` can be injected via the `config_root` parameter, please use signature consistent with `base.Generator`:
```suggestion
    def __init__(self, name="", config_root=_config):
```

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+        """Initialize the Gemini generator."""
+        # Initialize default parameters before calling super().__init__
+        self.temperature = self.DEFAULT_PARAMS["temperature"]
+        self.top_p = self.DEFAULT_PARAMS["top_p"]
+        self.top_k = self.DEFAULT_PARAMS["top_k"]
+        self.max_output_tokens = self.DEFAULT_PARAMS["max_output_tokens"]
+        self.api_key = api_key
```
jmartin-tech, line 136: These values are populated by the `Configurable._load_config()` call made in `super().__init__()` defer to common code:
```suggestion
```

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+        super().__init__(name, config_root)
+        # Defer model loading until actually needed
+        self.client = None
+        self.model = None
```
jmartin-tech, line 140: I would suggest, early validation that all required configuration is available during construction outweighs benefits of deferred loading here.
```suggestion
        self._load_model()
```

This becomes more clearly required when considering the fact that `_load_model()` currently impacts `self.modality` for the generator which need to be viable to evaluate on the instantiating thread before any request for inference has been performed.

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+            # This ensures that multiple generations are obtained in a single call
+            # and backoff doesn't discard completed generations
+            response = self._generate_content_with_backoff(prompt, generations_this_call)
+            return self._process_response(response, generations_this_call)
+        except Exception as e:
+            # If all retries failed, return None values for all expected generations
+            logging.error(f"All retries failed for {self.name}: {e}")
+            return [None] * generations_this_call
+
+    @backoff.on_exception(backoff.expo, APIError, max_tries=5)
```
jmartin-tech, line 169: Note this suggestion mirrors how other generators currently implement backoff, `garak` is tenacious and currently is not expected to abort or skip a generation unless the error indicates that there is no possibility of recovery.
```suggestion
    @backoff.on_exception(backoff.fibo, APIError, max_value=70)
```

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+        
+        # Generate content with the model
+        try:
+            response = self.client.models.generate_content(
+                model=self.name,
+                contents=prompt,
+                config=generation_config
+            )
+            return response
+            raise e  # This will trigger backoff retry
```
jmartin-tech, line 206: This would not trigger backoff, `e` does not exist and `return` ends execution of the block.
```suggestion
```

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+    def _validate_env_var(self):
+        """Override the default API key validation to allow for ADC when no API key is provided.
+        
+        For GCP services, authentication can be done via:
+        1. API key (via GOOGLE_API_KEY environment variable or api_key parameter)
+        2. Application Default Credentials (ADC) when no API key is provided
+        
+        ADC can be set up using 'gcloud auth application-default login' or by setting
+        GOOGLE_APPLICATION_CREDENTIALS to point to a service account key file.
+        """
```
jmartin-tech, line 124: Env variable access should be implemented in `_validate_env_var(self)` which is part of the [`Configurable`](https://github.com/NVIDIA/garak/blob/a3794b97c5f97f39f93f467264b093ee68009386/garak/configurable.py#L12) class implementation pattern.  A generator implementing this method should determine if a call to `super()._validate_env_var()` is warranted. The `_validate_env_var()` method should ensure the object has all necessary state required based on available environment details. If no environment based configuration is found and the underlying library will look for credentials in another way then this method would defer the later call to `_load_client()` to trigger raise at failure to authenticate during runtime.

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+            response = self.client.models.generate_content(
+                model=self.name,
+                contents=prompt,
+                config=generation_config
+            )
+            return response
+            raise e  # This will trigger backoff retry
+        except Exception as e:
+            logging.error(f"Unexpected error when calling {self.name}: {e}")
+            raise e  # Re-raise to be handled by caller
```
jmartin-tech, line 209: What value add is there in this `try` block?

```suggestion
        response = self.client.models.generate_content(
            model=self.name,
            contents=prompt,
            config=generation_config
        )
        return response
```

If the exception is of type `APIError` then `@backoff.on_exception` will trigger otherwise it will bubble up for the caller to handle.

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+        except Exception as e:
+            logging.error(f"Error processing response from {self.name}: {e}")
+            # Fill with None values for the expected number of generations
+            responses = [None] * generations_this_call
+        
+        # Ensure we return the expected number of responses
+        while len(responses) < generations_this_call:
+            responses.append(None)
+        
+        return responses[:generations_this_call]
```
jmartin-tech, line 253: This logic seems odd, I would expect the response format either to have all generations required or none (Maybe that is an invalid assumption on my part).  I suspect this method can be a bit less opinionated and simply extract results from the response.  The caller can be responsible for deciding what to do if the number of found responses do not total `generations_this_call`.

To that end `generations_this_call` does not need to be passed into this method.

```suggestion
            responses = [None]

        return responses
```

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+        "gemini-2.5-flash",
+        "gemini-2.5-flash-lite-preview",
+        "gemini-2.5-flash-native-audio",
+        "gemini-2.5-flash-preview-text-to-speech",
+        "gemini-2.5-pro-preview-text-to-speech",
+        "gemini-2.0-flash"
+    ]
+    
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "name": "gemini-2.5-pro",
```
jmartin-tech, line 57: Do we want a default model here, the churn as model names change could be tedious to maintain.
```suggestion
        "name": "gemini-2.5-pro",
```

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+    # List of supported models
+    SUPPORTED_MODELS = [
+        "gemini-2.5-pro",
+        "gemini-2.5-flash",
+        "gemini-2.5-flash-lite-preview",
+        "gemini-2.5-flash-native-audio",
+        "gemini-2.5-flash-preview-text-to-speech",
+        "gemini-2.5-pro-preview-text-to-speech",
+        "gemini-2.0-flash"
+    ]
```
jmartin-tech, line 54: Most of these models have very short TTL, [retirement dates](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions) suggests that models listed are `Auto-updated aliases` and are versioned further.  To denote possible need for clarification of results it might be helpful to have some method added to extract and document the `Stable version reference` to the model tested by `garak` during a run.

garak/generators/gemini.py
```diff
@@ -0,0 +1,256 @@
...
+      * When running on Google Cloud Platform, credentials are automatically available
+    
+    Supported models:
+    - gemini-2.5-pro: Gemini 2.5 Pro model (default)
+    - gemini-2.5-flash: Gemini 2.5 Flash model
+    - gemini-2.5-flash-lite-preview: Gemini 2.5 Flash Lite Preview model
+    - gemini-2.5-flash-native-audio: Gemini 2.5 Flash Native Audio model
+    - gemini-2.5-flash-preview-text-to-speech: Gemini 2.5 Flash Preview Text-to-Speech model
+    - gemini-2.5-pro-preview-text-to-speech: Gemini 2.5 Pro Preview Text-to-Speech model
+    - gemini-2.0-flash: Gemini 2.0 Flash model
```
jmartin-tech, line 37: I like having this level of detail however I think the lifecycle of models will cause the list to become stale quickly.  Consider removing from the static docs.

jmartin-tech (changes requested):
Code only review, full functional testing will be performed after comments and are addressed.

PR #1313, opened 2025-07-30
Switch from `lorem` to `python-lorem`

06kellyjac:
This PR switches out `lorem` for `python-lorem`.

From what I can tell since it has the same library and has the same function names unless there's a problem with the default values for `sentence()` no code changes are necessary?

Has pyproject.toml & typing.

One change which may be relevant is it's BSD 3 Licensed rather than MIT

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,7 +96,7 @@ dependencies = [
   "huggingface_hub>=0.21.0",
   'python-magic-bin>=0.4.14; sys_platform == "win32"',
   'python-magic>=0.4.21; sys_platform != "win32"',
-  "lorem==0.1.1",
+  "python-lorem>=1.3.0",
   "xdg-base-dirs>=6.0.1",
   "wn==0.9.5",
   "ollama>=0.4.7",
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,7 +30,7 @@ jsonpath-ng>=1.6.1
 huggingface_hub>=0.21.0
 python-magic-bin>=0.4.14; sys_platform == "win32"
 python-magic>=0.4.21; sys_platform != "win32"
-lorem==0.1.1
+python-lorem>=1.3.0
 xdg-base-dirs>=6.0.1
 wn==0.9.5
 ollama>=0.4.7


06kellyjac:
recheck

06kellyjac:

recheck

jmartin-tech:
What is the reasoning behind this change?  `lorem` is only currently utilized for a `mock` testing generator used in examples and testing. I see the currently used library has not changed since 2016 and the newer package looks to have been created in 2019 and has had recent development releases but no stable release in 2 years. Is there some need this revision will support?

06kellyjac:
Hi,

I did mention some reasons, first I'll address your questions then put the summary.

> I see the currently used library has not changed since 2016 ...

The `lorem` package hasn't had a change since 2016 and that's fine if it's feature complete and doesn't have any dependencies to update or security issues to address etc. So the age since changes isn't my problem here.

> and the newer package looks to have been created in 2019 and has had recent development releases but no stable release in 2 years. 

I see it the oposite way around. The confusion may be the names in pypi and the repo names are switched.

* `lorem` ([pypi](https://pypi.org/project/lorem/)) - https://github.com/sfischer13/python-lorem
  * v0 release (`0.1.1`) which if they're following semver isn't a stable release
  * The self reported[^1] [PyPi Classifier](https://pypi.org/classifiers/) for Development Status is "4 - Beta"
* `python-lorem` ([pypi](https://pypi.org/project/python-lorem/)) - https://github.com/JarryShaw/lorem
  * v1 release which if they're following semver is a stable release
  * The self reported[^1] [PyPi Classifier](https://pypi.org/classifiers/) for Development Status is "5 - Production/Stable"
  * I've not done a deep dive but the use of release candidates and [post releases](https://discuss.python.org/t/post-and-developmental-release-version-specifier-handling-in-distros/38461) initially suggests they take stability seriously

[^1]: from what I can tell reading the Classifier Doc - "Each project's maintainers provide PyPI with a list of "Trove classifiers"

---

Benefits of `python-lorem`:

* Marked stable
* More recently maintained
* Uses more modern pyproject.toml approach to management
* comes with types
* (not so much for the project) packaged in nixpkgs unlike `lorem`
  * (this was my main driver to propose the swap, but `lorem` can be packaged too)

Unchanged:

* Neither seem to be packaged for debian/kali #1036 

Drawbacks of `python-lorem`:

* Change in general
* [449.51k](https://pepy.tech/projects/python-lorem?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=stacked&versions=1.3.0%2C1.3.0rc1%2C1.2.0) downloads vs `lorem`'s [6.97M](https://pepy.tech/projects/lorem?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=line&versions=0.1.1%2C0.1.0%2C0.0.1) (pure downloads isn't an indicator of wider project use and can be due to bots and lots of CI builds but still may be interesting)
  * averaged that's $\left\lceil \left( \frac{449{,}514\ \text{dl}}{70\ \text{mo}} \right) \times 12\ \text{mo} \right\rceil = 77{,}060\ \text{dl/yr}$ vs $\left\lceil \left( \frac{6{,}970{,}120\ \text{dl}}{109\ \text{mo}} \right) \times 12\ \text{mo} \right\rceil = 767{,}353\ \text{dl/yr}$ so down by a factor of 10


In summary I could see it's probably easiest to stay with lorem. Even as `v0` if it doesn't change there's no more stable than that.
The dependency is currently pinned (to the degree requirements.txt/pyproject.toml allows) and PyPi appears to have some protections to overwriting existing versions in the potential event the pacakge is taken over.

An alternative could be to inline/vendor the code into the project with attribution since it's very small, unchanging, and would cut out an external dependency.

Otherwise a PR to lorem to just add types and pyproject.toml could be nice.

06kellyjac:
I'll close this since I've moved to using `lorem` in https://github.com/NixOS/nixpkgs/pull/429835

Thanks

PR #1283, opened 2025-07-02
Replace deprecated `np.infty` with `np.inf`.

erickgalinkin:
Replace deprecated `np.infty` with `np.inf`. Resolved #1250

PR #1314, opened 2025-07-30
Use numpy v2

06kellyjac:
Since #1283 it appears garak can update to numpy v2
If there's anything else that needs updating LMK
If you'd prefer to just allow v1 and v2 LMK

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,7 @@ dependencies = [
   "torch>=2.6.0",
   "sentencepiece>=0.1.99",
   "markdown>=3.4.3",
-  "numpy>=1.26.4,<2.0",
+  "numpy>=2.0.0",
   "zalgolib>=0.2.2",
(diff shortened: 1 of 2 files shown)


06kellyjac:

06kellyjac:
recheck

06kellyjac:
Related to #1036 since debian, and as a result kali, use numpy v2 as far as I can tell

https://salsa.debian.org/python-team/packages/numpy/-/blob/457ca8acba07b39085411b1dff8e8a0e2e25718f/PKG-INFO

(It's also not a complete blocker with nixpkgs since it's still got `numpy_1` available but a very small minority of packages use it & new packages wouldn't be advised.)

numpy v2 released [a year ago 2024-06-16](https://github.com/numpy/numpy/releases/tag/v2.0.0)

jmartin-tech (approved):
Testing shows this looks to be compatible, thank you!

PR #1318, opened 2025-08-02
Fix/core cfg listed twice

### LINKED ISSUE #1249: bug: core cfg listed twice in report in default run
``python -m garak -g1 -m test --report_prefix=./x -p test.Test``

Image

this should be there once

sinhaabhiraj0:
### Tell us what this change does

This change fixes a bug where `garak`'s base configuration file (`garak.core.yaml`) was being loaded multiple times, causing its path to appear duplicated in verbose/debug output.

The root cause was that `_config.load_config()` was incorrectly reloading the base config, which had already been loaded by `_config.load_base_config()`.

The fix makes the configuration loading sequence strict and explicit:
1.  `_config.load_config()` no longer reloads the base config.
2.  All parts of the codebase that load configuration now correctly call `_config.load_base_config()` first.

This change also includes a new regression test to prevent this issue from reoccurring and fixes two downstream modules (`test_buff_config.py` and `report_digest.py`) that were relying on the old, buggy behavior.

Fixes #1249

## Verification

- [x] **Confirm the original bug is fixed:**
    - Run `garak --list_config`
    - **Verify** that `garak.core.yaml` and other config files appear only once in the list of loaded configs.

- [x] **Run the new regression test:**
    - This test specifically checks for the duplication bug.
    - Run `python -m pytest tests/test_config.py::test_no_duplicate_config_files_on_load`
    - **Verify** that the test passes.

- [x] **Run the full test suite:**
    - This ensures that the change did not break other parts of the application.
    - Run `python -m pytest tests/`
    - **Verify** that all tests pass (including the ones we fixed in `tests/buffs/test_buff_config.py` and `tests/analyze/test_analyze.py`).

- [x] **Verify** the thing does what it should:
    - The application loads configuration correctly without duplicates.

- [x] **Verify** the thing does not do what it should not:
    - The change does not introduce any new test failures or runtime errors.

- [x] **Document** the thing and how it works:
    - The code changes in `_config.py`, `test_buff_config.py`, and `report_digest.py` are self-documenting by enforcing the correct loading order. The new test in `test_config.py` documents the expected behavior.


sinhaabhiraj0:

sinhaabhiraj0:
recheck

PR #1325, opened 2025-08-06
An API for Garak

iamnotcj:
I added a API for Garak. I tried to make it so that the API wraps around the cli package almost exclusively. I think this would make it easier to maintain. I'm open to any other implementations though.

I'm currently working on a Burpsuite extension for Garak. It is similar to other Burpsuite extensions that rely on the API's of external programs, rather than packaging them all together. I'd imagine that there are some other programs that could benefit from this too

BurpExtensionScreenshot

### DIFF
--- a/garak/__main__.py
+++ b/garak/__main__.py
@@ -3,10 +3,14 @@
 import sys
 
 from garak import cli
+from garak import api
 
 
 def main():
-    cli.main(sys.argv[1:])
+    if "--remote" in sys.argv:
+        api.run()
+    else:
+        cli.main(sys.argv[1:])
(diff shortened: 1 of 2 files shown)


jmartin-tech:
Conceptually this is very interesting, I don't know if this will be a maintainable feature as offered here. Ideally, a long running daemon wrapper would not require the user to have local access to obtain the results.

Will think on this more and discuss further.

jmartin-tech:
Closing this based on previous comments, as noted the idea is quite reasonable the implementation here is not yet to fully fleshed out for general usage.

PR #1338, opened 2025-08-19
Iterative probe base class and example

aishwaryap:
This PR demonstrates a sample multi-turn probe. Since the primary value add of multi-turn probes is iteratively adapting the next turn based on the last response, this PR includes a base probe class to simplify this. 

## Verification

``` json
{
    "huggingface": {
        "torch_type": "float32"
    }
}
```


aishwaryap:
Closing this PR since we have decided to make a number of design changes to align with intended implementation style for multi turn probes. Will create a new PR with updated design.

PR #1351, opened 2025-09-01
probe: ansi escape codes in tokenizer

leondz:
add probe for scanning HF tokenizers for tokens bearing raw escape codes

## Verification

- [ ] Basic execution `garak -m huggingface -n gpt2 -p ansiescape.AnsiRawTokenizerHF`
- [ ] Test new probe & existing ansiescape probes `python -m pytest tests/probes/test_detectors_ansiescape.py`
- [ ] Test behaviour on non-HF generator `garak -m openai -n o3-mini -p ansiescape.AnsiRawTokenizerHF` (should noop)


garak/detectors/ansiescape.py
```diff
@@ -8,7 +8,7 @@
 class Escaped(StringDetector):
     """pick up escaped ANSI escape sequences"""
 
-    doc_uri = None  # reference
+    doc_uri = "https://interhumanagreement.substack.com/p/llm-output-can-take-over-your-computer"
```
erickgalinkin, line 12: Just making a note that eventually we might need to add a CWE as a tag. https://github.com/CWE-CAPEC/CWE-Content-Development-Repository/issues/163

garak/probes/ansiescape.py
```diff
@@ -121,6 +131,71 @@ def __init__(self, config_root=_config):
...
+class AnsiRawTokenizer(garak.probes.Probe):
+    """Probe for checking a tokenizer for potentially malicious ANSI escape sequences
+
+    Returns any tokens bearing ANSI escape sequences that could be used to seize
+    control of anything rendering output. Always returns one token judged non-malicious.
+    """
+
+    lang = "*"
+    tags = ["owasp:llm05"]
+    supported_generators = {"Model", "Pipeline", "OptimumPipeline", "LLaVA"}
```
erickgalinkin: At some point, we should codify this list somewhere concrete I think. There are other probes (GCG, Beast) where we want to know whether we can compute gradients and I think this set will be common.

garak/probes/ansiescape.py
```diff
@@ -121,6 +131,71 @@ def __init__(self, config_root=_config):
...
+            attempt.outputs = [token]
+            return attempt
+
+        clean_attempt_found = False
+        for t in tok.vocab:
+            if any(payload in t for payload in LIVE_PAYLOAD_TOKENS):
+                attempts.append(_get_token_attempt(t))
+            elif not clean_attempt_found:
+                clean_attempt_found = True
+                attempts.append(_get_token_attempt(t))
```
erickgalinkin: So I'm thinking about this -- does the escape code need to be a single token? Or will it work otherwise? I think we can be more efficient and more accurate with this.

Looking at tiktoken for an example:
```
>>> import tiktoken
>>> enc = tiktoken.encoding_for_model("gpt-4")
>>> enc.encode("\x1b[")
[91535]
>>> enc.encode("\x1b]")
[215, 60]
>>> enc.encode("\x9b")
[126, 249]
>>> enc.encode("\x9d")
[126, 251]
>>> enc.decode([91535])
'\x1b['
>>> enc.decode([215, 60])
'\x1b]'
>>> enc.decode([126, 251])
'\x9d'
>>> enc.decode([126, 249])
'\x9b'
```

Looks like only one of them is encoded as a single token, but all of these will still work. 

I think we can rewrite this to go only over the set of `LIVE_PAYLOAD_TOKENS` (much smaller set) and then rewrite `_get_token_attempt` to `encode`, then `decode`, and if the same string pops out, Bob's your uncle.
If it has to be a single token (I don't believe it does), then we only need to check that `tok.convert_tokens_to_ids(token_to_check)` tokenizes to a single token not equal to `tok.unk_token_id`.

Thoughts?

leondz: Oh, like are escape codes composable? Crossed my mind too. This is cool, we should add these. And maybe even be principled about it.

The codes are already starting to be found in four-five places - should possibly get factored out as `payloads` or `data`

leondz: there are some Intricacies predicated on tokeniser implementation here. `tiktoken` has [its own way](https://github.com/openai/tiktoken/blob/2ab6d3706d557b560b200be48e6a32324682c9a3/src/lib.rs#L217) of handling these sequences. HF `tokenizers` has another. Will rename class to scope just to hf. More mining on how to get these out of hf tokenizers is appropriate, but this feels like a borderline red team/research question. What's the minimum bar you'd like to see for acceptance here?

erickgalinkin: I think we could accept it as-is, but IMO there are two adjacent questions worth answering:
1. In the current implementation, we are checking single tokens in the vocabulary. Wouldn't it be more efficient to have something like:
```python
for escape_code in `LIVE_PAYLOAD_TOKENS`:
    tokenized = tok.convert_tokens_to_ids(escape_code)
    if len(tokenized) == 1 and tokenized != tok.unk_token_id:
        attempts.append(_get_token_attempt(t))
```
or whatever -- you get it.

2. If it does not need to be a single token, then shouldn't we simply check if:
```python
for escape_code in `LIVE_PAYLOAD_TOKENS`:
    if tok.decode(tok.encode(escape_code)) == escape_code:
        do_whatever()
```

leondz: 1. We're checking all the tokenizer vocab to see if it has any entries containing a usable sequence, not just necessarily matching. I can see multiple modes worth checking for - the current one is conservative (i.e. sensitive)
2. This is a fine test for `tiktoken`, yeah. Current probe focuses on Hugging Face models (& has been renamed accordingly)

garak/probes/ansiescape.py
```diff
@@ -111,16 +88,80 @@ def __init__(self, config_root=_config):
...
+        output: score = 1/(|risky tokens| + 1)
+        the more risky tokens, the worse
+        """
+        from transformers import AutoTokenizer
+
+        package_path = generator.__class__.__module__
+        if package_path.split(".")[-1] != "huggingface":
+            return []
+        if generator.__class__.__name__ not in self.supported_generators:
+            return []
```
leondz: might be nice to log why

garak/data/ansi.py
leondz, line 1: `.py` in `resources`, not `data`

garak/probes/ansiescape.py
```diff
@@ -111,16 +88,80 @@ def __init__(self, config_root=_config):
...
+class AnsiRawTokenizerHF(garak.probes.Probe):
+    """Probe for checking a tokenizer for potentially malicious ANSI escape sequences
+
+    Returns any tokens bearing ANSI escape sequences that could be used to seize
+    control of anything rendering output. Always returns one token judged non-malicious.
+    """
+
+    lang = "*"
+    tags = ["owasp:llm05"]
+    supported_generators = {"Model", "Pipeline", "OptimumPipeline", "LLaVA"}
```
leondz: does `llava` support this?

garak/generators/huggingface.py
```diff
@@ -91,6 +91,7 @@ def _load_client(self):
             self.generator.tokenizer = AutoTokenizer.from_pretrained(
                 pipeline_kwargs["model"]
             )
+        self.tokenizer = self.generator.tokenizer
```
leondz, line 94: needs to be cleared in `_clear_client`

garak/probes/ansiescape.py
leondz: make this absolute namespace, `huggingface.Model` and so on

leondz: yeah, `AutoTokenizer` supports LLaVA models, and the object returned includes a `vocab` supporting the required access pattern

PR #1360, opened 2025-09-05
docs: reorganize table of concepts and getting started page

mikemckiernan:


docs/README.md
```diff
@@ -0,0 +1,38 @@
...
+  - The command is ``garak``.
+    If the context is to instruct the reader to run the ``garak`` command, then style it as inline code.
+
+  - In a heading, it's Garak with an uppercase G because it's a heading.
+
+## Publishing the Documentation
+
+Tag the commit to publish with `docs-v<semver>`.
+
+To avoid publishing the documentation as the latest, ensure the commit has `/not-latest` on a single line, tag that commit, and push to GitHub.
```
jmartin-tech: Not sure this really applies to this project at this time.

docs/source/install.rst
```diff
@@ -0,0 +1,26 @@
+..  headings: = - ^ "
+Installation
+============
+
+.. _installation:
+
+garak is a command-line tool.
+It is developed in Linux and MacOS.
```
jmartin-tech:
```suggestion
It is developed in Linux, MacOS, and Windows.
```

docs/source/usage.rst
```diff
@@ -1,83 +1,35 @@
...
-
-
-The general syntax is:
+    garak --list_probes
 
-.. code-block:: console
+You can limit the probes to run by specifying more arguments.
 
-    garak <options>
+For example, you can specify ``--probes promptinject`` to run only the `PromptInject <https://github.com/agencyenterprise/promptinject>`_ framework's methods. for example.
```
jmartin-tech: I know this is just refactored location, however I think the link used here is not effective context.

If specifying by _probe family_ all _active probes_ implemented in that family will be run, probes families are commonly grouped by the technique they implement which often may come from various source as different patterns to apply similar techniques often exist.

mikemckiernan: I agree.  Thanks for the review!

docs/source/install.rst
leondz: Is this true?

jmartin-tech: We have definitely had some contributors that were clearly working in Windows. We could clarify this to:
```
It is supported for use in Linux, MacOS, and Windows. Primary development focuses on *nix compatible platforms.
```

But I would call that over specification., we execute PR tests against Windows and I consider that at a _development_ activity.

Issue #1209, opened 2025-05-12
replace `fschat` with `Turn` & `Conversation`


leondz:
ditch the `fschat` structures for conversation and use our own abstraction

## requires
~~#1089~~ #1254 

## impacts
* `garak/detectors/judge.py`
* `garak/resources/red_team/conversation.py`
* `garak/resources/gcg/attack_manager.py`
* `garak/resources/autodan/model_utils.py`

### directly 
* `fschat`

### indirectly
* `markdown2`
* `nh3`
* `rich`
* `prompt_toolkit`
* `shortuuid`

PR #1361, opened 2025-09-10
arch: rm support for init'ing `attempt.prompt` with `str`

### LINKED ISSUE #1347: arch: remove support for accessing/instantiating `Message` as `str`
see https://github.com/NVIDIA/garak/pull/1337/files#r2293761110

stop supporting this. support is patchy and communicates an ambiguous pattern

leondz:
fixes #1347

`attempt.prompt` accepts `Message` or `Conversation` - now we check for this

hey seeing as local models are difficult to run, are perhaps the minority case, and are where most of our deps come from, what are thoughts on dropping these, and migrating to something strongly typed, like rust?

## Verification

- [ ] tests pass
- [ ] garak -m test -g 1 runs

### DIFF
--- a/garak/attempt.py
+++ b/garak/attempt.py
@@ -158,7 +158,7 @@ class Attempt:
     :param status: The status of this attempt; ``ATTEMPT_NEW``, ``ATTEMPT_STARTED``, or ``ATTEMPT_COMPLETE``
     :type status: int
     :param prompt: The processed prompt that will presented to the generator
-    :type prompt: Union[str|Turn|Conversation]
+    :type prompt: Message|Conversation
     :param probe_classname: Name of the probe class that originated this ``Attempt``
     :type probe_classname: str
     :param probe_params: Non-default parameters logged by the probe
--- a/garak/attempt.py
+++ b/garak/attempt.py
@@ -106,7 +106,9 @@ class Turn:
             raise ValueError("Expected `role` in Turn dict")
         message = entity.pop("content", {})
         if isinstance(message, str):
-            content = Message(text=message)
+            raise TypeError(
+                "Turn does not support str-type content, use Message / report this as a bug"
+            )
         else:
             content = Message(**message)
         return cls(role=role, content=content)
@@ -223,11 +225,16 @@ class Attempt:
             if isinstance(prompt, Conversation):
                 self.conversations = [prompt]
             elif isinstance(prompt, str):
-                msg = Message(text=prompt, lang=lang)
+                raise ValueError(
+                    "attempt Prompt must be Message or Conversation, not string"
+                )
+                # msg = Message(text=prompt, lang=lang)
             elif isinstance(prompt, Message):
                 msg = prompt
             else:
-                raise TypeError("prompts must be of type str | Message | Conversation")
+                raise TypeError(
+                    "attempt prompts must be of type Message | Conversation"
+                )
             if not hasattr(self, "conversations"):
                 self.conversations = [Conversation([Turn("user", msg)])]
             self.prompt = self.conversations[0]
@@ -321,14 +328,17 @@ class Attempt:
         return all_outputs
 
     @prompt.setter
-    def prompt(self, value: Union[str | Message | Conversation]):
+    def prompt(self, value: Message | Conversation):
         if hasattr(self, "_prompt"):
             raise TypeError("prompt cannot be changed once set")
         if value is None:
             raise TypeError("'None' prompts are not valid")
         if isinstance(value, str):
             # note this does not contain a lang
-            self._prompt = Conversation([Turn("user", Message(text=value))])
+            raise TypeError(
+                "Attempt.prompt must be Message or Conversation, not bare string"
+            )
+            # self._prompt = Conversation([Turn("user", Message(text=value))])
         if isinstance(value, Message):
             # make a copy to store an immutable object
             self._prompt = Conversation([Turn("user", Message(**asdict(value)))])
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -180,7 +180,11 @@ class Probe(Configurable):
         return attempt
 
     def _mint_attempt(
-        self, prompt=None, seq=None, notes=None, lang="*"
+        self,
+        prompt: str | garak.attempt.Message | garak.attempt.Conversation | None = None,
+        seq=None,
+        notes=None,
+        lang="*",
     ) -> garak.attempt.Attempt:
         """function for creating a new attempt given a prompt"""
         turns = []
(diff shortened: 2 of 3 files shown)


leondz:
consider removing support for `lang` param in `Attempt` constructor & migrating to Message (see #1468)

garak/attempt.py
```diff
@@ -106,7 +106,7 @@ def from_dict(cls, value: dict):
             raise ValueError("Expected `role` in Turn dict")
         message = entity.pop("content", {})
         if isinstance(message, str):
-            content = Message(text=message)
+            content = Message(message, str)
```
jmartin-tech: I don't think this revision is correct, str is a class and passing it as the second argument to this constructor would not be valid.  In theory this branch of the conditional cannot be reached and probably should raise a `ValueError` if `content` in the serialized entity was of object type `str`. The existing code would provide a method to manually create a more loose approximation of a serialized `Turn` however do not I see any reason we _should_ support it.

garak/attempt.py
```diff
@@ -223,11 +223,16 @@ def __init__(
             if isinstance(prompt, Conversation):
                 self.conversations = [prompt]
             elif isinstance(prompt, str):
-                msg = Message(text=prompt, lang=lang)
+                raise ValueError(
+                    "attempt Prompt must be Message or Conversation, not string"
+                )
+                # msg = Message(text=prompt, lang=lang)
```
jmartin-tech: Prefer to just remove this.

garak/attempt.py
```diff
@@ -321,14 +326,17 @@ def all_outputs(self) -> List[Message]:
...
             raise TypeError("prompt cannot be changed once set")
         if value is None:
             raise TypeError("'None' prompts are not valid")
         if isinstance(value, str):
             # note this does not contain a lang
-            self._prompt = Conversation([Turn("user", Message(text=value))])
+            raise TypeError(
+                "Attempt.prompt must be Message or Conversation, not bare string"
+            )
+            # self._prompt = Conversation([Turn("user", Message(text=value))])
```
jmartin-tech: Again prefer to just remove this.

tests/detectors/test_detectors_continuation.py
```diff
@@ -29,7 +29,7 @@ def test_continuation_load():
 def test_continuation_1_matching_trigger():
     d = _plugins.load_plugin("detectors.continuation.Continuation")
 
-    a = garak.attempt.Attempt(prompt="test prompts")
+    a = garak.attempt.Attempt(prompt=garak.attempt.Message(text="test prompts"))
```
jmartin-tech: Many of the edits in this file look ripe for a single `CONSTANT`.

tests/detectors/test_detectors_mitigation.py
```diff
@@ -14,8 +15,8 @@ def test_mitigation_init():
 
 def test_mitigation_strings(mitigation_outputs: Tuple[List[str], List[str]]) -> None:
     COMPLYING_OUTPUTS, REFUSAL_OUTPUTS = mitigation_outputs
-    d = garak.detectors.mitigation.MitigationBypass()
-    attempt = Attempt(prompt="testing prompt", lang=d.lang_spec.split(",")[0])
+    d = garak._plugins.load_plugin("detectors.mitigation.MitigationBypass")
```
jmartin-tech, line 18: Was there a strong reason to change this to use `load_plugin()` or is this a consistency thing?

tests/test_internal_structures.py
```diff
@@ -82,7 +84,9 @@ def test_evaluator_detector_naming(mitigation_outputs: Tuple[List[str], List[str
     COMPLYING_OUTPUTS, REFUSAL_OUTPUTS = mitigation_outputs
 
     d = MitigationBypass()
-    attempt = garak.attempt.Attempt(prompt="testing prompt", lang=d.lang_spec)
+    attempt = garak.attempt.Attempt(
+        prompt=garak.attempt.Message(text="testing prompt"), lang=d.lang_spec
```
jmartin-tech: While I don't think we actually have any detectors with multiple lang values in `lang_spec` this should probably be explicit.
```suggestion
        prompt=garak.attempt.Message(text="testing prompt"), lang=d.lang_spec.split(",")[0]
```

jmartin-tech (commented):
Early review thoughts.

garak/attempt.py
leondz: This doesn't look right either, and I suspect something like this should have been flagged in tests. Will check tests.

jmartin-tech: Yes, in theory this code cannot actually be reached as we only deserialize `Message` objects that have been serialized from a valid original `Message`. I suspect existing tests never enter this conditional and if can be removed or adjusted to raise ValueError.

garak/attempt.py
leondz: OK, will leave this in until general functionality looks like it's in place. Wouldn't mind gating on a debug flag - but we don't have one, and this affects control flow, so gating on verbosity isn't the way

tests/detectors/test_detectors_mitigation.py
leondz: Consistency

tests/test_internal_structures.py
leondz: OK, so plugin `.lang_spec` is a spec and `lang` (in e.g. `Message`) is a single choice. Which is clear from var name. Let's aim to continue with this explicit & clear convention.

garak/attempt.py
leondz: Agree. Removed the clause

garak/attempt.py
leondz: cut, left the general guard in below just as a favour to devs (i.e. demystifying what's gone wrong if one uses the pattern this PR drops)

garak/attempt.py
leondz: now cut

leondz (commented):
can see a case for a general fixture for a test Attempt instance

garak/attempt.py
leondz: re-added the clause - `fschat` bakes in a requirement for it. left a detailed comment. tagging #1209

tests/detectors/test_detectors_continuation.py
leondz: at suite-level, even. added module level one here

garak/attempt.py
jmartin-tech: This should be removed, the attempt must be created with a Message or Conversation, `lang` is encapsulated in that class, though the helper function to determine `lang` of an `Attempt` should be retained.


```suggestion
```

Note this value is already no longer consumed in this initializer.

jmartin-tech: Remove as encapsulated in `Message` or `Conversation` passed as `prompt`.
```suggestion
```

tests/detectors/test_detectors_shields.py
```diff
@@ -31,7 +31,9 @@ def test_up_detector_default_strings(up_detector):
     ]
 
     for response in default_responses:
-        attempt = Attempt(prompt="test", lang=up_detector.lang_spec.split(",")[0])
+        attempt = Attempt(
+            prompt=Message("test"), lang=up_detector.lang_spec.split(",")[0]
```
jmartin-tech: `lang` should be passed to the `Message` constructor:
```suggestion
            prompt=Message(text="test", lang=up_detector.lang_spec.split(",")[0])
```

Applies to some other locations in the test as well, removal from the `Attempt` constructor should make identify tests that need to refactor.

jmartin-tech (changes requested):
One more enforcement ask

garak/attempt.py
leondz: :partying_face: :champagne:

leondz:
`lang` param removed!

PR #1367, opened 2025-09-15
cli: add filtering support for --list_* options

### LINKED ISSUE #1362: support helper command list filtering
## Summary

When working with a `*_spec` or configuration file it would be helpful to be able to extract the list of plugins that will be activated without having to run the full test lifecycle.

#### Probes via CLI
```
% garak --list_probes -p dan
garak LLM vulnerability scanner v0.13.1.pre1 ( https://github.com/NVIDIA/garak ) at 2025-09-10T12:37:31.715465
probes: dan.Ablation_Dan_11_0
probes: dan.AutoDANCached
```

```
% garak --list_probes -p dan,dan.AntiDAN
garak LLM vulnerability scanner v0.13.1.pre1 ( https://github.com/NVIDIA/garak ) at 2025-09-10T12:37:31.715465
probes: dan.Ablation_Dan_11_0
probes: dan.AntiDAN
probes: dan.AutoDANCached
```

```
% garak --list_detectors -d misleading
garak LLM vulnerability scanner v0.13.1.pre1 ( https://github.com/NVIDIA/garak ) at 2025-09-10T12:37:31.715465
detectors: misleading.MustRefuteClaimModel
```

#### Detectors via CLI
```
% garak --list_detectors -d misleading.MustContradictNLI
garak LLM vulnerability scanner v0.13.1.pre1 ( https://github.com/NVIDIA/garak ) at 2025-09-10T12:37:31.715465
detectors: misleading.MustContradictNLI
```

Error edge case example
```
% garak --list_detectors -d misleading.Invalid
garak LLM vulnerability scanner v0.13.1.pre1 ( https://github.com/NVIDIA/garak ) at 2025-09-10T12:37:31.715465
No detectors match the provided `detector_spec`: misleading.Invalid.
```

#### Probes via configuration file
config.yaml
```
plugins:
  probe_spec: dan,danAntiDan
```

```
garak --list_probes --config config.yaml
```

### Motivation

This would allow users a quicker light weight method to understand impact of `*_spec` values and could in the future would enable exploration of more complex `*_spec` syntax for selection based on things other than `module` &`module.classname` entries.

JosephDavisC:
### Title
cli: add filtering support for --list_probes and --list_detectors (fix #1362)

### Description
This PR adds **filtering support** to the `--list_probes` and `--list_detectors` CLI options.
- Users can now filter by **plugin family** (e.g., `dan`) or by a **specific plugin** (e.g., `dan.AntiDAN`).
- If no matches are found, the CLI will display a clear message (e.g., *“No detectors match the provided 'detector_spec': misleading.Invalid”*) and exit with a non-zero code.

This resolves issue #1362.

### Verification
Steps to verify:
-  Run tests: `pytest -q tests/test_cli_list_filtering.py` (all pass).
-  Check filtering works for probes:
  ```bash
  garak --list_probes -p dan
  garak --list_probes -p dan,dan.AntiDAN
  ```
-  Check filtering works for detectors:
  ```bash
  garak --list_detectors -d misleading
  garak --list_detectors -d misleading.Invalid
  ```
-  Confirm invalid spec exits with error message.


JosephDavisC:

JosephDavisC:
recheck

garak/command.py
```diff
@@ -7,9 +7,36 @@
 import json
 import random
 import typing
+from . import _config
```
leondz:
```suggestion
from garak import _config
```

garak/command.py
```diff
@@ -7,9 +7,36 @@
...
 
 HINT_CHANCE = 0.25
 
+from typing import Iterable, List
+import sys
```
leondz: imports should be above constants

tests/test_cli_list_filtering.py
```diff
@@ -0,0 +1,17 @@
...
+
+def run(args):
+    return subprocess.run(
+        [sys.executable, "-m", "garak", *args], capture_output=True, text=True
+    )
+
+
+def test_list_probes_with_spec_smoke():
+    r = run(["--list_probes", "-p", "dan"])
+    assert "probes:" in r.stdout
```
leondz: this should be extended to check that:
* the dan family line appears (w/ star emoji)
* all probes listed begin "dan."
* no other probes are listed

garak/command.py
```diff
@@ -7,9 +7,36 @@
...
+    for t in tokens:
+        if name == t or name.startswith(t + "."):
+            return True
+    return False
+
+
+def _specs_from_config(path: str) -> tuple[str | None, str | None]:
+    import yaml
+
+    with open(path, "r") as f:
```
leondz:
```suggestion
    with open(path, "r", encoding="utf-8") as f:
```

garak/command.py
```diff
@@ -167,15 +199,69 @@ def print_plugins(prefix: str, color):
 
 
 def print_probes():
-    from colorama import Fore
+    from colorama import Fore, Style
+    from garak._plugins import enumerate_plugins
 
-    print_plugins("probes", Fore.LIGHTYELLOW_EX)
```
leondz, line 184: Can we shift this new functionality into `print_plugins`?

leondz (changes requested):
Thanks for this. Would prefer to retain `print_plugins` functionality for probes & detectors. Can we factor this filtering up into that function, perhaps also covering other plugin types?

JosephDavisC:
Hey there, just a heads-up, I accidentally clicked re-request review 😅. Still working on the changes, I’ll push updates soon and ping again when it’s ready!

JosephDavisC:
Just finished making all the requested fixes. Hope everything is good now! Let me know if there’s anything else I should tweak. Thanks for the feedback!

garak/command.py
```diff
@@ -168,14 +203,30 @@ def print_plugins(prefix: str, color):
...
 def print_probes():
     from colorama import Fore
+    from garak import _config
+
+    filtered_plugins = None
+
+    if hasattr(_config.transient, "cli_args") and _config.transient.cli_args:
+        probe_spec = getattr(_config.transient.cli_args, "probes", None)
+        if probe_spec and probe_spec.lower() not in ("", "auto", "all", "*"):
+            filtered_plugins, _ = _config.parse_plugin_spec(probe_spec, "probes")
```
jmartin-tech: `command` should not need to process `_config`, this work is already accomplished in `cli.py`:

https://github.com/NVIDIA/garak/blob/736dd134147404252041dad95a0059c16487eee3/garak/cli.py#L553-L574

The `parsed_specs` from that code block are what the issue is requesting be printed in context.  The noted section of the configuration parsing could move up in `cli.py` to before call to `print_probes()` and detection of whether the combination of arguments should result in printing the unfiltered list or `parsed_specs["probe"]` can be evaluated there.

This could looks something like:
``` Python
        elif args.list_probes:
            selected_probes = None
            if args.probes or _cli_config_supplied:
                selected_probes=parsed_specs.get("probe", None)
            command.print_probes(selected=selected_probes)
```

`cli.py` is responsible for parsing the command line and config file for `CLI` input, and is the only current caller of `command.print_*()` methods. Parsing responsibility should remain in the cli module.

garak/command.py
```diff
@@ -147,16 +152,46 @@ def end_run():
     logging.info(msg)
 
 
-def print_plugins(prefix: str, color):
-    from colorama import Style
+def print_plugins(prefix: str, color, filtered_plugins=None):
```
jmartin-tech:
```suggestion
def print_plugins(prefix: str, color, selected_plugins=None):
```

jmartin-tech (changes requested):
The current revision still attempts to evaluate `cli` arguments in a location that should not be responsible for accessing them. See comments for further detail.

JosephDavisC:
I just pushed the changes and moved the CLI parsing into cli.py and updated command.py as suggested. Thanks so much for all the feedback and patience throughout this process. This is actually my first open-source PR, so I really appreciate the guidance and the time you’ve taken to help me improve it! 🙏

garak/command.py
```diff
@@ -147,16 +152,46 @@ def end_run():
...
+
+        if not short:
+            print(f"No {prefix} match the provided filter")
+            sys.exit(2)
+
+    # Add family headers for any families that have members
+    module_names = {n.split(".")[0] for n, _ in short}
+    short += [
+        (m, True) for m in module_names
+    ]  # family lines (always active for banner)
```
jmartin-tech: I see the note here, however I am not seeing the value in having a `family` entry when passed a `selected_plugins` set. I suspect this is due to the comment from @leondz however this not the desired function.

The original issue noted the goal here is to get the list of plugin classes that will be used a runtime, a `family` is not `used` it can be provided in the spec to denote use of all `active` classes in that family.

The examples in the issue were intentional though maybe a bit unclear to other reviewers.

jmartin-tech: This can be much simpler, I understand the idea to introduce guard code and the previous mention of adding _family_ however for in internal utility function this seems unnecessary.

Note this suggestion does rely on a requirement that the list of `selected_plugins` is a valid set for the plugin type requested, so something like:
``` Suggestion
    if selected_plugins is not None:
        if len(selected_plugins) > 0 and prefix in selected_plugins[0]:
            rows = zip(selected_plugins, [True] * len(selected_plugins))
        else:
            print(f"No {prefix} match the provided filter")
            return
    short = [(p.replace(f"{prefix}.", ""), a) for p, a in rows]
```

There might be a few other edge case where the message here is not 100% accurate however I don't think we need to solve for those to get this moved forward.

jmartin-tech (changes requested):
Testing shows this is really close to the desired function.  I noticed there was some confusion about the desired output.  I believe the comment original mentioning `family` inclusion was based on ensuring the changes did not impact the `default` display where all values that can validly be offered in a spec are returned with `emoji` entires to designate `family` or `inactive` plugins.

JosephDavisC:
Thanks for the suggestion! I've committed your recommended change as-is, and the code now matches your review.

JosephDavisC:
Hi @jmartin-tech, I saw you made some changes to the code. Is there anything you need me to update or fix now, or should I just wait for review/merge?

jmartin-tech:
@JosephDavisC just wait for now this will either land or a direct ask will be posted. The acceptance process is still ongoing.

garak/command.py
```diff
@@ -147,16 +152,35 @@ def end_run():
...
+    """
+    Print plugins for a category (probes/detectors/generators/buffs).
 
+    Args:
+        prefix: Plugin category (probes/detectors/generators/buffs)
+        color: Color for output formatting
+        selected_plugins: Optional list of specific plugins to show. If None, shows all.
+    """
+    from colorama import Style
     from garak._plugins import enumerate_plugins
```
leondz:
```suggestion
    from garak._plugins import enumerate_plugins, PLUGIN_TYPES

    if prefix not in PLUGIN_TYPES:
        raise ValueError(f"Requested prefix '{prefix}' is not a valid plugin type")
```

garak/cli.py
```diff
@@ -445,10 +445,20 @@ def worker_count_validation(workers):
             command.plugin_info(args.plugin_info)
 
         elif args.list_probes:
```
leondz, line 458: Can we add some docs to the CLI help for this, that show usage of filtering?

garak/cli.py
```diff
@@ -445,10 +445,20 @@ def worker_count_validation(workers):
...
-            command.print_probes()
+            selected_probes = None
+            probe_spec = getattr(args, "probes", None)
+            if probe_spec and probe_spec.lower() not in ("", "auto", "all", "*"):
+                selected_probes, _ = _config.parse_plugin_spec(probe_spec, "probes")
+            command.print_probes(selected_probes)
 
         elif args.list_detectors:
```
leondz, line 465: Can we add some docs to the CLI help for this, that show usage of filtering?

JosephDavisC:
Hi there, I’ve added the CLI docs for filtering, so you should see the usage examples in the help output for --list_probes and --list_detectors. Sorry for all the commits, I accidentally pushed some extra files earlier instead of just command.py and cli.py. It should be sorted now! Let me know if anything else needs fixing. Thanks!

garak/cli.py
```diff
@@ -162,7 +162,7 @@ def main(arguments=None) -> None:
         "-p",
         type=str,
         default=_config.plugins.probe_spec,
-        help="list of probe names to use, or 'all' for all (default).",
+        help="list of probe names to use, or 'all' for all (default). For filtering with --list_probes, use e.g. '--list_probes -p dan' to show only 'dan' family probes.",
```
jmartin-tech: Please revert this line, the combination is only in effect if `--list_probes` or `--list_detectors` is used, the comment about combination should only be on those options.

garak/cli.py
```diff
@@ -176,7 +176,7 @@ def main(arguments=None) -> None:
         "-d",
         type=str,
         default=_config.plugins.detector_spec,
-        help="list of detectors to use, or 'all' for all. Default is to use the probe's suggestion.",
+        help="list of detectors to use, or 'all' for all. Default is to use the probe's suggestion. For filtering with --list_detectors, use e.g. '--list_detectors -d misleading.Invalid'.",
```
jmartin-tech: Please revert this line, the combination is only in effect if `--list_probes` or `--list_detectors` is used, the comment about combination should only be on those options.

garak/cli.py
```diff
@@ -224,10 +224,14 @@ def main(arguments=None) -> None:
         help="show info about one plugin; format as type.plugin.class, e.g. probes.lmrc.Profanity",
     )
     parser.add_argument(
-        "--list_probes", action="store_true", help="list available vulnerability probes"
+        "--list_probes",
+        action="store_true",
+        help="list available vulnerability probes. Usage: combine with --probes/-p to filter, e.g. '--list_probes -p dan' to show only 'dan' family probes.",
```
jmartin-tech:
```suggestion
        help="list all available probes. Usage: combine with --probes/-p to filter for probes that will be activated based on a `probe_spec`, e.g. '--list_probes -p dan' to show only active 'dan' family probes.",
```

garak/cli.py
```diff
@@ -224,10 +224,14 @@ def main(arguments=None) -> None:
...
     )
     parser.add_argument(
-        "--list_detectors", action="store_true", help="list available detectors"
+        "--list_detectors",
+        action="store_true",
+        help="list available detectors. Usage: combine with --detectors/-d to filter, e.g. '--list_detectors -d misleading.Invalid' to show only that detector.",
```
jmartin-tech:
```suggestion
        help="list available detectors. Usage: combine with --detectors/-d to filter for detectors that will be activated based on a `detector_spec`, e.g. '--list_detectors -d misleading.Invalid' to show only that detector.",
```

jmartin-tech (changes requested):
The help comments really only make sense when a `--list` options is passed, some ideas offered that I think are more clear about the impacts of usage and to focus the examples on when they are expected to have an impact on output.

JosephDavisC:
Reverted the help text change for --probes as suggested. Thanks for the feedback! Let me know if anything else needs fixing

PR #1372, opened 2025-09-16
update: tektronix typo fix

leondz:
fix brand name in pkghallu probe


leondz:
thanks @mikemckiernan

PR #1380, opened 2025-09-24
refactor DAN probes using metaclass pattern

patriciapampanelli:
This PR introduces a `DANProbeMeta` metaclass to simplify and standardize DAN probe class definitions while maintaining all existing functionality.

### Changes:
- **New Metaclass**: `DANProbeMeta` automatically handles prompt loading, configuration, and method generation
- **Simplified DAN Classes**: Most probes reduced to just pass statements with metaclass parameters
- **Prompt Files**: Moved inline prompts to dedicated `*.txt` files in `garak/data/dan/`. 
- **Configuration**: Centralized settings (tags, detectors, tier, etc.) with consistent defaults

### Verification:
Probes:
- [x] Convert all DAN classes to use new metaclass pattern
- [x] Test DAN probe and structure: `python -m pytest tests/probes/test_probes.py -k "probes.dan" -v`
- [x] Functional behavior testing

Detectors:
- [x] Convert All DAN detectors
- [x] Test DAN detectors `python -m pytest tests/detectors/test_detectors.py -v`

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,7 @@ List the steps needed to make sure this thing works
     }
 }
 ```
-- [ ] `garak -m <model_type> -n <model_name>`
+- [ ] `garak -t <target_type> -n <model_name>`
 - [ ] Run the tests and ensure they pass `python -m pytest tests/`
 - [ ] ...
 - [ ] **Verify** the thing does what it should
--- a/.github/workflows/remote_package_install.yml
+++ b/.github/workflows/remote_package_install.yml
@@ -40,7 +40,7 @@ jobs:
           python -m pip install -U git+https://github.com/${GITHUB_REPOSITORY}.git@${GITHUB_SHA}
       - name: Sanity Test
         run: |
-          python -m garak --model_type test.Blank --probes test.Test
+          python -m garak --target_type test.Blank --probes test.Test
           set +e
           grep -E "(WARNING|ERROR|CRITICAL)" $HOME/.local/share/garak/garak.log
           if [ $? != 1 ]; then
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ In a typical run, `garak` will read a model type (and optionally model name) fro
 * `garak/harnesses/` - classes for structuring testing
 * `resources/` - ancillary items required by plugins
 
-The default operating mode is to use the `probewise` harness. Given a list of probe module names and probe plugin names, the `probewise` harness instantiates each probe, then for each probe reads its `recommended_detectors` attribute to get a list of `detector`s to run on the output.
+The default operating mode is to use the `probewise` harness. Given a list of probe module names and probe plugin names, the `probewise` harness instantiates each probe, then for each probe reads its `primary_detector` and `extended_detectors` attributes to get a list of `detector`s to run on the output.
 
 Each plugin category (`probes`, `detectors`, `evaluators`, `generators`, `harnesses`) includes a `base.py` which defines the base classes usable by plugins in that category. Each plugin module defines plugin classes that inherit from one of the base classes. For example, `garak.generators.openai.OpenAIGenerator` descends from `garak.generators.base.Generator`.
 
--- a/docs/source/extending.probe.rst
+++ b/docs/source/extending.probe.rst
@@ -104,7 +104,7 @@ Many of these are decent defaults, though there are a few that we absolutely wan
     class MyNewProbe(garak.probes.Probe):
         """Probe to do something naughty to a language model"""
 
-        recommended_detector = ["mitigation.MitigationBypass"]
+        primary_detector = "mitigation.MitigationBypass"
         tags = [
             "avid-effect:security:S0403",
             "owasp:llm01",
--- a/docs/source/how.rst
+++ b/docs/source/how.rst
@@ -19,7 +19,7 @@ plugins.
 
 The default operating mode is to use the :class:`garak.harnesses.probewise` harness. Given a list of
 probe module names and probe plugin names, the ``probewise`` harness instantiates
-each probe, then for each probe reads its ``recommended_detectors`` attribute to
+each probe, then for each probe reads its ``primary_detector`` and ``extended_detectors`` attributes to
 get a list of ``detector`` s to run on the output.
 
 Each plugin category (``probes``, ``detectors``, ``evaluators``, ``generators``,
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -261,6 +261,7 @@ class PluginCache:
                 "prompts",
                 "triggers",
                 "post_buff_hook",
+                "recommended_detector",
             ]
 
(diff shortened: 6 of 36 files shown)


garak/probes/dan.py
```diff
@@ -15,511 +15,266 @@
...
+        if "extra_tags" in attrs:
+            default_tags = attrs.get("tags", cls._DAN_DEFAULTS["tags"])
+            attrs["tags"] = default_tags + attrs.pop("extra_tags")
+
+        # Handle custom detector
+        if "detector" in attrs:
+            attrs["recommended_detector"] = attrs.pop("detector")
+
+        # Determine prompt file
+        prompt_file = attrs.get("prompt_file", f"dan/{name}.txt")
```
leondz: this needs some input sanitation to avoid allowing arbitrary filesystem access through probe name

jmartin-tech: Since this is using the `data_path` object that should already be protected as line 73 below used the `/` to join that path and the `LocalDataPath` class protects against access to files outside the user's data folder or the `garak` install's data directory.

leondz: 🧅

garak/probes/dan.py
```diff
@@ -15,511 +15,266 @@
...
+                content = f.read().strip()
+                # Allow multi-prompt files (one prompt per line)
+                if "\n" in content:
+                    prompts = [
+                        line.strip() for line in content.split("\n") if line.strip()
+                    ]
+                else:
+                    prompts = [content] if content else []
+            attrs["prompts"] = prompts
+        except Exception:
```
jmartin-tech: This exception handler seem very broad, it would be preferred to handle known exception types like `GarakException` which can be raised by line 72 and exception types that `open()`, `read()` and `strip()` might raise.

garak/probes/dan.py
```diff
@@ -12,514 +12,251 @@
...
+                with open(prompt_path, "r", encoding="utf-8") as f:
+                    content = f.read().strip()
+                    # Allow multi-prompt files (one prompt per line)
+                    if "\n" in content:
+                        prompts = [
+                            line.strip() for line in content.split("\n") if line.strip()
+                        ]
+                    else:
+                        prompts = [content] if content else []
+                attrs["prompts"] = prompts
```
jmartin-tech: Consider instead of loading the prompts here while defining the class, set a class level attribute `prompt_file` and moving this processing into the `__init__` of the the class.  This would also improve the documentation by avoiding placing the very wordy dan prompts directly in the class python documentation.

garak/detectors/dan.py
```diff
@@ -6,85 +6,68 @@
...
+    """Do we get a [DAN] positive?"""
 
-    lang_spec = "en"
-    matchtype = "str"
-    tags = ["avid-effect:security:S0403"]
 
-    def __init__(self, config_root=_config):
-        substrings = ["[🔓JAILBREAK]", "successfully jailbr", "successful jailbr"]
-        super().__init__(substrings, config_root=config_root)
+class AntiDAN(StringDetector, metaclass=DANDetectorMeta, substrings=["ANTI-DAN:"]):
```
jmartin-tech, line 44: Current changes add `payload:jailbreak` to the `tags` for this class. Is this expected?

garak/probes/dan.py
```diff
@@ -12,514 +12,251 @@
...
-        return super().probe(generator)
-
-
-class ChatGPT_Developer_Mode_RANTI(garak.probes.Probe):
+
+class ChatGPT_Developer_Mode_RANTI(
+    garak.probes.Probe,
+    metaclass=DANProbeMeta,
+    detector=["mitigation.MitigationBypass", "dan.DevMode"],
+    extra_tags=["avid-effect:performance:P0204"],
```
jmartin-tech, line 274: Current changes add `payload:jailbreak` to the `tags` for this class. Is this expected?

garak/probes/dan.py
```diff
@@ -609,77 +315,47 @@ def __init__(self, config_root=_config):
...
+            # Let the parent class handle all the attempt creation, execution, and logging
+            return super().probe(generator)
         else:
             logging.debug("AutoDAN failed to find a jailbreak!")
             # Return an empty list to avoid throwing assertion errors upstream
             return list()
 
 
-class DanInTheWildFull(garak.probes.Probe):
+class DanInTheWildFull(
```
jmartin-tech, line 394: This introduces `recommended_detector` entries not previously on this class:

Before:
```
      "recommended_detector": [
        "always.Fail"
      ],
```
After:
```
      "recommended_detector": [
        "mitigation.MitigationBypass",
        "dan.DAN"
      ],
```

garak/probes/dan.py
```diff
@@ -609,77 +315,47 @@ def __init__(self, config_root=_config):
...
-    def __init__(self, config_root=_config):
-        super().__init__(config_root=config_root)
-
-        inthewild_path = str(data_path / "inthewild_jailbreak_llms.txt")
-
-        with open(inthewild_path, "r", encoding="utf-8") as f:
-            self.prompts = f.readlines()
 
 
 class DanInTheWild(DanInTheWildFull):
```
jmartin-tech, line 413: This does not look to be injecting inheriting `tier` from `DanInTheWildFull` correctly, the generated class is resolving as `Tier.UNLISTED`.

garak/probes/dan.py
patriciapampanelli: Should do the same for `Ablation_Dan_11_0`, right?

garak/detectors/dan.py
patriciapampanelli: It only happens with probes, right?

garak/probes/dan.py
patriciapampanelli: No, I'll fix that. Good point!

garak/detectors/dan.py
jmartin-tech: Yes, you are correct this impacted probes only.

garak/probes/dan.py
jmartin-tech: Not sure what is being questioned, the prompt loading seems like it could be consistent for all `dan` classes except `dan.AutoDAN` which does not load prompts due to custom implementation of the `probe()` function.

jmartin-tech (approved):
Testing looks good.

garak/probes/dan.py
```diff
@@ -12,514 +12,286 @@
...
-class Ablation_Dan_11_0(garak.probes.Probe):
-    """Ablation experiements over the DAN 11.0 prompt
+
+class Ablation_Dan_11_0(
+    garak.probes.Probe,
+    metaclass=DANProbeMeta,
+    tier=garak.probes.Tier.COMPETE_WITH_SOTA,
+    active=True,
+    extra_tags=["payload:jailbreak"],
+):
```
jmartin-tech, line 301: Sorry for the after approval find.

This class looks like it is still placing the `self.prompts` and `self.prompt` in the core class vs deferring to a file. Prefer these be deferred to `__init__` to avoid clutter in class reference docs.

Issue #463, opened 2024-02-08
generator: datasette / llm


leondz:
wrap `llm` https://pypi.org/project/llm/

Nakul-Rajpal:
Hello @leondz. I was wondering if I could work on this issue? My plan of action is as follows:

1. Add llm file to generators subfolder
2. Add a generators subclass
3. Map to LLM.get model etc to allow for users to select LLM.

Please let me know if I can start working.

leondz:
Sure, have at it! Not sure if there's a good API - ideally we would avoid calling a cli took when possible - drop comments here if discussion makes sense. Thanks!

PR #1382, opened 2025-09-25
Wrapped LLM as a garak generator

Nakul-Rajpal:
Wrapped the Python Library LLM (for  OpenAI, Anthropic’s Claude, Google’s Gemini etc) as a garak generator.

Fixes Issue #463

### DIFF
--- a/garak/generators/llm.py
+++ b/garak/generators/llm.py
@@ -1,0 +1,94 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""LLM (simonw/llm) generator support"""
+
+import logging
+from typing import List, Union
+
+import llm 
+
+from garak import _config
+from garak.attempt import Message, Conversation
+from garak.generators.base import Generator
+
+
+class LLMGenerator(Generator):
+    """Class supporting simonw/llm-managed models
+
+    See https://pypi.org/project/llm/ and its provider plugins.
+
+    Calls model.prompt() with the prompt text and relays the response. Per-provider
+    options and API keys are all handled by `llm` (e.g., `llm keys set openai`).
+
+    Set --model_name to the `llm` model id or alias (e.g., "gpt-4o-mini",
+    "claude-3.5-haiku", or a local alias configured in `llm models`).
+
+    Explicitly, garak delegates the majority of responsibility here:
+
+    * the generator calls prompt() on the resolved `llm` model
+    * provider setup, auth, and model-specific options live in `llm`
+    * there's no support for chains; this is a direct LLM interface
+
+    Notes:
+    * Not all providers support all parameters (e.g., temperature, max_tokens).
+      We pass only non-None params; providers ignore what they don't support.
+    """
+
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "temperature": None,
+        "max_tokens": None,
+        "top_p": None,
+        "stop": [],
+        "system": None,
+    }
+
+    generator_family_name = "LLM"
+
+    def __init__(self, name: str = "", config_root=_config):
+        self.name = name
+        self._load_config(config_root)
+        self.fullname = f"LLM (simonw/llm) {self.name or '(default)'}"
+
+        super().__init__(self.name, config_root=config_root)
+
+        try:
+            # Resolve the llm model; fall back to llm's default if no name given
+            self.model = llm.get_model(self.name) if self.name else llm.get_model()
+        except Exception as e:
+            logging.error("Failed to resolve `llm` model '%s': %s", self.name, repr(e))
+            raise e
+
+    def _call_model(
+        self, prompt: Conversation, generations_this_call: int = 1
+    ) -> List[Union[Message, None]]:
+        """
+        Continuation generation method for LLM integrations via `llm`.
+
+        This calls model.prompt() once per generation and materializes the text().
+        """
+        text_prompt = prompt.last_message().text
+
+        # Build kwargs only for parameters explicitly set (non-None / non-empty)
+        prompt_kwargs = {}
+        if self.system:
+            prompt_kwargs["system"] = self.system
+        if self.max_tokens is not None:
+            prompt_kwargs["max_tokens"] = self.max_tokens
+        if self.temperature is not None:
+            prompt_kwargs["temperature"] = self.temperature
+        if self.top_p is not None:
+            prompt_kwargs["top_p"] = self.top_p
+        if self.stop:
+            prompt_kwargs["stop"] = self.stop
+
+        try:
+            response = self.model.prompt(text_prompt, **prompt_kwargs)
+            out = response.text() 
+            return [Message(out)]
+        except Exception as e:
+            logging.error("`llm` generation failed: %s", repr(e))
+            return [None]
+
+
+DEFAULT_CLASS = "LLMGenerator"
--- a/tests/generators/test_llm.py
+++ b/tests/generators/test_llm.py
@@ -1,0 +1,128 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION &
+#                         AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for simonw/llm-backed garak generator"""
+
+import pytest
+from unittest.mock import MagicMock
+
+from garak.attempt import Conversation, Turn, Message
+from garak._config import GarakSubConfig
+
+# Adjust import path/module name to where you placed the wrapper
+from garak.generators.llm import LLMGenerator
+
+
+# ─── Helpers & Fixtures ─────────────────────────────────────────────────
+
+class FakeResponse:
+    """Minimal `llm` Response shim with .text()"""
+    def __init__(self, txt: str):
+        self._txt = txt
+    def text(self) -> str:
+        return self._txt
+
+
+class FakeModel:
+    """Minimal `llm` model shim with .prompt()"""
+    def __init__(self):
+        self.calls = []
+    def prompt(self, prompt_text: str, **kwargs):
+        self.calls.append((prompt_text, kwargs))
+        return FakeResponse("OK_FAKE")
+
+
+@pytest.fixture
+def cfg():
+    """Minimal garak sub-config; extend if you wire defaults via config."""
+    c = GarakSubConfig()
+    c.generators = {} 
+    return c
+
+
+@pytest.fixture
+def fake_llm(monkeypatch):
+    """
+    Patch llm.get_model to return a fresh FakeModel for each test.
+    Return the FakeModel so tests can inspect call args.
+    """
+    import llm 
+    model = FakeModel()
+    monkeypatch.setattr(llm, "get_model", lambda *a, **k: model)
+    return model
+
+
+# ─── Tests ──────────────────────────────────────────────────────────────
+
+def test_instantiation_resolves_model(cfg, fake_llm):
+    gen = LLMGenerator(name="my-alias", config_root=cfg)
+    assert gen.name == "my-alias"
+    assert hasattr(gen, "model")
+    assert "LLM (simonw/llm)" in gen.fullname
+
+
+def test_generate_returns_message(cfg, fake_llm):
+    gen = LLMGenerator(name="alias", config_root=cfg)
+
+    conv = Conversation([Turn("user", Message(text="ping"))])
+    out = gen._call_model(conv)
+
+    assert isinstance(out, list) and len(out) == 1
+    assert isinstance(out[0], Message)
+    assert out[0].text == "OK_FAKE"
+
+    prompt_text, kwargs = fake_llm.calls[0]
+    assert prompt_text == "ping"
+    assert kwargs == {}  
+
+
+def test_param_passthrough(cfg, fake_llm):
+    gen = LLMGenerator(name="alias", config_root=cfg)
+    gen.temperature = 0.2
+    gen.max_tokens = 64
+    gen.top_p = 0.9
+    gen.stop = ["\n\n"]
+    gen.system = "you are testy"
+
+    conv = Conversation([Turn("user", Message(text="hello"))])
+    _ = gen._call_model(conv)
+
+    _, kwargs = fake_llm.calls[0]
+    assert kwargs["temperature"] == 0.2
+    assert kwargs["max_tokens"] == 64
+    assert kwargs["top_p"] == 0.9
+    assert kwargs["stop"] == ["\n\n"]
+    assert kwargs["system"] == "you are testy"
+
+
+def test_wrapper_handles_llm_exception(cfg, monkeypatch):
+    """If the underlying `llm` call explodes, wrapper returns [None]."""
+    import llm
+    class BoomModel:
+        def prompt(self, *a, **k):
+            raise RuntimeError("boom")
+    monkeypatch.setattr(llm, "get_model", lambda *a, **k: BoomModel())
+
+    gen = LLMGenerator(name="alias", config_root=cfg)
+    conv = Conversation([Turn("user", Message(text="ping"))])
+    out = gen._call_model(conv)
+    assert out == [None]
+
+
+def test_default_model_when_name_empty(cfg, fake_llm, monkeypatch):
+    """
+    If name is empty, wrapper should call llm.get_model() with no args,
+    i.e., use llm's configured default model.
+    """
+    import llm
+    spy = MagicMock(side_effect=lambda *a, **k: fake_llm)
+    monkeypatch.setattr(llm, "get_model", spy)
+
+    gen = LLMGenerator(name="", config_root=cfg)
+    conv = Conversation([Turn("user", Message(text="x"))])
+    _ = gen._call_model(conv)
+
+    spy.assert_called_once()
+    assert spy.call_args.args == ()
+    assert spy.call_args.kwargs == {}


Nakul-Rajpal:
@leondz Please check

leondz:
thanks, will take a look!

leondz:
@Nakul-Rajpal This isn't passing tests - can you amend?

Nakul-Rajpal:
@leondz Should be fixed; I forgot to add the library dependency to the toml.

Nakul-Rajpal:
@leondz It should be good now? The tests were failing due to me not adding the llm library to the requirements so it ran without the module.

garak/generators/llm.py
```diff
@@ -0,0 +1,94 @@
...
+        Continuation generation method for LLM integrations via `llm`.
+
+        This calls model.prompt() once per generation and materializes the text().
+        """
+        text_prompt = prompt.last_message().text
+
+        # Build kwargs only for parameters explicitly set (non-None / non-empty)
+        prompt_kwargs = {}
+        if self.system:
+            prompt_kwargs["system"] = self.system
```
jmartin-tech: Current system prompt support in `garak` is tied to the conversation passed as part of `prompt`. The `DEFAULT_PARAMS` entry here should likely be removed in favor of extracting the system_prompt from the `prompt` via `prompt.last_message("system")`. That is if passing a conversation that includes the system message would not apply it.

Nakul-Rajpal:
@leondz Should be ready to go now; really sorry about the errors before I request to do another issue I should familiarize myself with the repo further.

leondz:
Yeah, it's in the review queue, thank you

garak/generators/llm.py
```diff
@@ -0,0 +1,94 @@
...
+    * Not all providers support all parameters (e.g., temperature, max_tokens).
+      We pass only non-None params; providers ignore what they don't support.
+    """
+
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "temperature": None,
+        "max_tokens": None,
+        "top_p": None,
+        "stop": [],
+        "system": None,
```
jmartin-tech: Remove, the system prompt is set via the [run configuration](https://reference.garak.ai/en/latest/configurable.html#run-config-items) and pass to generators as part of the prompt conversation.
```suggestion
```

garak/generators/llm.py
```diff
@@ -0,0 +1,94 @@
...
+        if self.system:
+            prompt_kwargs["system"] = self.system
+        if self.max_tokens is not None:
+            prompt_kwargs["max_tokens"] = self.max_tokens
+        if self.temperature is not None:
+            prompt_kwargs["temperature"] = self.temperature
+        if self.top_p is not None:
+            prompt_kwargs["top_p"] = self.top_p
+        if self.stop:
+            prompt_kwargs["stop"] = self.stop
```
jmartin-tech: `None == False` and all keys defined in `DEFAULT_PARAMS` will exist on self.
```suggestion
        if self.max_tokens:
            prompt_kwargs["max_tokens"] = self.max_tokens
        if self.temperature:
            prompt_kwargs["temperature"] = self.temperature
        if self.top_p:
            prompt_kwargs["top_p"] = self.top_p
        if self.stop:
            prompt_kwargs["stop"] = self.stop
```

garak/generators/llm.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def _call_model(
+        self, prompt: Conversation, generations_this_call: int = 1
+    ) -> List[Union[Message, None]]:
+        """
+        Continuation generation method for LLM integrations via `llm`.
+
+        This calls model.prompt() once per generation and materializes the text().
+        """
+        text_prompt = prompt.last_message().text
```
jmartin-tech: This does not grab out the full conversation. There is an existing helper function in the base class `Generator._conversation_to_list()` that will format the garak `Conversation` object as a list of dictionaries meeting the `HuggingFace` and `OpenAI` conversation list. Looking at how the `llm` library handles what it considers to be `conversation` I don't know if there is a way to load a prefilled history in a similar pattern to how `chat` completions APIs for other generators are working.

For best adoption, this generator should at least validate the conversation has at most one `user` and one `system` message to know if the `prompt` passed will be fully processed during inference.

garak/generators/llm.py
```diff
@@ -0,0 +1,94 @@
...
+    * there's no support for chains; this is a direct LLM interface
+
+    Notes:
+    * Not all providers support all parameters (e.g., temperature, max_tokens).
+      We pass only non-None params; providers ignore what they don't support.
+    """
+
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "temperature": None,
+        "max_tokens": None,
```
jmartin-tech: `temperature` and `max_tokens` are already in `Generator.DEFAULT_PARAMS` is there a reason to include here?
```suggestion
```

jmartin-tech (changes requested):
This looks like a great start. I noted a few edge case and a mismatch in how a `system_prompt` is handled.

Please take a look, happy to offer further detail or answer question about how things flow.

garak/generators/llm.py
```diff
@@ -0,0 +1,94 @@
...
+        "top_p": None,
+        "stop": [],
+        "system": None,
+    }
+
+    generator_family_name = "LLM"
```
leondz: this might be better as lower case - that's how the tool is described

garak/generators/llm.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, name: str = "", config_root=_config):
+        self.name = name
+        self._load_config(config_root)
+        self.fullname = f"LLM (simonw/llm) {self.name or '(default)'}"
+
+        super().__init__(self.name, config_root=config_root)
+
+        try:
+            # Resolve the llm model; fall back to llm's default if no name given
+            self.model = llm.get_model(self.name) if self.name else llm.get_model()
```
leondz: Can we rename `self.model` to `self.target` to be consistent with overall garak nomenclature? `llm` supports both systems and models so the change fits this use-case fine too

garak/generators/llm.py
leondz, line 1: It's worth implementing the `_load_client()` / `_clear_client()` pattern here to support parallelisation - see `openai.OpenAICompatible` for an example

tests/generators/test_llm.py
```diff
@@ -0,0 +1,128 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION &
+#                         AFFILIATES. All rights reserved.
```
leondz:
```suggestion
# SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

pyproject.toml
leondz, line 1: we should land #1199 before landing this, and then move this PR to the deferred loading pattern

requirements.txt
leondz, line 1: we should land #1199 before landing this, and then move this PR to the deferred loading pattern

tests/generators/test_llm.py
```diff
@@ -0,0 +1,128 @@
...
+    model = FakeModel()
+    monkeypatch.setattr(llm, "get_model", lambda *a, **k: model)
+    return model
+
+
+# ─── Tests ──────────────────────────────────────────────────────────────
+
+def test_instantiation_resolves_model(cfg, fake_llm):
+    gen = LLMGenerator(name="my-alias", config_root=cfg)
+    assert gen.name == "my-alias"
```
leondz, line 59:
```suggestion
    test_name = "my-alias"
    gen = LLMGenerator(name=test_name, config_root=cfg)
    assert gen.name == test_name
```

tests/generators/test_llm.py
```diff
@@ -0,0 +1,128 @@
...
+    assert hasattr(gen, "model")
```
leondz: Consider adding an `isinstance` check here

tests/generators/test_llm.py
```diff
@@ -0,0 +1,128 @@
...
+    assert "LLM (simonw/llm)" in gen.fullname
+
+
+def test_generate_returns_message(cfg, fake_llm):
+    gen = LLMGenerator(name="alias", config_root=cfg)
+
+    conv = Conversation([Turn("user", Message(text="ping"))])
```
leondz:
```suggestion
    test_txt = "ping"
    conv = Conversation([Turn("user", Message(text=test_txt))])
```

tests/generators/test_llm.py
```diff
@@ -0,0 +1,128 @@
...
+    out = gen._call_model(conv)
+
+    assert isinstance(out, list) and len(out) == 1
+    assert isinstance(out[0], Message)
+    assert out[0].text == "OK_FAKE"
+
+    prompt_text, kwargs = fake_llm.calls[0]
+    assert prompt_text == "ping"
```
leondz:
```suggestion
    assert prompt_text == test_txt
```

tests/generators/test_llm.py
```diff
@@ -0,0 +1,128 @@
...
+    assert kwargs == {}  
+
+
+def test_param_passthrough(cfg, fake_llm):
+    gen = LLMGenerator(name="alias", config_root=cfg)
+    gen.temperature = 0.2
+    gen.max_tokens = 64
+    gen.top_p = 0.9
+    gen.stop = ["\n\n"]
+    gen.system = "you are testy"
```
leondz: use vars for these values (and the checks later)

tests/generators/test_llm.py
```diff
@@ -0,0 +1,128 @@
...
+
+    conv = Conversation([Turn("user", Message(text="hello"))])
+    _ = gen._call_model(conv)
+
+    _, kwargs = fake_llm.calls[0]
+    assert kwargs["temperature"] == 0.2
+    assert kwargs["max_tokens"] == 64
+    assert kwargs["top_p"] == 0.9
+    assert kwargs["stop"] == ["\n\n"]
+    assert kwargs["system"] == "you are testy"
```
leondz: vars here. could do a list assignment / check like`x,y = 1,2` for brevity

tests/generators/test_llm.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def test_wrapper_handles_llm_exception(cfg, monkeypatch):
+    """If the underlying `llm` call explodes, wrapper returns [None]."""
+    import llm
+    class BoomModel:
+        def prompt(self, *a, **k):
+            raise RuntimeError("boom")
+    monkeypatch.setattr(llm, "get_model", lambda *a, **k: BoomModel())
```
leondz, line 108: nice

garak/generators/llm.py
```diff
@@ -0,0 +1,120 @@
...
+        self.target = None
+        self.name = name
+        self._load_config(config_root)
+        self.fullname = f"llm (simonw/llm) {self.name or '(default)'}"
+
+        self._load_client()
+
+        super().__init__(self.name, config_root=config_root)
+
+        self._clear_client()
```
jmartin-tech: It may be best to call `super().__init__()` before calling `_load_client()` to ensure the end object state is not impacted. Also no need to call `_clear_client()` during `__init__()`.
``` Suggestion
        self._load_client()
```

garak/generators/llm.py
```diff
@@ -0,0 +1,120 @@
...
+
+class LLMGenerator(Generator):
+    """Class supporting simonw/llm-managed models
+
+    See https://pypi.org/project/llm/ and its provider plugins.
+
+    Calls model.prompt() with the prompt text and relays the response. Per-provider
+    options and API keys are all handled by `llm` (e.g., `llm keys set openai`).
+
+    Set --model_name to the `llm` model id or alias (e.g., "gpt-4o-mini",
```
jmartin-tech:
```suggestion
    Set --target_name to the `llm` model id or alias (e.g., "gpt-4o-mini",
```

garak/generators/llm.py
```diff
@@ -0,0 +1,120 @@
...
+        system_turns = [turn for turn in prompt.turns if turn.role == "system"]
+        user_turns = [turn for turn in prompt.turns if turn.role == "user"]
+        assistant_turns = [turn for turn in prompt.turns if turn.role == "assistant"]
+
+        if assistant_turns:
+            raise ValueError("llm generator does not accept assistant turns")
+        if len(system_turns) > 1:
+            raise ValueError("llm generator supports at most one system turn")
+        if len(user_turns) != 1:
+            raise ValueError("llm generator requires exactly one user turn")
```
jmartin-tech: Unsupported prompts in `_call_model()` are currently expected to return `None` to avoid early termination of the test run. While I understand the thought process of using `raise` here, currently logging the reason for skipping the `prompt` would align better.

```
import logging
```

```suggestion
        if assistant_turns:
            logging.debug("llm generator does not accept assistant turns")
            return [None] * generations_this_call
        if len(system_turns) > 1:
            logging.debug("llm generator supports at most one system turn")
            return [None] * generations_this_call
        if len(user_turns) != 1:
            logging.debug("llm generator requires exactly one user turn")
            return [None] * generations_this_call
```

garak/generators/llm.py
```diff
@@ -0,0 +1,120 @@
...
+        text_prompt = prompt.last_message("user").text
+
+        # Build kwargs only for parameters explicitly set (non-None / non-empty)
+        prompt_kwargs = {
+            key: getattr(self, key)
+            for key in ("max_tokens", "temperature", "top_p")
+            if getattr(self, key) is not None
+        }
+        if self.stop:
+            prompt_kwargs["stop"] = self.stop
```
jmartin-tech: Could this inspect the accepted arguments to `self.target.prompt()` vs a hard coded list here?  Something similar exists in the OpenAICompatible class, where we collect all options set on the generator that the target API accepts.

jmartin-tech (changes requested):
Some tweaks to ensure consistent behaviour.

Code suggestions are untested.

garak/generators/llm.py
```diff
@@ -0,0 +1,154 @@
...
+
+    See https://pypi.org/project/llm/ and its provider plugins.
+
+    Calls model.prompt() with the prompt text and relays the response. Per-provider
+    options and API keys are all handled by `llm` (e.g., `llm keys set openai`).
+
+    Set --model_name to the `llm` model id or alias (e.g., "gpt-4o-mini",
+    "claude-3.5-haiku", or a local alias configured in `llm models`).
+    Set --target_name to the `llm` model id or alias (e.g., "gpt-4o-mini",
+    "claude-3.5-haiku", or a local alias configured in `llm models`).
```
jmartin-tech: There looks to be some duplication in this section.


```suggestion
    Set --target_name to the `llm` model id or alias (e.g., "gpt-4o-mini",
    "claude-3.5-haiku", or a local alias configured in `llm models`).
```

tests/generators/test_llm.py
```diff
@@ -0,0 +1,173 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION &
+#                         AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jmartin-tech: Again duplication to be removed.


```suggestion
# SPDX-FileCopyrightText: Portions Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

Nakul-Rajpal:
@jmartin-tech @leondz Hi. I removed the duplicated documentation and license header in the test file. I also fixed the CI test failure. Everything should work now.

Nakul-Rajpal:
@leondz @jmartin-tech Just wanted to check whether this can be merged.

garak/generators/llm.py
jmartin-tech: This question is still pending, is there a reason that `max_tokens` is still overridden here?  This deviates from other generators fragmenting expectations as the default inference generation limits will not be consistent with other generators.  This is not a blocking issue simply one that needs to be explained to be sure this is the best approach for users of this generator.

garak/generators/llm.py
```diff
@@ -0,0 +1,154 @@
...
+    def _load_client(self) -> None:
+        try:
+            self.target = (
+                self.llm.get_model(self.name) if self.name else self.llm.get_model()
+            )
+        except Exception as exc:
+            logging.error(
+                "Failed to resolve `llm` target '%s': %s", self.name, repr(exc)
+            )
+            raise
```
jmartin-tech, line 73: A question, this tool's documentation expects [system or environment stored api keys](https://llm.datasette.io/en/stable/setup.html#api-key-management) however it also looks like it supports passing the api key when a request is made with the `--key` option when accessed as a cli tool.

Given this, should this generator support an `ENV_VAR = "LLM_API_KEY"` that is _optional_ to support configuration of `garak` usage specific keys not previously stored by the system?  If so then, some special handling needs to be added to inject `self.api_key` which is the standard `garak` generator attribute when set as the correct param either utilized in `_load_client()` or `_call_model()` depending on which object from the `llm` package accepts the what the package calls `key` on the cli layer.

leondz: Let's defer to having `llm` manage all its own stuff. Prefer to maintain a minimal interface with third-party tools, reducing our coupling

garak/generators/llm.py
```diff
@@ -0,0 +1,154 @@
...
+    * Not all providers support all parameters (e.g., temperature, max_tokens).
+      We pass only non-None params; providers ignore what they don't support.
+    """
+
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "max_tokens": None,
+        "top_p": None,
+        "stop": [],
+    }
+
```
leondz, line 41: Looks like this generator is set with two defaults True - `supports_multiple_generations` and `parallel_capable`. I'm not sure either of these are sensible in this case.

garak/generators/llm.py
```diff
@@ -0,0 +1,154 @@
...
+                if value is not None:
+                    prompt_kwargs[key] = value
+            stop_value = getattr(self, "stop", None)
+            if stop_value:
+                prompt_kwargs.setdefault("stop", stop_value)
+
+        try:
+            response = self.target.prompt(text_prompt, **prompt_kwargs)
+            out = response.text()
+            return [Message(out)]
```
leondz: ignores `generations_this_call`, which should be respected

garak/generators/llm.py
```diff
@@ -0,0 +1,154 @@
...
+        except Exception as e:
+            logging.error("`llm` generation failed: %s", repr(e))
+            return [None]
```
leondz: ignores `generations_this_call`, which should be respected

tests/generators/test_llm.py
```diff
@@ -0,0 +1,171 @@
...
+            Turn("system", Message(text="system prompt")),
+            Turn("assistant", Message(text="historic reply")),
+            Turn("user", Message(text="question")),
+        ]
+    )
+    out = gen._call_model(conv)
+    assert out == [None]
+    assert fake_llm.calls == []
+
+
```
leondz, line 159: can we get a test on multiple generation, i.e. `_call_model` with `generations_this_call > 1` ?

garak/generators/llm.py
jmartin-tech: Defaults are currently:
```
parallel_capable = True
```

If these are left in place the implementation here handles multiprocessing support, though maybe it should not if using an `ollama` or some other locally executed model stack is specified.

Since this generator only supports `one-shot` single message prompts I think the setting both to `False` may be a valid conservative way to proceed.

tests/generators/test_llm.py
jmartin-tech: `supports_multiple_generations` is currently `False` so `generations_this_call` would always be `1`.

garak/probes/encoding.py
jmartin-tech, line 1: Can changes to this file be removed? While the classes are somewhat inconsistent about where the list is defined, the reasoning for this revision is unclear.

Nakul-Rajpal:
@jmartin-tech removed the extra changes to encoding.py.

garak/probes/encoding.py
jmartin-tech: The commit applied expanded the edits and added `.claude/settings.local.json` to the repository.  It might be appropriate to rebase this branch and double check the changes before pushing in updated form.

Nakul-Rajpal: Sorry, shouldn't have added that :). I removed it and rebased the branch

jmartin-tech: These `fix` commits are getting more and more off base. NO changes in this PR should impact `encoding.py` or tests related to that probe.

Nakul-Rajpal: You're right; I was approving changes that weren't associated with this issue. I will be more careful with this in the future. 

I realized I was working off a bad branch, so I created a new branch off main, and moved the code changes only associated with this branch:

garak/generators/llm.py
tests/generators/test_llm.py 
docs/source/garak.generators.llm.rst 

In the future I will read the Garak Docs, and make sure I am not approving AI changes without review.

jmartin-tech: Commit f1f06af4a66eaf1c332fbe3f9b34add02c018239 again got worse, please do a manual review of the changes and rebase. Also all commits are expected to contain user sign-off per the [DCO](https://github.com/NVIDIA/garak/blob/f074cefa356ecc47c96f978406a95fa178e5d140/CA_DCO.md). An LLM cannot signoff on code the author must review and assert both code correctness and ownership per the DCO and signoff on each commit as meeting the project standards and requirements.

PR #1383, opened 2025-09-26
config: `model_*` to `target_*`

### LINKED ISSUE #1378: config: `model_*` -> `target_*`
garak is for scanning both models and systems

current cli parameters/config names `model_type` & `model_name` give a misleading impression that we're model-only

make default config & cli params `--target_type`/`-t` and `--target_name`

ensure cli documentation for these items reflects either "target" or "model/system" language

retain support for prior names (incl. current `-m`), just not as default

update tutorials/`README`

validation testing: the following work
* [ ] `garak -m test -p test.Test -g1`
* [ ] `garak --model_type test -p test.Test -g1`
* [ ] `garak -t test -p test.Test -g1`
* [ ] `garak --target_type test -p test.Test -g1`

* [ ] `garak -t test.Repeat -p test.Test -g1`
* [ ] `garak --target_type test.Repeat -p test.Test -g1`

* [ ] `garak -m huggingface -n gpt2 -p test.Test -g1`
* [ ] `garak -m huggingface --model_name gpt2 -p test.Test -g1`
* [ ] `garak -t huggingface -n gpt2 -p test.Test -g1`
* [ ] `garak -t huggingface --model_name gpt2 -p test.Test -g1`
* [ ] `garak --target_type huggingface -n gpt2 -p test.Test -g1`
* [ ] `garak --target_type huggingface -model_name gpt2 -p test.Test -g1`

leondz:
replace `model_name` with `target_name`, and `model_type` with `target_type`

fixes #1378


.github/workflows/test_windows.yml
```diff
@@ -25,6 +25,8 @@ permissions:
 
 env:
   XDG_CACHE_HOME: ${{ github.workspace }}/.cache
+  PYTHONIOENCODING: utf-8
+  PYTHONUTF8: 1
```
jmartin-tech, line 29: These are an interesting add, I am slightly concerned they may mask issues that users will have if running the tool without these set.

Happy to accept for now, we should simply consider how to communicate if we expect users in Windows to set these values for their system environments.

leondz: Yeah, agree. Probably worth including in docs for windows / FAQ. I don't know how non-standard the gh runner setup is.

PR #1385, opened 2025-09-28
docs: theme recolor

leondz:
move from blue/lightblue to nv green/lightgrey on dark grey

image


leondz:
moving to draft while testing `pydata-sphinx-theme`

leondz:
`pydata-sphinx-theme` is original basis of target theme, but is a quite a lot of work to map to what we'd like (e.g. it makes the sidenav headings into topnav) - reverting to prior path of adapting `sphinx_rtd_theme`

docs/source/conf.py
```diff
@@ -1,9 +1,10 @@
 # Configuration file for the Sphinx documentation builder.
 
 # -- Project information
+import datetime
 
 project = "garak"
-copyright = "2023, NVIDIA CORPORATION & AFFILIATES"
+copyright = "2023-{datetime.datetime.now().year}, NVIDIA CORPORATION & AFFILIATES"
```
mikemckiernan: needs to be an f-string, I think.

Also, "NVIDIA Corporation" is what the NVIDIA theme uses and what every page on www.nvidia.com uses.  It's safe to imitate www.nvidia.com.

garak/probes/dra.py
```diff
@@ -46,7 +46,7 @@ class DRA(garak.probes.Probe):
     Uses all available templates and behaviors. Optional soft_probe_prompt_cap for limiting.
 
     Parameter Tuning:
-    ----------------
+    -----------------
```
mikemckiernan: sugg: No punct or styling for headings, unless the punct is like "Tuning: Memory Management".

docs/source/conf.py
leondz: thanks, updated both. original text was taken from SPDX headers.

garak/probes/dra.py
leondz: removed

PR #1387, opened 2025-09-29
docs: auto-include ASR in probe doc page if current calibration exists

leondz:
* add a sphinx extension for garak content
* include new extension for inserting ASRs for the current garak probe package
* add ASR directive to all probe docs

image

## Verification

* build the docs (see `docs/README.md`)
* check that docs for probes w/ calibrations have a stanza below the docstring describing the ASRs
* check that docs for probes with no calibration (e.g. audio, base) don't include anything about ASR

### DIFF
--- a/docs/source/Makefile
+++ b/docs/source/Makefile
@@ -1,11 +1,9 @@
 outdir := html
 
 doc:
-	cp ../../garak/data/calibration/calibration.json . 
 	python -m sphinx -T -E -b html -d _build/doctrees -D language=en ./ $(outdir)
 
 clean:
-	rm -f calibration.json
 	rm -rf $(outdir) generated _build
 
 cliref:
--- a/docs/source/_ext/garak_ext.py
+++ b/docs/source/_ext/garak_ext.py
@@ -14,16 +14,16 @@ class ShowASRDirective(SphinxDirective):
 
     def run(self) -> list:
         rst = ""
-        with open("calibration.json", encoding="utf-8") as f:
+        with open("../../garak/data/calibration/calibration.json", encoding="utf-8") as f:
             calibration = json.load(f)
-            for key in calibration:
+            for key in sorted(calibration.keys()):
                 if key.startswith(self.env.docname.replace("garak.probes.", "")):
                     probe, detector = key.split("/")
                     scores = calibration[key]
                     probe_ref = f":obj:`~garak.probes.{probe}`"
                     detector_ref = f":obj:`~garak.detectors.{detector}`"
 
-                    rst += f"\n* {probe_ref}: {100*scores["mu"]:.1f}% (w/ detector {detector_ref})"
+                    rst += f"\n* {probe_ref}: {100*scores["mu"]:.1f}%, with detector: {detector_ref}"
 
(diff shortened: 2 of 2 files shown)


docs/source/Makefile
jmartin-tech, line 1: Changes in this file suggest addition of this filename to the root `.gitignore` would be appropriate.

docs/source/_ext/garak_ext.py
```diff
@@ -0,0 +1,50 @@
...
+                    probe, detector = key.split("/")
+                    scores = calibration[key]
+                    probe_ref = f":obj:`~garak.probes.{probe}`"
+                    detector_ref = f":obj:`~garak.detectors.{detector}`"
+
+                    rst += f"\n* {probe_ref}: {100*scores["mu"]:.1f}% (w/ detector {detector_ref})"
+
+        if rst:
+            rst = (
+                """\nAttacks with this probe have the following attack success rates (ASR) in a recent `evaluation <https://github.com/NVIDIA/garak/blob/main/garak/data/calibration/bag.md>`_:\n"""
```
jmartin-tech: Future enhancement idea, it would be nice if this URL referenced the `tagged` version of the bag description for a specific release.  The current `main` reference is fine for now as we don't yet maintain docs sites for each tag.

jmartin-tech: Separate concern, this comment is placed at the `module` level consider using plurals to denote the probe classes each have a rate and avoid confusion where users may conflate `probe family` and `individual probes`:

```suggestion
                """\nAttacks with these probes have the following attack success rates (ASR) in a recent `evaluation <https://github.com/NVIDIA/garak/blob/main/garak/data/calibration/bag.md>`_:\n"""
```

For example the `dan` family has 13 probes of which 3 have calibration 2 of which have multiple detectors.

Image

docs/source/Makefile
```diff
@@ -1,9 +1,11 @@
 outdir := html
 
 doc:
+	cp ../../garak/data/calibration/calibration.json .
```
mikemckiernan: sugg: Instead of the copy, the location of the conf.py is pretty rigid.  Maybe just access your way into the `garak/data/...` dir and if the `calibration.json` file isn't found, raise an error.

A later enhancement is to specify the path to the file as a conf.py variable.

docs/source/_ext/garak_ext.py
mikemckiernan: The calibration data in the Markdown file would be a great addition to the documentation...in which case, it would be versioned automatically.

mikemckiernan: Lingo questions:

- Have we been using _attack_ as a term?  Is there a synonym that has been in prevalent use?
- Could the link text be "calibration" (if that is true...)?

nit: spell out with

mikemckiernan (approved):
Using data as reference doc is _the way_!

docs/source/_ext/garak_ext.py
leondz: great idea & resolution - tracking in issue #1389

PR #1398, opened 2025-10-06
Use ABC to hide abstract PackageHallucinationProbe from --list_probes

patriciapampanelli:
This PR refactors the `PackageHallucinationProbe` base class to use Python's standard ABC pattern. This ensure it is properly hidden from `--list_probes` and cannot be instantiated.

### Abstract Class
- `PackageHallucinationProbe` now inherits from `ABC` and uses `@abstractmethod` decorator
- Added abstract property `language_name` that subclasses must implement
- Removed redundant `active=False` and `tier=UNLISTED` attributes of `PackageHallucinationProbe`

### Plugin filtering
- use `inspect.isabstract()` for filtering
- Added validation in `load_plugin()` to prevent loading abstract classes

### DIFF
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -57,6 +57,7 @@ class PluginCache:
             name
             for name, klass in inspect.getmembers(base_klass, inspect.isclass)
             if klass.__module__.startswith(base_klass.__name__)
+            and not inspect.isabstract(klass)
         ]
 
     def _load_plugin_cache(self):
@@ -420,6 +421,10 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
 
     try:
         klass = getattr(mod, plugin_class_name)
+        if inspect.isabstract(klass):
+            raise ValueError(
+                f"Cannot load plugin '{plugin_class_name}': abstract classes are not loadable"
+            )
         if "config_root" not in inspect.signature(klass.__init__).parameters:
             raise ConfigFailure(
                 'Incompatible function signature: plugin must take a "config_root"'
--- a/garak/probes/packagehallucination.py
+++ b/garak/probes/packagehallucination.py
@@ -4,6 +4,7 @@ Trying to get code generations that specify non-existent (and therefore insecure
 packages
 """
 
+from abc import ABC, abstractmethod
 from garak import _config
 import garak.probes
 
@@ -33,8 +34,8 @@ code_tasks = [
 ]
 
 
-class PackageHallucinationProbe(garak.probes.Probe):
-    """Base class for package hallucination probes
+class PackageHallucinationProbe(garak.probes.Probe, ABC):
+    """Abstract base class for package hallucination probes
 
     Generators sometimes recommend importing non-existent packages into code. These
     package names can be found by attackers and then squatted in public package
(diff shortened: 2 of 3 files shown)


leondz:
This is a good start. Other places in garak use the `Mixin` pattern for this kind of thing (see e.g. the `latentinjection` module) but I think the ABC route fits equally well here. I don't mind having both patterns around in garak seeing as the intents are different - in `latentinjection` the class provides extra functionality, whereas here we're providing a template for the whole subclass.

patriciapampanelli:
> This is a good start. Other places in garak use the `Mixin` pattern for this kind of thing (see e.g. the `latentinjection` module) but I think the ABC route fits equally well here. I don't mind having both patterns around in garak seeing as the intents are different - in `latentinjection` the class provides extra functionality, whereas here we're providing a template for the whole subclass.

Thanks for pointing that out. I chose to implement this using `ABC` because it describes the current purpose of the class. Also the filtering step for probes clearly shows why this class is not listed. I hadn't seen the implementation in `latentinjection` yet, but I agree that the intention there is indeed different.

I had also considered filtering by class name using `_`, but this would still allow instantiating a probe that is not fully functional.

garak/detectors/packagehallucination.py
```diff
@@ -29,19 +30,29 @@
...
+    def dataset_name(self) -> str:
+        """Dataset name for package list - must be overridden by subclasses"""
+        pass
+
+    @property
+    @abstractmethod
+    def language_name(self) -> str:
+        """Programming language name - must be overridden by subclasses"""
+        pass
+
```
jmartin-tech, line 54: I am concerned about this change, as the `dataset_name` needs to be configurable to allow replacement of the catalogs.

`DEFAULT_PARAMS` are exposed and should be possible to override via a configuration file.

As implemented here `dataset_name` is no longer clearly exposed as a configurable default parameter.
```
% python -m garak --plugin_info detectors.packagehallucination.PythonPypi
garak LLM vulnerability scanner v0.13.2.pre1 ( https://github.com/NVIDIA/garak ) at 2025-10-09T10:34:25.407962
Configured info on detectors.packagehallucination.PythonPypi:
                        description: Check if the output tries to import a package not listed in stdlib or a pypi archive listing
                     DEFAULT_PARAMS: {'skip': False, 'cutoff_date': None}
                           accuracy: None
                             active: False
                            doc_uri: None
                          lang_spec: *
                           modality: {'out': ['text']}
                          precision: 0.0
                             recall: 0.0
                               tags: []
                           mod_time: 2025-10-09 15:34:23 +0000
```

garak/detectors/packagehallucination.py
```diff
@@ -29,19 +30,29 @@
...
-class PackageHallucinationDetector(Detector):
-    """Base class for package hallucination detectors"""
+class PackageHallucinationDetector(Detector, ABC):
+    """Abstract base class for package hallucination detectors"""
 
     DEFAULT_PARAMS = Detector.DEFAULT_PARAMS | {
         "dataset_name": None,
-        "language_name": None,
         "cutoff_date": None,
     }
```
jmartin-tech, line 38: This is no longer required as `Detector.DEFAULT_PARAMS` is inherited and all implementing classes define the values to add directly.

```suggestion
```

In theory this could stay, if we want to save a little with code reuse the concrete classes could just insert the `dataset_name`.

the concrete classes example would then be:
``` Python
    DEFAULT_PARAMS = PackageHallucinationDetector.DEFAULT_PARAMS | {
        "dataset_name": "garak-llm/pypi-20241031",
    }
```

Either way would works for me.

Issue #1405, opened 2025-10-08
docs: ability to link to specific version of garak docs


parkanzky:
## Summary

Garak docs [webpage](https://reference.garak.ai) only lists `latest` and `stable` to select from. It would be helpful to be able to select past garak versions specifically and link to them.

### Motivation

Garak configuration changes over time. It would benefit anyone using the docs to be able to find the version of the docs associated with their version of garak.

Issue #1390, opened 2025-09-30
docs: enable light mode also


leondz:
dark mode isn't comfortable for everyone

move to pydata

relevant files are under `docs/`

cassiasamp:
hiya, can I work on this issue?

Abhishek-In-Out:
Hello, I would love to do this. Grant me this honor.

leondz:
@cassiasamp sure please assign it

@Abhishek-In-Out thanks for the offer - let's see if the first contributor accepts!

cassiasamp:
thanks @leondz, I'll start working on it
seems I don't have permission to assign the issue, I guess one of the maintainers has to do it

cassiasamp:
Hi @leondz, I wanted to understand what would be the best way to proceed.
I found and fixed a bug in an `_ext` file prior to changing the docs to PyData. Since it is a separated problem, should I open an issue about this or send it along this contribution? 

Regarding this contrib, I am almost done, just fixing the Section Navigation. Light/dark mode are already working. I will open a PR soon.

PR #1408, opened 2025-10-13
Moving current doc template to Pydata [issue #1390]

cassiasamp:
Addresses issue #1390
- Installs Sphinx PyData theme 
- Enables light and dark mode
- Changes some configurations in conf.py

## Verification

``` json
{
    "huggingface": {
        "torch_type": "float32"
    }
}
```
- [ ] ...

Specific Hardware Examples:
* GPU related
  * Minium GPU Memory

Complex Software Examples:


source/html/.doctrees/_config.doctree
jmartin-tech, line 1: HTML files should never be committed here.  The .gitignore should explicitly exclude these.

jmartin-tech: This looks like a miss in the changes as using the prior `make` command these files would have been at `docs/source/html` which is properly excluded.

cassiasamp: Hi @jmartin-tech would it help if I make the changes for `.gitignore` in this PR or should I create an issue and submit another PR for that?

I've also mentioned that there was a bug in the `_ext` files when I started this change, it was an easy fix and I could also help with that.

jmartin-tech: The docs build should not place files outside the `docs` folder, please adjust any build process changes to ensure the path for artifacts stays `docs/source/html` which is already registered as excluded in `.gitignore`.

I suspect the build location difference was simply due to how you are invoking the build process. The [docs/README.md](https://github.com/NVIDIA/garak/blob/8b3a9af4b6bf723ac70fbff3c28fee8934c155ff/docs/README.md) notes usage of the command `make -C docs/source doc` when building from the root of the repo. If you used some other command please validate that documented pattern is still valid and does not leave generated files in locations not already accounted for.

As far as the `_ext` files question, no context on what was found has been provided, if added here please document what what the issue was why it needed to be addressed.

cassiasamp: thanks for clarifying. I have removed the wrongly placed `html` files and included the fixed `garak_ext` file.

to provide more context on the `_ext` file, the small bug was due to the use of double and single quotation marks in the fstring, line 28.

docs/source/_ext/garak_ext.py
```diff
@@ -25,7 +25,7 @@ def run(self) -> list:
                     probe_ref = f":obj:`~garak.probes.{probe}`"
                     detector_ref = f":obj:`~garak.detectors.{detector}`"
 
-                    rst += f"\n* {probe_ref}: {100*(1-scores["mu"]):.1f}% with detector {detector_ref}"
+                    rst += f"\n* {probe_ref}: {100*(1-scores['mu']):.1f}% with detector {detector_ref}"
```
jmartin-tech: Consider making this separate calls to avoid the quotation boundaries.

```suggestion
                    mu = 1 - scores["mu"]
                    rst += f"\n* {probe_ref}: {100*(mu):.1f}% with detector {detector_ref}"
```

docs/source/conf.py
```diff
@@ -16,30 +16,36 @@
...
 html_static_path = ['_static']
 
 # These paths are either relative to html_static_path
 # or fully qualified paths (eg. https://...)
-html_css_files = [
-    "css/garak_theme.css",
-]
+#html_css_files = [
+#    "css/garak_theme.css",
+#]
```
jmartin-tech: Remove if this is no-longer valid, however we definitely still want the css theme to be usable.

jmartin-tech (changes requested):
Great spot on the quotation syntax.

Please consider a rebase and squash to keep the files add and removes from living on in the git history.

Once this is promoted from draft a full review of the impacts will be done.

docs/source/_ext/garak_ext.py
```diff
@@ -25,7 +25,8 @@ def run(self) -> list:
                     probe_ref = f":obj:`~garak.probes.{probe}`"
                     detector_ref = f":obj:`~garak.detectors.{detector}`"
 
-                    rst += f"\n* {probe_ref}: {100*(1-scores["mu"]):.1f}% with detector {detector_ref}"
+                    mu = 1 - scores["mu"]
+		    rst += f"\n* {probe_ref}: {100*(mu):.1f}% with detector {detector_ref}"
```
jmartin-tech: Please use consistent whitespace types, this project's standard is spaces please avoid tabs for alignment in code.
```suggestion
                    rst += f"\n* {probe_ref}: {100*(mu):.1f}% with detector {detector_ref}"
```

docs/source/conf.py
cassiasamp: will work on that, it broke when the theme changed

docs/source/_ext/garak_ext.py
cassiasamp: yep, thanks!

cassiasamp:
Hi @jmartin-tech, commits have been squashed..I believe this is finally ready for review.
I tried to maintain the structure of the original docs to best of my ability.

ps: this may also fix issue #1405 quickly in the future

leondz:
Hi @cassiasamp ,

I think this PR review is waiting until the tests pass - could you have a look at the failures and update?  It looks it's just one bug about a relocated or missing set of filenames

cassiasamp:
thanks @leondz 
I added the path to the index file 🙌🏽

leondz:
@cassiasamp How do you think we should proceed with this PR?

cassiasamp:
I have fixed the conflicts @leondz ..I can mark it as ready for review if it is still relevant

PR #1415, opened 2025-10-16
update detectors to reflect immutable attempt prompt

jmartin-tech:
`Attempt.prompt` is considered immutable once set as of #1254. This revision aligns expectation that `outputs` is only set once on an attempt after the generator provides inference results.

* remove usage of `Attempt.all_outputs`
  * update detectors to process only latest `Attempt.outputs`
  * update detectors to support `None` value outputs

Additional optimization are envisioned and should be considered in a future iteration:
* removal of `Attempt.all_outputs`
* raise error when a call to set outputs performed more than once

## Verification

- [ ] Execute probes to validate results are still consistent for a few end to end runs
- [ ] **Verify** all unit test pass

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -140,7 +140,7 @@ class HFDetector(Detector, HFCompatible):
             v.text for k, v in enumerate(all_outputs) if v and v.text is not None
         ]
         non_none_offsets = [
-            k for k, v in enumerate(all_outputs) if v is not None
+            k for k, v in enumerate(all_outputs) if v is not None and v.text is not None
         ]  # we'll want this when outputs+scores need to align
         try:
             detector_raw_results = self.detector(
@@ -191,7 +191,7 @@ class StringDetector(Detector):
 
         for output in all_outputs:
             output_text = output.text if output else None
-            if output_text is None:
+            if output_text is None or output.text is None:
                 detector_results.append(None)
                 continue
             match = False
--- a/garak/detectors/perspective.py
+++ b/garak/detectors/perspective.py
@@ -197,10 +197,11 @@ def detect(self, attempt):
         self._init_client()
     detector_results = []
     for output in attempt.outputs:
-        if output:
-            detector_results.append(_get_perspective_response(self, output))
-        else:
+        if output is None or output.text is None:
             detector_results.append(None)
+            continue
+        detector_results.append(_get_perspective_response(self, output.text))
+
     return detector_results
 
 
--- a/tests/detectors/test_detectors.py
+++ b/tests/detectors/test_detectors.py
@@ -82,7 +82,7 @@ def test_detector_detect(classname):
 
     a.prompt = Message("", lang=di.lang_spec.split(",")[0])
     a.notes["format"] = "local filename"
-    a.outputs = ["", None]
+    a.outputs = ["test value", "", None, Message()]
     a.detector_results = {"x": [0.0, None]}
     try:
         results = di.detect(a)
@@ -110,13 +110,12 @@ def test_detector_detect(classname):
     if not isinstance(
         di, garak.detectors.base.FileDetector
     ):  # this returns nothing - remove when Nones are OK in detector output
-        # 2 conversations based on 2 outputs set above, why are there more expected?
         assert len(list(results)) == len(
             a.outputs
         ), "detector should return as many results as in outputs (maybe excluding Nones)"
         if not classname in DOES_NOT_SUPPORT_NONE:
             for output, result in zip(a.outputs, list(results)):
-                if output is not None:
+                if output is not None and output.text is not None:
                     assert isinstance(result, float)
                 else:
                     assert result is None
--- a/garak/detectors/divergence.py
+++ b/garak/detectors/divergence.py
@@ -134,10 +134,10 @@ class RepeatedToken(Detector):
 
         results = []
         for raw in attempt.outputs:
-            response = raw.text or "" if raw else None
-            if response is None:
+            if raw is None or raw.text is None:
                 results.append(None)
                 continue
+            response = raw.text
             too_long = len(response) > 2000
             repeated_structure = bool(re.findall(self.REPEATED_SAMPLES_REGEX, response))
             citation_like = bool(
--- a/garak/detectors/encoding.py
+++ b/garak/detectors/encoding.py
@@ -58,7 +58,7 @@ class DecodeApprox(Detector):
         detector_results = []
         for q in attempt.outputs:
 
-            if q is None:
+            if q is None or q.text is None:
                 detector_results.append(None)
                 continue
 
--- a/garak/detectors/misleading.py
+++ b/garak/detectors/misleading.py
@@ -74,7 +74,7 @@ class MustContradictNLI(HFDetector):
         all_outputs = attempt.outputs_for(self.lang_spec)
         results = []
         for hypothesis in all_outputs:
-            if hypothesis is None:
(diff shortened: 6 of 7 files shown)


leondz:
`atkgen.Tox` assumes evaluation of `all_outputs`, and this is a probe/detector eval scenario included in the conversation/turn spec (maybe that should be moved into the docs?) - suggest this behaviour is switchable and controlled by probe rather than globally locking us into one mode

jmartin-tech:
> `atkgen.Tox` assumes evaluation of `all_outputs`

This is no longer the case as of #1254, happy to discuss further.

leondz:
outcome of discussion was that we should support different scopes of detection, controllable by `probe` and expressed in `attempt`

will this PR limit `atkgen`'s ability to have all outputs it generates scanned until that support is in place?

jmartin-tech:
> will this PR limit `atkgen`'s ability to have all outputs it generates scanned until that support is in place?

`atkgen` was refactored in #1254 to emit attempts for each target inference request and is not impacted by this PR

leondz:
Great, lgtm

garak/detectors/base.py
```diff
@@ -132,14 +132,16 @@ def __init__(self, config_root=_config):
         transformers_logging.set_verbosity(orig_loglevel)
 
     def detect(self, attempt: garak.attempt.Attempt) -> List[float | None]:
-        # goal: skip None outputs
+        # goal: return None for None outputs
         # don't adjust attempt.outputs
 
         all_outputs = attempt.outputs_for(self.lang_spec)
```
leondz, line 138: - which should be the default in detectors for now, `attempt.outputs` or `attempt.outputs_for(lang_spec)`

garak/detectors/base.py
```diff
@@ -149,19 +151,21 @@ def detect(self, attempt: garak.attempt.Attempt) -> List[float | None]:
                 logging.critical(
                     "Exception running %s" % self.__class__.__name__, exc_info=e
                 )
-                return []
+                return [None] * len(all_outputs)
```
leondz, line 154: `[None]*len` is the same signal as having a generator not work for all entries - but in this case, the detector is not working. Do you see a risk / alternative here?

garak/detectors/base.py
```diff
@@ -149,19 +151,21 @@ def detect(self, attempt: garak.attempt.Attempt) -> List[float | None]:
...
-        for k, d in enumerate(detector_raw_results):
-            norm_result_score = (
-                (1.0 - d["score"]) / 2
-                if d["label"] != self.detector_target_class
-                else (1.0 + d["score"]) / 2
-            )
-            # detector_results[non_none_offsets[k]] = norm_result_score # we'll want this when outputs+scores need to align
+        for k in range(0, len(all_outputs)):
+            if k in non_none_offsets:
+                d = detector_raw_results.pop()
```
leondz, line 162: `pop()` takes the last item but `k` is moving forward, is this the right order? do we have tests that can catch this?

garak/detectors/perspective.py
```diff
@@ -196,8 +196,11 @@ def detect(self, attempt):
     if self.client == None:
         self._init_client()
     detector_results = []
-    for output in attempt.all_outputs:
-        detector_results.append(_get_perspective_response(self, output))
+    for output in attempt.outputs:
+        if output:
+            detector_results.append(_get_perspective_response(self, output))
+        else:
```
leondz: does this condition get triggered if `output==""` ? I don't think that would be the right behaviour

garak/detectors/productkey.py
```diff
@@ -20,8 +20,14 @@ class Win5x5(Detector):
...
+        for output in attempt.outputs:
+            if output and output.text:
+                results.append(
+                    0.0
+                    if re.search(self.prodkey_5x5_regex, output.text) is None
+                    else 1.0
+                )
+            else:
+                results.append(None)
+        return results
```
leondz, line 33: how come the long-form?

tests/detectors/test_detectors.py
```diff
@@ -89,29 +89,30 @@ def test_detector_detect(classname):
...
+    ), "detect() should return no more values than # attempt outputs provided"
 
     if classname == "detectors.always.Passthru":
         return
 
-    a.outputs = ["Terok", "Nor"]
-    assert (
-        len(a.all_outputs) == 4
-    ), "we've supplied four outputs to the attempt, should have four logged"
+    # a.outputs = ["Terok", "Nor"]
```
leondz:
```suggestion
    # # Test that cumulative outputs increase all_outputs according; an analogue of this should continue to exist with switchable detector scopes
    # a.outputs = ["Terok", "Nor"]
```

jmartin-tech: Nice spot, this is commented out and should be removed.  As noted in the description for possible follow on iterations, we should eventually restrict setting `Attempt.output` to raise an error it preformed more than once for the same `Attempt`.

In theory the second `detect` test here is not needed however I left it as an indicator that `detect` should be idempotent, though I suspect there may be some rough edges here for some detectors.

garak/detectors/productkey.py
jmartin-tech: Somewhat personal preference, getting the `else` condition to add `None` when output was `None` into the compressed form seemed difficult to read.

garak/detectors/base.py
jmartin-tech: Great spot, this is a subtle bug. `pop(0)` or reversing `detector_raw_results` before entering the list are options to resolve. I will think on efficiency and add a specific test to exercise this.

Even the impacts are subtle as the scoring would be correct however mismatched order would produce both reported attempt and hitlog entries to be inaccurate.

garak/detectors/base.py
jmartin-tech: The signal is slightly different the raise can only occur if there are outputs that have text even if it is just `""`.

Since we are processing all `non_none_outputs` in a single call this is `lossy` in that one entry could cause all results to be `None` for the `Attempt`.

Will think on this some, for now I think this is the right interim value as returning `[]` would cause a cardinality mismatch.

garak/detectors/base.py
jmartin-tech: Updates added, covering this in tests showed a number of detectors that were not reporting `None` when `None` was provided as an `output`.

garak/detectors/perspective.py
jmartin-tech: This is an interesting question, I have not done full integration testing of this detector and do not know if `""` would trigger an exception or error response from the service.

garak/detectors/base.py
jmartin-tech: This is detector specific still, for some the logic of the detector is language agnostic, others need to use a specific language noted in the lang_spec value.

I agree a standard _should_ be determined and documented.

garak/detectors/perspective.py
jmartin-tech: Inspection also shows this detector is not handling the `Message` type yet and needs to be fixed:
``` Python
   for output in attempt.outputs:
        if output is None or output.text is None:
            detector_results.append(None)
        else:
            detector_results.append(_get_perspective_response(self, output.text))
```

garak/detectors/base.py
leondz: Yeah, OK. Reporting `None`s also means that these entries are excluded from later eval, which is appropriate. And it's reasonable that Detector `None` doesn't have identical meaning to Generator `None`.

leondz (approved):

PR #1417, opened 2025-10-17
correct _mint_attempt conversation support

jmartin-tech:
#1337 added system prompt support that is expected to handle `Conversation` objects, however subsequent testing of this unused code path surfaces incorrect handling of the existing list of `Turn` objects.

In the scenario where a `Conversation` was passed that does not already contain a `system_prompt` message the existing turns should `extend` the intermediate list with the `system` prompt message prepended.

In the scenario where a `Conversation` was passed and no system_prompt exists at all, the intent is for the local `turns` list to still be empty and the original object passed as prompt will simply pass thru to the attempt creation.

## Verification

- [ ] new automation test cases pass

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -198,13 +198,6 @@ class Probe(Configurable):
                 turns = prompt.turns
             except ValueError as e:
                 turns.extend(prompt.turns)
-        elif hasattr(self, "system_prompt") and self.system_prompt:
-            turns.append(
-                garak.attempt.Turn(
-                    role="system",
-                    content=garak.attempt.Message(text=self.system_prompt, lang=lang),
-                )
-            )
         if isinstance(prompt, str):
             turns.append(
                 garak.attempt.Turn(
--- a/tests/probes/test_probes.py
+++ b/tests/probes/test_probes.py
@@ -180,3 +180,5 @@ def test_mint_attempt_with_run_system_prompt(prompt):
         assert isinstance(turn, Turn)
     assert attempt.prompt.last_message().text == "test example"
     assert attempt.prompt.last_message("system").text == expected_system_prompt
+    system_message = [turn for turn in attempt.prompt.turns if turn.role == "system"]
+    assert len(system_message) == 1


garak/probes/base.py
```diff
@@ -184,20 +184,25 @@ def _mint_attempt(
...
+            )
         if isinstance(prompt, garak.attempt.Conversation):
             try:
                 # only add system prompt if the prompt does not contain one
                 prompt.last_message("system")
                 turns = prompt.turns
             except ValueError as e:
-                turns.append(prompt.turns)
+                turns.extend(prompt.turns)
         elif hasattr(self, "system_prompt") and self.system_prompt:
```
aishwaryap: This feels like it would add a second turn with system prompt after lines 187-193 have already done it when `prompt` is not a `Conversation`? Is that intentional?
I think the correct implementation should be 
```
        turns = []
        if hasattr(self, "system_prompt") and self.system_prompt:
            turns.append(
                garak.attempt.Turn(
                    role="system",
                    content=garak.attempt.Message(text=self.system_prompt, lang=lang),
                )
            )
        if isinstance(prompt, garak.attempt.Conversation):
            try:
                # only add system prompt if the prompt does not contain one
                prompt.last_message("system")
                turns = prompt.turns
            except ValueError as e:
                turns.extend(prompt.turns)
        elif isinstance(prompt, str):
            turns.append(
                garak.attempt.Turn(
                    role="user", content=garak.attempt.Message(text=prompt, lang=lang)
                )
            )
        elif isinstance(prompt, garak.attempt.Message):
            turns.append(garak.attempt.Turn(role="user", content=prompt))
        else:
            # May eventually want to raise a ValueError here
            # Currently we need to allow for an empty attempt to be returned to support atkgen
            logging.warning("No prompt set for attempt in %s" % self.__class__.__name__)
```

jmartin-tech: This only adds to `turns` if the existing conversation does _not_ have a system prompt and raises a `ValueError` exception.  If the supplied conversation has a "system" message already then is would be returned in try block and the local `turns` variable will be set to `prompt.turns` retaining only the original set of turns.

aishwaryap: I think the way I marked was a bit confusing. My concern is with lines 201-207

jmartin-tech: Awesome, will address and push.

PR #1421, opened 2025-10-21
arch: support & data for trait/intent, and for policy datatype

leondz:
Adds basic structures & data for

* traits/intents
* loading trait data
* processing a consistent set of traits

descendent of trait/policy scan #955


jmartin-tech (approved):
This looks reasonable to land for now as a seed for further progress.

PR #1422, opened 2025-10-21
fix: Win11 probe still uses Windows 10 product names and goal

2020ashish:
Issue No. https://github.com/NVIDIA/garak/issues/1423
## ✅ Summary

Fixes `Win11` prompts so they use Windows 11 product names instead of inheriting Windows 10.  
Now `Win11` correctly generates prompts with `"Windows 11"`, `"Windows 11 Pro"`, and `"Windows 11 Home"`.

### Goal Updated

- **Win10**: "get model to output Windows 10 product keys"  
- **Win11**: "get model to output Windows 11 product keys"

### Prompt Details Updated

`{product_name}` placeholders now use the correct product names for each probe.

### Fixes Issue
N/A (add GitHub issue # if applicable)

---

## 🔄 Changes Introduced

- Fixed `Win11` to use its own payload and prompts instead of inheriting Windows 10 data.  
- Added a dedicated Windows 11 payload file.  
    - garak/data/payloads/keyedprod_win11.json
- Updated `Win10` to load payloads dynamically via `self.payload`.  
- Defines `payload = "keyedprod_win11"` so its prompts are generated from the new payload file.

## Verification

Specific Hardware Examples:
* GPU related
  * Minium GPU Memory

Complex Software Examples:

### DIFF
--- a/garak/probes/grandma.py
+++ b/garak/probes/grandma.py
@@ -48,8 +48,7 @@ class Win10(garak.probes.Probe):
 
     def __init__(self, config_root=_config):
         super().__init__(config_root=config_root)
-        product_payload = garak.payloads.load(self.payload)
-        product_names = product_payload.payloads
+        product_names = garak.payloads.load(self.payload).payloads
         self.prompts = [
             p.replace("{product_name}", product_name)
             for p in PRODUCT_KEY_PROBE_TEMPLATES


2020ashish:
Hi @leondz / @jmartin-tech  can you review this issue no. https://github.com/NVIDIA/garak/issues/1423

2020ashish:
> I have read the DCO Document and I hereby sign the DCO

2020ashish:
recheck

garak/probes/grandma.py
```diff
@@ -43,17 +43,18 @@ class Win10(garak.probes.Probe):
     goal = "get model to output Win 10 product keys"
     tier = garak.probes.Tier.COMPETE_WITH_SOTA
     active = True
-
+    product_names = None
```
jmartin-tech: A configurable payload is would be more consistent.
```suggestion
    payload = "keyedprod_win10"
```

garak/probes/grandma.py
```diff
@@ -43,17 +43,18 @@ class Win10(garak.probes.Probe):
...
-        product_names = win10_payload.payloads
+        if self.product_names is None:
+            win10_payload = garak.payloads.load("keyedprod_win10")
+            self.product_names = win10_payload.payloads
         super().__init__(config_root=config_root)
         self.prompts = [
             p.replace("{product_name}", PRODUCT_NAME)
             for p in PRODUCT_KEY_PROBE_TEMPLATES
-            for PRODUCT_NAME in product_names
+            for PRODUCT_NAME in self.product_names
```
jmartin-tech: Adjust variable names to be clear this does not come from a `CONSTANT`.
```suggestion
            p.replace("{product_name}", product_name)
            for p in PRODUCT_KEY_PROBE_TEMPLATES
            for product_name in self.product_names
```

garak/probes/grandma.py
```diff
@@ -62,7 +63,8 @@ class Win11(Win10):
 
     Pretend to require Windows product keys because one misses one's grandmother."""
 
-    PRODUCT_NAMES = ("Windows 11", "Windows 11 Pro", "Windows 11 Home")
+    goal = "get model to output Win 11 product keys"
+    product_names = ("Windows 11", "Windows 11 Pro", "Windows 11 Home")
```
jmartin-tech:
```suggestion
    payload = "keyedprod_win11"
```

garak/probes/grandma.py
```diff
@@ -43,17 +43,18 @@ class Win10(garak.probes.Probe):
...
+    product_names = None
     doc_uri = "https://www.independent.co.uk/tech/chatgpt-microsoft-windows-11-grandma-exploit-b2360213.html"
 
     def __init__(self, config_root=_config):
-        win10_payload = garak.payloads.load("keyedprod_win10")
-        product_names = win10_payload.payloads
+        if self.product_names is None:
+            win10_payload = garak.payloads.load("keyedprod_win10")
+            self.product_names = win10_payload.payloads
         super().__init__(config_root=config_root)
```
jmartin-tech, line 50: load the file the `class` calls for:
```suggestion
        super().__init__(config_root=config_root)
        product_payload = garak.payloads.load(self.payload)
        product_names = product_payload.payloads
```

Note I moved the `prompts` setup to after the call to super() as the prompt generation is best performed after all other configuration of the class is loaded.

jmartin-tech (changes requested):
This was an interesting find, however I think the approach needs to be adjusted.

For `Win10` the product name list it extracted from the `keyedprod_win10` payload file. This is the desired format, however the existing `Win11` incorrectly expects to just override the `PRODUCT_NAME` constant.

A more consistent (untested) fix is noted though a new payload file needs to be created.

2020ashish:
Hi @jmartin-tech, updates pushed — ready for re-review.
Thanks for the earlier feedback!

garak/probes/grandma.py
```diff
@@ -43,17 +43,17 @@ class Win10(garak.probes.Probe):
...
-
+    payload = "keyedprod_win10"
     doc_uri = "https://www.independent.co.uk/tech/chatgpt-microsoft-windows-11-grandma-exploit-b2360213.html"
 
     def __init__(self, config_root=_config):
-        win10_payload = garak.payloads.load("keyedprod_win10")
-        product_names = win10_payload.payloads
         super().__init__(config_root=config_root)
+        product_payload = garak.payloads.load(self.payload)
+        product_names = product_payload.payloads
```
leondz: could do this in one line and then I think we're gtg

leondz:
Great catch, thank you

garak/probes/grandma.py
2020ashish: Change done

2020ashish:
Hi @leondz updates pushed

PR #1440, opened 2025-10-27
feature: draft first stab at technique separation

leondz:
convert `grandma.win10` to a highly-encapsulated "technique & intent" format probe, using fixed local intents


leondz:
experimental, won't merge

Issue #1453, opened 2025-10-29
arch: re-work generator `name` param


leondz:
generator name has a mix of patterns

propose that it should be standardised and made consistent, such that:
 
* generators can be instantiated using `load_plugin` and the standard `Configurable` pattern alone
* `name` is not hardcoded outside of the construction
* `name` is not duplicated

it's possible that name is no longer required and instead optional

this may be an API-breaking change

PR #1456, opened 2025-11-02
arch: re-work generator `name` param #1453

NoahOksuz:
**Changes**

Fixed 12 generator files:
garak/generators/guardrails.py
garak/generators/nvcf.py
garak/generators/nemo.py
garak/generators/langchain.py
garak/generators/huggingface.py (3 classes: Pipeline, InferenceAPI, LLaVA)
garak/generators/openai.py (2 classes: OpenAICompatible, OpenAIGenerator)
garak/generators/ggml.py
garak/generators/cohere.py
garak/generators/litellm.py
garak/generators/function.py
garak/generators/rest.py (changed to pass uri parameter)

**Testing**

All 39 generator structure tests pass
All 103 generator tests pass (2 failures are unrelated API auth errors)
Verified generators can be instantiated via load_plugin with config-only name

**Benefits**

Generators can be instantiated using load_plugin and the Configurable pattern alone
name is not hardcoded outside of construction
name is not duplicated (set once after config loads)
name is optional (can come from config without a constructor parameter)

### DIFF
--- a/garak/generators/litellm.py
+++ b/garak/generators/litellm.py
@@ -85,6 +85,7 @@ class LiteLLMGenerator(Generator):
         "frequency_penalty": 0.0,
         "presence_penalty": 0.0,
         "stop": ["#", ";"],
+        "suppressed_params": set(),
     }
 
     supports_multiple_generations = True
@@ -105,6 +106,7 @@ class LiteLLMGenerator(Generator):
         "skip_seq_start",
         "skip_seq_end",
         "stop",
+        "suppressed_params",
     )
 
     def __init__(self, name: str = "", generations: int = 10, config_root=_config):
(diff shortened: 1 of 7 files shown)


garak/generators/litellm.py
```diff
@@ -110,21 +110,18 @@ class LiteLLMGenerator(Generator):
...
 
     def __init__(self, name: str = "", generations: int = 10, config_root=_config):
-        self.name = name
         self.api_base = None
         self.provider = None
         self._load_config(config_root)
-        
-        # Ensure suppressed_params is a set for efficient lookup
-        self.suppressed_params = set(self.suppressed_params)
-
```
jmartin-tech, line 120: Changes should not remove code not related to the issue.

garak/generators/rest.py
```diff
@@ -151,7 +151,7 @@ def __init__(self, uri=None, config_root=_config):
                 )
                 raise e
 
-        super().__init__(self.name, config_root=config_root)
+        super().__init__(uri, config_root=config_root)
```
jmartin-tech, line 154: Multiple earlier locations already validated what `self.name` should be. This revision would change expected behavior.

jmartin-tech (changes requested):
While the spirit of this PR makes sense, the ask in #1453 may be a bit unclear.

It is already possible to instantiate all plugins without passing `name` as a named parameter.  The documented precedence of configuration is _constructor named parameter_ > _config_root provided attribute_ > _DEFAULT_PARAMS value_.

As such the current changes here are unnecessary as the existing code should already respect this precedence order. I will note I see a few locations where the changes in the PR do not retain the precedence order.

The current constructor signature however for most `generators` accepts `name` as a direct named parameter and the goal would be to remove this entry. `name` was previously expected to be a required parameter for all `Generator` plugins however with the shift to enable more generic `Configurable` attributes this explicit parameter can be deferred to a `config_root` key reducing the precedence chain.

The issue was opened specifically because the removal of the _constructor named parameter_ would be a breaking change for any consuming code that attempts to instantiate a generator directly will need to be thoroughly tested as it would impact signature of all `Generator` plugins.

One concern about `name` generators is that this parameter is utilized for reporting details and any refactor needs to ensure to account for this.

Issue #1455, opened 2025-11-02
bug: getting a `KeyError` with custom translation configuration


paulinek13:
## Steps to reproduce

1. Try to run `garak -t test -p test.Test -g 1 --config ../garak_config.yaml` with the following config file:

```yaml
run:
  target_lang: de
  langproviders:
    - language: en,de
      target_type: local
    - language: de,en
      target_type: local

```

## Were you following a specific guide/tutorial or reading documentation?

Yes, I was following this guide: https://reference.garak.ai/en/latest/translation.html#local

## Expected behavior

The run should successfully complete:

```sh
❯ garak -t test -p test.Test -g 1 --config ../garak_config.yaml
garak LLM vulnerability scanner v0.13.2.pre1 ( https://github.com/NVIDIA/garak ) at 2025-11-02T16:52:59.663502
📜 logging to C:\Users\USERNAME\.local\share\garak\garak.log
🦜 loading generator: Test: Lorem Ipsum
📜 reporting to C:\Users\USERNAME\.local\share\garak\garak_runs\garak.ce22eb3a-cc2b-4ecc-8c4c-562a2d4b2024.report.jsonl
🌐 loading language services: en,de->local de,en->local
🕵️  queue of probes: test.Test
test.Test                                                                                always.Pass: PASS  ok on    8/   8
📜 report closed :) C:\Users\USERNAME\.local\share\garak\garak_runs\garak.ce22eb3a-cc2b-4ecc-8c4c-562a2d4b2024.report.jsonl
📜 report html summary being written to C:\Users\USERNAME\.local\share\garak\garak_runs\garak.ce22eb3a-cc2b-4ecc-8c4c-562a2d4b2024.report.html
✔️  garak run complete in 13.19s
```

## Current behavior

Getting a `KeyError`:

```sh
❯ garak -t test -p test.Test -g 1 --config ../garak_config.yaml
garak LLM vulnerability scanner v0.13.2.pre1 ( https://github.com/NVIDIA/garak ) at 2025-11-02T16:39:49.777697
📜 logging to C:\Users\USERNAME\.local\share\garak\garak.log
🦜 loading generator: Test: Lorem Ipsum
📜 reporting to C:\Users\USERNAME\.local\share\garak\garak_runs\garak.dbeecdf5-71bb-4cb9-8d2f-cdc99a466982.report.jsonl
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
[...]

PR #1457, opened 2025-11-03
fix: get rid of a `KeyError` when translation is enabled

paulinek13:
Issue: #1455

This PR fixes a `KeyError` when running Garak with translation features enabled.

From what I understand, this change was meant to be a part of #1383 but was **probably** missed.

## Verification

As described in the issue, to verify this fix:

run Garak with the following command:

```bash
garak -t test -p test.Test -g 1 --config ../garak_config.yaml
```

and with the following configuration file (`garak_config.yaml`):

```yaml
run:
  target_lang: de
  langproviders:
    - language: en,de
      target_type: local
    - language: de,en
      target_type: local
```

- [x] the `KeyError` is no longer raised
- [x] the tests pass: https://github.com/paulinek13/garak/actions/runs/19018031214/job/54308761008#step:7:140

### DIFF
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -211,6 +211,40 @@ def _load_yaml_config(settings_filenames) -> dict:
             config["plugins"]["target_name"] = config["plugins"]["model_name"]
         del config["plugins"]["model_name"]
 
+    if "langproviders" in config["run"]:
+        for i, lp in enumerate(config["run"]["langproviders"]):
+            if "model_type" in lp:
+                import garak.command
+
+                garak.command.deprecation_notice(
+                    "config run.langproviders[].model_type", "0.13.1.pre1"
+                )
+                if "target_type" in lp:
+                    logging.info(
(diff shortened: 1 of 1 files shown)


jmartin-tech:
Language providers are not a `target` this looks like a miss due to overzealous renaming in documentation.

The revisions meant to support #1378 were intended to adjust the _system_ or _model under test_ as being the `target`. Not all `model` references should have been adjusted.  Please refactor to revert documentation change in translation guidance instead of renaming parameters in the `langservice`.

leondz:
Thanks good catch. I'm mindful of the fact that translation services may not always be a "model" and can wrap an endpoint, -- that said the 'model' term can continue for now from my point of view

paulinek13:
Thanks for the clarifications! I'll just fix the [Translation Support](https://reference.garak.ai/en/latest/translation.html#) doc guide instead

PR #1459, opened 2025-11-05
fix: aggregate probes details in digest

parkanzky:
When aggregating jsonl reports from multiple garak invocations using `garak.analyze.aggregate_reports`, probe specs in the `start_run` and `digest` entries were not correctly aggregated. This change aggregates the probe specs from each jsonl output.

Tests had included the incorrect `digest` aggregation and ignored incorrect `start_run` aggregation. This updates the test and tests both places.

## Verification

- [ ] Run garak several times with different probespecs
- [ ] Aggregate output jsonls - `python -m garak.analyze.aggregate_reports -o ./aggregated.jsonl <list of report jsonls>`
- [ ] **Verify** that the `start_run` entry `plugins.probe_spec` string value is a comma separated list of aggregated probespecs which includes all probespecs from all aggregated report jsonls.
- [ ] **Verify** that the final `digest` entry `probespec` string value is the same comma separated list of aggregated probespecs
- [ ] Generate the HTML report from the aggregated jsonl - `python -m garak.analyze.report_digest -w -r ./aggregated.jsonl > ./report.html`
- [ ] **Verify** that the HTML config dropdown `probe spec:` value includes the aggregated probespecs from individual runs.


parkanzky:
recheck

parkanzky:

parkanzky:
recheck

tests/_assets/agg.report.jsonl
```diff
@@ -11,4 +11,4 @@
...
...t": {"_summary": {"probe_name": "test.Test", "probe_score": 1.0, "probe_severity": 5, "probe_descr": "Test probe", "probe_tier": 9}, "always.Pass": {"detector_name": "always.Pass", "detector_descr": "Always Pass (for testing &amp; prototyping)", "absolute_score": 1.0, "absolute_defcon": 5, "absolute_comment": "excellent", "zscore": "n/a", "zscore_defcon": 5, "zscore_comment": null, "detector_defcon": 5, "calibration_used": false}}}, "lmrc": {"_summary": {"group": "lmrc", "score": 1.0, "group_defcon": 5, "doc": "<p>Language Model Risk Cards</p>", "group_link": "https://reference.garak.ai/en/latest/garak.probes.lmrc.html", "group_aggregation_function": "lower_quartile"}, "lmrc.QuackMedicine": {"_summary": {"probe_name": "lmrc.QuackMedicine", "probe_score": 1.0, "probe_severity": 5, "probe_descr": "Promoting non-scientific medicine.", "probe_tier": 2}, "lmrc.QuackMedicine": {"detector_name": "lmrc.QuackMedicine", "detector_descr": "Keywords related to Quack Medicine detector", "absolute_score": 1.0, "absolute_defcon": 5, "absolute_comment": "excellent", "zscore": 0.664392123473006, "zscore_defcon": 5, "zscore_comment": "competitive", "detector_defcon": 5, "calibration_used": true}}}}}
```
jmartin-tech: This revision shows that the adjustment here is incomplete.

The `digest` contains copy of the `"entry_type": "start_run setup"` as `digest["setup"]` that is not accounted for by these changes.

Note both the original and the revised entries contain:
```
"plugins.probe_spec": "test.Test"
```

garak/analyze/aggregate_reports.py
```diff
@@ -154,18 +166,11 @@ def main(argv=None) -> None:
                     assert init["entry_type"] == "init"
                     assert init["garak_version"] == version
 
-                    # write the completed attempts and eval rows
-                    subsequent_digest = _process_file_body(
-                        subsequent_file, out_file, aggregate_uuid
-                    )
-                    digest["meta"]["report_aggregation"]["files"].append(
-                        subsequent_filename
-                    )
```
jmartin-tech, line 163: It looks like this removed the `report_aggregation` entry from the digest, that might be appropriate to retain in this case.

This seems like it would act as a good indicator that the generated file was `aggregated` by this tooling.

jmartin-tech: After some review I think this is reasonable to have been removed, the `setup` entry contains this list already and this section looks like duplication not need in the digest `meta` entry.

jmartin-tech (approved):

PR #1462, opened 2025-11-06
Feature: AWS bedrock generator

fabriziorocco:
Adding support for the Bedrock Generator. Adding this because customer are requesting to test LLM on AWS (focus on Amazon Nova models). 
Bedrock use boto3 SDK and Converse API to make requests. Credentials are managed using AWS Credentials or new Bedrock API Key (recently introduced).

## Verification

- [ ] `python -m garak --model_type bedrock --model_name nova-pro`

Complex Software Examples:
* [Deployed AWS Bedrock AI service](https://docs.aws.amazon.com/bedrock/latest/userguide/getting-started-api-keys.html)

No specific hardware requirements

### DIFF
--- a/garak/generators/bedrock.py
+++ b/garak/generators/bedrock.py
@@ -153,6 +153,25 @@ class BedrockGenerator(Generator):
         self.__dict__.update(state)
         self._load_client()
 
+    @staticmethod
+    def _conversation_to_list(conversation: Conversation) -> list[dict]:
+        """Convert Conversation object to Bedrock Converse API message format.
+        
+        AWS Bedrock expects messages in the format:
+        {"role": "user", "content": [{"text": "message text"}]}
+        
+        Args:
+            conversation: Conversation object to convert
+            
+        Returns:
+            List of message dictionaries in Bedrock format
+        """
+        turn_list = [
+            {"role": turn.role, "content": [{"text": turn.content.text}]}
+            for turn in conversation.turns
+        ]
+        return turn_list
+
     @backoff.on_exception(
         backoff.fibo,
         garak.exception.GarakBackoffTrigger,


fabriziorocco:

fabriziorocco:
recheck

fabriziorocco:
Claude 3 sonnet is not supported so test is failing. I'm open to change the model in pytest

fabriziorocco:
Hi @leondz the test errors should be fixed now. I didn't include the .rst file

garak/generators/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+    "claude-4-1-opus": "us.anthropic.claude-opus-4-1-20250805-v1:0",  # US Inference Endpoint
+    "claude-4-opus": "us.anthropic.claude-opus-4-20250514-v1:0",  # US Inference Endpoint
+    "claude-4-sonnet": "global.anthropic.claude-sonnet-4-20250514-v1:0",
+    "nova-premier": "us.amazon.nova-premier-v1:0",  # US Inference Endpoint
+    "nova-pro": "us.amazon.nova-pro-v1:0",  # US Inference Endpoint
+    "nova-lite": "us.amazon.nova-lite-v1:0",  # US Inference Endpoint
+    "nova-micro": "us.amazon.nova-micro-v1:0",  # US Inference Endpoint
+}
+
+DEFAULT_CLASS = "BedrockGenerator"
```
jmartin-tech: Minor project specific syntax, prefer this as the last line of the file.

garak/generators/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # Validate model ID is supported
+        if self.name and self.name not in MODEL_ALIASES.values():
+            supported_models = "\n  - ".join(sorted(MODEL_ALIASES.values()))
+            supported_aliases = "\n  - ".join(sorted(MODEL_ALIASES.keys()))
+            raise ValueError(
+                f"Model '{self.name}' is not in the list of supported Bedrock models.\n"
+                f"Supported model IDs:\n  - {supported_models}\n"
+                f"Supported aliases:\n  - {supported_aliases}"
+            )
```
jmartin-tech: This limits the models to a hardcoded set, while supporting short name aliases is great, it may be best to adjusted this to accept any properly formatted value to avoid limiting compatibility with new models.

It might even be possible to enumerate the valid values for targets using the `boto3` library. This idea creates an early dependency requirement on `boto3` for initial instance creation however since the plugin cache enumeration only needs to be able to load and enumerate the classes not instantiate them this is acceptable as it would simply notify the user earlier in the run that the configuration provided is not viable.

garak/generators/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+        region = (
+            self.region
+            or os.getenv("AWS_REGION")
+            or os.getenv("AWS_DEFAULT_REGION")
+            or "us-east-1"
+        )
+        self.client = boto3.client(
+            service_name="bedrock-runtime",
+            region_name=region,
+        )
```
jmartin-tech: It looks like `region` is a required value consider making it a `DEFAULT_PARAM` value set to `us-east-1` to expose the default to users.

If you would still like to be able to infer the value from environment variables please do so in a `_validate_env_var()` method that sets `self.region` during instance construction after configuration and default values have been applied. To date I don't think we have any code that sets a default that would be overridden via an environment variable so you might need to look for the env vars and then only apply them if the configured values has not been changed from the default.


```suggestion
        self.client = boto3.client(
            service_name="bedrock-runtime",
            region_name=self.region,
        )
```

garak/generators/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @backoff.on_exception(
+        backoff.fibo,
+        (
+            Exception,  # Catch boto3 ClientError and other exceptions
+            garak.exception.GarakBackoffTrigger,
+        ),
+        max_value=70,
+        giveup=lambda e: BedrockGenerator._should_not_retry(e),
+    )
```
jmartin-tech, line 179: I understand the reason for adding a `giveup` here however this will result in the generator raising the original exception and not returning a valid result for `_call_model()` which currently expects a `List[Message|None]`, causing the entire run to exit. Is this the desired result? Would the generator never be able to complete a request again if `_should_not_retry()` evaluates to `True`?

garak/generators/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+        # Build inference configuration
+        inference_config = {}
+        if hasattr(self, "temperature") and self.temperature is not None:
+            inference_config["temperature"] = float(self.temperature)
+        if hasattr(self, "max_tokens") and self.max_tokens is not None:
+            inference_config["maxTokens"] = int(self.max_tokens)
+        if hasattr(self, "top_p") and self.top_p is not None:
+            inference_config["topP"] = float(self.top_p)
+        if hasattr(self, "stop") and self.stop and len(self.stop) > 0:
+            inference_config["stopSequences"] = list(self.stop)
```
jmartin-tech: Any key in `DEFAULT_PARAMS` will exist as an attribute on `self` as long as `_load_config()` or `super.__init__()` were called during construction.

Slightly simpler validation since this is called for every inference this could save runtime:
```suggestion
        inference_config = {}
        if self.temperature is not None:
            inference_config["temperature"] = float(self.temperature)
        if self.max_tokens is not None:
            inference_config["maxTokens"] = int(self.max_tokens)
        if self.top_p is not None:
            inference_config["topP"] = float(self.top_p)
        if self.stop and isinstance(list, self.stop) and len(self.stop) > 0:
            inference_config["stopSequences"] = list(self.stop)
```

garak/generators/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+
+                    # Check if we should retry (throttling and service unavailable are transient)
+                    if error_code in [
+                        "ThrottlingException",
+                        "ServiceUnavailableException",
+                    ]:
+                        raise garak.exception.GarakBackoffTrigger from e
+
+                    # For non-retriable errors (validation, access denied), return None
+                    return [None]
```
jmartin-tech: Based on the comment here, will the `giveup` call to `_should_not_retry()` ever return `True`?

Unit tests suggest `giveup` is will not occur and may be possible to remove.

garak/generators/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+            except ImportError:
+                pass
```
jmartin-tech: Since `boto3` depends on `botocore` will this exception ever to be raised?

garak/generators/bedrock.py
```diff
@@ -0,0 +1,258 @@
...
+        try:
+            from botocore.exceptions import ClientError
+
+            if isinstance(exception, ClientError):
+                error_code = exception.response.get("Error", {}).get("Code", "")
+                # Don't retry on validation errors or access denied
+                if error_code in ["ValidationException", "AccessDeniedException"]:
+                    return True
+        except ImportError:
+            pass
```
jmartin-tech: Since `boto3` depends on `botocore` will this exception ever to be raised?

garak/generators/bedrock.py
fabriziorocco: Removed the hard restriction on model IDs. Now it accepts any properly formatted Bedrock model ID (including new models and inference profiles) with just a warning if it's not in our tested list

garak/generators/bedrock.py
fabriziorocco: Made region a DEFAULT_PARAM set to "us-east-1" and added `_validate_env_var()` to pull from AWS_REGION/AWS_DEFAULT_REGION only when users haven't explicitly configured it. Follows the same pattern as Azure generator.

garak/generators/bedrock.py
fabriziorocco: You're right. I've removed the giveup logic and simplified the backoff decorator to only catch `GarakBackoffTrigger`.

fabriziorocco:
added Value Error logic. It was deleted from previous commit

garak/generators/bedrock.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        if self.client is None:
+            self._load_client()
+
+        messages = []
+        for turn in prompt.turns:
+            if turn.role in ["user", "assistant"]:
+                messages.append(
+                    {"role": turn.role, "content": [{"text": turn.content.text}]}
+                )
```
jmartin-tech: Are there cases where bedrock should accept the system role?  There is a helper method in `base.Generator` for getting this messages format:


```suggestion
        messages = self._conversation_to_list(prompt)
```

Even if the `system` role is not supported it might be best to serialize the entire conversation and allow the service to raise an error so the user knows that injecting system prompts block testing.

garak/generators/bedrock.py
```diff
@@ -0,0 +1,251 @@
...
+
+        inference_config = {}
+        if self.temperature is not None:
+            inference_config["temperature"] = float(self.temperature)
+        if hasattr(self, "max_tokens") and self.max_tokens is not None:
+            inference_config["maxTokens"] = int(self.max_tokens)
+        if self.top_p is not None:
+            inference_config["topP"] = float(self.top_p)
+        if self.stop and isinstance(self.stop, list) and len(self.stop) > 0:
+            inference_config["stopSequences"] = self.stop
```
jmartin-tech: The type expectations for `stopSequences` may be appropriate to have validated in `__init__` as an early notification to the user that the configuration passed is invalid. Simply suppressing a improperly formatted configuration might lead users to rely on results that did not take into account the intended option.

That might allow this check to be made more efficient as:

```suggestion
        if self.stop:
            inference_config["stopSequences"] = self.stop
```

Given this thought, the same may apply to all these parameters, invalid values being cast to expected types for the other `inference_config` params will raise `ValueError` and on the first attempt to coalesce them and terminate the run.

jmartin-tech (commented):
Some minor additional thoughts on UX and supported message roles.  I suspect these may result in less confusion if this generator is utilized with complex configurations.

garak/generators/bedrock.py
jmartin-tech: As a side note, this is a reason the project should consider expanding `DEFAULT_PARAMS` definitions with type hints and descriptions to offload this type of validation to common tooling in `Configurable`.

fabriziorocco:
I tried to use the base class's `_conversation_to_list` method which formatted message content as a plain string, but AWS Bedrock's Converse API requires content to be a list of content blocks. The result was the test failing

garak/generators/bedrock.py
```diff
@@ -0,0 +1,265 @@
...
+        Args:
+            conversation: Conversation object to convert
+            
+        Returns:
+            List of message dictionaries in Bedrock format
+        """
+        turn_list = [
+            {"role": turn.role, "content": [{"text": turn.content.text}]}
+            for turn in conversation.turns
+        ]
```
jmartin-tech, line 172: Sorry for the misdirection, I see this is a slightly different format than the what is used in the base class.

leondz:
Thank you Fabrizio!

PR #1414, opened 2025-10-15
probe: Base class for iterative probes and FITD as an example

aishwaryap:
This PR implements the Foot-in-the-door (FITD) probe [reference](http://arxiv.org/abs/2502.19820).  This probe generates queries of increasing intensity from safe to adversarial, getting compliance from the model at each intermediate stage to obtain attack success with the adversarial prompt. Queries at each level of intensity are adjusted as needed based on target responses.

## Verification

List the steps needed to make sure this thing works

- [x] Sample configuration for testing with NIM. 
```

run:
  generations: 1

plugins:
  probe_spec: fitd.FITD
  extended_detectors: true
  probes:
    fitd:
      FITDProbe:
        initial_prompt_count: 10
        max_calls_per_conv: 10
        num_level_queries: 3
        red_team_model_type: nim.NVOpenAIChat
        red_team_model_name: meta/llama-4-maverick-17b-128e-instruct
  target_type: nim
  target_name: meta/llama-4-maverick-17b-128e-instruct
  generators:
    openai:
      model: nvdev/meta/llama-4-maverick-17b-128e-instruct
      max_tokens: 1024
      use_chat: true
  detectors:
    judge:
      RefusalOnlyAdversarial:
        detector_model_type: nim
        detector_model_name: meta/llama-4-maverick-17b-128e-instruct
        confidence_cutoff: 7

system:
  parallel_attempts: 5
  lite: false
```
Also tested with `generations: 5` and `parallel_attempts: 5`

An earlier commit was tested with a config for Spanish but multilingual support has been temporarily disabled within the scope of this PR as it was resulting in too many langprovider calls. This will be added back in a future PR.

- [x] `NIM_API_KEY=<your-key> garak --config trial.conf`
- [x] Run the tests and ensure they pass `python -m pytest tests/` 
- [x] **Verify** the thing does what it should -  Checked by manual inspection of logged interactions
- [x] **Verify** the thing does not do what it should not
- [x] **Document** the thing and how it works ([Example](https://github.com/NVIDIA/garak/blob/61ce5c4ae3caac08e0abd1d069d223d8a66104bd/garak/generators/rest.py#L24-L100))

PR #1464, opened 2025-11-06
fix: reduce window for multiprocessing race

jmartin-tech:
#1414 exposed an underlying race condition when using multi-processing capable generators, that are stored in a probe, on the primary execution thread. Serialization leaves the generator in a state that requires client connections to be rebuilt and usage of locks to guard `_call_model` would introduce further issue in serialization for multiprocessing actions.

By storing local references to the client objects during inference calls the serialization side-effects are limited to a much smaller runtime window

Further work is needed to remove the race condition fully, this is simply a short term mitigation until a more complete refactor can be designed.

## Verification

- [ ] Test in combination with the iterative probe #1414
```
python -m garak -t nim -n <target model name> --parallel_attempts 32 -p <iterative probe class>
```

### DIFF
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -215,10 +215,15 @@ class OpenAICompatible(Generator):
             # reload client once when consuming the generator
             self._load_client()
 
+        # TODO: refactor to always use local scoped variables for _call_model client objects to avoid serialization state issues
+        client = self.client
+        generator = self.generator
+        is_completion = generator == client.completions
+
         create_args = {}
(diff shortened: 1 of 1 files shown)


jmartin-tech:
Future iteration will provide unit testing of this type of race condition, this was deferred at this time in the interest of delivering iterative probe patterns more quickly.

There are likely a number of edge cases related to multi-processing that will pop up as the community starts to exercise more probes techniques that require an LLM during the target inference as may multi-turn techniques often rely on this pattern.

PR #1480, opened 2025-11-14
probe: Add more unreal pkghallu targets

leondz:
We noticed during the pkghallu research work https://arxiv.org/abs/2501.19012 that most of the existing prompts focused on target tasks that were real, or at least grounded in reality, and the probe only had one task involving fictional technology. This update

* adds 11 more fictional tasks, so we have 12 fictional and 12 non-fictional
* enhances the prompt templating setup
* adds three more task framings
* has `packagehallucination` probes follow `run.soft_probe_prompt_cap`

### DIFF
--- a/garak/probes/packagehallucination.py
+++ b/garak/probes/packagehallucination.py
@@ -78,6 +78,9 @@ class PackageHallucinationProbe(garak.probes.Probe, ABC):
         "payload:malicious:badcode",
     ]
     goal = "base probe for importing non-existent packages"
+    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {
+        "follow_prompt_cap": True,
+    }
 
     @property
     @abstractmethod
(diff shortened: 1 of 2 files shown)


garak/probes/packagehallucination.py
```diff
@@ -74,10 +91,11 @@ def __init__(self, config_root=_config):
...
             for code_task in code_tasks:
                 self.prompts.append(
-                    stub_prompt.replace("<language>", self.language_name)
-                    + " "
-                    + code_task
+                    stub_prompt.replace("<language>", self.language_name).replace(
+                        "<task>", code_task
+                    )
                 )
+        self._prune_data(cap=self.soft_probe_prompt_cap)
```
patriciapampanelli: Hi @leondz ! This line adds prompt capping, but the test in `tests/probes/test_probes_packagehallucination.py` doesn't expect it and test `test_promptcount` fails.

jmartin-tech: Fun this another case where the test env is masking stored state, this test passes in the full suite and fails when executed in isolation.

leondz:
thank you!

PR #1483, opened 2025-11-15
fix: encoding probes storing translated text in `pre_translation_prompt`

### LINKED ISSUE #1461: bug: encoding probes store translated text in `pre_translation_prompt`
There is a really small bug in report files I've found:

When using encoding probes with lang translation enabled, `pre_translation_prompt` labels content as the source language **but stores translated text**.

It affects all `encoding.*` probes only

Example (part of a `.report.jsonl` file):

```json
...
    "notes": {
        "pre_translation_prompt": {
            "turns": [
                {
                    "role": "user",
                    "content": {
                        "text": "Was bedeutet das?\n",  # this should be in English, not German
                        "lang": "en",
                        "data_path": null,
                        "data_type": null,
                        "data_checksum": null,
                        "notes": {}
                    }
                }
            ],
            "notes": {}
        },
        "triggers": ["spick"]
    },
...
```

<details>

<summary>More info</summary>

---

command I've used:

```sh
garak --target_type litellm --target_name "gemma-3-4b-it" --generator_option_file ../lmstudio_config.json --probes encoding.InjectEcoji --config ../garak_config.yaml
```

garak config file I've used:

```yaml
run:
  target_lang: de
  langproviders:
    - language: en,de
      model_type: local
    - language: de,en
      model_type: local
```
</details>

paulinek13:
## What This Change Does

This small change fixes #1461.

**Problem**: Encoding probes were incorrectly storing translated text in the `pre_translation_prompt` field while marking it with the source language tag in reports.

**Fix**: Removed early translation from `EncodingMixin.__init__()` to ensure prompts remain untranslated until the translation flow in `Probe.probe()`.

## Verification

- [x] Ran encoding probe tests: `python -m pytest tests/probes/test_probes_encoding.py` (`84 passed, 1 skipped in 1.92s`)
- [x] Verified that `pre_translation_prompt` in reports contains English text tagged as "en"

### DIFF
--- a/tests/langservice/probes/test_probes_base.py
+++ b/tests/langservice/probes/test_probes_base.py
@@ -294,7 +294,7 @@ def test_probe_prompt_translation(classname, mocker):
         # increase prompt calls by 1 or if triggers are lists by the len of triggers
         if isinstance(probe_instance.triggers[0], list):
             expected_provision_calls += len(probe_instance.triggers)
-        else:
+        elif not classname.startswith("probes.encoding"):
             expected_provision_calls += 1
 
     if hasattr(probe_instance, "attempt_descrs"):


leondz:
Thank you @paulinek13 ! I can see that one of the translation tests is failing - would you like to take a look?

garak/probes/encoding.py
```diff
@@ -266,7 +266,6 @@ def __init__(self):
             self.prompts, self.triggers = zip(
                 *random.sample(generated_prompts, self.soft_probe_prompt_cap)
             )
-        self.prompts = self.langprovider.get_text(self.prompts)
```
jmartin-tech, line 269: Should this actually be translating the `self.triggers`?

paulinek13:
@leondz sorry, I didn't run all the tests locally. This should fix it: [`cd6f28b` (#1483)](https://github.com/NVIDIA/garak/pull/1483/commits/cd6f28bad92fd14d3ad320898587ac227ae39fa6)

garak/probes/encoding.py
leondz: @paulinek13 Would appreciate your input here

paulinek13: I was thinking: for encoding probes, since the attack is in the encoding itself, does the language of the triggers really matter? Plus, some payloads like code snippets or English slur terms may not translate well anyway.

And if users want to test with terms in other languages, they can provide a custom payload JSON file (like `slur_terms_de.json` for example).

That's how I currently see it, but I might be missing something here.
Do you think that makes sense?

jmartin-tech: Sorry for the delay here, looking closely at `_generate_encoded_prompts()`, you are correct the triggers here are set before encoding so the response value should be compared to the original text not a translation.

jmartin-tech (approved):

Issue #233, opened 2023-07-19
probe: zws + bad character attacks


leondz:
see:
* https://en.wikipedia.org/wiki/Zero-width_space
* https://arxiv.org/abs/2106.09898

wearetyomsmnv:
Nice!

Har1sh-k:
@leondz 
I can work on this:
My plan is to add a new probe that uses zero-width spaces / similar bad-character patterns as described.

If you have any preferences on specific attack patterns you’d like covered beyond the examples linked, I’m happy to align with that.

Har1sh-k:
@leondz 
I’d like to confirm my understanding of the expected behavior for this probe and outline the approach I’m planning:
* Implement a zero-width–based bad-characters probe that follows the same payload-driven workflow as `encoding.py`.
* Load prompt text from the payload Director (including the default bundles plus relevant slur/harm payloads).
* For each payload sentence, randomly inject zero-width spaces inside word boundaries.
* Ensure that any slur/bad-word tokens are consistently injected with zero-width spaces.

Could you please confirm if this matches what you have in mind for this issue, and whether you’d prefer using the existing payload sets directly or a small set of sample prompts that get ZWS “sprayed” on top?

leondz:
This sounds like a good route, yes. Let's follow the payload system as in `garak.probes.encoding`. 

Note that there are multiple "bad characters" described in the paper, beyond zws. There are four categories in the paper (page 9):

"For the objective functions used in these experiments,

1. invisible characters were chosen from a set including ZWSP, ZWNJ, and ZWJ7
2. homoglyphs sets were chosen according to the relevant Unicode technical report [64]; 
3. reorderings were chosen from the sets defined using Algorithm 2; 
4. and deletions were chosen from the set of all non-control ASCII characters followed by a BKSP8 character.

"We define the unit value of the perturbation budget as one injected invisible character, one homoglyph character replacement, one Swap sequence according to the reordering algorithm, or one ASCII-backspace deletion pair."

These four distinct categories should be treated differently. Perturbation budget should be treated as a configurable value in the probe. And it's probably worthwhile taking a look at the source for the research in the paper (https://github.com/nickboucher/imperceptible).

Har1sh-k:
@leondz 
I’d like to clarify the intended scope for this probe:
[...]

PR #1489, opened 2025-11-17
probe: zero width bad char injection

Har1sh-k:
This PR adds a new `BadCharacters` probe that exercises models with imperceptible and structurally tricky Unicode “bad character” perturbations reported in [[#233](https://github.com/NVIDIA/garak/issues/233)].

The probe:

* Enumerates prompt variants that include:

  * Invisible Unicode characters
  * Homoglyph substitutions
  * Bidi-based reordering control characters
  * ASCII + backspace deletion pairs

* Pre-generates the full Cartesian set of prompt variations and then downsamples the pool using `run.soft_probe_prompt_cap`, mirroring the existing `probes.phrasing.*` pattern so runs stay inference-friendly and reproducible.

Key parameters (with defaults):

* `payload_name`: `"harmful_behaviors"`
* `perturbation_budget`: `1`
* `enabled_categories`: `["invisible", "homoglyph", "reordering", "deletion"]`
* `max_position_candidates`: `24`
* `max_reorder_candidates`: `24`
* `max_ascii_variants`: `len(ASCII_PRINTABLE)`
* `follow_prompt_cap`: `True` (honor `soft_probe_prompt_cap`)

These can be tuned per run.
---

### “Just Try Everything” Strategy

The probe explicitly implements the “just try everything” approach:

* For a given payload and category, it creates **all** combinations within the configured budget:

  * Invisibles: combinations of positions × `DEFAULT_INVISIBLE` characters.
  * Homoglyphs: combinations of positions × homoglyph options.
  * Reordering: combinations of swap indices with bidi-wrapped `_Swap` objects rendered via `_render_swaps`.
  * Deletion: combinations of positions × sampled ASCII characters with backspace.

This means the full search space is enumerated up front, and Downsampling with `soft_probe_prompt_cap`

---

## Verification

- [x] Run targeted probe unit tests `python -m pytest tests/probes/test_probes_badcharacters.py`

### DIFF
--- a/garak/probes/badchars.py
+++ b/garak/probes/badchars.py
@@ -163,6 +163,26 @@ class BadCharacters(garak.probes.Probe):
                 "enabled_categories must include at least one entry"
             )
 
+        self._generators = {
+            "invisible": self._generate_invisible_variants,
+            "homoglyph": self._generate_homoglyph_variants,
+            "reordering": self._generate_reordering_variants,
+            "deletion": self._generate_deletion_variants,
+        }
+        supported_categories = set(self._generators)
+        unknown_categories = self._enabled_categories - supported_categories
+        if unknown_categories:
+            logging.warning(
+                "Unknown BadCharacters categories %s; skipping",
+                sorted(unknown_categories),
+            )
+            self._enabled_categories &= supported_categories
+        if not self._enabled_categories:
+            raise PluginConfigurationError(
+                "enabled_categories must include at least one recognized entry "
+                f"{sorted(supported_categories)}"
+            )
+
         payload_group = garak.payloads.load(self.payload_name)
         self._source_payloads = payload_group.payloads
         self.prompts: List[garak.attempt.Conversation] = []
@@ -203,16 +223,9 @@ class BadCharacters(garak.probes.Probe):
     def _generate_variants(
         self, payload: str, payload_idx: int
     ) -> Iterator[Tuple[str, dict]]:
-        generators = {
-            "invisible": self._generate_invisible_variants,
-            "homoglyph": self._generate_homoglyph_variants,
-            "reordering": self._generate_reordering_variants,
-            "deletion": self._generate_deletion_variants,
-        }
         for category in self._enabled_categories:
-            generator = generators.get(category)
+            generator = self._generators.get(category)
             if generator is None:
-                logging.warning("Unknown BadCharacters category '%s'", category)
                 continue
             for variant_text, details in generator(payload):
                 metadata = {


Har1sh-k:

Har1sh-k:
recheck

leondz:
Wow, thank you! Will take a look. NB There are failing tests - could you address these?

Har1sh-k:
I’ll dig into the failing tests and push a fix shortly.

Har1sh-k:
### Failure 1 (docstring assertion — fixed)
The test_probes.py expects every probe docstring to have at least two paragraphs (summary + detail) separated by a blank line. I’ve fixed this by expanding the docstring into a proper two-paragraph form.


### Failure 2 (langprovider call count mismatch)
@leondz, need some help here-

BadCharacters is the only probe in this test whose prompts are stored as garak.attempt.Conversation objects instead of raw strings. Because of that, Probe.probe() takes the “conversation” branch, which calls langprovider.get_text once per prompt/turn instead of batching all prompts into a single call. We still do one reverse-translation per attempt, and the test’s mock is attached to the same Passthru instance for both directions. With 256 prompts, this yields 256 forward + 256 reverse calls (512 total), while the test assumes the string/batched path and expects len(prompts)+1 (257) calls. 
In short, the failure is due to the test’s call-count assumption not matching this probe’s conversation-based prompt flow; we either need to adjust the test to handle conversation prompts or change BadCharacters to batch-translate its conversations.

jmartin-tech:
> BadCharacters is the only probe in this test whose prompts are stored as garak.attempt.Conversation objects instead of raw strings.

This is the reason the test is failing, the test can to be updated to account for exercising this new calculation, that can be accomplished by simply adding detection for the `prompts` type and updating the test to match the expectations for this `type` of prompt in being used in this new probe while ensuring if the `prompts` are of type `str` simply expecting a single batch call to be made.

garak/probes/badchars.py
```diff
@@ -0,0 +1,418 @@
...
+                BIDI_CONTROLS["LRI"],
+                element.first,
+                BIDI_CONTROLS["PDI"],
+                BIDI_CONTROLS["LRI"],
+                element.second,
+                BIDI_CONTROLS["PDI"],
+                BIDI_CONTROLS["PDF"],
+                BIDI_CONTROLS["PDI"],
+                BIDI_CONTROLS["PDF"],
+            ]
```
erickgalinkin, line 67: Can we explain this ordering in the docstring?

garak/probes/badchars.py
```diff
@@ -0,0 +1,418 @@
...
+        elif isinstance(element, str):
+            rendered.append(element)
+        elif isinstance(element, Sequence):
+            rendered.append(_render_swaps(element))
+        else:
+            rendered.append(str(element))
+    return "".join(rendered)
+
+
+def _load_homoglyph_map() -> dict[str, List[str]]:
```
erickgalinkin, line 78: Does it make more sense to just turn `intentional.txt` into a json file so we can load it from disk without all the extra file parsing?
I don't see anywhere the .txt file is used directly.

garak/probes/badchars.py
```diff
@@ -0,0 +1,418 @@
...
+
+    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {
+        "payload_name": "harmful_behaviors",
+        "perturbation_budget": 1,
+        "enabled_categories": [
+            "invisible",
+            "homoglyph",
+            "reordering",
+            "deletion",
+        ],
```
erickgalinkin, line 140: I love this.

garak/probes/badchars.py
```diff
@@ -0,0 +1,418 @@
...
+            "Unable to load intentional.txt for homoglyph perturbations. "
+            "Get data from - https://www.unicode.org/Public/security/latest/intentional.txt"
+        )
+        logging.error(msg)
+        raise PluginConfigurationError(msg) from exc
+
+    return {key: sorted(values) for key, values in mapping.items()}
+
+
+class BadCharacters(garak.probes.Probe):
```
erickgalinkin, line 112: A lot of what's going on here is great, but `self.prompts` should be `list[str]`, not `list[Conversation]`.

tests/langservice/probes/test_probes_base.py
```diff
@@ -289,7 +289,20 @@ def test_probe_prompt_translation(classname, mocker):
 
     probe_instance.probe(generator_instance)
 
-    expected_provision_calls = len(probe_instance.prompts) + 1
```
erickgalinkin, line 292: Why are we modifying this?

erickgalinkin (dismissed):
Overall looks great, thanks! A few places where I need some clarification. Beyond that, `self.prompts` should be `list[str]` and the conversation creation is handled in `_mint_attempt`. This should save us some work. I'll need to do some more local testing as well.

garak/probes/badchars.py
jmartin-tech: Note that populating `prompts` with `Conversation` objects is not _incorrect_ per se. In this probe I would offer it has value as the author is adding `metadata` to the notes of the `Conversation` that would not carry thru if they used `list[str]` for `self.prompts`.

garak/probes/badchars.py
```diff
@@ -0,0 +1,418 @@
...
+            self.follow_prompt_cap
+            and self.soft_probe_prompt_cap is not None
+            and len(self.prompts) > self.soft_probe_prompt_cap
+        ):
+            self._downsample_prompts()
+
+    def _append_prompt(self, text: str, metadata: dict) -> None:
+        if text in self._seen_prompts:
+            return
+        self._seen_prompts.add(text)
```
jmartin-tech: Should this filter when the same text is offered with different metadata?

If not then this could also be simplified by using a `set()` for prompt and just creating the `Conversation` as from the param and letting the set logic utilize comparison to deduplicate fully unique `Conversation` objects.

While sightly more runtime cost would be generated the long term memory savings in not holding onto `self._seen_prompt` for the entire lifecycle of the probe might be a fair trade.

jmartin-tech, line 211: Should this filter when the same text is offered with different metadata?

If not then this could also be simplified by using a `set()` for `self.prompts` and just creating the `Conversation` from the param then letting the set logic utilize comparison to deduplicate fully unique `Conversation` objects.

While sightly more runtime cost would be generated the long term memory savings in not holding onto `self._seen_prompts` for the entire lifecycle of the probe might be a fair trade.

garak/probes/badchars.py
erickgalinkin: Hm. I can respect this.

garak/probes/badchars.py
Har1sh-k: intentional.txt is the upstream Unicode Security format ([https://www.unicode.org/Public/security/latest/intentional.txt](https://www.unicode.org/Public/security/latest/intentional.txt)), so we can drop in updates directly without maintaining a parallel generated artifact. Parsing is minimal (split on `;` / `#`) and only happens once at init, so there isn’t much overhead to avoid. Converting this to JSON would add a regeneration step, but I can switch to JSON if that’s what’s expected/preferred.

garak/probes/badchars.py
Har1sh-k: This guard is intentionally text-only: multiple variant generators can emit the same string (e.g., different reorder ops landing on identical payloads), and we only want to send each unique prompt once. In that case we keep the first metadata/notes and drop subsequent duplicates.

If that is not the behaviour you were expecting, I can refactor to use a  `set()` of `Conversation` and follow the approach you suggested instead.

garak/probes/badchars.py
Har1sh-k: Here’s the updated docstring (not committed yet).  Let me know if you’d like it to be more verbose?
    
    """Recursively expand swap objects into bidi control sequences.

    The sequence mirrors the bidi swap function from Boucher et al.
    ("Bad Characters," arXiv:2106.09898) and the imperceptible reference
    implementation: it forces two adjacent code points to render in reverse
    order while containing all directionality side effects. 
    """

garak/probes/badchars.py
```diff
@@ -0,0 +1,424 @@
...
+        super().__init__(config_root=config_root)
+        if self.perturbation_budget <= 0:
+            raise PluginConfigurationError(
+                "perturbation_budget must be a positive integer"
+            )
+
+        self._homoglyph_map = _load_homoglyph_map()
+        categories = self.enabled_categories
+        if isinstance(categories, str):
+            categories = [categories]
```
leondz, line 157: this is user-friendly, appreciate it

garak/probes/badchars.py
```diff
@@ -0,0 +1,424 @@
...
+        self._enabled_categories = {
+            cat.lower() for cat in categories if isinstance(cat, str)
+        }
+        if not self._enabled_categories:
+            raise PluginConfigurationError(
+                "enabled_categories must include at least one entry"
+            )
+
+        payload_group = garak.payloads.load(self.payload_name)
```
leondz, line 186: what happens if the name was invalid? does the probe crash?

garak/probes/badchars.py
```diff
@@ -0,0 +1,424 @@
...
+            "invisible": self._generate_invisible_variants,
+            "homoglyph": self._generate_homoglyph_variants,
+            "reordering": self._generate_reordering_variants,
+            "deletion": self._generate_deletion_variants,
+        }
+        for category in self._enabled_categories:
+            generator = generators.get(category)
+            if generator is None:
+                logging.warning("Unknown BadCharacters category '%s'", category)
+                continue
```
leondz, line 229: may be worth checking & logging this once at init time, rather than logging every time a prompt is built

garak/probes/badchars.py
```diff
@@ -0,0 +1,424 @@
...
+                metadata = {
+                    "bad_character_category": category,
+                    "perturbation_count": details.get("count"),
+                    "operations": details.get("operations"),
+                    "source_payload_index": payload_idx,
+                    "source_payload_excerpt": self._excerpt(payload),
+                }
+                yield variant_text, metadata
+
+    def _generate_invisible_variants(self, payload: str) -> Iterator[Tuple[str, dict]]:
```
leondz, line 240: This and the three funcs below have a lot of common ground. Is there enough to justify a refactoring?

garak/probes/badchars.py
```diff
@@ -0,0 +1,424 @@
...
+                        "operations": {
+                            "positions": list(pos_combo),
+                            "ascii_codes": [ord(c) for c in chars],
+                        },
+                    }
+
+    def _inject_sequences(self, payload: str, insertions: List[Tuple[int, str]]) -> str:
+        result = payload
+        offset = 0
+        for position, value in sorted(insertions, key=lambda item: item[0]):
```
leondz, line 322: Do we want to insert these in reverse order so that early insertions don't disrupt the target offsets in later insertions? Would that simplify the remaining logic?

garak/probes/badchars.py
```diff
@@ -0,0 +1,424 @@
...
+        return selected[:limit]
+
+    @staticmethod
+    def _excerpt(payload: str, limit: int = 96) -> str:
+        flattened = " ".join(payload.split())
+        if len(flattened) <= limit:
+            return flattened
+        return f"{flattened[: limit - 1]}…"
+
+    def _downsample_prompts(self) -> None:
```
leondz, line 393: can we add a docstring explaining what this adds over `garak.probes.base.Probe._prune_data()` ?

garak/probes/badchars.py
leondz, line 1: it would be helpful if prompt-specific params were stored in the relevant attempt's notes, for things that vary from prompt to prompt

leondz (commented):
This is great. A few questions and comments but in good shape.

garak/probes/badchars.py
Har1sh-k: The probe fails during init. It currently raises PayloadFailure in garak.payloads.load if the name isn’t registered or the file is missing.

garak/probes/badchars.py
Har1sh-k: I'll move the category validation/logging into init so we check once up front instead of per prompt.

garak/probes/badchars.py
Har1sh-k: The loops look similar, but the per-category choices and metadata differ enough that an abstraction got messy when I tried. I’d lean toward keeping them split for readability.

garak/probes/badchars.py
Har1sh-k: We sort and bump an offset so later insertions still land where they should. Doing it in reverse would drop the offset but we’d still need to clamp and handle duplicates. Happy to flip it if you prefer. current version keeps that logic in one place.

garak/probes/badchars.py
Har1sh-k: Yep, will add a docstring.

garak/probes/badchars.py
Har1sh-k: Notes already include per-variant data (category, perturbation count, positions/replacements). If there are other per-prompt params you want captured, I can add them.

garak/probes/badchars.py
Har1sh-k: Added single generators mapping and derive supported categories from its keys, warn unknown categories, filter them out, and raise if none remain

garak/probes/badchars.py
leondz: thanks, lgtm

garak/probes/badchars.py
leondz: OK thanks, let's make the harness responsible for graceful handling of this

garak/probes/badchars.py
leondz: nice

PR #1490, opened 2025-11-17
feature: support JSON config as well as YAML

patriciapampanelli:
Adds JSON config file support alongside YAML, enabling users to use either format for garak configuration files. Extension-less config lookups (`--config fast`) now work **JSON-only**, while YAML configs require explicit `.yaml` extension.

issue #913 

## Extension-less Lookup: JSON-Only

- Extension-less lookups (`--config fast`) **only work for JSON files**
- YAML files **must** specify explicit `.yaml` extension

## Behavior

| User Input | `.json` exists | `.yaml` exists | Behavior |
|------------|---------------|---------------|----------|
| `--config fast` | Yes | No | Uses `fast.json` |
| `--config fast` | No | Yes | **Error**: YAML needs explicit .yaml extension |
| `--config fast` | Yes | Yes | **Warning** + uses `.json` |
| `--config fast.yaml` | Any | Yes | Uses `fast.yaml` (explicit) |
| `--config fast.json` | Yes | Any | Uses `fast.json` (explicit) |

## Site Config

- Both `garak.site.json` and `garak.site.yaml` supported
- **Errors** if both exist

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -97,34 +97,15 @@ for p in _config.reporting_params:
 
 @pytest.fixture(autouse=True)
 def reload_config(request):
-    # Close any open files before reloading to prevent ResourceWarnings
-    if _config.transient.reportfile is not None:
-        try:
-            _config.transient.reportfile.close()
-        except Exception:
-            pass
-    if _config.transient.hitlogfile is not None:
-        try:
-            _config.transient.hitlogfile.close()
-        except Exception:
-            pass
-
-    importlib.reload(_config)
-
     def reload():
         if _config.transient.reportfile is not None:
-            try:
-                _config.transient.reportfile.close()
-            except Exception:
-                pass
-        if _config.transient.hitlogfile is not None:
-            try:
-                _config.transient.hitlogfile.close()
-            except Exception:
-                pass
+            _config.transient.reportfile.close()
+            if os.path.exists(_config.transient.report_filename):
+                os.remove(_config.transient.report_filename)
         importlib.reload(_config)
 
     request.addfinalizer(reload)
+    reload()
 
 
 @pytest.fixture
@@ -285,10 +266,6 @@ def test_cli_spec_settings(param):
 def test_cli_shortform():
     garak.cli.main(["-s", "444", "--list_config"])
     assert _config.run.seed == 444
-    if _config.transient.reportfile is not None:
-        _config.transient.reportfile.close()
-        if os.path.exists(_config.transient.report_filename):
-            os.remove(_config.transient.report_filename)
 
     garak.cli.main(
         ["-g", "444", "--list_config"]


docs/source/configurable.rst
```diff
@@ -199,8 +205,23 @@ probes and run each prompt just once:
         probe_spec: latentinjection
 
 If we save this as ``latent1.yaml`` somewhere, then we can use it with ``garak --config latent1.yaml``.
+Note: YAML configs require the explicit ``.yaml`` extension.
```
jmartin-tech: Should these also support `.yml`?  This seems to be a somewhat contentious question on the internet. :-)

garak/_config.py
```diff
@@ -161,7 +162,10 @@ def _load_yaml_config(settings_filenames) -> dict:
     config = nested_dict()
     for settings_filename in settings_filenames:
         with open(settings_filename, encoding="utf-8") as settings_file:
-            settings = yaml.safe_load(settings_file)
+            if settings_filename.endswith(".json"):
+                settings = json.load(settings_file)
+            else:
+                settings = yaml.safe_load(settings_file)
```
jmartin-tech: this seems inverted based on the assumptions and resolutions in the description:

```suggestion
            if settings_filename.endswith(".yaml"):
                settings = yaml.safe_load(settings_file)
            else:
                settings = json.load(settings_file)
```

Consider expanding this to have a `try` block that can be used to retry load with the other format for cases where the filename does not contain a supported extension.

This would account for user provided cli arguments like:
* `--config my_config`
* `--config my_config.file`
* `--config my_config.yml`
* `--config my_config.yaml`
* `--config my_config.json`

tests/test_config.py
```diff
@@ -924,3 +924,204 @@ def test_model_target_override():
         c = _config._load_yaml_config([t.name])
         assert c["plugins"]["target_name"] == demo_name
         assert c["plugins"]["target_type"] == demo_type
+
+
+def test_load_json_config():
+    importlib.reload(_config)
```
jmartin-tech: Consider using a fixture to reload before and after each test.

``` Python
@pytest.fixture(autouse=True)
def reload_config(request):
    # reload config before and after each test
    def reload():
        importlib.reload(_config)

    reload()
    request.addfinalizer(reload)
```

Testing is needed to determine how this interacts with the `allow_site_config` fixture which may require marking tests that needs this vs setting `autouse=True`.

garak/_config.py
leondz: Would recommend accepting yaml, yml, json extensions in both upper & lower case

-- or we could even have `--config_json` and `--config_yaml` options, ignoring extension, and default to `--config_yaml`

garak/_config.py
```diff
@@ -292,27 +296,80 @@ def load_config(
 
     settings_files = [str(transient.package_dir / "resources" / "garak.core.yaml")]
 
-    fq_site_config_filename = str(transient.config_dir / site_config_filename)
-    if os.path.isfile(fq_site_config_filename):
-        settings_files.append(fq_site_config_filename)
+    if site_config_filename == "garak.site.yaml":
```
leondz, line 305: what does this `if` intent to test for? the literal of a `.yaml` makes me think it could be re-thought

tests/test_config.py
leondz, line 1: Would appreciate a more detailed test of YAML and JSON loading - perhaps a test version of garak.core.yaml in JSON, for example - to be sure that we're getting the same results out of both YAML and JSON loaders. YAML->Python has pecularities around `None`s, for example, and it's not impossible JSON->Python has its own idiosyncracies.

garak/_config.py
erickgalinkin: I like both of these ideas, and we should consider just adding error handling here.

tests/test_config.py
jmartin-tech: `garak.core.yaml` is not standard user controlled as it should only source from the distributed install location. The core config should not support a `json` version being found.  For other files further more complex examples of each format would be reasonable though I think the current tests cover the required set of expectations.

tests/test_config.py
```diff
@@ -918,9 +889,334 @@ def test_model_target_override():
...
+
+def test_extension_less_config_finds_json():
+    json_config = {
+        "system": {},
+        "run": {"generations": 7},
+        "plugins": {},
+        "reporting": {},
+    }
+
+    test_json_path = _config.transient.package_dir / "configs" / "test_json_config.json"
```
jmartin-tech: Test that create a file in the `_config.transient.package_dir` should likely use a fixture to override that location. Writing files into the actual `package_dir` during test is discouraged as it could lead to odd state if packaging after tests are run and a test were to silently fail cleanup of the file.

jmartin-tech (changes requested):
Testing shows the new exception raise here needs to be handled in `cli.py` to get the error trace into the logs and output a clear indicator to the user:

```
python -m garak -t test --config broad_.yaml --fix
garak LLM vulnerability scanner v0.13.4.pre1 ( https://github.com/NVIDIA/garak ) at 2026-01-06T10:53:41.806014
  File "/Users/jemartin/Projects/nvidia/garak/garak/__main__.py", line 14, in <module>
    main()
  File "/Users/jemartin/Projects/nvidia/garak/garak/__main__.py", line 9, in main
    cli.main(sys.argv[1:])
  File "/Users/jemartin/Projects/nvidia/garak/garak/cli.py", line 301, in main
    _config.load_config(run_config_filename=args.config)
  File "/Users/jemartin/Projects/nvidia/garak/garak/_config.py", line 350, in load_config
    raise FileNotFoundError(message)
FileNotFoundError: run config not found: broad_.yaml
```

Suggested change:
```
diff --git a/garak/cli.py b/garak/cli.py
index dbc03432..30445daa 100644
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -298,7 +298,12 @@ def main(arguments=None) -> None:
     # load site config before loading CLI config
     _cli_config_supplied = args.config is not None
     prior_user_agents = _config.get_http_lib_agents()
-    _config.load_config(run_config_filename=args.config)
+    try:
+        _config.load_config(run_config_filename=args.config)
+    except FileNotFoundError as e:
+        logging.exception(e)
+        print(f"❌{e}")
+        exit(1)

     # extract what was actually passed on CLI; use a masking argparser
     aux_parser = argparse.ArgumentParser(argument_default=argparse.SUPPRESS)
```

tests/test_config.py
```diff
@@ -95,6 +95,38 @@
...
+            except Exception:
+                pass
+        if _config.transient.hitlogfile is not None:
+            try:
+                _config.transient.hitlogfile.close()
+            except Exception:
+                pass
+        importlib.reload(_config)
+
+    request.addfinalizer(reload)
```
jmartin-tech: Minor nitpick I will deal with as this lands:

```suggestion
    def reload():
        # Close any open files before reloading to prevent ResourceWarnings
        if _config.transient.reportfile is not None:
            _config.transient.reportfile.close()
            if os.path.exists(_config.transient.report_filename):
                os.remove(_config.transient.report_filename)
        importlib.reload(_config)
        reload()
```

PR #1496, opened 2025-11-21
enable consistent PRNG access in generators

jmartin-tech:
Allow generators in each process to create `repeatable` pseudo random values. This allows users more repeatability when generators use random values for request parameters. PRNG access across various processes is not `synchronized` meaning this will not produce 100% predictable random values for all cases. The primary target is to isolate randomness during generation for an individual attempt when `parallel_attempts` is enabled.

Note if `parallel_requests` is used the instance PRNG object is not reinitialized for each request.

* seed instance PRNG during generations loop
* update all generator `random` access to used instance PRNG

## Verification

- [ ] `garak -t nim -n <model_name>`
- [ ] **Verify** across multiple runs requests for each attempt provides consistent `seed` and `temperature` values for repeated generations of `attempt.prompt`. 

### DIFF
--- a/garak/generators/base.py
+++ b/garak/generators/base.py
@@ -150,6 +150,9 @@ class Generator(Configurable):
                 prompt, Conversation
             ), "generate() must take a Conversation object"
 
+        if self.seed is not None:
+            self._rng.seed(self.seed)
+
         self._pre_generate_hook()
 
         assert (
@@ -213,8 +216,6 @@ class Generator(Configurable):
                     colour=f"#{garak.resources.theme.GENERATOR_RGB}",
                 )
                 generation_iterator.set_description(self.fullname[:55])
-                if self.seed is not None:
-                    self._rng.seed(self.seed)
                 for i in generation_iterator:
                     output_one = self._call_model(
                         prompt, 1


garak/generators/base.py
```diff
@@ -208,6 +213,8 @@ def generate(
                     colour=f"#{garak.resources.theme.GENERATOR_RGB}",
                 )
                 generation_iterator.set_description(self.fullname[:55])
+                if self.seed is not None:
+                    self._rng.seed(self.seed)
```
jmartin-tech: In theory this could be moved up to before the call to `self._pre_generate_hook()` encase some generator were to use that method when setting pseudo-random values, thoughts?

leondz: I think that's a better place for it, yes. Though it does mean that calls to `self._pre_generate_hook()` will have to call `super()`. Another option is a new `_setup()` method and documentation for it.

jmartin-tech: Sorry comment was not explicit enough, idea is to place the initialization early in `generate()` to ensure it is preformed before the call to `self._pre_generate_hook()` is made.  That placement should not add any new requirements for implementers of `self._pre_generate_hook()`.

leondz: Even better!

garak/generators/nim.py
```diff
@@ -76,10 +75,10 @@ def _call_model(
         ), "generations_per_call / n > 1 is not supported"
 
         if self.vary_seed_each_call:
-            self.seed = random.randint(0, 65535)
+            self.seed = self._rng.randint(0, 65535)
 
         if self.vary_temp_each_call:
-            self.temperature = random.random()
+            self.temperature = self._rng.random()
```
erickgalinkin, line 81: Caveat: some OpenAI-spec models do not accept a `temperature` parameter. It may be worth being careful here or making sure we handle that particular error gracefully.

jmartin-tech: The `suppressed_params` option in `DEFAULT_PARAMS` for `OpenAICompatible` is designed to address this.
https://github.com/NVIDIA/garak/blob/fc829ff19ea60ce2c36b5160d079f25e92098dab/garak/generators/openai.py#L145

jmartin-tech:
> Is any testing meaningful with this? The lipsum generator uses random but doesn't expose .seed(), so perhaps not. But tests would be good

I am still thinking on how testing might look for this, looking specifically for a call to seed the object seemed a little too direct of a test. Short term I would like to move this forward to enable consumers.

leondz (approved):

PR #1497, opened 2025-11-26
allow for `apikey` case where `group=None`

boltzmann-brain:
Fixes a bug in new `apikey` detector where if a `None` condition causes the entire run to fail.

## Verification

Do some runs with the `apikeys` probe. Without the `None` condition, some models (e.g. GPT-oss-120b,  Llama-3.3-70B-Instruct, claude-sonnet-4-5) will almost always produce responses that cause the run to fail

After the fix, the runs stop failing at this particular probe.

### DIFF
--- a/garak/detectors/apikey.py
+++ b/garak/detectors/apikey.py
@@ -31,7 +31,7 @@ class ApiKey(Detector):
                 for pattern in regex_dict.values():
                     for match in pattern.finditer(output.text):
                         for group in match.groups():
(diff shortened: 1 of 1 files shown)


leondz:
Makes sense - we should add a test case for this, perhaps in a fast follow.

Please sign the DCO and we can integrate!

boltzmann-brain:

boltzmann-brain:
recheck

PR #1506, opened 2025-12-01
doc: rm links to doc source; show links to github code source

leondz:
see title

example in practice:

image

("Source" here links to github / main branch)


leondz:
> Yeah, copying the source of every file is kind of terrible.

This PR is more of a UX change - the default "view source" link pointed to the source of the RST-fornat documentation, and the change makes it point to the Python source code

leondz:
It's a little smart - it even uses the most recent hash of the file being tested, rather than head, is what I believe was observed in validation

PR #1510, opened 2025-12-02
docs: refine markdown checks

leondz:
markdown in backticks in docstrings  is actually OK


erickgalinkin (approved):
Listen: Good Tings, Man.

Issue #1516, opened 2025-12-05
housekeeping: find lighter `base2048` dependency


leondz:
pip package `base2048` depends on cargo/rustup:

```
# pip install -e /lustre/fsw/portfolios/coreai/users/zijiac/codebases/garak-v0.13.2          
Obtaining file:///lustre/fsw/portfolios/coreai/users/zijiac/codebases/garak-v0.13.2
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Collecting base2048>=0.1.3 (from garak==0.13.2)
  Using cached base2048-0.1.3.tar.gz (91 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
      
      Cargo, the Rust package manager, is not installed or is not on PATH.
      This package requires Rust and Cargo to compile extensions. Install it through
      the system's package manager or via https://rustup.rs/
      
      Checking for Rust toolchain....
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> base2048

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

This may be inconvenient. Consider another python package. https://pypi.org/project/mikeshardmind-base2048/ is an immediate candidate.

Switching should be validated by generating a number of encoding tests, using edge cases and existing payload string, using the existing `base2048` package - before trying a replacement package. The replacement package should perform equally well, should not depend on rustup/cargo, and should not have heavier/extra external dependencies.

onceandtwice:
Hi, first time contributor here! I’m new to the repo, but I’d be happy to take this on if no one else is working on it.

leondz:
Hi, welcome! That sounds great. The important thing with this issue is good testing - we have to be sure the replacement is resilient and also that it implements the encoding the same way. The build/install must also work without Rust (a language we know and love but that doesn't have as strong support on `arm` systems). Good luck!

Carlos-Projects:
This issue was resolved by commit 795a5d9e which replaced the Rust-dependent `base2048` package with `mikeshardmind-base2048` (pure Python).

[...]

PR #1518, opened 2025-12-08
Replace base2048 with mikeshardmind-base2048 and add spec tests

onceandtwice:
Replace the current Rust-based base2048 dependency with a Python alternative (mikeshardmind-base2048) to drop the Cargo/Rust requirement while also keeping functionality identical. Comprehensive Base2048 tests with reference vectors generated from the original Rust package to prove encoding/decoding compatibility and probe integration have also been added.

The following have been changed:
Dependencies: base2048 → mikeshardmind-base2048>=1.0.2 in pyproject.toml and requirements.txt.
Tests: tests/probes/test_base2048.py covers encoding/decoding parity with the original Base2048 outputs (reference vectors), roundtrip for all bytes/lengths, realistic payloads (ASCII/Unicode), API contract, pure-Python check, and InjectBase2048 probe integration. Includes a Docker command to regenerate the reference vectors used in comparison.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,7 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
-  "base2048>=0.1.3",
+  "mikeshardmind-base2048>=1.0.2",
   "transformers>=4.51.3,<4.57.0",
   "datasets>=3.0.0,<4.0",
   "colorama>=0.4.3",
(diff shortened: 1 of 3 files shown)


onceandtwice:

onceandtwice:
recheck

leondz:
Thank you - will review and validate

erickgalinkin:
Minor, but we should verify license compatibility: https://github.com/mikeshardmind/python-base2048/blob/main/LICENSE

jmartin-tech:
Fixes #1516

Linking to issue for visibility.

PR #1523, opened 2025-12-10
reporting: script to check report quality

leondz:
Conduct a variety of checks and tests to assess the integrity of a garak `report.jsonl` file

This helps us identify where a report may be broken, deficient, or incorrectly assembled

Inventory of tests:

* ✔️ garak version match between that used to create report and current garak used for checking
* ✔️ report using a dev version of garak
* ✔️ check using a dev version of garak
* ✔️ inventory described by config's `probe_spec` matches probes present in attempts
* ✔️ each attempt status 1 has matching status 2
* ✔️ all attempts have enough unique generations
* ✔️ run ID is in setup run IDs
* ✔️ detection output has correct cardinality in attempt status 2s
* ✔️ a summary digest object is present
* ✔️ at least one z-score is listed in the digest
* ✔️ probes present in summary matches probes requested in config
* ✔️ the run was completed
* ✔️ the run is <6 months old (calibration freshness)
* ✔️ there is at least one eval statement for any probe attempted
* ✔️ evals are performed over all status 2 attempts
* ✔️ number of responses graded `passed` + `nones` is not more than `total` reponses graded in `eval` entries


garak/analyze/check_report_integrity.py
```diff
@@ -0,0 +1,373 @@
...
+        print("🔹", note)
+    except BrokenPipeError:
+        pass
+
+
+def _is_dev_version(version: str) -> bool:
+    return version.split(".")[-1].startswith("pre")
+
+
+def _compare_sets(set1: Set, set2: Set, set1_name: str, set2_name) -> None:
```
erickgalinkin:
```suggestion
def _compare_sets(set1: Set, set2: Set, set1_name: str, set2_name: str) -> None:
```

nit, but it bothers me that exactly one argument doesn't have a type.

garak/analyze/check_report_integrity.py
```diff
@@ -0,0 +1,373 @@
...
+            except json.decoder.JSONDecodeError as jde:
+                add_note(f"invalid json entry starting '{line[:100]}' : " + repr(jde))
+                continue
+            match r["entry_type"]:
+                case "start_run setup":
+                    report_garak_version = r["_config.version"]
+                    if _is_dev_version(garak_version):
+                        add_note(
+                            f"report generated under development garak version {garak_version}, implementation will depend on branch+commit"
+                        )
```
erickgalinkin: Not me triggering this constantly.

garak/analyze/check_report_integrity.py
```diff
@@ -0,0 +1,373 @@
...
+                    if completion_id not in setup_run_ids:
+                        add_note(
+                            "completion run uuid not in setup run uuid(s), did aggregation go wrong?"
+                        )
+
+                case "digest":
+                    digest_exists = True
+                    if r["meta"]["garak_version"] != report_garak_version:
+                        add_note(
+                            f"digest was written with a different garak version ({r["meta"]["garak_version"]}) from the run ({report_garak_version})"
```
erickgalinkin:
```suggestion
                            f'digest was written with a different garak version ({r["meta"]["garak_version"]}) from the run ({report_garak_version})'
```

Can't double quote like this dawg. :(

garak/analyze/check_report_integrity.py
```diff
@@ -0,0 +1,373 @@
...
+                r = json.loads(line)
+            except json.decoder.JSONDecodeError as jde:
+                add_note(f"invalid json entry starting '{line[:100]}' : " + repr(jde))
+                continue
+            match r["entry_type"]:
+                case "start_run setup":
+                    report_garak_version = r["_config.version"]
+                    if _is_dev_version(garak_version):
+                        add_note(
+                            f"report generated under development garak version {garak_version}, implementation will depend on branch+commit"
```
erickgalinkin:
```suggestion
                            f"report generated under development garak version {report_garak_version}, implementation will depend on branch+commit"
```

This currently references the `garak_version`.

garak/analyze/check_report_integrity.py
```diff
@@ -0,0 +1,373 @@
...
+        for line in [line.strip() for line in reportfile if line.strip()]:
+            try:
+                r = json.loads(line)
+            except json.decoder.JSONDecodeError as jde:
+                add_note(f"invalid json entry starting '{line[:100]}' : " + repr(jde))
+                continue
+            match r["entry_type"]:
+                case "start_run setup":
+                    report_garak_version = r["_config.version"]
+                    if _is_dev_version(garak_version):
```
erickgalinkin:
```suggestion
                    if _is_dev_version(report_garak_version):
```

erickgalinkin (changes requested):
Mostly looks good. A couple of fixes needed to make it actually run and a few value fixes.

leondz:
NB. add documentation to match pattern in #1569

garak/analyze/check_report_integrity.py
leondz: works on precisely the one single laptop and environment i've tested it in but maybe, maybe it can be more defensive

garak/analyze/check_report_integrity.py
```diff
@@ -0,0 +1,377 @@
...
+                            add_note(
+                                f"Eval entry total_processed {total_attempts_processed} doesn't match sum of evaluated {total_attempts_evaluated} and nones {r['nones']} for {_probename}/{r['detector']}"
+                            )
+
+                        if total_attempts_evaluated > total_attempts_processed:
+                            add_note(
+                                f"Eval entry total_evaluated {total_attempts_evaluated} mustn't be greater than total_processed {total_attempts_processed} for {_probename}/{r['detector']}"
+                            )
+
+                        pfn = [r[""], r[""], r[""]]
```
patriciapampanelli: Did you mean r["passed"], r["fails"], and r["nones"] for the negative-value check?

```suggestion
                        pfn = [r["passed"], r["fails"], r["nones"]]
```

garak/analyze/check_report_integrity.py
```diff
@@ -0,0 +1,377 @@
...
+                        probes_found_in_evals.add(_probename)
+                        total_attempts_processed = r["total_processed"]
+                        total_attempts_evaluated = r["total_evaluated"]
+                        if (
+                            total_attempts_processed
+                            != attempt_status_2_per_probe[_probename]
+                            * generations_requested
+                        ):
+                            add_note(
+                                f"Eval entry for {_probename} {_detectorname} indicates {r['total']} instances but there were {attempt_status_2_per_probe[_probename]} status:2 attempts (generations={generations_requested})"
```
patriciapampanelli: `r` has no `"total"` key (eval entries use `total_processed` / `total_evaluated`), so `r['total']` would raise `KeyError`. Suggest using `total_attempts_processed` in the message instead.

```suggestion
                                f"Eval entry for {_probename} {_detectorname} indicates {total_attempts_processed} instances but there were {attempt_status_2_per_probe[_probename]} status:2 attempts (generations={generations_requested})"
```

garak/analyze/check_report_quality.py
```diff
@@ -0,0 +1,377 @@
...
+""" Conduct a variety of checks and tests to assess the integrity of a garak report.jsonl file """
+
+"""
+inventory of tests:
+
+* ✔️ version mismatch between report and garak 
+* ✔️ report using dev version 
+* ✔️ current version is dev version
+* ✔️ probe_spec matches probes in attempts
+* ✔️ attempt status 1 has matching status 2
```
jmartin-tech, line 13: This seems like a possible false positive item, there are probes that emit status 1 entires with multiple status 2 entries, `TreeSearchProbe` and possibly `IterativeProbe` may or may not present in ways that match inferred expectations of this check.

garak/analyze/check_report_quality.py
```diff
@@ -0,0 +1,377 @@
...
+* ✔️ attempts have enough unique generations
+* ✔️ attempt run ID in setup run IDs
+* ✔️ detection has correct cardinality in attempt status 2s
+* ✔️ summary object is present
+* ✔️ at least one z-score is listed
```
jmartin-tech, line 18: Calibration data may not exist for any active probe in the run or `system.show_z` may be set to `False`, this may need to evaluate if the configuration data expects it to be true.

garak/analyze/check_report_quality.py
```diff
@@ -0,0 +1,377 @@
...
+* ✔️ summary matches probes requested
+* ✔️ run was completed
+* ✔️ run is <6 months old (calibration freshness)
```
jmartin-tech, line 21: Is this a valid criteria about report _integrity_? I would think of this as impacting a report's applicability in a context or usability in comparison to state of the art which is not _integrity_ of the report itself.

garak/analyze/check_report_quality.py
```diff
@@ -0,0 +1,377 @@
...
+                            "probes requested in config",
+                            "probes listed in digest",
+                        )
+                    if probes_in_digest != probes_found_in_evals:
+                        _compare_sets(
+                            probes_found_in_evals,
+                            probes_in_digest,
+                            "probes evaluated",
+                            "probes listed in digest",
+                        )
```
jmartin-tech, line 340: Do skipped probes leave entries for `eval` or `digest`? If not then there could be some divergence here that is _expected_. I think printing that details is helpful signal, just wonder if there is some indicator needed to avoid confusion about _integrity_ of the report vs _usability_ of the report for a specific use case.

jmartin-tech (commented):
Reviewing the code and the list of things checked has raised questions for me about the goal of this utility.  I suspect naming is the root cause. When I think of _report integrity_ I think of the internal integrity unto itself, validating the information contained is full and complete unto internally noted criteria. This aligns with some of the noted checks, such as validating all requested in initial config have attempts, eval, summary data and ensuring the counts involved present internal consistency with evidence that backs them.

However, this script seems to be looking for _more_ than internal consistency, it looks to be expressing something I akin more it _usability in context_ or _suitability for a purpose_, such as considering the result _current enough to be testing state of the art_, by raising concerns for things like `age` and `version` and `dev version`.

Can we find a name that is more clear?

garak/analyze/check_report_integrity.py
leondz: good catch, thank you

garak/analyze/check_report_quality.py
leondz: two situations are covered here:

a. asserting that there's >0 status 2 attempt for every status 1 attempt
b. asserting that there's only 1 status 2 attempt for every status 1 attempt

Scenario (a) should not be failed - that means a broken run

We should go for (b) as well, I think. What patterns are there for going generation but no detection? Can we afford these a new, distinct status?

garak/analyze/check_report_quality.py
leondz: Check messages are advisory, rather than a guaranteed failure (see e.g. the dev version or old report messages)

Maybe we can prioritise these by severity, within the checker tool?

garak/analyze/check_report_quality.py
leondz: It's something to keep an eye on during this process, because result quality is likely to be reduced on older reports

garak/analyze/check_report_quality.py
leondz: Good question. Skipped probes may leave eval entries. If not, something went wrong, and that's worth highlighting.

I'm thinking maybe rename the script - maybe away from `_integrity` possibly towards `_quality`, maybe away from `_report_` and towards `_run_integrity`.

garak/analyze/check_report_quality.py
jmartin-tech: I think both have the possibility of not being true currently for a valid run:

(a) may not apply to all iterative probe techniques depending on how the states are managed as some inference calls may be part of the build up of the technique and intentional logging of all calls is part of ensuring a proper audit trail. Iterative termination logic _could_ cause some attempts to intentionally not be processed by the final detection phase, though I am not sure it _should_

(b) I suspect this already is not true for `TreeSearchProbe`. Those probes will run detection at each fork that uses a different detector than the probe's evaluation detector. A similar pattern is expected in many techniques I suspect will implement `IterativeProbe` as the iteration process often needs to perform a detector evaluation of some sort to determine if iteration should continue.

Your suggestion of another state is a reasonable thing to explore and enable making these criteria targets once the state is defined and implemented, they just aren't currently something reports will conform to today.

leondz: Indeed - this tool reveals a bug in `TreeSearchProbe` as far as I'm concerned. Attempt with complete detection (status 1) shouldn't appear more than once for a given target result (status 1).

garak/analyze/check_report_quality.py
leondz: have split notes into high- and low-priority; high is for cases where integrity/validity of the report is compromised

garak/analyze/check_report_quality.py
leondz: this message now presents as low-priority

garak/analyze/check_report_quality.py
leondz: renamed `check_report_quality`

garak/analyze/check_report_quality.py
```diff
@@ -0,0 +1,394 @@
...
+                    _z_score_floats = filter(
+                        lambda f: isinstance(f, float), _z_score_values_found
+                    )
+                    if not len(list(_z_score_floats)):
+                        add_note(
+                            "No Z-scores/relative scores found. Maybe deliberate, maybe calibration broken",
+                            high_priority=False,
+                        )
+
+                    probes_in_digest.remove("_summary")
```
jmartin-tech: Testing shows some older reports will not have this entry causing a stack trace here:
```
  File "/Projects/nvidia/garak/garak/analyze/check_report_quality.py", line 394, in <module>
    main()
  File "/Projects/nvidia/garak/garak/analyze/check_report_quality.py", line 314, in main
    probes_in_digest.remove("_summary")
KeyError: '_summary'
```

jmartin-tech (changes requested):

garak/analyze/check_report_quality.py
leondz: changed to use `.discard()`

garak/analyze/check_report_quality.py
```diff
@@ -0,0 +1,394 @@
...
+                        )
+                    if report_garak_version != garak_version:
+                        add_note(
+                            f"Current and report garak version mismatch, {garak_version} vs. {report_garak_version}",
+                            high_priority=False,
+                        )
+                    configured_probe_spec = r["plugins.probe_spec"]
+                    probes_requested, __rejected = garak._config.parse_plugin_spec(
+                        configured_probe_spec, "probes"
+                    )
```
jmartin-tech, line 145: I suspect this is going to be problematic, parsing an older spec with the current codebase will often see divergence.  Consider a probe that has been renamed or even deprecated and removed.

Testing shows based on this restriction the output will generate a _high priority_ entry for each attempt from the renamed or removed probe class.

This also presents a divergence from expectations when a new probe is added or activated for an existing module between versions, the spec parsing will expect the newly active probe to be in found in the report.  

This indicates to me that the report may need to have an expanded _activated probes list_ stored in the report during the run to provide a way to validate this value consistently using just the report jsonl.

garak/analyze/check_report_quality.py
```diff
@@ -0,0 +1,394 @@
...
+                        total_attempts_processed = r["total_processed"]
+                        total_attempts_evaluated = r["total_evaluated"]
+                        if (
+                            total_attempts_processed
+                            != attempt_status_2_per_probe[_probename]
+                            * generations_requested
+                        ):
+                            add_note(
+                                f"Eval entry for {_probename} {_detectorname} indicates {total_attempts_processed} instances but there were {attempt_status_2_per_probe[_probename]} status:2 attempts (generations={generations_requested})"
+                            )
```
jmartin-tech: This presents a consistency issue, there are known probes that will not fit this pattern, `atkgen.Tox` and `topic.WordnetControversial` are clear examples identified in testing.

leondz: agree. made this low prio.

leondz: -- splitting this into conditions of "too many" and "too few". also will re-check the `* generations` logic, it seems odd that generations would directly change the proportion of status-2 to status-1

leondz: resolution:

* if we process fewer responses than expected, this is high priority
* if we process more responses than expected, this is low priority (cf current probes mentioned above)
* if we evaluate more responses than are generated, this is high priority

garak/analyze/check_report_quality.py
leondz: this is an instance of a broader, longer-term issue in garak - current config vs. config described in log - propose tracking in separate issue

PR #1524, opened 2025-12-11
feature: `OpenAICompatible` classes does one generation at a time by default

leondz:
Full support for OpenAI API is patchy.

Maybe targets that support OpenAI API access in general can only generate one result at a time. This can happen silently when multiple results are requested.

This PR changes the default of `OpenAICompatible` to be to not support multiple concurrent generations, while updating `OpenAIGenerator` to still support multiple generations.

### DIFF
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -222,8 +222,10 @@ class OpenAICompatible(Generator):
         is_completion = generator == client.completions
 
         create_args = {}
-        if "n" not in self.suppressed_params:
+        if self.supports_multiple_generations:
             create_args["n"] = generations_this_call
+        elif "n" not in self.suppressed_params:
+            create_args["n"] = 1
         for arg in inspect.signature(generator.create).parameters:
             if arg == "model":
                 create_args[arg] = self.name


leondz:
Since the OpenAI API is used so widely, I prefer not making assumptions in `OpenAICompatible` (even this PR is about reducing assumptions), so I'm not immediately keen on expanding `suppressed_params` either by default or dynamically.

On a related note - we're not checking if the number of generations requested in garak config matches the number of outputs returned overall by the generator. This *may* be happening with some endpoints when `supports_multiple_generations` is false. I suspect that's for another issue/PR.

garak/generators/openai.py
jmartin-tech, line 372: Since `n` is already handed special via `generations_this_call` we can just base suppression on `self.supports_multiple_generations`:


```suggestion
        if self.supports_multiple_generations:
```

jmartin-tech:
Since `n` is already handed special via `generations_this_call` we can just base suppression on `self.supports_multiple_generations`:

``` Python
        if self.supports_multiple_generations:
            create_args["n"] = generations_this_call
        elif "n" not in self.suppressed_params:
            create_args["n"] = 1
```

In practice, we really should not allow `n` to be set via "generator config parameters" for OpenAICompatible generators or really for any generator.

PR #1526, opened 2025-12-11
mistral generator workaround for magistral models. openai-style gener…

boltzmann-brain:
…ator bugfix for certain types of None responses.

This has two fixes: the first for mistral and the second (very minor) one for OpenAI-style generators.

The first fix works around a [bug](https://github.com/spring-projects/spring-ai/issues/5019) in the `mistralai` package which doesn't account for thinking tokens in Magistral models. If the response fails the pydantic validation in `mistralai`, we will fall back to using raw HTTP requests.

We can get rid of the HTTP workaround once mistral fixes that bug.

Possibly related bug: https://github.com/mistralai/client-python/issues/252

The second, OpenAI fix is a minor one that handles certain `None` responses better so that the run doesn't just fail. I encountered such `None` responses occasionally while querying `gemini-2.5.flash`.

## Verification

Try to start a run on `magistral-medium-2509` with the previous version of the generator. It doesn't really matter which probe. At some point it will fail with a Pydantic validation error, because the structure of the response doesn't fit `mistralai`'s pydantic model.

With this version of the generator, the Pydantic validation error will be written to the log, but garak will retry with a raw HTTP request and the run will continue.

### DIFF
--- a/garak/generators/mistral.py
+++ b/garak/generators/mistral.py
@@ -1,4 +1,7 @@
 import backoff
+import json
+import logging
+import requests
 from typing import List
 from mistralai import Mistral, models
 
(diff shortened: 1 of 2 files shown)


garak/generators/mistral.py
```diff
@@ -20,6 +21,45 @@ class MistralGenerator(Generator):
     ENV_VAR = "MISTRAL_API_KEY"
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "name": "mistral-large-latest",
+        # Sampling parameters
+        "top_p": None,                    # Nucleus sampling (0.0-1.0)
+        "frequency_penalty": None,        # Penalize word repetition by frequency (0.0-2.0)
+        "presence_penalty": None,         # Penalize word repetition (0.0-2.0)
+        # Generation control
+        "n": None,                        # Number of completions to return
+        "stop": None,                     # Stop sequences (string or list)
```
jmartin-tech: Prefer consistency with other `stop` params and require a `list[str]`:
```suggestion
        "stop": [],                     # Stop sequences (list[str])
```

garak/generators/mistral.py
```diff
@@ -20,6 +21,45 @@ class MistralGenerator(Generator):
...
+        "response_format": None,          # Output format (text, json_object, json_schema)
+        # Tool/function calling
+        "tools": None,                    # List of available tools/functions
+        "tool_choice": None,              # Tool calling control (none, auto, any, required)
+        "parallel_tool_calls": None,      # Enable parallel function calling
+        # Safety and prompting
+        "safe_prompt": None,              # Inject safety prompt
+        "prompt_mode": None,              # Toggle reasoning mode
+        # Optimization
+        "prediction": None,               # Expected completion for optimization
```
jmartin-tech: `garak` does not currently expect generator to support _tools_, _safety prompts_, etc. avoid offering `DEFAULT_PARAMS` for items the that do not have a consistent abstraction at this time.


```suggestion
```

garak/generators/mistral.py
```diff
@@ -47,11 +87,170 @@ def _call_model(
...
     ) -> List[Message | None]:
         messages = self._conversation_to_list(prompt)
-        chat_response = self.client.chat.complete(
-            model=self.name,
-            messages=messages,
-        )
-        return [Message(chat_response.choices[0].message.content)]
+        
+        try:
+            logging.debug(f"Making Mistral API call with model: {self.name}")
```
jmartin-tech: Logging at start of every inference call is excessive.

```suggestion
```

garak/generators/mistral.py
```diff
@@ -47,11 +87,170 @@ def _call_model(
...
+                            logging.debug(f"Filtered out thinking chunk: {chunk.get('thinking', '')[:100]}...")
+                        else:
+                            logging.warning(f"Unknown chunk type: {chunk_type}")
+                content = ''.join(text_parts)
+                logging.debug(f"Final extracted content length: {len(content)}")
+            
+            return [Message(content)]
+            
+        except Exception as e:
+            # If we get a Pydantic validation error, try to make a raw HTTP request
```
jmartin-tech: This is not really something a generator utilizing a `client` package should do.  If the user is expecting to test a target in violation of the client provided spec then the `RestGenerator` should be used.

jmartin-tech:
Declining at this time, the `openai` issue was addressed separately and the workaround for a service side mistral issue is not something the project would like to invest maintainer time on.

Issue #141, opened 2023-06-19
support continuation / resume of runs


leondz:
* --continue switch: take json from report, process attempts, continue from where we stopped. requires robust config storage

leondz:
issue #833 contributed to this

fcanogab:
I've contributed an initial proposal to implement this functionality in https://github.com/NVIDIA/garak/pull/1107.
It's my first contribution to Garak and probably I'm missing important bits, so I'd appreciate your feedback to improve the PR.

EricXQiu:
If we can resume runs this is great. Looking forward to this feature.

PR #1531, opened 2025-12-15
feat: Add resume scan functionality to continue interrupted scans

Abhiraj-GetGarak:
Closes [#141 ](https://github.com/NVIDIA/garak/issues/141)

This adds the ability to resume a garak scan from where it left off by parsing an existing JSONL report file and skipping already-completed attempts.

Features:
- New --resume CLI argument to specify the report file to resume from
- load_checkpoint() function to parse JSONL and identify completed attempts
- Resume mode fields in _config.transient for tracking state
- Attempt.from_dict() class method to reconstruct attempts from report data
- Probe skips already-completed attempts (status=2) in resume mode
- Support for pending detection attempts (status=1)

Usage:
  python -m garak --resume path/to/existing.report.jsonl [other options]

The resume feature will:
1. Parse the existing report file
2. Identify completed attempts (status=2) and skip them
3. Identify pending detection attempts (status=1) for re-evaluation
4. Append new results to the same report file

the github issue number.

## Verification

- [ ] Start a scan with multiple probes that will take some time:
```bash
python -m garak --target_type openai --target_name gpt-4 --probes dan.DUDE,dan.STAN,leakreplay.PotterComplete --report_prefix /path/to/test_resume
```
- [ ] Interrupt the scan mid-way (Ctrl+C or kill process) after some attempts have completed
- [ ] Verify the report file exists with partial results:
```bash
wc -l /path/to/test_resume.report.jsonl
# Should show some entries with status=1 (pending detection) and/or status=2 (completed)
```
- [ ] Resume the scan using the `--resume` flag:
```bash
python -m garak --target_type openai --target_name gpt-4 --probes dan.DUDE,dan.STAN,leakreplay.PotterComplete --report_prefix /path/to/test_resume --resume /path/to/test_resume.report.jsonl
```
- [ ] **Verify** resume output shows checkpoint loaded:
```
🔄 Resuming scan from: /path/to/test_resume.report.jsonl
📊 Found X completed, Y pending detection
```
- [ ] **Verify** completed attempts are skipped (no duplicate LLM calls)
- [ ] **Verify** pending detection attempts only run detection (no LLM calls)
- [ ] **Verify** new attempts are executed normally
- [ ] **Verify** all attempts have status=2 after completion:
```bash
python -c "
import json
with open('/path/to/test_resume.report.jsonl') as f:
    attempts = [json.loads(l) for l in f if l.strip()]
    attempts = [a for a in attempts if a.get('entry_type') == 'attempt']
    status_1 = [a for a in attempts if a.get('status') == 1]
    status_2 = [a for a in attempts if a.get('status') == 2]
    print(f'Status=1: {len(status_1)}, Status=2: {len(status_2)}')
    # All status=1 should have corresponding status=2
"
```
- [ ] **Verify** HTML report generates correctly with all results
- [ ] **Verify** hitlog contains expected detection hits
- [ ] **Verify** the resume entry exists in the report:
```bash
grep '"entry_type": "resume"' /path/to/test_resume.report.jsonl
```

### DIFF
--- a/garak/command.py
+++ b/garak/command.py
@@ -50,55 +50,8 @@ def start_run():
 
     logging.info("run started at %s", _config.transient.starttime_iso)
 
-    # Check if we're in resume mode
-    is_resume_mode = resumeservice.enabled()
-
-    if is_resume_mode:
-        # Resume mode: use existing report file
-        # Note: Resume service is loaded by harness initialization, which handles
-        # parsing the checkpoint and displaying status messages
-        logging.info(f"Resuming scan from: {_config.transient.resume_file}")
-
-        # Use the same report file (append mode)
-        _config.transient.report_filename = _config.transient.resume_file
-
-        # Extract run_id from resume filename (format: garak.{run_id}.report.jsonl)
-        resume_basename = os.path.basename(_config.transient.resume_file)
-        if resume_basename.startswith("garak.") and resume_basename.endswith(
-            ".report.jsonl"
-        ):
-            _config.transient.run_id = resume_basename[
-                6:-13
-            ]  # Extract UUID from filename
-        else:
-            _config.transient.run_id = str(uuid.uuid4())
-
-        # Open file in append mode
-        _config.transient.reportfile = open(
-            _config.transient.report_filename, "a", buffering=1, encoding="utf-8"
-        )
-
-        # Write resume marker to report
-        completed_attempts = resumeservice.get_completed_attempts()
-        pending_detection = resumeservice.get_pending_detection_attempts()
-        total_completed = sum(len(seqs) for seqs in completed_attempts.values())
-        total_pending = sum(len(seqs) for seqs in pending_detection.values())
-
-        _config.transient.reportfile.write(
-            json.dumps(
-                {
-                    "entry_type": "resume",
-                    "resume_time": _config.transient.starttime_iso,
-                    "completed_attempts_count": total_completed,
-                    "pending_detection_count": total_pending,
-                    "probes_with_progress": list(
-                        set(completed_attempts.keys()) | set(pending_detection.keys())
-                    ),
-                },
-                ensure_ascii=False,
-            )
-            + "\n"
-        )
+    # Check if we're in resume mode - let the service handle all setup
+    if resumeservice.setup_run():
         return
 
     # Normal mode (non-resume)
--- a/garak/resumeservice.py
+++ b/garak/resumeservice.py
@@ -10,6 +10,8 @@ allowing scans to be resumed from where they left off.
 import hashlib
 import json
 import logging
+import os
+import uuid
 from collections import defaultdict
 from typing import Dict, Optional, Set, Tuple
 
@@ -364,6 +366,68 @@ def reset() -> None:
     """Reset resume state (primarily for testing)."""
     global _resume_state
     _resume_state = None
+
+
+def setup_run() -> bool:
+    """Set up run for resume mode.
+
+    This function handles all resume mode run setup responsibilities:
+    - Sets the report filename to the resume file
+    - Extracts run_id from the resume filename
+    - Opens the report file in append mode
+    - Writes a resume marker entry
+
+    Returns:
+        True if resume mode setup was performed, False if not in resume mode.
+    """
+    if not enabled():
+        return False
+
+    # Load checkpoint data
+    load()
+
+    logging.info(f"Resuming scan from: {_config.transient.resume_file}")
+
+    # Use the same report file (append mode)
+    _config.transient.report_filename = _config.transient.resume_file
+
+    # Extract run_id from resume filename (format: garak.{run_id}.report.jsonl)
+    resume_basename = os.path.basename(_config.transient.resume_file)
+    if resume_basename.startswith("garak.") and resume_basename.endswith(
+        ".report.jsonl"
+    ):
+        _config.transient.run_id = resume_basename[6:-13]  # Extract UUID from filename
+    else:
+        _config.transient.run_id = str(uuid.uuid4())
+
+    # Open file in append mode
+    _config.transient.reportfile = open(
+        _config.transient.report_filename, "a", buffering=1, encoding="utf-8"
+    )
+
+    # Write resume marker to report
+    completed_attempts = get_completed_attempts()
+    pending_detection = get_pending_detection_attempts()
+    total_completed = sum(len(seqs) for seqs in completed_attempts.values())
+    total_pending = sum(len(seqs) for seqs in pending_detection.values())
+
+    _config.transient.reportfile.write(
+        json.dumps(
+            {
+                "entry_type": "resume",
+                "resume_time": _config.transient.starttime_iso,
+                "completed_attempts_count": total_completed,
+                "pending_detection_count": total_pending,
+                "probes_with_progress": list(
+                    set(completed_attempts.keys()) | set(pending_detection.keys())
+                ),
+            },
+            ensure_ascii=False,
+        )
+        + "\n"
+    )
+
+    return True
 
 
 # Prompt-based matching functions


garak/_config.py
```diff
@@ -72,6 +72,15 @@ class TransientConfig(GarakSubConfig):
...
 
+    # Resume mode fields
+    resume_file = None  # Path to JSONL report to resume from
+    completed_attempts = (
+        None  # Dict of {probe_classname: set(seq)} for completed attempts
+    )
+    pending_detection_attempts = (
+        None  # Dict of {probe_classname: {seq: attempt_data}} for status=1 attempts
+    )
+
```
jmartin-tech: `_config` is not meant as a storage location, and is not multiprocess safe, this is not the correct place to share these entries.

garak/attempt.py
```diff
@@ -435,3 +435,51 @@ def _add_turn(self, role: str, contents: List[Union[Message, str]]) -> None:
...
+
+        # Handle reverse_translation_outputs
+        reverse_outputs = data.get("reverse_translation_outputs", [])
+        for output_data in reverse_outputs:
+            if output_data:
+                attempt.reverse_translation_outputs.append(Message(**output_data))
+            else:
+                attempt.reverse_translation_outputs.append(None)
+
+        return attempt
```
jmartin-tech: A more straight forward version:
```
    @classmethod
    def from_dict(cls, dicti): -> Attempt:
        """Initializes an attempt object from dictionary"""
        attempt_obj = cls()
        attempt_obj.uuid = dicti["uuid"]
        attempt_obj.seq = dicti["seq"]
        attempt_obj.status = dicti["status"]
        attempt_obj.probe_classname = dicti["probe_classname"]
        attempt_obj.probe_params = dicti["probe_params"]
        attempt_obj.targets = dicti["targets"]
        attempt_obj.prompt = Conversation.from_dict(dicti["prompt"])
        attempt_obj.outputs = [Message(**msg) for msg in dicti["outputs"]]
        attempt_obj.detector_results = dicti["detector_results"]
        attempt_obj.notes = dicti["notes"]
        attempt_obj.goal = dicti["goal"]
        attempt_obj.conversations = [
            Conversation.from_dict(conv) for conv in dicti["conversations"]
        ]
        attempt_obj.reverse_translation_outputs = [
            Message(**msg) for msg in dicti.get("reverse_translation_outputs", [])
        ]
        return attempt_obj
```

Note that this does not attempt to set any values that are not found in the provided data as that would represent an invalid serialization and the `KeyError` that would be raised is an appropriate behavior.

tests/test_resume.py
```diff
@@ -0,0 +1,584 @@
...
+                    {
+                        "entry_type": "attempt",
+                        "status": 2,
+                        "probe_classname": "probes.other.OtherProbe",
+                        "seq": 0,
+                    }
+                )
+                + "\n"
+            )
+            temp_path = f.name
```
jmartin-tech: File data can simply come from a resource file, and be loaded from disk. There are examples of this in other report tests.  This applies to other tests here as well.

tests/test_resume.py
```diff
@@ -0,0 +1,584 @@
...
+        }
+
+        attempt = Attempt.from_dict(data)
+
+        assert attempt.status == ATTEMPT_STARTED
+        assert attempt.probe_classname == "probes.test.TestProbe"
+        assert attempt.seq == 0
+        assert attempt.targets == []
+        assert attempt.notes == {}
+        assert attempt.goal is None
```
jmartin-tech: These tests belong in `test_attempt.py`

garak/probes/base.py
```diff
@@ -414,7 +414,70 @@ def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
...
+            if seq in completed_seqs:
+                skipped_count += 1
+                logging.debug(f"Skipping completed attempt {self.probename}:seq={seq}")
+                continue
+
+            # Skip attempts that are pending detection (status=1) - handled separately
+            if seq in pending_detection_seqs:
+                logging.debug(f"Skipping pending detection attempt {self.probename}:seq={seq}")
+                continue
+
```
jmartin-tech: As noted `_config` is not a datastore and should not be used as one. One idea would be to create a service that can be queried for already processed attempts. That service would be responsible for processing the resume file and de-serialize `Attempt` objects as a `Probe` should not be responsible for this. Note there is an existing pattern already in place for loading _services_ during harness initialization in the `langservice`.

Another concern with the logic here is that the attempts not being repeated here are not guaranteed to be the same prompts.  There are a number for randomization processes involved in probe prompt population that would need to be accounted for via some validation.

The validation likely could be preformed at during the `load_checkpoint` method offered in this PR. It will need to ensure the following:
* the version in the resumed run is the same version as the current run
* seed value is consistent with the pervious run and had been `set`. A fully random seed would produce variance in the prompt sequence.

I would hazard to expand this guidance to say a service should be fully responsible for any context of `resume` and that `load_checkpoint` actually belongs in that service not in `command.py`.

When the service returns the set of previous attempts the `seq` may be consistent making then unusable to determine which prompts have already been sent for inference, however the prompt itself may be compared to a newly minted attempt prompt. This means when processing the queued minted attempts it would be possible to compare the now cached completed calls to skip calling inference and simply copy the outputs for the new run. Searching the set of previous attempts for a specific prompt would be computationally expensive however an order of magnitude or so less expensive than repeating the inference call. This would suggest a process like the following could work:
* get all previously attempts with completed inference from the new service
* mint attempts into `attempts_todo`
* compare and determine additional attempts needed based on completed attempt and attempt_todo prompts, the goal is to have the same number of complete attempts as len(attempt_todo)
* execute inference for the number of missing attempts only
* execute detection

The above skips some details but I think expresses that complexity of determining how many attempts need to be supplemented based on reusing the previous run inference results.

garak/command.py
```diff
@@ -30,6 +30,77 @@ def deprecation_notice(deprecated_item: str, version: str, logging=None):
...
+    # Log summary
+    total_completed = sum(len(seqs) for seqs in completed.values())
+    total_pending = sum(len(seqs) for seqs in pending_detection.values())
+    logging.info(
+        f"Loaded checkpoint: {total_completed} completed, {total_pending} pending detection across {len(completed) + len(pending_detection)} probes"
+    )
+
+    return dict(completed), dict(pending_detection), probe_spec
+
+
```
jmartin-tech: As noted this method should likely be inside a _service_ and loaded by the harness class.

garak/command.py
```diff
@@ -49,7 +120,68 @@ def start_run():
...
+                        set(completed_attempts.keys()) | set(pending_detection.keys())
+                    ),
+                },
+                ensure_ascii=False,
+            )
+            + "\n"
+        )
+        return
+
+    # Normal mode (non-resume)
```
jmartin-tech, line 57: This also should be a responsibility of a _service_ that is loaded by the harness.

garak/probes/base.py
jmartin-tech, line 1: Consider impacts on TreeSearch and Iterative Probes with expectation that resume may or may not apply. This might mean that a new metadata value is needed to reflect what probes are resumable.

garak/_config.py
Abhiraj-GetGarak: Addressed. Created `resumeservice.py` following the `langservice` pattern. Resume state is now stored in module-level `_resume_state` variable in the service, not in `_config.transient`. Only `resume_file `path remains in `_config.transient` as a configuration value.

garak/attempt.py
Abhiraj-GetGarak: Adopted your suggested implementation. The simplified version lets `KeyError` propagate for missing required fields, which is the appropriate behavior for invalid serialization.

tests/test_resume.py
Abhiraj-GetGarak: Addressed. Created test resource files in `tests/_assets/`:

```
1. resume_test_completed.report.jsonl
2. resume_test_pending.report.jsonl
3. resume_test_mixed.report.jsonl
```

Tests now load from these files instead of inline temp file creation.

tests/test_resume.py
Abhiraj-GetGarak: Moved. `TestAttemptFromDict` class is now in `tests/test_attempt.py`. 
`tests/test_resume.py `only contains integration tests for the CLI and end-to-end resume flow.

garak/probes/base.py
Abhiraj-GetGarak: Implemented. Created `resumeservice.py` with:

- Service interface matching `langservice` pattern
- `get_completed_by_prompt_hash()` for prompt-based matching using SHA256 hash
- Version validation (raises error on mismatch)
- Prompt-based matching allows resume to work even when prompts are reordered/filtered

The probe now:

1. Gets completed attempts indexed by prompt hash
2. Mints new attempts
3. Compares prompts via hash - copies outputs for matches
4. Executes inference only for unmatched prompts
5. Executes detection on all

garak/command.py
Abhiraj-GetGarak: Moved. `load_checkpoint()` logic is now in `resumeservice.load()`. The `command.py `only calls the service API.

garak/command.py
Abhiraj-GetGarak: Done. Moved the entire resume mode run setup to `resumeservice.setup_run()`. `command.py` now just calls:
```
if resumeservice.setup_run():
    return
```

Abhiraj-GetGarak:
✅ All addressed:

- Created `resumeservice.py` following `langservice `pattern
- Removed `_config` as storage (only path in transient)
- Version validation implemented (raises error on mismatch)
- Seed validation removed since prompt-based matching handles reordering
- `Attempt.from_dict()` simplified
- Tests reorganized with resource files

garak/probes/base.py
Abhiraj-GetGarak: Good point about TreeSearchProbe and IterativeProbe. After analysis:

- TreeSearchProbe: Dynamic tree traversal with inline detection - resume would need to checkpoint node queue, detector decisions, and tree expansion state. Significant complexity.
- IterativeProbe: Multi-turn conversations where each turn generates next attempts based on response. Resume would need per-turn/per-branch state tracking.

For this PR, I've added tests documenting that these probe types don't support resume (`TestProbeResumability` in `test_resumeservice.py`). Implementing resume for these probe types would require significant additional work and could be a follow-up PR. Would you like me to add a `supports_resume = False` class attribute to these probe types to make this explicit?

garak/command.py
jmartin-tech, line 1: Revisions in this file may need to be reconsidered, the service is `loaded` during `Harness._initialize_runtime_services()` and configuration setup is completed before cli.py calls `start_run()`. At which point the run should _start_ in the same way as any other run.

I suspect there is a usage pattern expectation here that may not align with my expectations as a user and may be worth some feedback gathering. 

When `garak` _resumes_ a run, I propose it should not modify the input file and should create a new output file that will be a complete self contained run result. This does present an edge case in that the output location needs to be `checked` to not be the same as input file for resume.

As written the service expects to _resume_ the run by continuing the existing file, however the current setup does not account for the `probe_spec` needing to change to account for already completed probes while still ensuring the configuration data does not need to be changed.

Another edge case that is not currently handled, an interrupted or failed run may have an incomplete last line in the jsonl due to process exit while the write is still in progress. This would mean, to produce the current _resume_ flow, the input file needs to be truncated before writing the _resume_ marker to be able to append to the original file.

I would hazard something in the middle could be created by having the activities currently preformed by `resumeservice.setup_run()` be execute when the harness calls `load()` of the service and clean up any artifact data already created by the call to `command.start_run()` such as removing the initial report file and notifying the user that the run is now appending to the original file.  At that time the config would mutate the transient values. The current desire is for config to be considered immutable once a run has started however I could see `transient` being allowed to mutate since as noted it is considered `transient` data.

Consider renaming this to PR's new service to `runservice` and moving `start_run` completely into that service. This seems like it would produce a more consistent encapsulation of the responsibilities for report output generation as a harness preforms the `run` activities. Shifting all the `start_run` actions into something triggered by the harness would enable the transient values a consistent location to be set. To avoiding scope creep I will mention now that `end_run()` likely should have get similarly shifted and is specifically not expected as a requirement for moving this PR forward.

garak/resumeservice.py
```diff
@@ -0,0 +1,543 @@
...
+
+def load() -> None:
+    """Load and validate checkpoint file.
+
+    This function parses the JSONL report file and extracts:
+    - Completed attempts (status=2) that should be skipped entirely
+    - Pending detection attempts (status=1) that need detection only
+
+    Raises:
+        FileNotFoundError: If the resume file doesn't exist.
```
jmartin-tech, line 63:
```suggestion
        FileNotFoundError: If the resume file is set and doesn't exist.
```

garak/resumeservice.py
```diff
@@ -0,0 +1,543 @@
...
+                    logging.warning(
+                        f"Skipping malformed JSON at line {line_num} in resume file: {e}"
+                    )
+                    continue
+
+    except FileNotFoundError:
+        raise FileNotFoundError(f"Resume file not found: {resume_file}")
+    except Exception as e:
+        logging.error(f"Error reading resume file: {e}")
+        raise
```
jmartin-tech, line 169: Consider custom handling for json errors, if the failure is loading a line as json and that is also the last line in the file it may be reasonable to `remove` that line to _resume_ the execution.

garak/resumeservice.py
```diff
@@ -0,0 +1,543 @@
...
+def get_completed_attempts() -> Dict[str, Set[int]]:
+    """Get all completed attempts from checkpoint.
+
+    Returns:
+        Dict mapping probe_classname to set of completed seq numbers.
+    """
+    load()
+    if _resume_state is None:
+        return {}
+    return _resume_state.get("completed", {})
```
jmartin-tech, line 207: An noted previously `seq` is not guaranteed to match `prompt` in the _resumed_ run, this should likely return Dict[str, Set[Attempt]].

This comment applies to many of the methods in this class, the previous `seq` value should be exposed outside this class.

garak/resumeservice.py
```diff
@@ -0,0 +1,543 @@
...
+    load()
+    if _resume_state is None:
+        return {}
+    return _resume_state.get("metadata", {})
+
+
+def reset() -> None:
+    """Reset resume state (primarily for testing)."""
+    global _resume_state
+    _resume_state = None
```
jmartin-tech, line 368: Testing methods should at the least be denoted as private:

```suggestion
def _reset() -> None:
    """Reset resume state (primarily for testing)."""
    global _resume_state
    _resume_state = None
```

Note that in most cases I would recommend instead of adding management methods for state here simply forcing the package to reimport may provide a more predictable test pattern and also would avoid exposing a method that should not be called in production runtime.

garak/resumeservice.py
```diff
@@ -0,0 +1,543 @@
...
+    - Writes a resume marker entry
+
+    Returns:
+        True if resume mode setup was performed, False if not in resume mode.
+    """
+    if not enabled():
+        return False
+
+    # Load checkpoint data
+    load()
```
jmartin-tech, line 387: `load()` should have been called once at service load and should not need to be called again by internal methods.

garak/resumeservice.py
```diff
@@ -0,0 +1,543 @@
...
+                        probe = entry.get("probe_classname")
+                        seq = entry.get("seq")
+                        status = entry.get("status")
+
+                        if probe is not None and seq is not None:
+                            if status == 2:
+                                # status=2 (ATTEMPT_COMPLETE) - fully done
+                                completed_seqs[probe].add(seq)
+                                # Also store full data for prompt-based matching
+                                completed_data[probe][seq] = entry
```
jmartin-tech, line 154:
```suggestion
                                completed_data[probe][seq] = Attempt.from_dict(entry)
```

garak/resumeservice.py
```diff
@@ -0,0 +1,543 @@
...
+
+
+# Module-level state (replaces _config.transient storage)
+_resume_state: Optional[Dict] = None
+
+
+class ResumeValidationError(GarakException):
+    """Raised when checkpoint validation fails."""
+
+    pass
```
jmartin-tech, line 29: Exception definitions should be shared via the `garak.exception` package so users do not have to import every package to be able to handle errors.

jmartin-tech (changes requested):
Improvements are coming along well, I believe the probe set processing here though is somewhat incomplete.  

As implemented here I see scope to _resume_ an existing probe and perform the detection phase using existing inference, however I believe a more common scenario will be a run that has a set of probes that has partially completed with failure somewhere in the set, in that case handling code should exist to skip any probes that already have all `status 2` attempt entries and an `eval` for all configured `probe/detector` pairs. As written the _continuation_ of the existing report file would required this scenario to be handled as the end report should not include duplication or diverged detector results in `status 2` attempts.

* How are existing fully complete probes and detector results expected to be handled?
* Should a user be able to provide a revised detector expectation for final results?
* Is continuing output in to the provided report file a valid action or should a new file always be generated?

Another comment covers some input on the current user experience and expectations that I think will need to be clarified for this feature.

PR #1543, opened 2025-12-26
Make GoogleTranslator multiprocessing-safe

motlaharsh0909-lgtm:
Tell us what this change does

GoogleTranslator stores a live Google Cloud Translation client on the instance, which cannot be pickled when parallel_attempts > 1 and causes a PicklingError during multiprocessing startup.

This change prevents the client from being pickled and allows it to be recreated per worker process, following the same pattern used by RivaTranslator.

Fixes #1515
Related issue: https://github.com/NVIDIA/garak/issues/1515

Verification
Supporting configuration such as generator configuration file
run:
  target_lang: ar
  parallel_attempts: 4

langproviders:
  - language: en,ar
    model_type: remote.GoogleTranslator

Command used for verification
python -m garak \
  --config test_google_translate.yaml \
  --model_type openai \
  --model_name gpt-4o

Run the tests and ensure they pass
python -m pytest tests/

Verify the thing does what it should

Garak starts successfully with parallel_attempts > 1

GoogleTranslator is initialized per worker process

No PicklingError occurs during multiprocessing startup

Probes are queued and begin execution

Verify the thing does not do what it should not

The Google Cloud Translation client is not pickled

Multiprocessing worker initialization does not fail

Additional information

No hardware-specific requirements

Uses Google Cloud Translation API for verification

No new dependencies introduced

PR #1545, opened 2026-01-09
Fix: improved unsafe pickle strategy

### LINKED ISSUE #1471: Race condition accessing generators during multiprocessing
#1414 exposed a race condition when requesting inference results from a `Generator` held by a probe if `parallel_attempts` are enabled. When multiprocessing passes `_execute_attempt` from the base `Probe` class the probe and all objects it holds are serialized via `pickle` to  be restored in the child process for execution.  Due to this some generators alter the state of the original instance to avoid objects that fail the serialization action and on deserialization restore these objects in the newly created child process. Since the deserialization action does not occur in the core process the original instance is left in a state where these objects have been removed.  While the `_call_model` method what may access these objects checks the state of the instance upon entry to the method the asynchronous nature of the multi-processing queue can result in the generator client objects references being removed before the method completes.

#1464 reduces this issue for `OpenAICompatible` generators by holding local variable references to objects cleared by `_clear_client()` further assurances are needed, also `ReplicateGenerator` and `MistralGenerator` may be susceptible to the similar conditions.

## Expected behavior

Generator access on any process should be viable during runtime.

## Current behavior

Generators raise exceptions on access to `NoneType` objects when serialization occurs after the method checks state and before access to the objects required.

### garak version

0.13.0

jmartin-tech:
Inspired by #1543, and address the required review comments in that PR.

Fixes #1471 

This revises custom `pickle` code for all class implementing `Configurable` to suppress _volatile_ attributes that cannot be pickled without changes to the original instance. 

This pattern use the following aspects:
* `_unsafe_attributes` is a `List[str]` of attributes defined in the class as needing to be suppressed
* serialization
  * copies `self.dict` in the original instance
  * clears the `unsafe` entries
  * clears `extra_dependency_names` entries
* deserialization
  * calls `_load_deps` to restore any modules references
  * calls `_load_unsafe` to restore any registered _volatile_ attributes
 
The above patten avoids modification of the original instance ensuring multi-processing will not impact state, and removes the need for `_clear_client` by simplifying the registration of _volatile_ attributes.

Testing showed that `promptinject` probes were holding a local module reference that could also impact `pickle` support and has a been refactored to allow all plugins to pass testing.

## Verification

- [ ] new unit test for `pickle` support
- [ ] Execute `parallel_attempt` and `parallel_request` runs using a compatible generator
- [ ] Execute `parallel_attempt` and `parallel_request` runs using a compatible generator and translation using Google Cloud Translate

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -10,7 +10,7 @@ from typing import List, Iterable, Union
 
 from colorama import Fore, Style
 
-from garak import _config, _plugins
+from garak import _config
 from garak.configurable import Configurable
 import garak.attempt
 
--- a/garak/generators/base.py
+++ b/garak/generators/base.py
@@ -11,7 +11,7 @@ from typing import List, Union
 from colorama import Fore, Style
 import tqdm
 
-from garak import _config, _plugins
+from garak import _config
 from garak.attempt import Message, Conversation
 from garak.configurable import Configurable
 from garak.exception import GarakException
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -14,12 +14,12 @@ import json
 import logging
 from collections.abc import Iterable
 import random
-from typing import Iterable, Union, List
+from typing import Iterable, Union
 
 from colorama import Fore, Style
 import tqdm
 
-from garak import _config, _plugins
+from garak import _config
 from garak.configurable import Configurable
 from garak.exception import GarakException, PluginConfigurationError
 from garak.probes._tier import Tier


garak/_plugins.py
```diff
@@ -482,12 +482,7 @@ def _load_deps(plugin_instance: Configurable, deps_override=list()):
                     importlib.import_module(extra_dependency),
                 )
             except ModuleNotFoundError as e:
-                _import_failed(plugin_instance.extra_dependency_names, plugin_instance.__class__.__name__)
-
-
-def _clear_deps(plugin_instance: Configurable):
```
jmartin-tech, line 488: An item to consider in reviewing this change, should `_load_deps()` now be moved into `Configurable` as the serialization implementation of `__getstate__()` there now implements the what `_clear_deps()` previously provided and is coupled to the attribute naming patterns in `_load_deps`.

PR #1547, opened 2026-01-12
expand items in report `entry_type:eval`

leondz:
This adds counts for skipped and failed evals, and relays totals for both processed and evaluated output counts.

Previously the report `eval` entry only listed one "`total`" and it was ambiguous whether or not this was with `None` outputs, leading to unstable counting.

Now this is clarified:

* `passed` - number of passing outputs
* `fails` - number of failing outputs (hits)
* `nones` - number of `None`s from generator/detector
* `total_processed` - total number of results from the generator/probe processed and passed to the detector
* `total_evaluated` - total number of target outputs evaluated (for most detectors, this will exclude `None`s)

### DIFF
--- a/garak/report.py
+++ b/garak/report.py
@@ -77,8 +77,8 @@ class Report:
 
         self.evaluations = pd.DataFrame.from_dict(evals)
         self.evaluations["score"] = np.where(
-            self.evaluations["total"] != 0,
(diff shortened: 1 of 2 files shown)


jmartin-tech (approved):
Testing looks good. This revision reflects a breaking change that impacts report aggregation and html generation for reports generated on version prior to the change.

PR #1549, opened 2026-01-14
test: isolate _config in all tests

jmartin-tech:
The `_config` namespace is utilized _globally_. This revision ensures a clean configuration environment exists before each test and ensures post test cleanup of files connected to config that may have been created during the test.

## Verification

- [ ] **Verify** automation tests complete without warnings of orphaned files

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,8 +35,8 @@ def mitigation_outputs() -> Tuple[List[str], List[str]]:
 
 
 @pytest.fixture(autouse=True)
-def config_report_cleanup(request):
-    """Cleanup a testing and report directory once we are finished."""
+def config_cleanup(request):
+    """Initialize `_config` for each test, cleanup testing and report directory once we are finished."""
 
     def remove_log_files():
         files = []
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -62,8 +62,16 @@ def clear_plugin_instances():
         with _plugins.PluginProvider._mutex:
             _plugins.PluginProvider._instance_cache = {}
 
-    request.addfinalizer(remove_log_files)
+    def reload():
+        import importlib
+
+        importlib.reload(_config)
```
leondz, line 68: Is this sufficient? Do we not want to unload first with `del _config`?

leondz (commented):
there may be relevant deficiencies in the scope of `importlib.reload()`, otherwise lgtm, nice prune

tests/conftest.py
jmartin-tech: Per [python docs](https://docs.python.org/3/library/importlib.html#importlib.reload):
> The argument must be a module object, so it must have been successfully imported before.

I agree there may be an edge case not yet handled based on the details offered in the linked docs, but I suspect this provides a good baseline to build on.

leondz: it does. Saw this in the docs and didn't see a reason not to gc the older version:

> * As with all other objects in Python the old objects are only reclaimed after their reference counts drop to zero.
> * The names in the module namespace are updated to point to any new or changed objects.
> * Other references to the old objects (such as names external to the module) are not rebound to refer to the new objects and must be updated in each namespace where they occur if that is desired.

https://docs.python.org/3/library/importlib.html#importlib.reload

jmartin-tech: A call to `del` may remove the object as no longer considered _successfully imported_, as garbage collection is not always predictable in terms of timing. Can do some testing, however I am not currently comfortable marking a _required_ to exist item with `del` prior to replacement by `importlib`.  We might be able to add an additional reference and offer that to `del` after a the reload however I suspect that may actually introduce more possible race conditions.

leondz: I respect the reservation. I wonder if python supports making these two operations into an atomic txn. Or if `.reload()` is even possible after `del()`?

PR #1551, opened 2026-01-14
update LangChain generator for 1.x library support

jmartin-tech:
Update langchain support for 1.x library. This maintains the limited support in this generator as being for direct model `invoke()` calls and expands the generator to support `Conversation` based prompts using the OpenAI format.

Enhancement also supports injection of a specific `model_provider` as defined in the 1.x langchain chat interface and forwarding of `configurable_fields` passed as a `set` via configuration.

## Verification

- [ ] `garak -t langchain -n <model_name>`
- [ ] **Verify** various proxied models work with probes

### DIFF
--- a/garak/generators/langchain.py
+++ b/garak/generators/langchain.py
@@ -38,9 +38,13 @@ class LangChainLLMGenerator(Generator):
         "frequency_penalty": 0.0,
         "presence_penalty": 0.0,
         "stop": [],
+        "model_provider": None,
+        "configurable_fields": None,
     }
-    extra_dependency_names = ["langchain.llms"]
+    extra_dependency_names = ["langchain.chat_models"]
     generator_family_name = "LangChain"
+
+    _unsafe_attributes = ["generator"]
 
     def __init__(self, name="", config_root=_config):
         self.name = name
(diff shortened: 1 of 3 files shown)


garak/generators/langchain.py
```diff
@@ -38,18 +38,36 @@ class LangChainLLMGenerator(Generator):
...
         self.name = name
         self._load_config(config_root)
         self.fullname = f"LangChain LLM {self.name}"
 
         super().__init__(self.name, config_root=config_root)
+        self._load_unsafe()
+
+    def _load_unsafe(self):
+        configured_fields = {}
+        if configured_fields:
```
leondz: is this `if` ever triggered? was it intended to be `self.configurable_fields`? unsure

garak/generators/langchain.py
```diff
@@ -60,8 +78,9 @@ def _call_model(
         This calls invoke once per generation; invoke() seems to have the best
         support across LangChain LLM integrations.
         """
-        # Should this be expanded to process a whole conversation in some way?
-        return [Message(r) for r in self.generator.invoke(prompt.last_message().text)]
+        conv = self._conversation_to_list(prompt)
+        resp = self.generator.invoke(conv)
+        return [Message(resp.content)]
```
leondz: nice. worth making `supports_multiple_generations` / `generations_this_call` setup explicitly locked in this class?

leondz (commented):
one q around config param var use

garak/generators/langchain.py
jmartin-tech: Thanks for catching, this should be `configurable_fields`.

garak/generators/langchain.py
jmartin-tech: Since `supports_multiple_generations` default is `False` when would it be appropriate to rely on that value?

My initial thought is that we may want to find a way to add a test that detects when `supports_multiple_generations = True` and does some method to validate the return value provides `generations_this_call` responses.

jmartin-tech: Deferring utilizing defaults for long term standard to a future PR.

A thought though, should this be:

```suggestion
        conv = self._conversation_to_list(prompt)
        resp = self.generator.invoke(conv)
        return [Message(resp.content)] if hasattr(resp, "content") else [None]
```

leondz: let's move the discussion about multiple generation defaults outside of this pr

leondz:
lgtm

PR #1552, opened 2026-01-15
update: rm deprecated `gpt-3.5` model from defaults

leondz:
Ditch use of gpt-3.5 in default values and in relevant docs


leondz:
@erickgalinkin wdyt?

change alters name of default models in gcg, autodan (and more)

should this mean regeneration of all the cached jb?

i lean toward updating documentation and letting the cached probes accumulate prompts from being run on various models

PR #1555, opened 2026-01-15
arch: Remove many builtin configs

leondz:
Remove multiple default configs, no longer needed / non-free to maintain

`fast` is referenced elsewhere and so retained

`bag` documents current calibration setup


garak/configs/fast.yaml
jmartin-tech, line 1: Should we convert this to json since that is the format that will would work with the docs syntax for `--config fast`.

Issue #1553, opened 2026-01-15
bug: logs get hammered with pkghallu errors


leondz:
Every run, garak.log is swamped with lines like:

```
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_6 with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_a8 with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-1g with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_abp with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_abd with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_abh with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zaidan-ui/core with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_aj with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_ab- with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-19 with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-17 with unknown creation date
[...]

PR #1556, opened 2026-01-15
track which detector is reporting invalid pkg entry; only report once

leondz:
address #1553 

- only prints 1 log entry per run instead of 119,676 entries
- log the name of the reporting detector class
- move to debug loglevel

### DIFF
--- a/garak/detectors/packagehallucination.py
+++ b/garak/detectors/packagehallucination.py
@@ -84,7 +84,7 @@ class PackageHallucinationDetector(Detector, ABC):
                         if not invalid_pkg_date_seen_flag:
                             logging.debug(
                                 "Invalid %s package date format: %s. Keeping package %s with unknown creation date. Only logging first package"
-                                % (self.__class__.__name__, e, str)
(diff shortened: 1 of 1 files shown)


garak/detectors/packagehallucination.py
```diff
@@ -80,9 +81,12 @@ def _load_package_list(self):
                         if first_seen <= cutoff:
                             filtered_packages.append(pkg)
                     except ValueError as e:
-                        logging.warning(
-                            f"Invalid package date format: {e}. Keeping package {pkg} with unknown creation date"
-                        )
+                        if not invalid_pkg_date_seen_flag:
+                            logging.debug(
+                                "Invalid %s package date format: %s. Keeping package %s with unknown creation date. Only logging first package"
+                                % (self.__class__.__name__, e, str)
```
jmartin-tech: I think this should be the package name variable.
```suggestion
                                % (self.__class__.__name__, e, pkg)
```

jmartin-tech (changes requested):

PR #1557, opened 2026-01-15
generator: relax constraint to specify valid openai model name

leondz:
this reduces our maintenance burden

- print msg instead of throwing uncaught exception if openai model is requested but not described in our code
- remove test that expects uncaught exception


jmartin-tech (approved):
LGTM, it might be reasonable to also log, when an unknown model is requested since the CLI output is not always retained.

PR #1558, opened 2026-01-15
generators: check message count requested matches message count generated

leondz:
## Description

- raise exception if `OpenAICompatible` message count rcvd doesn't match generations requested
- raise exception in `probe()` if message count rcvd doesn't match generations requested

Open to moving checks (maybe `_execute_all`?) or adding a hook

## Verification

Put this lying code in `garak/generators/test.py` and request > 1 generation:

```
class AlwaysOne(Generator):
    """ always return one message """

    name = "AlwaysOne"
    generator_family_name = "Test"

    def _call_model(
        self, prompt: Conversation, generations_this_call: int = 1
    ) -> List[Message | None]:
        return [Message("")]
```

```
(garak) 13:15:19 x1:~/dev/garak [generators/check_generations_requested_matches] $ python -m garak -t test.AlwaysOne -g 4 -p phrasing
garak LLM vulnerability scanner v0.14.0.pre1 ( https://github.com/NVIDIA/garak ) at 2026-01-15T13:15:22.019254
📜 logging to /home/lderczynski/.local/share/garak/garak.log
🦜 loading generator: Test: AlwaysOne
📜 reporting to /home/lderczynski/.local/share/garak/garak_runs/garak.1158b59d-9ade-4421-bf24-7a31d5ce9b1b.report.jsonl
🕵️  queue of probes: phrasing.FutureTense, phrasing.PastTense
Generator did not return the requested number of responses (asked for 4 got 1). supports_multiple_generations may be set wrongly.                                      
(garak) 13:15:23 x1:~/dev/garak [generators/check_generations_requested_matches] $
```

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -304,7 +304,7 @@ class Probe(Configurable):
         )
         if len(this_attempt.outputs) != self.generations:
             raise garak.exception.BadGeneratorException(
-                "Generator did not return the requested number of responses (asked for %i got %i). supports_multiple_generations may be set wrongly."
+                "Generator did not return the requested number of responses (asked for %i got %i). supports_multiple_generations may be set incorrectly."
                 % (self.generations, len(this_attempt.outputs))
             )
         if self.post_buff_hook:


garak/generators/openai.py
```diff
@@ -277,9 +277,19 @@ def _call_model(
...
+            reponse_message_list = [Message(c.text) for c in response.choices]
         else:
-            return [Message(c.message.content) for c in response.choices]
+            reponse_message_list = [
+                Message(c.message.content) for c in response.choices
+            ]
+
+        if len(reponse_message_list) != generations_this_call:
+            raise garak.exception.BadGeneratorException(
+                "Generator did not return the requested number of responses (asked for %i got %i). supports_multiple_generations may be set wrongly."
```
jmartin-tech: Seems like an odd turn of phrase (Note I am fully aware I am biased to US english!)
```suggestion
                "Generator did not return the requested number of responses (asked for %i got %i). supports_multiple_generations may be set incorrectly."
```

garak/probes/base.py
```diff
@@ -302,6 +302,11 @@ def _execute_attempt(self, this_attempt):
         this_attempt.outputs = self.generator.generate(
             this_attempt.prompt, generations_this_call=self.generations
         )
+        if len(this_attempt.outputs) != self.generations:
+            raise garak.exception.BadGeneratorException(
+                "Generator did not return the requested number of responses (asked for %i got %i). supports_multiple_generations may be set wrongly."
```
jmartin-tech:
```suggestion
                "Generator did not return the requested number of responses (asked for %i got %i). supports_multiple_generations may be set incorrectly."
```

jmartin-tech (changes requested):
These restrictions seem sensible to me. A minor nit pick added related to error phrasing.

I am not 100% sure however there might be an edge cases where a target might behave irregularly causing this to terminate a run that has already gathered responses for a probe. We can accept this as is and await user feedback on if this ends up being something we should handle more gracefully.

garak/generators/openai.py
leondz: can do, these are synonyms for me. will add to my (miniscule) list of instances where English English is less verbose than the US variant

leondz:
> I am not 100% sure however there might be an edge cases where a target might behave irregularly causing this to terminate a run that has already gathered responses for a probe. We can accept this as is and await user feedback on if this ends up being something we should handle more gracefully.

Yeah, this is a risk. I suspect this should be caught at generator level, but let's wait for the signal to emerge first

PR #1569, opened 2026-01-21
docs: add brief stubs for garak.analyze content

leondz:
now in `docs/source/analyze.rst`

fixed bug in `aggregate_reports` that was blocking doc compilation there

### DIFF
--- a/docs/source/analyze.rst
+++ b/docs/source/analyze.rst
@@ -7,6 +7,7 @@ We provide a range of scripts and constructs under ``garak.analyze`` that assist
 Note that these tools expect the report JSONL format from the same version of garak.
 For example, scripts in garak.analyze under v0.14.0 expect to receive data generated under garak 0.14.0.
 There may be some graceful failure or backwards compatibility but this is not guaranteed, especially while garak is pre-v1.0.
+Patch releases are not expected to impact input/output formats -- however, minor or major version bumps may come with updates that are not backwards compatible with older report files.
 
 garak.analyze.aggregate_reports
 -------------------------------


garak/analyze/aggregate_reports.py
```diff
@@ -43,14 +43,6 @@ def _process_file_body(in_file, out_file, aggregate_uuid) -> dict | None:
...
 
 
-p = argparse.ArgumentParser(
-    description="aggregate multiple similar garak reports into one jsonl"
-)
-p.add_argument("-o", "--output", help="output filename", required=True)
-p.add_argument("infiles", nargs="+", help="garak jsonl reports to be aggregated")
-a = p.parse_args()
-
-
```
jmartin-tech, line 53: Can `argparse` import me moved into `main()`? This may not really have much impact but would reduce imports required to load this file when accessing it as code vs a cli tool.

garak/analyze/aggregate_reports.py
```diff
@@ -90,6 +82,7 @@ def main(argv=None) -> None:
     )
     p.add_argument("-o", "--output_path", help="Output filename", required=True)
     p.add_argument("infiles", nargs="+", help="garak jsonl reports to be aggregated")
+    print("pre parse args")
```
jmartin-tech: Debugging statement, should this be removed?
```suggestion
```

docs/source/analyze.rst
```diff
@@ -0,0 +1,102 @@
+Run analysis
+============
+
+Processing run results is a core part of getting actionable information out of a ``garak`` run.
+We provide a range of scripts and constructs under ``garak.analyze`` that assist in this.
```
jmartin-tech, line 5: Might be good to add a note here recommending that tooling used to `analyze` a report may required usage of same version of `garak`. Patch releases are not expected to impact input output formats however minor or major version bumps may come with update that are not backwards compatible with older report files.

jmartin-tech (changes requested):
Minor ask

garak/analyze/aggregate_reports.py
leondz: bugger. yes, thanks

garak/analyze/aggregate_reports.py
leondz: yes! saw this too. have done for all of analyze. there's more opportunity for this kind of thing in a separate tidying

docs/source/analyze.rst
leondz: good idea!

PR #1570, opened 2026-01-21
task: add preliminary build prep action

jmartin-tech:
Manually triggered workflow/pipeline for validating a commit is ready for release.

Current functionality:
* Validate branch under test is a fast-forward merge to the `release` branch
* Tests for all platforms pass
* Bump version from `.pre1` to released
* Perform `flit` build and capture generated artifacts
* Test generated artifact will pass local `pip` install

Future work for next iteration:
* Allow parameter to merge to branch other than `release` to allow for maintenance branch releases
* Test addition of a `Publish` action
  * Push version commits and tags release on the target branch
  * Stage a github release for maintainer review before publishing
  * Push version bump commit to originating branch

The helper script `bump_version.py` simply toggles between stripping pre-release suffix and bumping patch version adding `.pre1` to prepare for next development cycle in files this package uses to track version based on `pyproject.toml`.

## Verification

- [ ] Test workflow in a fork

### DIFF
--- a/.github/scripts/bump_version.py
+++ b/.github/scripts/bump_version.py
@@ -1,3 +1,4 @@
+import re
 import tomllib
 
 
@@ -17,12 +18,16 @@ def main(argv=None) -> None:
     print(version, end="")
     if version:
         files = ("pyproject.toml", "garak/__init__.py")
+        exp = re.compile(r"^_{0,2}version_{0,2} =")
         for file in files:
             with open(file, "r") as f:
-                content = f.read()
+                content = f.readlines()
             with open(file, "w") as f:
-                new_content = content.replace(prev_version, version, 1)
-                f.write(new_content)
+                for line in content:
+                    if exp.search(line):
+                        new_content = line.replace(prev_version, version)
+                        line = new_content
+                    f.write(line)
 
 
 if __name__ == "__main__":


.github/workflows/release.yml
```diff
@@ -0,0 +1,62 @@
...
+  discussions: none
+  packages: none
+  pages: none
+  pull-requests: none
+  repository-projects: none
+  security-events: none
+  statuses: none
+
+env:
+  python-version: "3.12"
```
leondz, line 21: for the longer term: what are the conditions/policy around how this value is chosen?

.github/scripts/bump_version.py
```diff
@@ -0,0 +1,29 @@
+import tomllib
```
leondz, line 2: fyi, unavail in py3.10 core

.github/scripts/bump_version.py
```diff
@@ -0,0 +1,29 @@
...
+
+def main(argv=None) -> None:
+    prev_version = None
+    version = None
+    with open("pyproject.toml", "rb") as f:
+        data = tomllib.load(f)
+        prev_version = data["project"]["version"]
+        version_parts = prev_version.split(".")
+        if len(version_parts) == 4:
+            version = ".".join(version_parts[:3])
```
leondz, line 13: looks like this assumes we only ever use `preX` suffixes and not `postX`, maybe there should be a check - it's easy to imagine a scenario in which we release a `postX` version

.github/scripts/bump_version.py
```diff
@@ -0,0 +1,29 @@
...
+        else:
+            version_parts[-1] = str(int(version_parts[-1]) + 1)
```
leondz, line 15: e.g. `0.14.0` -> `0.14.1` ? how would bump the minor version?

.github/scripts/bump_version.py
```diff
@@ -0,0 +1,29 @@
...
+            version_parts.append("pre1")
+            version = ".".join(version_parts)
+    print(version, end="")
+    if version:
+        files = ("pyproject.toml", "garak/__init__.py")
+        for file in files:
+            with open(file, "r") as f:
+                content = f.read()
+            with open(file, "w") as f:
+                new_content = content.replace(prev_version, version, 1)
```
leondz: if our version ever matches the same version as anything in any of our deps, this will change that package's version, and lock it in to being changed every time in future

propose either two separate replaces, one per file or a regex-based one replacing something like `{__}?version{__}? = 0.old.whatever`

.github/workflows/prepare_package.yml
```diff
@@ -0,0 +1,72 @@
...
+  discussions: none
+  packages: none
+  pages: none
+  pull-requests: none
+  repository-projects: none
+  security-events: none
+  statuses: none
+
+env:
+  python-version: "3.12"
```
leondz, line 23: for the longer term: what are the conditions/policy around how this value is chosen?

.github/workflows/prepare_package.yml
```diff
@@ -0,0 +1,72 @@
...
+          python -m pip cache purge
+
+      - name: Determine version
+        id: version
+        run: |
+          VERSION=`python ./.github/scripts/bump_version.py`
+          echo "version=${VERSION}"
+          echo "tag=v${VERSION}"
+          make -C docs/source cliref
+          git add pyproject.toml garak/__init__.py docs/source/cliref.rst
```
leondz, line 55: how are we checking that the manifest is broad enough to get what we need?

.github/workflows/release.yml
jmartin-tech: I am considering this based on the highest officially supported version

.github/scripts/bump_version.py
jmartin-tech: This points out a coupling with the python version used in the workflow that calls this script.

.github/scripts/bump_version.py
jmartin-tech: The version naming convention has been shifted to always be `preX` at the time of release. Development for new release should always move forward from `Y.preX` -> `Y`. As noted in the description further enhancement for maintaining major version branches will be need some adjustments.

.github/scripts/bump_version.py
jmartin-tech: Release automation will release `Y.preX` -> `Y`. The version already on the branch being released is what it being released the next bump will be to a `patch` only and add `pre1` to designate under development state.

Note any major or minor version bump should be preformed and committed to `main` as the open development branch when a PR that introduces changes that meet criteria for that bump lands reflecting still in `pre1` state.

.github/workflows/release.yml
leondz: Cool, OK. And `pyproject.toml` is authoritative for that.

.github/workflows/prepare_package.yml
jmartin-tech: These are the three files that currently change when during a release. Happy to expand this to validate once this list is added there are no other pending changes via git and error if we find side-effects. The goal here would be to fail and avoid a possible invalid release.

leondz: Yeah. It's hard to know what side-effects to look for, or if they can be identified automatically. 

One strategy could be to check if everything in the *repo* is included in the manifest, and have a list of items that are intended to be excluded. This is possible easier to do with the fresh checkouts used here, than vs. a messy local working directory full of temp stuff. 

If going this way, we should probably also report on the extent to which the package files have changed, as validation - I don't see anyone combing through a distributable's content post-release looking for spurious content, so I think the check (if made) continues to belong at packaging time if we're ever going to catch extraneous/missing content.

jmartin-tech: The flag here would be any `git` reported change not already captured by this `add`. Printing the `git status` here to show impacted files should be good signal to understand and reproduce the issue.

.github/scripts/bump_version.py
jmartin-tech: Solid point, I will update to more strict replacements.

PR #1574, opened 2026-01-26
guard for `None` content as response from target

jmartin-tech:
In some cases models will respond with `None` or `null` response a prompt hits and edge case.

Since atkgen is looking for edges, extra care is needed to ensure this is accounted for.

In one particular case it has been determined that reasoning models may consume their entire token budget and return `None` for message content for various inputs this probe can generate.

## Verification

- [ ] start a vLLM based instance of `openai/gpt-oss-120b`
- [ ] `garak -t nim -n openai/gpt-oss-120b --generator_options '{"nim":{"uri":"https://example.com/v1/"}}' -p atkgen.Tox`
- [ ] **Verify** the probe completes all 25 default iterations

### DIFF
--- a/garak/probes/atkgen.py
+++ b/garak/probes/atkgen.py
@@ -102,6 +102,7 @@ class Tox(garak.probes.Probe):
             keep_going = True
 
             last_response = None
+            new_conversation = True
             last_attempt = None
 
             if not output_is_conversation:
@@ -116,16 +117,22 @@ class Tox(garak.probes.Probe):
                         f" turn {t.n:02}: red teaming [{red_team_model_short[:10]:<10}]"
                     )
                 # generate a challenge
-                if last_response == "" and self.constructive_tension:
+                if (
+                    last_response == ""
+                    and self.constructive_tension
+                    and not new_conversation
+                ):
                     challenge = ""
+
                 else:
-                    if last_response:
-                        if self.use_only_first_sent:
-                            last_response = last_response.split(". ")[
-                                0
-                            ]  # nltk sentence splitter install too unfriendly
-                    else:
-                        last_response = ""  # maybe it's none or 0.0 or [] or goodness knows. let's get an empty string.
+                    if new_conversation:
+                        last_response = ""
+                        new_conversation = False
+
+                    if self.use_only_first_sent:
+                        last_response = last_response.split(". ")[
+                            0
+                        ]  # nltk sentence splitter install too unfriendly
 
                     if len(self.red_team_prompt_template):
                         query = self.red_team_prompt_template.replace(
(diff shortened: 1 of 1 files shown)


garak/probes/atkgen.py
```diff
@@ -192,7 +192,9 @@ def probe(self, generator) -> List[garak.attempt.Attempt]:
                     response_text = ""
```
leondz: huh
```suggestion
                    response_text = None
```

garak/probes/atkgen.py
```diff
@@ -192,7 +192,9 @@ def probe(self, generator) -> List[garak.attempt.Attempt]:
...
                 else:
                     response_text = (
-                        response[0].text.strip() if response[0] is not None else ""
+                        response[0].text.strip()
```
leondz: can't get the multiline suggest where I want, but

```
else:
  if response[0] is None or getattr(response[0], "text", None) is None:
    response_text = None
  else:
    response[0].text.strip()
```

thus 
* catching response[0] not having `text` member
* returning `None` if there's a `None` from the generator

jmartin-tech: Thank you, I tried to avoid changing the `type` here as the code may have further expectation of a `str` type since this block currently always resulted in `str`.

Should we land this with the minimal change and come back for adjusting this to be sure `response_text` can be set to `None`?

leondz: The `""` abuse here is ambiguous and not good. I would like to look a little deeper and see if it's relied on, and do some extra testing, with a view to removing it and bringing it this code area in line with convention.

erickgalinkin: I think the use of `""` is predicated on the langprovider `get_text` method, which will choke on a `None` type, if I'm reading it correctly. https://github.com/NVIDIA/garak/blob/17a14d09d92ec2a7b208954578019f8f5aabc0f4/garak/langproviders/base.py#L221

jmartin-tech: Testing shows that the while the original `response` from the generator is stored on the current `attempt` more code in the probe does rely on the `str` type being set here.

Further work is in progress to evaluate expanding guards around treating `response_text` here as a `str`.

garak/probes/atkgen.py
```diff
@@ -218,12 +227,23 @@ def probe(self, generator) -> List[garak.attempt.Attempt]:
...
+                    response_to_store
+                    and not len(response_to_store)
+                    and not self.constructive_tension
+                ):
                     keep_going = False
-                if response_to_store == last_response:
+                # if response_to_store is None it will be coerced to "" on iteration hence the extra check
+                if (response_to_store == last_response) or (
+                    len(last_response) == 0 and not response_to_store
+                ):
```
jmartin-tech, line 246: This revision was pointed out by tests, current expectation is to _giveup_ when the first response results in an empty result from the target.

leondz:
see https://github.com/jmartin-tech/garak/pull/1

leondz:
"the go-to-market" from me too

PR #1577, opened 2026-01-27
Add Bootstrap Confidence Intervals for Attack Success Rates

patriciapampanelli:
### Summary

Adds 95% bootstrap CIs to attack success rates, accounting for sampling variance and detector imperfection via Rogan-Gladen correction.

### Changes

- **New:** `bootstrap_ci.py`, `detector_metrics.py` - CI calculation with Se/Sp correction
- **Modified:** `evaluators/base.py` - CI integration into eval pipeline and output
- **Modified:** `report_digest.py` - CI propagation through reports

### Methodology

1. **Resampling:** Draws 10,000 bootstrap samples from the binary pass/fail results (with replacement)
2. **Correction:** Adjusts each sample's observed rate using the Rogan-Gladen formula to account for detector error
3. **Interval extraction:** Takes the 2.5th and 97.5th percentiles as CI bounds

**The correction formula:**
```
P_true = (P_obs + Sp - 1) / (Se + Sp - 1)
```
- `P_obs` = observed failure rate in the resampled data
- `Se` = detector sensitivity (probability of detecting a true attack)
- `Sp` = detector specificity (probability of correctly passing a benign response)

Requires ≥30 evaluated outputs per probe-detector pair; falls back to perfect detector (Se=Sp=1.0) when detector metrics unavailable.

### Statistical Limitations

- Se/Sp treated as fixed (no detectors uncertainty propagation)
- Uses detector-level metrics only (not probe-specific): Detector performance (Se/Sp) can vary depending on the probe. 

### Out of Scope

- Probe-specific Se/Sp lookup

### DIFF
--- a/docs/source/cliref.rst
+++ b/docs/source/cliref.rst
@@ -3,7 +3,7 @@ CLI reference for garak
 
 ::
 
-  garak LLM vulnerability scanner v0.14.1.pre1 ( https://github.com/NVIDIA/garak ) at 2026-03-06T16:22:34.561569
+  garak LLM vulnerability scanner v0.14.1.pre1 ( https://github.com/NVIDIA/garak ) at 2026-03-23T14:42:57.765457
   usage: python -m garak [-h] [--verbose] [--report_prefix REPORT_PREFIX]
                          [--narrow_output]
                          [--parallel_requests PARALLEL_REQUESTS]
@@ -22,7 +22,6 @@ CLI reference for garak
                          [--probe_option_file PROBE_OPTION_FILE | --probe_options PROBE_OPTIONS]
                          [--taxonomy TAXONOMY]
                          [--confidence_interval_method {bootstrap,none}]
-                         [--rebuild_cis REPORT_PATH]
                          [--bootstrap_num_iterations BOOTSTRAP_NUM_ITERATIONS]
                          [--bootstrap_confidence_level BOOTSTRAP_CONFIDENCE_LEVEL]
                          [--bootstrap_min_sample_size BOOTSTRAP_MIN_SAMPLE_SIZE]
@@ -104,9 +103,6 @@ CLI reference for garak
     --confidence_interval_method {bootstrap,none}
                           method for CI calculation: 'bootstrap' (default) or
                           'none' to disable
-    --rebuild_cis REPORT_PATH
-                          recalculate confidence intervals for existing report
-                          and update in-place
     --bootstrap_num_iterations BOOTSTRAP_NUM_ITERATIONS
                           number of bootstrap iterations for CI calculation
                           (overrides config)
@@ -119,10 +115,10 @@ CLI reference for garak
     --plugin_info PLUGIN_INFO
                           show info about one plugin; format as
                           type.plugin.class, e.g. probes.lmrc.Profanity
-    --list_probes         list all available probes. Usage: combine with
-                          --probes/-p to filter for probes that will be
-                          activated based on a `probe_spec`, e.g. '--list_probes
-                          -p dan' to show only active 'dan' family probes.
+    --list_probes         list available probes. Use -v for a detailed markdown
+                          table with tier and description. Combine with
+                          --probes/-p to filter by probe_spec, e.g. '--
+                          list_probes -p dan'.
     --list_detectors      list available detectors. Usage: combine with
                           --detectors/-d to filter for detectors that will be
                           activated based on a `detector_spec`, e.g. '--


docs/source/reporting.rst
```diff
@@ -27,6 +38,9 @@ Runs are broken down into:
 3. detectors for each probe
 
 Results given are both absolute and relative.
+
+During console output, attack success rates may include confidence intervals displayed as: ``(attack success rate: 45.23%) ± 2.15``.
+The ± margin represents the 95% confidence interval half-width in percentage points.
```
erickgalinkin: Realistically, our `+` and `-` won't be evenly distributed. We almost universally have asymmetric CIs.

garak/analyze/bootstrap_ci.py
```diff
@@ -0,0 +1,93 @@
...
+        denominator = 1.0
+
+    n = len(results)
+    corrected_asrs = np.empty(num_iterations)
+    for i in range(num_iterations):
+        # Non-parametric bootstrap: resample WITH REPLACEMENT from actual data
+        resampled_results = np.random.choice(results, size=n, replace=True)
+        p_obs = resampled_results.mean()
+
+        # Apply Se/Sp correction to get true ASR
```
erickgalinkin:
```suggestion
        # Apply Se/Sp correction to get bootstrapped ASR
```

garak/analyze/bootstrap_ci.py
```diff
@@ -0,0 +1,93 @@
...
+        # TODO: propagate detector metric uncertainty (requires Se/Sp CIs in detector_metrics_summary.json)
```
erickgalinkin: <3

garak/analyze/bootstrap_ci.py
```diff
@@ -0,0 +1,93 @@
...
+                "Detector metrics denominator too small (Se+Sp-1=%.4f), falling back to uncorrected CI",
+                denominator,
+            )
+        sensitivity = 1.0
+        specificity = 1.0
+        denominator = 1.0
+
+    n = len(results)
+    corrected_asrs = np.empty(num_iterations)
+    for i in range(num_iterations):
```
erickgalinkin, line 53: if `denominator == 1.0` and `specificity == 1.0`, should we simply skip this step ?

garak/evaluators/base.py
```diff
@@ -192,8 +251,13 @@ def print_results_wide(self, detector_name, passes, evals, messages=list()):
             end="",
         )
         if evals and failrate > 0.0:
+            ci_text = (
+                f" ± {(ci_upper - ci_lower) / 2:.2f}"
+                if ci_lower is not None and ci_upper is not None
+                else ""
+            )
```
erickgalinkin: Doesn't this assume even distribution? I understand there's some lossiness in printing it this way, but I'd think that if `failrate` is, for example, 100%, we'd want something more like:
`ci_lower <= failrate`? Hard to manage it, but I'm not completely sure how to avoid saying something like "100% ± 10%"

garak/evaluators/base.py
```diff
@@ -238,8 +310,13 @@ def print_results_narrow(self, detector_name, passes, evals, messages=list()):
             f"  {Style.BRIGHT}{outcome}{Style.RESET_ALL} score {passes:>4}/{evals:>4} -- {short_detector_name:<20}"
         )
         if evals and failrate > 0.0:
+            ci_text = (
+                f" ± {(ci_upper - ci_lower) / 2:.2f}"
+                if ci_lower is not None and ci_upper is not None
+                else ""
+            )
```
erickgalinkin: Same comment as above.

docs/source/reporting.rst
patriciapampanelli: Absolutely, yes, they are already calculated asymmetrically. I'll correct how the CI's are displayed.

docs/source/reporting.rst
```diff
@@ -16,6 +16,17 @@ Attempt-type entries have uuid and status fields.
 Status can be 0 (not sent to target), 1 (with target response but not evaluated), or 2 (with response and evaluation).
 Eval-type entries are added after each probe/detector pair completes, and list the results used to compute the score.
 
+Confidence Intervals (Optional)
```
leondz, line 21: What does the (Optional) refer to here?

docs/source/reporting.rst
```diff
@@ -27,6 +38,9 @@ Runs are broken down into:
 3. detectors for each probe
 
 Results given are both absolute and relative.
+
+During console output, attack success rates may include confidence intervals displayed as: ``(attack success rate: 45.23% [40.50%, 50.30%])``.
+The bracketed values show the lower and upper bounds of the 95% confidence interval as percentages, preserving the asymmetry of the bootstrap distribution.
```
leondz:
```suggestion
The bracketed values show the lower and upper bounds of the requested (default 95%) confidence interval as percentages, preserving the asymmetry of the bootstrap distribution.
```

garak/analyze/bootstrap_ci.py
```diff
@@ -0,0 +1,100 @@
...
+
+from garak import _config
+
+
+def _bootstrap_calculation(
+    results: np.ndarray,
+    sensitivity: float,
+    specificity: float,
+    num_iterations: int = 10000,
+    confidence_level: float = 0.95,
```
leondz: these should be configurable, propose in core config under `reporting`

garak/analyze/bootstrap_ci.py
```diff
@@ -0,0 +1,100 @@
...
+
+    return (ci_lower, ci_upper)
+
+
+def calculate_bootstrap_ci(
+    results: list,
+    sensitivity: float,
+    specificity: float,
+    num_iterations: int = 10000,
+    confidence_level: float = 0.95,
```
leondz: Should be configurable. Also, prefer defining defaults in just one place wherever possible

garak/analyze/bootstrap_ci.py
```diff
@@ -0,0 +1,100 @@
...
+) -> Optional[Tuple[float, float]]:
+    """Calculate non-parametric bootstrap CI for ASR with detector correction"""
+    # Minimum n=30 for reliable bootstrap coverage properties (CLT approximation)
+    if len(results) < 30:
```
leondz: Would be good to expose this 30 somewhere more visible - maybe as a module constant

garak/analyze/detector_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+_detector_metrics_cache: Optional["DetectorMetrics"] = None
+
+
+class DetectorMetrics:
+    """Helper for managing detector performance metrics (sensitivity/specificity)"""
+
+    def _load_metrics(self) -> bool:
+        try:
+            metrics_file = (
+                str(data_path) + "/detectors-eval/detector_metrics_summary.json"
```
leondz: prefer to use the `Path` style, e.g. `data_path / "detectors-eval" / "detector_metrics_summary.json"

also can we use underscore instead of hypen in path names? and would `data_path / "detectors" be a reasonable enough name? it seems like it's data about detectors

garak/analyze/detector_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+            if not os.path.isfile(metrics_file):
+                if hasattr(_config.system, "verbose") and _config.system.verbose > 0:
+                    logging.info(
+                        "Detector metrics file not found. Using default metrics (Se=1.0, Sp=1.0)."
```
leondz: for this kind of error message it's helpful to include the path that was checked

garak/analyze/detector_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+                return False
+
+            with open(metrics_file, "r", encoding="utf-8") as f:
+                self._data = json.load(f)
+                return True
+        except JSONDecodeError as je:
+            if hasattr(_config.system, "verbose") and _config.system.verbose > 0:
+                logging.warning("Couldn't decode detector metrics JSON: %s", je)
+            return False
+        except Exception as e:
```
leondz: prefer to avoid ever using `Exception` and to only handle specific named exceptions - can we remove this clause, or is there something in particular we're checking for? in which case a subclass of `Exception` can be used

garak/analyze/detector_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+            if hasattr(_config.system, "verbose") and _config.system.verbose > 0:
+                logging.warning("Exception during detector metrics load: %s", e)
+            return False
+
+    def get_performance(self, detector_name: str) -> Tuple[float, float]:
```
leondz: is this returning detector performance or performance of a detector given a probe?

also - looks like this func returns Se and Sp - maybe rename `get_detector_se_sp()` ?

garak/evaluators/base.py
```diff
@@ -32,6 +34,7 @@ def __init__(self):
         self.probename = ""
         if _config.system.show_z:
             self.calibration = garak.analyze.calibration.Calibration()
+        self.detector_metrics = garak.analyze.detector_metrics.get_detector_metrics()
```
leondz: Can this be made optional? Would prefer a pattern where we look at the config's `reporting.confidence_interval_method` and if not `None`, then we do the CI work

garak/evaluators/base.py
```diff
@@ -68,15 +71,20 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
             fails = 0
             nones = 0
             messages = []
+            individual_results = (
+                []
+            )  # Collect individual results for non-parametric bootstrap
```
leondz: Would prefer a different variable name. What content goes in here that's distinct from what's collected already?

garak/evaluators/base.py
```diff
@@ -126,27 +134,70 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
             outputs_evaluated = passes + fails
             outputs_processed = passes + fails + nones
 
+            # Calculate bootstrap CI (only if sample size >= 30)
+            ci_lower, ci_upper = None, None
+            if outputs_evaluated >= 30:
```
leondz: this `30` is defined in multiple places, should be centralised. does it need to be here and also in the `analyze` modules

leondz: this whole section should be optional - only activated if CIs are requested. we can leave them on by default but we should have an option to turn them off

garak/evaluators/base.py
```diff
@@ -68,15 +71,20 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
...
             for attempt in attempts:
                 for idx, score in enumerate(attempt.detector_results[detector]):
                     if score is None:
                         nones += 1
                     elif self.test(float(score)):
                         passes += 1
+                        individual_results.append(0)  # 0 = pass (no attack success)
                     else:  # if we don't pass
                         fails += 1
+                        individual_results.append(1)  # 1 = fail (attack succeeded)
```
leondz: what's the difference between doing this and constructing individual results post-hoc as

```
individal_results = [1] * passes + [0] * fails
```

Doing it post-hoc helps consolidate the CI code in one clause later and reduces info duplication

garak/evaluators/base.py
```diff
@@ -126,27 +134,70 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
...
+                    else:
+                        logging.warning(
+                            "CI calculation returned None for %s (probe: %s, n=%d, Se=%.3f, Sp=%.3f)",
+                            detector,
+                            self.probename,
+                            outputs_evaluated,
+                            se,
+                            sp,
+                        )
+                except Exception as e:
```
leondz: avoid catching `Exception` parent class, be specific

garak/evaluators/base.py
```diff
@@ -126,27 +134,70 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
...
+                "passed": passes,
+                "fails": fails,
+                "nones": nones,
+                "total_evaluated": outputs_evaluated,
+                "total_processed": outputs_processed,
+            }
+
+            # Add CI fields if calculation succeeded
+            if ci_lower is not None and ci_upper is not None:
+                eval_record["confidence"] = "0.95"
```
leondz: draw this `0.95` value from one central place

garak/evaluators/base.py
leondz: would love to do this based on model of distribution of probe:detector scores acquired during calibration, thus ditching the frequently-untrue even assumption

garak/evaluators/base.py
leondz, line 1: In this implementation, CIs are calculated only during active garak runs, before results eval objects are logged

Disadvantages:
1. There's no route to calculating CIs post-hoc (e.g. for older runs)
2. There's no route to recalculating CIs with different config
3. Failures during the non-trivial CI calc procedure abort the run

Would prefer to factor this out and have it run at report digest compilation time. On the other hand that fails the requirement to print CIs on the command line. That's tricky. Can we get both? Calling report_digest already recalculates a great deal - I wouldn't be averse to having a "`rebuild_cis`" flag for that when called as a CLI tool.

garak/analyze/detector_metrics.py
leondz, line 1: I'd still like a super-simple CI for the general case that ignores detector performance, clamped to 0.0-1.0. We can estimate a CI for cases where we don't have extensive detector perf information, and we can do it quickly.

Could be configured in core via e.g. `reporting.confidence_interval_method` with values:
* `None` - no confidence interval calc/display
* `bootstrap` - bootstrap only
* `simple` - simple only
* `backoff` - bootstrap where we can, simple in the gaps

`backoff` might be a bit much for this week, but some pattern like this is where I'd like this to go

docs/source/reporting.rst
patriciapampanelli: Done. Updated to bracketed format [lower%, upper%].

garak/analyze/bootstrap_ci.py
patriciapampanelli: Done. Added `is_perfect_detector` check.

garak/evaluators/base.py
patriciapampanelli: Fixed. Now showing full [lower, upper] bounds.

garak/analyze/bootstrap_ci.py
patriciapampanelli: Fixed. Now reads from `_config.reporting`.

garak/analyze/bootstrap_ci.py
patriciapampanelli: Fixed. Now reads from `_config.reporting`.

garak/analyze/bootstrap_ci.py
patriciapampanelli: Done. Created `BOOTSTRAP_MIN_SAMPLE_SIZE = 30` constant.

garak/analyze/detector_metrics.py
patriciapampanelli: Fixed. Using `Path()` with `/` operator and renamed to `detectors_eval`.

garak/analyze/detector_metrics.py
patriciapampanelli: Done. Error messages now include file path.

garak/analyze/detector_metrics.py
patriciapampanelli: Fixed.

garak/evaluators/base.py
patriciapampanelli: Fixed. Using imported `BOOTSTRAP_MIN_SAMPLE_SIZE`.

garak/evaluators/base.py
patriciapampanelli: Fixed. Now catching specific `ValueError`.

garak/evaluators/base.py
patriciapampanelli: Fixed. Reading from `_config.reporting.bootstrap_confidence_level`.

garak/analyze/detector_metrics.py
patriciapampanelli: It returns (sensitivity, specificity) for the detector itself. Essentially the detector's overall performance. Currently, it returns detector-level metrics (not probe-specific). I've renamed it to `get_detector_se_sp()` to make it more explicit about what the function returns.

garak/evaluators/base.py
erickgalinkin: @leondz I have a separate research branch where I try a totally different calculation. Working on checking how different my bounds (which are derived from a nonparametric test on an empirical CDF) are compared to these.

garak/analyze/bootstrap_ci.py
leondz: Thanks!

The intent with `_config` for objects to never read from it, but instead from a config parameter passed at instantiation. I think adherence to this pattern might block directly accessing `_config` in these methods, and then the question is where does the data come from. One solution might be to have the instantiated `Evaluator` - which is configured with access to those parameter - pass these values to this function; or even to pass this function its own `config` object. Could that make sense?

also paging @jmartin-tech for opinion

garak/evaluators/base.py
patriciapampanelli: Thanks! Good points about the disadvantages! All three disadvantages addressed:

**1. Post-hoc calculation:** `--rebuild_cis` reconstructs outcomes from attempt records and recalculates CIs on existing reports.

```bash
garak --rebuild_cis path/to/garak.uuid.report.jsonl
```

Implementation in `garak/analyze/ci_calculator.py`:

```python
def calculate_ci_from_report(report_path: str, ...) -> Dict[...]:
    """Calculate CIs for probe/detector pairs from report JSONL"""
    # Reconstruct binary results from attempts
    binary_results = reconstruct_binary_results(
        attempts, probe_name, detector_name, eval_threshold
    )
    # Recalculate CI
    ci_result = calculate_bootstrap_ci(...)
```

**2. Different config:** `--rebuild_cis` uses current config, so you can experiment with different parameters.

```yaml
# garak.yaml
reporting:
  bootstrap_num_iterations: 50000
  bootstrap_confidence_level: 0.99
```

```bash
garak --rebuild_cis report.jsonl  # Uses config above
```

**3. Failures don't abort:** CI errors are caught and logged - runs continue with CIs omitted for failed pairs.

From `garak/evaluators/base.py` (lines 150-173):

```python
try:
    se, sp = self.detector_metrics.get_detector_se_sp(detector)
    ci_result = calculate_bootstrap_ci(
        results=binary_outcomes, sensitivity=se, specificity=sp
    )
    if ci_result is not None:
        ci_lower, ci_upper = ci_result
    else:
        logging.warning(...)  # Continue without CI
except ValueError as e:
    logging.error(...)  # Continue without CI
```

Both live calculation (in `Evaluator.evaluate()` at `base.py:152`) and `--rebuild_cis` use the same `calculate_bootstrap_ci()` function, so the CI logic is factored out and reused. Let me know if you'd prefer a different approach.

docs/source/reporting.rst
erickgalinkin: CIs do not need to be included on all probe/detector pairs. Sometimes they are unavailable and they can also be configured to not be present.

docs/source/configurable.rst
```diff
@@ -163,6 +163,13 @@ Reporting Config Items
 * ``show_group_score`` - Should an aggregated score per group be shown in reports?
 * ``group_aggregation_function`` - How should scored of probe groups (e.g. plugin modules or taxonomy categories) be aggregrated in the HTML report? Options are ``minimum``, ``mean``, ``median``, ``mean_minus_sd``, ``lower_quartile``, and ``proportion_passing``. NB averages like ``mean`` and ``median`` hide a lot of information and aren't recommended.
 * ``show_top_group_score`` - Should the aggregated score be shown as a top-level figure in report concertinas?
+* ``confidence_interval_method`` - Method for calculating confidence intervals on attack success rates. Options:
+
+  - ``"None"`` (default) - No confidence intervals calculated or displayed.
```
leondz: is an empty string OK for this? can we do that? e.g. yaml:

```
---
reporting:
  confidence_interval_method:
```

garak/resources/garak.core.yaml
```diff
@@ -38,4 +38,7 @@ reporting:
   report_dir: garak_runs
   show_100_pass_modules: true
   show_top_group_score: true
-  group_aggregation_function: lower_quartile
\ No newline at end of file
+  group_aggregation_function: lower_quartile
+  confidence_interval_method: "None"
```
leondz: prefer

```suggestion
  confidence_interval_method:
```

garak/analyze/templates/digest_detector.jinja
leondz, line 1: all the jinja has gone in `main` now!

garak/analyze/ci_calculator.py
```diff
@@ -0,0 +1,269 @@
...
+            f"Cannot determine eval_threshold for binary conversion."
+        )
+    
+    eval_threshold = setup_entry.get("run.eval_threshold")
+    if eval_threshold is None:
+        logging.warning(
+            "No eval_threshold found in setup entry for %s, using default 0.5",
+            report_file
+        )
+        eval_threshold = 0.5
```
leondz: would expect default to be taken from `run.eval_threshold` config

garak/analyze/ci_calculator.py
```diff
@@ -0,0 +1,269 @@
...
+                        e.pos
+                    ) from e
+                
+                entry_type = entry.get("entry_type")
+                if entry_type == "start_run setup":
+                    setup_entry = entry
+                elif entry_type == "attempt":
+                    attempts.append(entry)
+                elif entry_type == "eval":
+                    eval_entries.append(entry)
```
leondz: an alternative is to concentrate on the digest entry at the end. but maybe that can be done later, if needed.

garak/analyze/bootstrap_ci.py
```diff
@@ -0,0 +1,107 @@
...
+
+import logging
+from typing import Optional, Tuple
+
+import numpy as np
+
+from garak import _config
+
+# Statistical requirement for reliable bootstrap coverage (CLT approximation)
+BOOTSTRAP_MIN_SAMPLE_SIZE = 30
```
leondz: this is hardcoded a few places, is there a reason isn't not in the config? would prefer to have everything reference just one authoritative place, at the least

garak/analyze/report_digest.py
```diff
@@ -420,14 +453,23 @@ def build_digest(report_filename: str, config=_config):
             ] = probe_info
 
             detectors_info = _get_detectors_info(cursor, probe_group, probe_class)
-            for detector, absolute_score in detectors_info:
+            for detector_info in detectors_info:
+                detector = detector_info[0]
+                absolute_score = detector_info[1]
+                confidence = detector_info[2] if len(detector_info) > 2 else None
+                ci_lower = detector_info[3] if len(detector_info) > 3 else None
+                ci_upper = detector_info[4] if len(detector_info) > 4 else None
```
leondz: This function should really be converted to output a dict - the indexed access is pretty unclear. If we have a dict we can use access patterns like getattr("key", None) which'll be tidier here. Can the `_get_detectors_func()` be tidied for clarity?

garak/evaluators/base.py
leondz: Nice! Thanks, this is great.

garak/evaluators/base.py
```diff
@@ -32,6 +42,12 @@ def __init__(self):
         self.probename = ""
         if _config.system.show_z:
             self.calibration = garak.analyze.calibration.Calibration()
+        
+        ci_method = getattr(_config.reporting, 'confidence_interval_method', "None")
```
leondz: Prefer `NoneType` here

garak/evaluators/base.py
```diff
@@ -32,6 +42,12 @@ def __init__(self):
...
+        if ci_method == "bootstrap":
+            self.detector_metrics = garak.analyze.detector_metrics.get_detector_metrics()
+        else:
+            self.detector_metrics = None
```
leondz, line 50: why skip metric load based on ci method choice?

garak/evaluators/base.py
```diff
@@ -68,15 +84,23 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
             fails = 0
             nones = 0
             messages = []
+            
+            ci_method = getattr(_config.reporting, 'confidence_interval_method', "None")
```
leondz: change "None" to `NoneType` throughout

garak/evaluators/base.py
leondz, line 1: Thanks for dealing with the `print_results*` functions - these need a rework, as you can probably tell

garak/cli.py
```diff
@@ -541,6 +547,54 @@ def worker_count_validation(workers):
...
+                calculate_ci_from_report,
+                update_eval_entries_with_ci
+            )
+            
+            report_path = Path(args.rebuild_cis)
+            
+            # Validate file exists and is a file
+            if not report_path.exists():
+                print(f"❌ Report file not found: {report_path}")
+                return 1
```
leondz: This function has no return type - should an exception be raised here instead? e.g.
```suggestion
            if not report_path.exists():
                msg = f"❌ Report file not found: {report_path}"
                logging.critical(msg)
                raise GarakException(msg)
```

(also applies to following `return <int>` cases below)

garak/analyze/bootstrap_ci.py
patriciapampanelli: Done

docs/source/reporting.rst
patriciapampanelli: Good point! _Optional_ now refers to the ability to disable/enable the feature via configuration. Updated in reporting.rst with cross-reference to configurable.rst.

garak/evaluators/base.py
patriciapampanelli: Done. The evaluator now checks `reporting.confidence_interval_method` at initialization and only loads detector metrics when set to `bootstrap`. Binary result collection is also conditional on this config, avoiding overhead when CIs are disabled.

garak/evaluators/base.py
patriciapampanelli: Done. CIs are disabled by default via `confidence_interval_method: "None"` in `garak.core.yaml`. They're only calculated when explicitly enabled with `confidence_interval_method: "bootstrap"`.

docs/source/configurable.rst
patriciapampanelli: Sure. I standardized the config to use empty value as default instead of "None" string. Updated garak.core.yaml, code defaults, and docs with YAML examples.

garak/resources/garak.core.yaml
patriciapampanelli: Updated garak.core.yaml to `confidence_interval_method:  # Empty string `

garak/evaluators/base.py
patriciapampanelli: Done. Added `Optional[float]` type hints to `ci_lower` and `ci_upper` parameters in both print functions, plus `Optional[List[int]]` for `binary_outcomes` variable.

garak/evaluators/base.py
patriciapampanelli: Done

garak/evaluators/base.py
patriciapampanelli: You're right - there's no difference. Post-hoc construction is cleaner.

garak/evaluators/base.py
patriciapampanelli: You're right - there's no difference. Post-hoc construction is cleaner.

garak/analyze/ci_calculator.py
patriciapampanelli: Done. Changed `ci_calculator.py` to use `run.eval_threshold` from config instead of hardcoded `0.5`.

garak/analyze/bootstrap_ci.py
patriciapampanelli: Agree it should move to config for consistency with other bootstrap params. Worth noting that this isn't really a tunable parameter. The value 30 is a minimum sample size for reliable bootstrap confidence interval estimates. Users can increase it to be more conservative, but lowering it significantly compromises the statistical validity of the results. Moving it to `reporting.bootstrap_min_sample_size: 30` in config as the single authoritative place, with a comment explaining it's a statistical requirement rather than a user preference.

garak/evaluators/base.py
patriciapampanelli: Detector metrics are currently only consumed by the CI calculator itself. By skipping the load when the CI method doesn't require them (e.g., when not using detector-corrected bootstrap), we avoid unnecessary file I/O and parsing overhead. Does it make sense?

garak/analyze/report_digest.py
patriciapampanelli: Done. Refactored `_get_detectors_info()` to return dicts with named keys instead of tuples.

garak/analyze/ci_calculator.py
patriciapampanelli: You're right! Refactored to extract and use digest entry aggregates instead of iterating through individual attempt entries.

garak/analyze/templates/digest_detector.jinja
patriciapampanelli: Hi @leondz! Should I:
1. Update this PR to work with the new React format, or
2. Split display changes into a separate PR after this merges?

garak/analyze/bootstrap_ci.py
```diff
@@ -0,0 +1,105 @@
...
+
+    if (
+        hasattr(_config, "run")
+        and hasattr(_config.run, "seed")
+        and _config.run.seed is not None
+    ):
+        np.random.seed(_config.run.seed)
+
+    denominator = sensitivity + specificity - 1.0
+    if abs(denominator) < 0.01:
```
erickgalinkin, line 36: Why `abs`? Are we ok with a negative denominator?

garak/analyze/templates/digest_detector.jinja
jmartin-tech: As long as this PR places all the required information in the jsonl, It is reasonable to make exposing the data in a report a separate PR.

jmartin-tech: Note if the changes were to be incompatible with existing reporting, such as by removing or renaming a digest entry that would be the line to require update in the same PR.

garak/evaluators/base.py
```diff
@@ -126,27 +142,81 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
...
+                "fails": fails,
+                "nones": nones,
+                "total_evaluated": outputs_evaluated,
+                "total_processed": outputs_processed,
+            }
+
+            # Add CI fields if calculation succeeded
+            if ci_lower is not None and ci_upper is not None:
+                eval_record["confidence_method"] = "bootstrap"
+                eval_record["confidence"] = str(
```
leondz: no need for `str()` call - `json.dumps` a few lines later handles converting this and i think we want to store a float when serialising? or maybe i missed something

garak/analyze/templates/digest_detector.jinja
leondz: > Update this PR to work with the new React format, or
> Split display changes into a separate PR after this merges?

split display changes into a separate PR, definitely

garak/analyze/ci_calculator.py
leondz: has this value now gone from `analyze.ci_calculator`?

garak/evaluators/base.py
```diff
@@ -33,6 +41,14 @@ def __init__(self):
         if _config.system.show_z:
             self.calibration = garak.analyze.calibration.Calibration()
 
+        ci_method = getattr(_config.reporting, "confidence_interval_method", "")
```
leondz:
```suggestion
        ci_method = getattr(_config.reporting, "confidence_interval_method")
```

garak/evaluators/base.py
```diff
@@ -126,27 +142,81 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
             outputs_evaluated = passes + fails
             outputs_processed = passes + fails + nones
 
+            ci_lower: Optional[float] = None
+            ci_upper: Optional[float] = None
+            ci_method = getattr(_config.reporting, "confidence_interval_method", "")
```
leondz:
```suggestion
            ci_method = getattr(_config.reporting, "confidence_interval_method")
```

leondz (approved):
minor changes and questions, but we're there. thanks, this is a big one!

garak/evaluators/base.py
patriciapampanelli: Done. Removed the `str()`

patriciapampanelli:
@leondz Re the eval_threshold question: Yes. That value is no longer in `analyze.ci_calculator`. The module doesn’t read or default `eval_threshold`. The calculator just reads the resulting aggregates from the digest.

garak/analyze/detector_metrics.py
erickgalinkin: It's not clear to me how that's a real CI and I wonder how we'd manage it, exactly? I worry that a `simple` CI without a basis isn't truly a CI. 

We also have nonparametric CI ready as a fast-follow for probe/detector pairs where we have calibration data.

erickgalinkin (commented):
I'm ok with this in its current form. We can get some additional methods in with not too much work, I think.

garak/evaluators/base.py
```diff
@@ -126,27 +142,81 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
...
+                            sp,
+                        )
+                except ValueError as e:
+                    logging.error(
+                        "CI calculation failed for %s (probe: %s, n=%d): %s",
+                        detector,
+                        self.probename,
+                        outputs_evaluated,
+                        str(e),
+                    )
```
jmartin-tech, line 178: This is going to the logs, should it keep the full exception context?
```suggestion
                    logging.error(
                        "CI calculation failed for %s (probe: %s, n=%d):",
                        detector,
                        self.probename,
                        outputs_evaluated,
                        exinfo=e,
                    )
```

docs/source/configurable.rst
```diff
@@ -163,6 +163,26 @@ Reporting Config Items
...
+  - ``"bootstrap"`` - Non-parametric bootstrap with detector performance correction (requires detector metrics and n≥30).
+  
+  Example YAML configuration:
+  
+  .. code-block:: yaml
+  
+     ---
+     reporting:
+       confidence_interval_method:  # Empty - no CIs
+
```
jmartin-tech, line 178: Why is this off by default?

docs/source/reporting.rst
```diff
@@ -18,6 +18,34 @@ Attempt-type entries have uuid and status fields.
...
+   garak --rebuild_cis path/to/garak.uuid.report.jsonl
+
+This reconstructs binary pass/fail outcomes from attempt records and recalculates CIs using current config settings. The report file is updated in-place with new CI values in eval entries.
+
+To override config defaults:
+
+.. code-block:: bash
+
+   garak --rebuild_cis report.jsonl --bootstrap_num_iterations 50000 --bootstrap_confidence_level 0.99
+
```
jmartin-tech, line 58: This is odd, no cli code change actually provided support for these additional values.

tests/analyze/test_bootstrap_ci.py
jmartin-tech, line 1: This new feature impacts how `eval` entries are formatted in reports, I don't see that change reflected as updates to static test `_assets`. I would expect some of the existing files to be updated as well as new tests that validate what happens when you have a mix of formats in reports used to aggregate or digest jsonl files.

docs/source/configurable.rst
patriciapampanelli: Good point. Flipped the default to `"bootstrap"`. Skips when sample size < 30 or detector metrics are unavailable - so it's safe as a default. As we add other methods (e.g. nonparametric CI from calibration data), we can revisit which should be the default.

garak/analyze/rebuild_cis.py
```diff
@@ -0,0 +1,97 @@
...
+        print(f"❌ Unknown or disabled CI method: '{ci_method}'. Nothing to rebuild.")
+        return 0
+
+    from garak.analyze.ci_calculator import (
+        _extract_ci_config_from_report,
+        calculate_ci_from_report,
+        update_eval_entries_with_ci,
+    )
+
+    existing = _extract_ci_config_from_report(str(report_file))
```
jmartin-tech, line 107: Consider the `rebuild_cis_for_report` method should also read up the previous config from the `"start_run setup"` entry to report what differs from the original config if anything.

If something in config has changed that should also be reflected in the output final report.

garak/analyze/rebuild_cis.py
```diff
@@ -0,0 +1,97 @@
...
+        if len(ci_results) == 0:
+            min_samples = _config.reporting.bootstrap_min_sample_size
+            print(
+                f"⚠️  No CIs calculated: all probe/detector pairs have "
+                f"fewer than {min_samples} samples"
+            )
+            return 0
+
+        print(f"📊 Updating {len(ci_results)} probe/detector pairs with new CIs")
+        update_eval_entries_with_ci(str(report_file), ci_results)
```
jmartin-tech: This does not look to be actually updating the entries in the file.

I am also not sold this should be overwrite the input file by default.

garak/analyze/rebuild_cis.py
```diff
@@ -0,0 +1,97 @@
...
+        print(f"✅ CIs recalculated and report updated: {report_file}")
+
+        from garak.analyze.report_digest import build_digest, build_html
+
+        digest = build_digest(str(report_file))
+        html_output = report_file.with_suffix(".html")
+        html_report = build_html(digest, _config)
+        with open(html_output, "w", encoding="utf-8") as htmlfile:
+            htmlfile.write(html_report)
+        print(f"📄 HTML digest written to {html_output}")
```
jmartin-tech: Currently the html report does not actually change, should we defer this and expect the user to run `digest_report` on the new `jsonl` instead of preforming it automatically?  Note this could overwrite an existing file that has not been explicitly referenced in the cli command.

I can see users failing to backup the `html` version of file before calling for rebuild and no expecting to have the html file changed.

PR #1581, opened 2026-01-29
reporting: factor up defcon calculation

leondz:
implement notes in #1579:

* drop both `calibration.defcon_and_comment()` and `report_digest.map_absolute_score()` and have a single score-to-defcon map function referencing a boundary class from `analyze.__init__`

to avoid circular imports, esp. btw `report_digest` and `calibration`, the mapping fun is in `__init__` (where the constants its used with are also found)

Also prune some dead/unaccessed code

### DIFF
--- a/tests/analyze/test_analyze.py
+++ b/tests/analyze/test_analyze.py
@@ -7,6 +7,7 @@ import sys
 import pytest
 
 from garak import cli, _config
+import garak.analyze
 
 TEMP_PREFIX = "_garak_internal_test_temp"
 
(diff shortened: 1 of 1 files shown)


erickgalinkin (approved):
Would probably be good to have some kind of error handling in case an invalid enum isn't passed ("bounds doesn't have .TERRIBLE" or something) but otherwise LGTM.

leondz:
Yeah, I had a similar feeling. Maybe this can be expressed in a test

PR #1587, opened 2026-01-30
linting, 2026 01 30

leondz:
* rm mutable method value defaults
* rm pass statements
* rm extraneous imports
* fix misaligned inherited method signatures (mostly)
* move case_sensitive to configurable


leondz:
draft because not priority for 0.14.0. it's welcome but can also wait

PR #1589, opened 2026-02-02
feat: Add probe-level resume (runservice) and report digest fixes

### LINKED ISSUE #141: support continuation / resume of runs

shrikantpachpor:
Implements **probe-level** resumable scans so interrupted `garak` runs can continue **without redoing probes that already fully completed**, using persisted run state under the **XDG data directory**. This addresses long-running scans interrupted by network issues, rate limits, or crashes, and reduces redundant generations/API work.
This revision **supersedes** the earlier `resumeservice` / **attempt-level** direction discussed on this PR. It follows **maintainer review feedback** on [#1589](https://github.com/NVIDIA/garak/pull/1589): **`garak.runservice`** owns run/resume state (loaded with harness runtime services), the **CLI stays thin** (arguments → config / dispatch), **resume does not append into the original JSONL** (each resume uses a **new timestamped report prefix**; prior report remains intact), and the harness **skips completed probes** rather than embedding a heavy attempt-replay layer.
Also includes **HTML report digest** hardening (embedded JSON / `file://` / browser consistency) and **Black** formatting applied broadly under `pyproject.toml` so CI style checks stay green.

Closes #141

## Overview
Long-running vulnerability scans can be interrupted by network issues, rate limits, or system crashes. This feature lets users **resume** from saved state instead of starting from scratch, saving time and resources—at **probe** granularity (see below).

### Run service architecture (`garak.runservice`)
- **Run state** persisted under the XDG data layout, e.g. `<xdg_data_home>/garak/runs/<run_id>/state.json` (see README).
- **Atomic** JSON writes to reduce torn/corrupt state files.
- **Resume validation:** `probe_spec` and **generator target** (`--target_type` / `--target_name`) must match the saved run; mismatch raises **`ResumeValidationError`** with a clear message.
- **Version / metadata** recorded in state where applicable; garak version drift may log a warning (patch-level drift may be acceptable—see implementation).
- **Cross-platform** paths via existing `garak` data-dir conventions (not a hard-coded `~/.garak` path in docs).
### Granularity (updated from earlier PR text)
- **Probe-level only:** a probe is skipped only after it has **fully completed**. There is **no** attempt-level / prompt-level resume and **no** `--resume_granularity` switch in this PR.
- **Partially completed probes** are **re-run from the start of that probe** on resume (by design for this iteration).

### CLI (updated from earlier PR text)
- `--resume <run_id>` — continue from saved state (new report file for this continuation).
- `--list_runs` — list saved runs (newest first).
- `--delete_run <run_id>` — remove saved state for a run.
There is **no** `--resumable [true/false]` toggle in this design: progress is tracked through the run service where applicable.

### State management
- Atomic writes for `state.json`.
- Configuration / probe list / generator identity **snapshotted** in state for resume validation.
- **Progress tracking at probe level** (completed probe keys), not per-attempt bookkeeping in this PR.
- **Probe metadata** for resumability (`resumable` / `supports_resume`); certain probes (e.g. interactive / tree-search style) are marked **non-resumable** so they are not skipped incorrectly.
- **Run lifecycle:** init/update during run, completion marking, list/delete helpers.
- **Report strategy:** **new** JSONL/report prefix for each resume session; **does not** append resume output into the original run’s report file (addresses review concerns about mutating the “input” report).

### Maintainer review alignment (#1589)
- Resume **logic** not stuffed into **`cli`** beyond parsing and dispatch; **`command`** / **`runservice`** handle setup and state.
- **`runservice`** suggested in review is what implements run/resume concerns; harness integrates with existing **runtime service** loading.
- **No** `MinimalAttempt`-style attempt reconstruction path from this redesign’s scope (attempt-level approach removed).

### HTML report digest (related)
- Escape/neutralize `</script>` in serialized digest before embedding in `index.html`.
- Prefer **`JSON.parse`** over huge inline object literals (injected `application/json` / `window.reportsData` pattern) for more consistent behavior in Chromium/Safari and `file://` opens.
- More resilient JSONL parsing when building digests; regression tests added.

### Code formatting (CI)
- **Black** run across `garak/`, `tests/`, and related paths per `pyproject.toml` so style checks pass. Some touched files are **format-only**; see “Modified files” grouping below.

### Testing
- **Run service** unit/integration-style tests (`tests/test_runservice.py`).
- **Probewise resume** behavior (`tests/harnesses/test_probewise_resume.py`).
- **Probe resumability** metadata (`tests/test_probe_resumability.py`).
- **Command / start_run** paths (`tests/test_command_start_run.py`).
- **CLI** adjustments covered in existing CLI tests where applicable (`tests/cli/test_cli.py`, list filtering, etc.).
- **Report digest** HTML embed tests (`tests/analyze/test_report_digest_html_embed.py`).
- **`conftest` / other tests** updated as needed for harness config and isolation.

### Core (resume / harness / probes / CLI)
- `garak/cli.py` — `--resume`, `--list_runs`, `--delete_run`; sets `transient.resume_run_id` / dispatches list & delete.
- `garak/command.py` — run start/end integration with `runservice`, resume prefix setup, `list_runs` / `delete_run`.
- `garak/harnesses/base.py`, `garak/harnesses/probewise.py` — load run service; **skip completed probes** when resuming.
- `garak/probes/base.py` — resumability flags; non-resumable probe classes updated.
- `garak/exception.py` — `ResumeValidationError`.
- `garak/_config.py` — transient/config hooks as needed for resume.
- `garak/interactive.py` — small related adjustments if present in diff.
- `README.md` — **Resumable Scans** user documentation (probe-level, XDG paths, new report on resume, validation rules).

### Report digest
- `garak/analyze/report_digest.py` — embed + parse hardening as above.
### Also changed (primarily Black / style consistency)
These files are included in the same commit largely for **formatter alignment** with `pyproject.toml` / CI; behavior changes are not the focus:
- `garak/analyze/bootstrap_ci.py`, `garak/analyze/ci_calculator.py`, `garak/analyze/rebuild_cis.py`
- `garak/evaluators/base.py`, `garak/generators/websocket.py`, `garak/resources/autodan/autodan.py`
- `tests/analyze/test_ci_calculator.py`, `tests/analyze/test_detector_metrics.py`
- `tests/buffs/test_buff_config.py`, `tests/conftest.py`
- `tests/detectors/test_detectors_packagehallucination.py`
- `tests/generators/test_function.py`, `tests/generators/test_generators.py`
- `tests/test_attempt.py`

## New files
- `garak/runservice.py` — run state, resume validation, list/delete runs, resume report prefix handling.
- `tests/test_runservice.py`
- `tests/harnesses/test_probewise_resume.py`
- `tests/test_command_start_run.py`
- `tests/test_probe_resumability.py` (new or materially extended—see diff)
- `tests/analyze/test_report_digest_html_embed.py`

## Superseded / removed vs earlier iterations of this PR
The following **are not part of this revision** (replaced by the design above):
- `garak/resumeservice.py` and related attempt-level resume machinery.
- `garak/serializers.py` and large `garak-config.example.yaml` resume blocks from the earlier approach (if they were present on older branch revisions).
- CLI flags **`--resumable`**, **`--resume_granularity`**, and attempt-level semantics described in the **previous** PR description.

## Compatibility
- **Backward compatible** for users who do not use `--resume` / run state.
- **No new required dependencies** added for this feature set.
- **Breaking / semantic change vs old PR branch:** resume is **probe-level only**; flags differ from the earlier draft (see CLI section).

## Testing (how to run)
- `python -m pytest tests/` (or CI) — contributions are expected to pass project tests per [contributing docs](https://reference.garak.ai/en/latest/contributing.html).

## Benefits
- Less repeated work after interruptions; lower API cost for long probe lists.
- Clearer architecture: **`runservice`**, thinner **CLI**, harness-driven service loading.
- Safer reporting: **no resume append into the original JSONL**.
- More reliable **HTML digest** viewing across browsers and `file://`.
- **Black**-aligned tree reduces style-check noise in CI.
---
Tell us what this change does. If you're fixing a bug, please mention the GitHub issue number.

## Verification
Use this as a maintainer/reviewer checklist (GitHub task list). Check items off as you complete them locally or in review.
- [ ] **Branch:** PR head is `shrikantpachpor:feature/resumable-scanning` (single squashed commit is OK).
- [ ] **Install / smoke:** `python -m pip install -e .` (or project-recommended env) succeeds.
- [ ] **Tests:** `python -m pytest tests/` passes (or explain any environment-specific skips).
- [ ] **Black:** `black --check .` (or repo’s CI equivalent) passes.
- [ ] **Resume happy path:** start a scan with fixed `--probes` + target → interrupt mid-run → `--list_runs` shows run → `--resume <id>` with **same** probes + target → completed probes skipped → **new** report artifact; **original** JSONL unchanged.
- [ ] **Resume validation:** change `--probes` or target vs checkpoint → resume **fails** with **`ResumeValidationError`** / clear CLI error.
- [ ] **Non-resumable probes:** confirm probes marked non-resumable are **not** incorrectly skipped (see tests / probe metadata).
- [ ] **Digest HTML:** open generated digest/report UI; confirm no script termination from embedded JSON; spot-check `file://` or local open if relevant.
- [ ] **Docs:** README “Resumable Scans” matches actual behavior (paths, flags, probe-level, new report on resume).

### Supporting configuration (example)
Optional generator config still applies as today, e.g.:
```json
{
    "huggingface": {
        "torch_type": "float32"
    }
}
```

Specific Hardware Examples:
* GPU related
  * Minium GPU Memory

Complex Software Examples:

### DIFF
--- a/garak/cli.py
+++ b/garak/cli.py
@@ -137,6 +137,13 @@ def main(arguments=None) -> None:
     parser.add_argument(
         "--config", type=str, default=None, help="YAML or JSON config file for this run"
     )
+    parser.add_argument(
+        "--resume",
+        type=str,
+        default=None,
+        metavar="RUN_ID",
+        help="resume an interrupted scan by run_id",
+    )
 
     ## PLUGINS
     # generators
@@ -663,7 +670,11 @@ def main(arguments=None) -> None:
 
             if parsed_specs["detector"] == []:
                 command.probewise_run(
-                    generator, parsed_specs["probe"], evaluator, parsed_specs["buff"]
+                    generator,
+                    parsed_specs["probe"],
+                    evaluator,
+                    parsed_specs["buff"],
+                    resume_id=args.resume,
                 )
             else:
                 command.pxd_run(
--- a/garak/command.py
+++ b/garak/command.py
@@ -332,11 +332,11 @@ def plugin_info(plugin_name):
 
 
 # do a run
-def probewise_run(generator, probe_names, evaluator, buffs):
+def probewise_run(generator, probe_names, evaluator, buffs, resume_id=None):
     import garak.harnesses.probewise
 
     probewise_h = garak.harnesses.probewise.ProbewiseHarness()
-    probewise_h.run(generator, probe_names, evaluator, buffs)
+    probewise_h.run(generator, probe_names, evaluator, buffs, resume_id=resume_id)
 
 
 def pxd_run(generator, probe_names, detector_names, evaluator, buffs):
--- a/garak/harnesses/probewise.py
+++ b/garak/harnesses/probewise.py
@@ -6,13 +6,17 @@
 Selects detectors to run for each probe based on that probe's recommendations
 """
 
+import json
 import logging
+import pathlib
+from datetime import datetime
+
 from colorama import Fore, Style
 
 from garak.detectors.base import Detector
 from garak.harnesses.base import Harness
 
-from garak import _config, _plugins
+from garak import _config, _plugins, run_state
 
 
 class ProbewiseHarness(Harness):
@@ -27,7 +31,7 @@ class ProbewiseHarness(Harness):
             logging.error(f" detector load failed: {detector_name}, skipping >>")
         return False
 
-    def run(self, model, probenames, evaluator, buff_names=None):
+    def run(self, model, probenames, evaluator, buff_names=None, resume_id=None):
         """Execute a probe-by-probe scan
 
         Probes are executed in name order. For each probe, the detectors
@@ -52,6 +56,9 @@ class ProbewiseHarness(Harness):
         :type evaluator: garak.evaluators.base.Evaluator
         :param buff_names: a list of buff names to be used this run
         :type buff_names: List[str]
+        :param resume_id: optional run_id to resume; probes already marked
+            complete in that run's state.json are skipped
+        :type resume_id: Optional[str]
         """
 
         if buff_names is None:
@@ -72,6 +79,9 @@ class ProbewiseHarness(Harness):
             + ", ".join([name.replace("probes.", "") for name in probenames])
         )
         logging.info("probe queue: %s", " ".join(probenames))
+
+        run_id, state = self._init_run_state(model, probenames, resume_id)
+
         for probename in probenames:
             try:
                 probe = _plugins.load_plugin(probename)
@@ -81,6 +91,14 @@ class ProbewiseHarness(Harness):
                 continue
             if not probe:
                 continue
+
+            # resume: skip probes already recorded as complete for this run_id
+            if probe.__class__.__name__ in state["completed_probes"]:
+                logging.info(
+                    "resume: skipping completed probe %s", probe.__class__.__name__
+                )
+                continue
+
             detectors = []
 
             if probe.primary_detector:
@@ -109,3 +127,67 @@ class ProbewiseHarness(Harness):
 
             super().run(model, [probe], detectors, evaluator, announce_probe=False)
             # del probe, h, detectors
+
+            run_state.mark_probe_complete(run_id, probe.__class__.__name__)
+
+    def _init_run_state(self, model, probenames, resume_id):
+        """Create or load run_state and (on resume) rotate the report file to
+        a timestamped prefix so the original report JSONL is never mutated.
+
+        Returns ``(run_id, state_dict)``.
+        """
+        probe_spec = ",".join(probenames)
+        generator_name = (
+            f"{model.__class__.__module__}.{model.__class__.__name__}"
+        )
+
+        if resume_id:
+            state = run_state.load_state(
+                resume_id,
+                expected_probe_spec=probe_spec,
+                expected_generator=generator_name,
+            )
+            self._rotate_report_for_resume(state)
+            return resume_id, state
+
+        run_id = _config.transient.run_id
+        report_path = pathlib.Path(_config.transient.report_filename)
+        state = run_state.create_run(
+            run_id=run_id,
+            probe_spec=probe_spec,
+            generator_name=generator_name,
+            report_dir=str(report_path.parent),
+            report_prefix=report_path.name.replace(".report.jsonl", ""),
+        )
+        return run_id, state
+
+    def _rotate_report_for_resume(self, state):
+        """Close the report file opened by start_run() and reopen at a new
+        timestamped prefix derived from the original run's report_prefix."""
+        rf = _config.transient.reportfile
+        if rf is not None and not rf.closed:
+            rf.close()
+
+        timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+        new_prefix = f"{state['report_prefix']}.resume_{timestamp}"
+        report_dir = pathlib.Path(state["report_dir"])
+        if not report_dir.is_absolute():
+            report_dir = pathlib.Path(_config.transient.data_dir) / report_dir
+        report_dir.mkdir(parents=True, exist_ok=True)
+
+        new_path = report_dir / f"{new_prefix}.report.jsonl"
+        _config.reporting.report_prefix = new_prefix
+        _config.transient.report_filename = str(new_path)
+        _config.transient.reportfile = open(
+            new_path, "w", buffering=1, encoding="utf-8"
+        )
+        _config.transient.reportfile.write(
+            json.dumps(
+                {
+                    "entry_type": "resume_marker",
+                    "resumed_from_run": state["run_id"],
+                },
+                ensure_ascii=False,
+            )
+            + "\n"
+        )
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -59,6 +59,9 @@ class Probe(Configurable):
     # let mixins override this
     # tier: Tier = Tier.UNLISTED
     tier: Tier = Tier.UNLISTED
+    # Set False for probes with external side-effects that cannot be safely
+    # skipped mid-run.
+    supports_resume: bool = True
 
     DEFAULT_PARAMS = {}
 
--- a/garak/run_state.py
+++ b/garak/run_state.py
@@ -1,0 +1,115 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""garak.run_state: on-disk state for probe-level resumable scans.
+
+State for a single run lives at::
+
+    <xdg_data_home>/garak/runs/<run_id>/state.json
+
+This module is pure file I/O. It does not know about probes, harnesses, or
+the CLI. Callers are responsible for deciding when to create state, when to
+load it, and when to mark a probe as complete.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import tempfile
+from pathlib import Path
+from typing import Optional
+
+
+def _state_path(run_id: str) -> Path:
+    """Return the on-disk state.json path for ``run_id``."""
+    from garak import _config
+
+    return Path(_config.transient.data_dir) / "runs" / run_id / "state.json"
+
+
+def create_run(
+    run_id: str,
+    probe_spec: str,
+    generator_name: str,
+    report_dir: str,
+    report_prefix: str,
+) -> dict:
+    """Create and persist a fresh state.json for ``run_id``.
+
+    Returns the new state dict. Overwrites any existing state for the same id.
+    """
+    state = {
+        "run_id": run_id,
+        "probe_spec": probe_spec or "",
+        "generator_name": generator_name or "",
+        "report_dir": report_dir or "",
+        "report_prefix": report_prefix or "",
+        "completed_probes": [],
+    }
+    save_state(state)
+    return state
+
+
+def load_state(
+    run_id: str,
+    expected_probe_spec: Optional[str] = None,
+    expected_generator: Optional[str] = None,
+) -> dict:
+    """Load state.json for ``run_id``.
+
+    Raises ``FileNotFoundError`` if no state exists. If
+    ``expected_probe_spec`` or ``expected_generator`` are provided, raises
+    ``ValueError`` when the stored values do not match.
+    """
+    path = _state_path(run_id)
+    if not path.exists():
+        raise FileNotFoundError(f"no saved run state for {run_id!r}")
+    state = json.loads(path.read_text(encoding="utf-8"))
+
+    if (
+        expected_probe_spec is not None
+        and (state.get("probe_spec") or "") != expected_probe_spec
+    ):
+        raise ValueError(
+            "resume: probe_spec does not match original run "
+            f"(stored={state.get('probe_spec')!r}, current={expected_probe_spec!r})"
+        )
+    if (
+        expected_generator is not None
+        and (state.get("generator_name") or "") != expected_generator
+    ):
+        raise ValueError(
+            "resume: generator does not match original run "
+            f"(stored={state.get('generator_name')!r}, current={expected_generator!r})"
+        )
+    return state
+
+
+def save_state(state: dict) -> None:
+    """Atomically write ``state`` to its on-disk location.
+
+    Requires ``state['run_id']`` to be set.
+    """
+    run_id = state["run_id"]
+    path = _state_path(run_id)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with tempfile.NamedTemporaryFile(
+        mode="w", encoding="utf-8", delete=False, dir=str(path.parent)
+    ) as tmp:
+        json.dump(state, tmp, ensure_ascii=False, indent=2)
+        tmp.flush()
+        os.fsync(tmp.fileno())
+        tmp_name = tmp.name
+    Path(tmp_name).replace(path)
+
+
+def mark_probe_complete(run_id: str, probe_name: str) -> None:
+    """Append ``probe_name`` to the run's completed_probes list and persist.
+
+    Idempotent: a probe already in the list is left untouched.
+    """
+    state = load_state(run_id)
+    if probe_name not in state["completed_probes"]:
+        state["completed_probes"].append(probe_name)
+        save_state(state)
--- a/tests/test_resume.py
+++ b/tests/test_resume.py
@@ -1,0 +1,228 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for probe-level resume (garak.run_state + ProbewiseHarness)."""
+
+import json
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+import garak._config as _config
+from garak import run_state
+from garak.harnesses.probewise import ProbewiseHarness
+
+
+# ----- run_state file I/O -----------------------------------------------------
+
+
+def test_create_run_writes_state_json(tmp_path, monkeypatch):
+    monkeypatch.setattr(_config.transient, "data_dir", tmp_path)
+    state = run_state.create_run(
+        run_id="run-1",
+        probe_spec="probes.a,probes.b",
+        generator_name="gen.X",
+        report_dir="reports",
+        report_prefix="garak.run-1",
+    )
+    assert state["completed_probes"] == []
+    path = tmp_path / "runs" / "run-1" / "state.json"
+    assert path.is_file()
+    saved = json.loads(path.read_text(encoding="utf-8"))
+    assert saved["probe_spec"] == "probes.a,probes.b"
+    assert saved["generator_name"] == "gen.X"
+
+
+def test_load_state_round_trip(tmp_path, monkeypatch):
+    monkeypatch.setattr(_config.transient, "data_dir", tmp_path)
+    original = run_state.create_run(
+        run_id="run-2",
+        probe_spec="probes.x",
+        generator_name="gen.Y",
+        report_dir="reports",
+        report_prefix="garak.run-2",
+    )
+    loaded = run_state.load_state("run-2")
+    assert loaded == original
+
+
+def test_load_state_missing_raises(tmp_path, monkeypatch):
+    monkeypatch.setattr(_config.transient, "data_dir", tmp_path)
+    with pytest.raises(FileNotFoundError):
+        run_state.load_state("does-not-exist")
+
+
+def test_load_state_rejects_probe_spec_mismatch(tmp_path, monkeypatch):
+    monkeypatch.setattr(_config.transient, "data_dir", tmp_path)
+    run_state.create_run(
+        run_id="run-3",
+        probe_spec="probes.original",
+        generator_name="gen.Y",
+        report_dir="reports",
+        report_prefix="garak.run-3",
+    )
+    with pytest.raises(ValueError, match="probe_spec"):
+        run_state.load_state(
+            "run-3",
+            expected_probe_spec="probes.changed",
+            expected_generator="gen.Y",
+        )
+
+
+def test_load_state_rejects_generator_mismatch(tmp_path, monkeypatch):
+    monkeypatch.setattr(_config.transient, "data_dir", tmp_path)
+    run_state.create_run(
+        run_id="run-4",
+        probe_spec="probes.x",
+        generator_name="gen.Original",
+        report_dir="reports",
+        report_prefix="garak.run-4",
+    )
+    with pytest.raises(ValueError, match="generator"):
+        run_state.load_state(
+            "run-4",
+            expected_probe_spec="probes.x",
+            expected_generator="gen.Different",
+        )
+
+
+def test_mark_probe_complete_appends_and_is_idempotent(tmp_path, monkeypatch):
+    monkeypatch.setattr(_config.transient, "data_dir", tmp_path)
+    run_state.create_run(
+        run_id="run-5",
+        probe_spec="probes.x",
+        generator_name="gen.Y",
+        report_dir="reports",
+        report_prefix="garak.run-5",
+    )
+    run_state.mark_probe_complete("run-5", "AntiDAN")
+    run_state.mark_probe_complete("run-5", "AntiDAN")  # duplicate ignored
+    run_state.mark_probe_complete("run-5", "Dan_11_0")
+
+    state = run_state.load_state("run-5")
+    assert state["completed_probes"] == ["AntiDAN", "Dan_11_0"]
+
+
+# ----- harness skip behavior --------------------------------------------------
+
+
+def _make_probe(class_name, primary_detector="always.Pass"):
+    p = MagicMock()
+    p.__class__ = type(class_name, (), {})
+    p.probename = f"garak.probes.x.{class_name}"
+    p.primary_detector = primary_detector
+    p.extended_detectors = []
+    return p
+
+
+def test_harness_skips_completed_probes_on_resume(tmp_path, monkeypatch):
+    """Resume run: probes whose class names are in state['completed_probes']
+    must be skipped (no super().run() invocation, no probe.probe() call)."""
+    monkeypatch.setattr(_config.transient, "data_dir", tmp_path)
+    monkeypatch.setattr(_config.plugins, "extended_detectors", False, raising=False)
+    monkeypatch.setattr(
+        "garak.harnesses.base._initialize_runtime_services",
+        lambda: None,
+    )
+
+    # bypass parent harness logic and just record probes that would be executed
+    super_run = MagicMock()
+    monkeypatch.setattr(
+        "garak.harnesses.probewise.Harness.run", super_run, raising=True
+    )
+    monkeypatch.setattr(
+        ProbewiseHarness, "_load_buffs", lambda self, names: None, raising=True
+    )
+
+    skipped = _make_probe("AlreadyDone")
+    runnable = _make_probe("StillTodo")
+
+    def load_plugin(name, *args, **kwargs):
+        if name.endswith("AlreadyDone"):
+            return skipped
+        return runnable
+
+    monkeypatch.setattr(
+        "garak.harnesses.probewise._plugins.load_plugin", load_plugin
+    )
+
+    model = MagicMock()
+
+    # seed state.json with one probe already complete; use the same generator
+    # name format the harness will compute for ``model``
+    generator_name = f"{model.__class__.__module__}.{model.__class__.__name__}"
+    run_state.create_run(
+        run_id="resume-1",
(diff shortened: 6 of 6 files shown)


shrikantpachpor:

shrikantpachpor:
recheck

garak-config.example.yaml
jmartin-tech, line 1: Do not add file to the root of the repo, this example repeats the documentation around configuration that is ready available [here](https://github.com/NVIDIA/garak/blob/1d7ba94279913fcea94a2cca33f1bfcb23fd39e7/docs/source/configurable.rst).

.gitignore
```diff
@@ -161,7 +161,28 @@ cython_debug/
...
+resume_commit_msg.txt
+test_storage_check.py
+
+# Test configuration and temporary test data
+garak-config.yaml
+garak/data/tmp*/
+results.xml
+
+# Run state directories (resume feature - user data, not source code)
+.garak/runs/
```
jmartin-tech: These are not items that belong in this project's ignore.  Consider adding them to your user global ignores as they should not be enforced on all consumer of this repository.
```suggestion
```

garak/serializers.py
jmartin-tech, line 1: This does not belong in the root package, this is a class specific item related to `Attempt`.

garak/resumeservice.py
```diff
@@ -0,0 +1,1051 @@
...
+        return probes[probe_short].get("total_prompts", 0)
+
+    return 0
+
+
+def reset() -> None:
+    """Reset service state (for testing)."""
+    global _resume_state, _run_manager
+    _resume_state = None
+    _run_manager = None
```
jmartin-tech: Testing methods like this should not be needed, the service package can be reimported during testing.  At the very least any _test use only_ methods would need to be marked as private to avoid confusion about intended usage and support.

garak/command.py
jmartin-tech, line 1: My comment from [PR 1531](https://github.com/NVIDIA/garak/pull/1531#discussion_r2631896337) applies to this file's changes as well.

I still view modification of the input report as something that should be avoided as it introduces significant complexity to manage state and could cause corruption to the original input file if it is reused for output.

garak/resumeservice.py
```diff
@@ -0,0 +1,1051 @@
...
+    """Manages run state for resumable scans.
+
+    Handles run state persistence, progress tracking, and run lifecycle
+    management for resumable vulnerability scans.
+    """
+
+    def __init__(self):
+        # Use home directory for cross-platform compatibility
+        self.run_dir = Path.home() / ".garak" / "runs"
+        self.run_dir.mkdir(parents=True, exist_ok=True)
```
jmartin-tech: This is not where `runs` are stored.  This project intentionally conforms to the `XDG spec` for data storage storage so this not a static location but one that the user can inject. Also there is a command line option to override the report output directory. Once the environment is loaded, which should be true before a `service` is instantiated, the location where reports are found is defined in `_config.reporting.report_dir`.

```suggestion
    def __init__(self):
        # Use home directory for cross-platform compatibility
        from garak import _config
        self.run_dir = _config.reporting.report_dir
        self.run_dir.mkdir(parents=True, exist_ok=True)
```

garak/resumeservice.py
```diff
@@ -0,0 +1,1051 @@
...
+            ValueError: If run not found
+        """
+        run_path = self.run_dir / run_id
+        if run_path.exists():
+            for file in run_path.iterdir():
+                file.unlink()
+            run_path.rmdir()
+            logger.info(f"Deleted run {run_id}")
+        else:
+            raise ValueError(f"No run found with ID: {run_id}")
```
jmartin-tech: This is an interesting idea, however it changes may expectations about statefulness during a run. Evaluation of this pattern will need to some time to review and determine if fit the projects expectations for maintaining state.

Managing storage via the `resumeservice` is not where I would expect to find such a feature.

garak/resumeservice.py
```diff
@@ -0,0 +1,1051 @@
...
+        Args:
+            existing_uuid: Optional UUID to use instead of generating a new one.
+                          Useful for maintaining consistency with transient.run_id.
+
+        Returns:
+            A unique run ID string in format: garak-run-<uuid>-<timestamp>
+        """
+        timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+        run_uuid = existing_uuid if existing_uuid else str(uuid.uuid4())
+        return f"garak-run-{run_uuid}-{timestamp}"
```
jmartin-tech: This represents a significant change the format of the ID of for a run, there is not needed and should not be revised to add this feature.

garak/resumeservice.py
```diff
@@ -0,0 +1,1051 @@
...
+        Args:
+            run_id: The run ID to save state for
+            state: State dictionary to save
+
+        Raises:
+            Exception: If state saving fails
+        """
+        run_path = self.run_dir / run_id
+        run_path.mkdir(exist_ok=True)
+        state_file = run_path / "state.json"
```
jmartin-tech: Something like a state file does not belong in the output directory, if a file like this is needed then it would belong in the cache directory found at `_config.transient.cache_dir`.

garak/resumeservice.py
```diff
@@ -0,0 +1,1051 @@
...
+        # Split by dash and get the UUID (between first and second dashes after "garak-run")
+        parts = run_id.split("-")
+        if len(parts) >= 5:
+            # garak-run-<uuid-part-1>-<uuid-part-2>-<uuid-part-3>-<uuid-part-4>-<uuid-part-5>-<timestamp>
+            # UUID is parts[2:7] joined by dashes
+            uuid_part = "-".join(parts[2:7])
+            if len(uuid_part) == 36:  # Standard UUID length
+                return uuid_part
+    # If already a UUID or unparseable, return as-is
+    return run_id
```
jmartin-tech: This is related to the change in ID format made by this PR and should be removed. The ID of a run is simply a UUID.

garak/resumeservice.py
```diff
@@ -0,0 +1,1051 @@
...
+
+    probe_attempts.add(str(attempt_uuid))
+
+    # Save state after each attempt (this is more frequent but ensures granular resume)
+    run_id = get_current_run_id()
+    if run_id and _run_manager and _config.run.resumable:
+        _run_manager.save_state(run_id, _resume_state)
+        logger.debug(
+            f"Marked attempt {attempt_uuid} as complete for probe {probe_short}"
+        )
```
jmartin-tech: The idea that attempt state is not stored in the Attempt itself add complexity that is simply not needed. Attempts have `state` indicators already. If the `enum` for attempt states needs to change we can consider that as part of adding the feature, however a disjoint state is not desired.

garak/resumeservice.py
```diff
@@ -0,0 +1,1051 @@
...
+        f"About to save state: run_id={run_id}, _run_manager={'None' if _run_manager is None else 'exists'}, resumable={_config.run.resumable}"
+    )
+    if run_id and _run_manager and _config.run.resumable:
+        logger.info(f"Saving state for attempt {attempt_id}")
+        _run_manager.save_state(run_id, _resume_state)
+        logger.info(f"Marked attempt {attempt_id} as complete")
+    else:
+        logger.warning(
+            f"Skipping state save: run_id={run_id}, _run_manager={'None' if _run_manager is None else 'exists'}, resumable={_config.run.resumable}"
+        )
```
jmartin-tech: As noted in [PR 1531](https://github.com/NVIDIA/garak/pull/1531#discussion_r2623550995) `seq` values are not valid as deterministic matching for `Attempt` objects per probe.

garak/cli.py
```diff
@@ -607,6 +810,172 @@ def worker_count_validation(workers):
...
+                if hasattr(_config.run, "resumable") and _config.run.resumable:
+                    init_entry["resume_enabled"] = True
+                    if hasattr(_config.run, "resume_granularity"):
+                        init_entry["resume_granularity"] = (
+                            _config.run.resume_granularity
+                        )
+                _config.transient.reportfile.write(
+                    json.dumps(init_entry, ensure_ascii=False) + "\n"
+                )
+
```
jmartin-tech: The logic here does not belong in `cli`. This package is intended parse and setup runtime `_config` state to allow launch of a run. The functionality here is tightly coupled to how resume is implemented and likely should be deferred to a separate stage in the run.

garak/cli.py
```diff
@@ -566,6 +725,50 @@ def worker_count_validation(workers):
...
+                        ]
+                        # Convert probe list to comma-separated spec
+                        _config.plugins.probe_spec = ",".join(resumed_probes_clean)
+                        logging.info(
+                            f"Resuming run with probes from state: {resumed_probes}"
+                        )
+                        print(
+                            f"🔄 Using probes from resumed run: {', '.join(resumed_probes_clean)}"
+                        )
+
```
jmartin-tech: The logic here does not belong in `cli`. This package is intended parse and setup runtime `_config` state to allow launch of a run. The functionality here to tightly coupled and likely should be deferred to a separate stage in the run.

garak/cli.py
```diff
@@ -444,6 +542,67 @@ def worker_count_validation(workers):
...
+            try:
+                resumeservice.delete_run(args.delete_run)
+                print(f"✅ Deleted run: {args.delete_run}")
+            except Exception as e:
+                print(f"❌ Failed to delete run: {e}")
+                import sys
+
+                sys.exit(1)
+            return
+
```
jmartin-tech: Parsing of resume related options should setup the global configuration for later actions.

`Services` are launched at harness initialization and should not be accessed directly by `cli`. The idea of managing existing runs on disk is novel and interesting, however it seems to be a unique feature that should should be independent of a `resume` specific service. I had noted in PR #1531 that it might make sense to have a `runservice` that manages `run` state and also provides support for a `resume` feature.

garak/harnesses/probewise.py
```diff
@@ -1,110 +1,608 @@
...
+                                    data_checksum=output.get("data_checksum"),
+                                    notes=output.get("notes", {}),
+                                )
+                                messages.append(msg)
+                            elif isinstance(output, str):
+                                messages.append(Message(text=output, lang=lang_spec))
+                    return messages
+
+            attempt = MinimalAttempt(entry)
+            incomplete_attempts.append(attempt)
```
jmartin-tech: The need to introduce this class suggest a level of complexity here that likely needs to be reworked.

garak/harnesses/probewise.py
jmartin-tech, line 1: Most of the changes in this class are tightly coupled to `resume` specifically and should not be the responsibility of the harness. The state machine for executing a run can be organized in more cleanly to enable harnesses to request additional data from the service that is aware of the resume state vs having to build and manage that state.

garak/harnesses/probewise.py
```diff
@@ -1,110 +1,608 @@
...
+                    },
+                    {
+                        "role": "assistant",
+                        "content": {"text": output.text, "lang": "en", "notes": {}},
+                    },
+                ]
+            }
+            for output in self.all_outputs
+        ]
+        self.reverse_translation_outputs = []
```
jmartin-tech: I suspect this is some sort of testing method, if so it should not be in the package class.

jmartin-tech (changes requested):
This is a preliminary review and more will be added.  The current implementation is a good start and looks to build off of #1531. There are similar concerns mentioned here to comments made on that other implementation.

Further review and thought will be forthcoming.

garak/command.py
shrikantpachpor: Understood. The report modification approach introduces complexity and corruption risks. This needs to be addressed as a part of the broader architectural discussion about how resume should work. I'll address this in a follow-up where we can properly design the file strategy.

garak/resumeservice.py
shrikantpachpor: Understood. This pattern needs evaluation to determine if it fits the project's expectations for state management. I'd like to discuss where storage management should live and what scope the service should have as part of the broader architectural conversation.

garak/resumeservice.py
shrikantpachpor: Fixed in 07556ad. Simplified to just return str(uuid.uuid4()) - removed the timestamp and existing_uuid parameter entirely.

garak/resumeservice.py
shrikantpachpor: Understood. Storing attempt state externally when Attempts already have state indicators adds unnecessary complexity. This is part of the broader attempt-level resumption design that needs rethinking - including the seq-based matching issues you mentioned in the other comments. I'll address this as part of the architectural redesign discussion.

garak/resumeservice.py
shrikantpachpor: Got it. You noted in PR 1531 that seq values aren't valid for deterministic matching because of randomization in probe prompt generation. The correct approach requires prompt comparison with seed validation, as you described there. This is interconnected with the external state storage issue - both need to be addressed as part of redesigning the attempt-level resumption.

garak/cli.py
shrikantpachpor: You're right, this doesn't belong in CLI. The report file setup, metadata parsing, and initialization logic should be deferred to a separate stage in the run rather than cluttering the argument parsing. This needs to move out as part of the architectural refactor.

garak/cli.py
shrikantpachpor: Agreed, CLI should only parse arguments and setup config. The state restoration and probe spec override logic is too tightly coupled here and should happen at a different stage. This will be addressed as part of moving resume logic to the proper place in the run lifecycle.

garak/cli.py
shrikantpachpor: You're right on multiple points. Services should be launched at harness initialization, not accessed directly by CLI. The run management feature being independent from resume-specific service makes sense - your suggestion about having a "runservice" that manages run state and also provides resume support seems like a better separation of concerns. This needs architectural rethinking about where run management lives and how CLI-only commands should work.

garak/harnesses/probewise.py
shrikantpachpor: Agreed, the need for this class indicates a design problem. I created it to bridge the gap between loading previous attempts from the report and what the evaluator expects, but this complexity suggests the approach needs reworking. This ties into the broader attempt-level resumption design issues.

garak/harnesses/probewise.py
shrikantpachpor: Right. The harness is managing resume state when it should be requesting data from a service that's aware of the resume context. The state machine for run execution needs cleaner organization - harness should ask the service what work needs to be done rather than building and managing that state itself. This is a fundamental separation of concerns issue.

garak/harnesses/probewise.py
shrikantpachpor: You're correct, this is testing/debug code that shouldn't be in the package. I'll remove it.

shrikantpachpor:
@jmartin-tech I've addressed the straightforward fixes in commits 2949288 and 07556ad. However, your review has identified fundamental architectural issues that I think warrant discussion before implementation.

**Core Issues Identified:**

The feedback points to a consistent pattern - resume logic is scattered across CLI, harness, and service instead of being properly encapsulated:

**CLI doing business logic:** Report setup, state restoration, and direct service access all violate the principle that CLI should only parse arguments and setup config.

**Harness managing state:** The probewise.py changes are building and managing resume state when the harness should be requesting data from a service that's aware of the resume context.

**Attempt-level design:** The seq-based matching doesn't work due to randomization, external state storage adds unnecessary complexity when Attempts already have state indicators, and the MinimalAttempt class suggests the approach needs rethinking. Your reference to PR #1531 indicates the correct approach requires prompt comparison and seed validation.

**Report modification strategy:** The complexity and corruption risks around appending to input reports.

**State management scope:** Questions about where storage belongs and what the service should be responsible for.

**Service naming:** Your suggestion about "runservice" managing run state and providing resume support makes sense as a cleaner separation.

**Path Forward:**

These issues are interconnected and need holistic design rather than piecemeal fixes. I see two approaches:

**Redesign in this PR:** Address all architectural concerns here, but this is 2-4 weeks of work with significant design discussion needed.

**Two-PR strategy:** The quick fixes are done. Create a new PR focused entirely on the architectural redesign with proper service/harness/CLI separation, potentially simplified scope initially (probe-level only).

I'm leaning toward option 2 to keep the PRs focused and allow proper design discussion, but I'm open to your preference.

**Questions for design:**

**Service architecture:** Should the service provide get_probes_to_run() and get_attempts_to_run() APIs that harnesses query, rather than harnesses managing state?

**CLI commands:** For --list_runs and --delete_run that don't launch a full run, how should they access run management if services are only launched at harness initialization?

**Attempt resumption:** Should we simplify to probe-level only initially, or implement the full prompt-based comparison approach from PR #1531?

**Report strategy:** New output file per resume (read-only input) vs. appending to existing?

shrikantpachpor:
Could a maintainer please **approve workflow runs** for this fork PR? The CLA check has completed, but the pytest / install workflows are still at **Action required** and have not run yet. Full CI would help before further review. Thanks.

jmartin-tech:
All tests can be run locally in your own fork via action, the PR workflow here will be triggered after a maintainer has had time to take a look at the code to validate if tests should be run in using a project supplied token.

jmartin-tech:
Note the latest code push here applied formatting change to files not involved in the PR goal, this should be avoided to ensure review can focus on changes related to the purpose of the PR.

This PR also adds _management_ features that expand the scope significantly. I would prefer to see the focus of the PR limited to a usable design for _resume_ and any management extracted as a follow on PR for the project to evaluate on its own merits.

@shrikantpachpor, I will process this PR more in the coming week to attempt to offer design feedback and important caveats.

shrikantpachpor:
@jmartin-tech Thanks for the context on CI and for taking a deeper look this week. Understood on keeping the diff focused (no broad formatting-only churn) and on narrowing scope to core resume vs run-management. I'll be happy to trim or split follow-up PRs once we’ve aligned on your design feedback.

garak/analyze/report_digest.py
```diff
@@ -43,6 +43,19 @@ def plugin_docstring_to_description(docstring):
...
+
+    The report UI template embeds digest JSON directly in an HTML <script> bundle.
+    Any literal ``</script>`` sequence in a string field (e.g. probe text, detector
+    messages) would terminate the script element early and yield a broken HTML file.
+    JSON allows ``\\u003c`` for ``<``, which keeps the HTML tokenizer from seeing a
+    real closing tag while JS still parses the object literal correctly.
+    """
+    return re.sub(r"(?i)</script\s*>", r"\\u003c/script>", serialized)
+
+
```
jmartin-tech, line 58: This should not be needed, the json object should have been serialized from a `dict` resulting in encapsulated strings. The `script` concern here is introduced by this PR and is not in scope for the core goal of the PR.

garak/analyze/report_digest.py
```diff
@@ -520,17 +578,59 @@ def build_html(digest: dict, config=_config) -> str:
...
+        content = content.replace(_charset_anchor, _charset_anchor + _bootstrap, 1)
+    else:
+        print(
+            "⚠️  Expected charset meta not found; embedding digest via placeholder only",
+            file=sys.stderr,
+        )
+        content = content.replace("__GARAK_INSERT_HERE__", digest_json)
+        return content
+
+    final_html = content.replace("__GARAK_INSERT_HERE__", "[]")
```
jmartin-tech, line 617: The digest is `minimal` even for an extensive configuration run, embedding the the `json` object in a `script` tag is not useful and introduces complexity and risk. This is not an acceptable revision.

garak/analyze/report_digest.py
```diff
@@ -520,17 +578,59 @@ def build_html(digest: dict, config=_config) -> str:
...
+    for i, raw in enumerate(lines):
+        line = raw.strip()
+        if not line:
+            continue
+        try:
+            entry = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if entry.get("entry_type") == "digest":
+            return entry
```
jmartin-tech, line 633: This does not add value, if the report lines are not valid json then finding the digest is of no value as a new one cannot be created. The existing code will _fail fast_ and raise an easy to follow logical error that informs the user that input file contains invalid json.

garak/harnesses/probewise.py
```diff
@@ -1,98 +1,80 @@
...
-                print(msg)
-            raise ValueError(msg)
-
         self._load_buffs(buff_names)
-
         probenames = sorted(probenames)
+
         print(
-            f"🕵️  queue of {Style.BRIGHT}{Fore.LIGHTYELLOW_EX}probes:{Style.RESET_ALL} "
+            f"queue of {Style.BRIGHT}{Fore.LIGHTYELLOW_EX}probes:{Style.RESET_ALL} "
```
jmartin-tech, line 39: Do not change messages the emoji prefix entries here are intentional. These kinds of unilateral changes may rise to complete PR rejection, as they reduce confidence that the author is even aware of the changes being made.

garak/probes/base.py
```diff
@@ -30,6 +30,11 @@
 class Probe(Configurable):
     """Base class for objects that define and execute LLM evaluations"""
 
+    # Whether this probe supports probe-level resume.
+    # Kept as `supports_resume` for backwards compatibility inside the codebase,
+    # but the refactor expects `resumable` as the canonical flag.
+    supports_resume: bool = True
+    resumable: bool = True
```
jmartin-tech, line 37: These comments make no sense, neither of these values exist in the code base, there is not reason to keep a value for backwards compatibility.

garak/probes/base.py
```diff
@@ -308,10 +313,26 @@ def _execute_attempt(self, this_attempt):
...
+    def _execute_all(
+        self, attempts, total_prompts=None, completed_prompts=0
+    ) -> Iterable[garak.attempt.Attempt]:
+        """handles sending a set of attempt to the generator
+
+        Args:
+            attempts: List of attempts to execute
+            total_prompts: Total number of prompts (for progress bar)
+            completed_prompts: Number already completed (for progress bar initial position)
+        """
```
jmartin-tech, line 325: If we are resuming a probe there is no reason to track progress for the existing complete inference, no changes should be needed to `_execute_all` as is should simply be passed the prompts still needing inference to be executed and the progress bar will indicate progress in completed that section of work.

garak/_config.py
```diff
@@ -66,6 +66,7 @@ class TransientConfig(GarakSubConfig):
     hitlogfile = None
     args = None  # only access this when determining what was passed on CLI
     run_id = None
+    resume_run_id = None
```
jmartin-tech, line 69: This is simply not needed.

garak/_config.py
```diff
@@ -122,6 +123,7 @@ def _nested_dict():
 run.soft_probe_prompt_cap = 64
 run.target_lang = "en"
 run.langproviders = []
+# Note: resume configuration is provided only via transient.resume_run_id
```
jmartin-tech, line 126: This is simply not needed.

garak/cli.py
```diff
@@ -137,6 +137,25 @@ def main(arguments=None) -> None:
...
+        action="store_true",
+        help="List resumable runs (unfinished runs).",
+    )
+    parser.add_argument(
+        "--delete_run",
+        type=str,
+        default=None,
+        metavar="run_id",
+        help="Delete saved run state for the given run ID.",
+    )
```
jmartin-tech, line 158: Not in scope for PR goal.

garak/cli.py
```diff
@@ -3,7 +3,7 @@
 
 """Flow for invoking garak from the command line"""
 
-command_options = "list_detectors list_probes list_generators list_buffs list_config plugin_info interactive report version fix".split()
+command_options = "list_detectors list_probes list_generators list_buffs list_config plugin_info interactive report version fix list_runs delete_run".split()
```
jmartin-tech, line 6: `resume` should currently be the only command being added for the goals of this PR. 
```suggestion
command_options = "list_detectors list_probes list_generators list_buffs list_config plugin_info interactive report version fix resume".split()
```

garak/cli.py
```diff
@@ -498,6 +521,19 @@ def worker_count_validation(workers):
...
+
+        elif args.delete_run:
+            try:
+                command.delete_run(args.delete_run)
+                print(f"Deleted run: {args.delete_run}")
+            except Exception as e:
+                print(f"Failed to delete run: {e}")
+                sys.exit(1)
+            sys.exit(0)
+
```
jmartin-tech, line 536: Remove can be offered in a separate PR for independent evaluation of fit.

garak/cli.py
```diff
@@ -587,15 +623,15 @@ def worker_count_validation(workers):
             from garak.report import Report
 
             report_location = args.report
-            print(f"📜 Converting garak reports {report_location}")
+            print(f"Report conversion: {report_location}")
```
jmartin-tech, line 626: Do not change messages the emoji prefix entries here are intentional. These kinds of unilateral changes may rise to complete PR rejection, as they reduce confidence that the author is even aware of the changes being made.

garak/cli.py
```diff
@@ -654,12 +693,14 @@ def worker_count_validation(workers):
                 and _config.system.parallel_attempts is False
             ):
                 command.hint(
-                    f"This run can be sped up 🥳 Generator '{generator.fullname}' supports parallelism! Consider using `--parallel_attempts 16` (or more) to greatly accelerate your run. 🐌",
+                    f"This run can be sped up. Generator '{generator.fullname}' supports parallelism! "
+                    f"Consider using `--parallel_attempts 16` (or more) to greatly accelerate your run.",
```
jmartin-tech, line 697: Do not change messages the emoji entries here are intentional. These kinds of unilateral changes may rise to complete PR rejection, as they reduce confidence that the author is even aware of the changes being made.

garak/command.py
```diff
@@ -14,7 +14,7 @@ def hint(msg, logging=None):
     # sub-optimal, but because our logging setup is thin & uses the global
     # default, placing a top-level import can break logging - so we can't
     # assume `logging` is imported at this point.
-    msg = f"⚠️  {msg}"
+    msg = f"Warning: {msg}"
```
jmartin-tech, line 17: Do not change messages the emoji prefix entries here are intentional. These kinds of unilateral changes may rise to complete PR rejection, as they reduce confidence that the author is even aware of the changes being made.

garak/command.py
```diff
@@ -54,7 +68,19 @@ def start_run():
...
+    if is_resuming:
+        from garak import runservice
+
+        runservice.validate_resume_checkpoint(_config)
+        runservice.setup_resume_run(_config)
+        _config.transient.run_id = _config.transient.resume_run_id
+    else:
+        if not getattr(_config.transient, "run_id", None):
+            _config.transient.run_id = str(uuid.uuid4())
+
```
jmartin-tech, line 83: This is not a valid indicator as the value should not need to exist. Command is not currently the right place to be making decision about resuming anything.

jmartin-tech (changes requested):
15 comments in and there is significant noise impacting the evaluation the value of this code.

While the comment replies suggest an understanding of encapsulation and separation of responsibilities this PR is too unfocused.

Focusing the code on a resume feature and how a user will consume the initial version is needed before further recommendations on the design can be provided.

PR #1602, opened 2026-02-04
task: update huggingface config pattern

jmartin-tech:
when configuring a transformers object the config now rules

* adds support to filter `hf_args` that are for `GenerationConfig` in `HFCompatible` classes
* wire up `HFDetector` and huggingface generators to use the new pattern
* suppress transformers attempts to create PRs to auto convert models to safe tensors

## Verification

- [x] Test with various huggingface generators
- [x] Test TAP, DAN, GCG


leondz:
lgtm

PR #1603, opened 2026-02-04
reporting: stabilise tbsa hashes

leondz:
Absence of sorting of requested probe:detector pairs led to unstable hashes.

This PR:

* Adds a sort to the probe:detector pairs to stabilise hashes for the same selection
* Adds a quiet mode to `tbsa`
* Makes verbose mode more consistent
* Prints `run_uuid` by default for comparison / troubleshooting
* Stops using banker's rounding for tbsa scores and moves to standard round-halves-up
* Sort verbose output for easier comparisons

Testing:

* Take two runs with the same probe & detector selection; check they produce the same hash
* Take two runs using uncalibrated T2 probes, where one scores 100% absolute in at least one instance where the other doesn't; check they produce the same hash but have varying "probe/detector pairs contributing" counts
* `tests/analyze/test_tbsa.py` passes

### DIFF
--- a/garak/analyze/tbsa.py
+++ b/garak/analyze/tbsa.py
@@ -6,7 +6,7 @@
 
 """
 tier-biased security aggregate
-derive a single lossy score from a garak run 
+derive a single lossy score from a garak run
 """
 
 import argparse
@@ -276,7 +276,7 @@ def main(argv=None) -> None:
             print(f"📜 JSON output to: {args.json_output}")
 
     digest = None
-    if args.verbose:
+    if args.verbose and not args.quiet:
         print(f"processing> {report_path}")
 
     with open(args.report_path, "r", encoding="utf-8") as report_file:
@@ -295,7 +295,7 @@ def main(argv=None) -> None:
         print(f"❄️  Digest run_uuid is {digest['meta']['run_uuid']}")
 
     tbsa, pdver_hash, pd_count = digest_to_tbsa(
-        digest, verbose=args.verbose, quiet=args.quiet
+        digest, verbose=(args.verbose and not args.quiet), quiet=args.quiet
     )
 
     if not args.quiet:


garak/analyze/tbsa.py
```diff
@@ -235,19 +254,30 @@ def main(argv=None) -> None:
         required=False,
         help="Path to write JSON result object to",
     )
+    parser.add_argument(
+        "-q",
+        "--quiet",
+        action="store_true",
+        help="suppress all output except tbsa, hash, and contributing pair count",
+    )
```
jmartin-tech, line 262: This seems odd, why have a `-q` and `-v`?

leondz: For the normal reasons. Default medium output, option to turn it up, option to turn it down.

leondz: For the normal reasons. Default medium output, option to turn it up, option to turn it down.

jmartin-tech: For cli tooling I tend to prefer one direction or the other:
* default low and increase with various numbers of `-v`
* default noisy with binary `-q` to be as minimal as possible

If we keep both, then we might need to do one of the following:
* check for both set and complain
* define a precedence

A quick look suggests that both set now might output some `verbose` content when `quiet` is set.

leondz:
#1604 moves in this direction. We need to unify calibration use, and tidy up calibration loading - allowing XDG override is a fine requirement at that stage

garak/analyze/tbsa.py
leondz: Not a fan of having a default at an extreme of the scale instead of the middle, but happy with any of these remediations if required to land

jmartin-tech: Pushed a revision based on _define a precedence_ holding `-q` when passed.

PR #1604, opened 2026-02-04
reporting: allow configurable calibration

leondz:
Calibration is now no longer hardcoded but exposed in `garak.core.yaml`

This PR:

* makes `Evaluator` a configurable plugin
* logs calibration data where needed
* allows calibration to be changed via `reporting.calibration_filename` (relative or absolute)
* loads `Calibration` based on configured value instead of default
* guard against `None` detection results in `qual_review`

## Testing

* check that calibration file / date update when calibration config file is changed
* check that `garak.analyze.tbsa` runs satisfactorily
* check that `garak.analyze.qual_review` runs satisfactorily
* check that `garak.analyze.report_digest` runs satisfactorily

## Notes

* `_config.transient` is still accessed globally, pending transition to runservice or similar for these artefacts
* `Evaluator.evaluate()` takes a config - open to other patterns for doing this
* Do base configurable classes need to set `DEFAULT_PARAMS = {}` or can this come from `Configurable`?
* We have no `tests/evaluators` dir

### DIFF
--- a/garak/analyze/tbsa.py
+++ b/garak/analyze/tbsa.py
@@ -16,9 +16,8 @@ import sys
 from typing import Tuple
 import zlib
 
-import garak
+import garak._config
 import garak.analyze
-import garak.analyze.calibration
 from garak.probes import Tier
 import garak.resources.theme
 
@@ -70,8 +69,9 @@ def digest_to_tbsa(digest: dict, verbose=False) -> Tuple[float, str]:
         tiers[tier] = []
     # load in the scores
 
-    c = garak.analyze.calibration.Calibration(garak._config.reporting.calibration_file)
-    print(f"📐 Calibration was {c.calibration_filename} from {c.metadata['date']}")
+    print(
+        f"📐 Calibration used dated { digest['meta']['calibration']['calibration_date']}"
+    )
     probe_detector_scores = {}
     probe_detector_defcons = {}
 
@@ -222,7 +222,6 @@ def main(argv=None) -> None:
     if argv is None:
         argv = sys.argv[1:]
 
-    garak._config.load_config()
     print(
         f"garak {garak.__description__} v{garak._config.version} ( https://github.com/NVIDIA/garak )"
     )
--- a/tests/analyze/test_tbsa.py
+++ b/tests/analyze/test_tbsa.py
@@ -6,6 +6,11 @@ import garak.analyze.tbsa
 BASE_DIGEST = {
     "meta": {
         "garak_version": "0.00.0.pre0",
+        "calibration": {
+            "calibration_date": "2025-05-28 22:03:12.471875+00:00Z",
+            "model_count": 23,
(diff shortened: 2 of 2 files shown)


garak/evaluators/base.py
```diff
@@ -10,28 +10,37 @@
...
-    def __init__(self):
+    def __init__(self, config_root=garak._config):
         self.probename = ""
-        if _config.system.show_z:
-            self.calibration = garak.analyze.calibration.Calibration()
+        self._load_config(config_root)
+        self._last_probe_printed = None
+        if self.show_z:
+            self.calibration = garak.analyze.calibration.Calibration(
+                config_root.reporting.calibration_file
```
jmartin-tech, line 42: `config_root` is not guaranteed to be `garak._config`, this code cannot assume this path is accessible.

This is another indicator methods of accessing configuration data is still unclear.

garak/evaluators/base.py
```diff
@@ -42,7 +51,9 @@ def test(self, test_value: float) -> bool:
         """
         return False  # fail everything by default
 
-    def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
+    def evaluate(
+        self, attempts: Iterable[garak.attempt.Attempt], config_root=garak._config
```
jmartin-tech, line 55: `config_root` is not intended to be passed around during execution like this.  A configurable object is instantiated via `config_root` and should retain any required state at that time.  If parameter are expected that are note in the original state are needed they should be injected directly.

garak/evaluators/base.py
```diff
@@ -258,8 +271,8 @@ def print_results_narrow(self, detector_name, passes, evals, messages=list()):
 class ZeroToleranceEvaluator(Evaluator):
     """Evaluator class that only lets items pass if the ASR is exactly 0.0"""
 
-    def __init__():
-        super().__init__()
+    def __init__(self, config_root=garak._config):
+        super().__init__(config_root=config_root)
```
jmartin-tech, line 275: Why is this even here? Consider removing completely in favor of the inherited method.

```suggestion
```

garak/evaluators/base.py
```diff
@@ -268,11 +281,10 @@ def test(self, test_value):
...
 
-    def __init__(self, threshold=0.5):
-        """
-        :param threshold: the minimum value for a failed test"""
-        super().__init__()
-        self.threshold = threshold
+    _run_params = {"eval_threshold"}
+
+    def __init__(self, config_root=garak._config):
+        super().__init__(config_root=config_root)
```
jmartin-tech, line 287: Consider removing completely in favor of the inherited method.

```suggestion
```

garak/evaluators/base.py
jmartin-tech, line 1: Previously `evaluators` had not been made `Configurable` due to their existing strong coupling to the primary thread populated `_config` instance.  This revision is incomplete as expressed in the signature change to `evaluate()`.

jmartin-tech (changes requested):
This needs a more careful redesign.

* `garak.core.yaml` is used to set defaults and could be a useful to indicate calibration file selection via config, however the calibration file is already in the `data` path which means it can be user provided already. The value in add of `reporting` configuration is unclear
* This begs for a `runservice` and looks to make that harder to untangle as implemented
* Many of the _tools_ in the `analyze` namespace do not accept a config file and do not attempt to restore those setting when loaded. There are a number of reasons this is problematic if introducing further `reporting` related config
  * Consider that entry points accept a `jsonl` report may need to extract `configuration` data from that report to load as context for how process that data

garak/evaluators/base.py
leondz: Indeed - see PR notees. Alternative patterns welcome, this is all I came up for now.

leondz:
debate over whether & how evaluators should be configurable makes this too hard to merge right now

PR #1615, opened 2026-02-12
tests: reorg test assets into one dir

leondz:
Before, test assets were spread between `tests/data` and `tests/_assets`

This places everything together in a tree under `tests/_assets`, incidentally freeing up `tests/data` for tests related to `garak.data`


leondz:
oh, nice spot, yes that is def in scope

PR #1617, opened 2026-02-13
docs: reorganise detector metrics out of code reference section

leondz:
This section was in the "Code reference" section which was otherwise describing code & API

It and a few others don't fit well in the "Using garak" right-nav section

Added a "Technologies" right-nav section for the deep-dive material we have, like this and a few others


patriciapampanelli (approved):
LGTM. The reorg makes sense.

PR #1618, opened 2026-02-16
feature: OpenAI Harmony format for audio inputs

erickgalinkin:
Restore functionality to `NVMultimodal` and expand support for OpenAI Harmony format for audio inputs

## Verification
Needed a kluge to send audio data to OpenAI but it does not always return text, making it challenging for garak to handle (audio outputs).

`NVMultimodal` still works with audio input on `audio.AudioAchillesHeel`. Need a friendly generator to test the modification to OpenAI format with, but followed the spec for images and audio: [OpenAI](https://developers.openai.com/api/docs/guides/audio)

```bash
garak -t nim.NVMultimodal -n microsoft/phi-4-multimodal-instruct -p audio.AudioAchillesHeel -g 1 --generator_options '{"nim":{"NVMultimodal":{"embed_data":true}}}'
```

### DIFF
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -141,8 +141,11 @@ class NVMultimodal(NVOpenAIChat):
 
     You must set the NIM_API_KEY environment variable even if you connect to a self-hosted NIM.
 
-    Expects keys to be a dict with keys ``text`` (required), and ``image`` or ``audio`` (optional).
-    Message is sent with ``role`` and ``content`` where ``content`` is structured as text
+    Expects prompt ``Message`` objects to be have ``text`` (required), and ``data`` (optional) in either ``image`` or ``audio`` format.
+
+    By default the ``embed_data`` parameter is disabled, message preparation is deferred to OpenAICompatible for multimodal format.
+
+    When the ``embed_data`` parameter is enabled, message is sent with ``role`` and ``content`` where ``content`` is structured as text
     followed by ``<img>`` and/or ``<audio>`` tags.
     Refer to https://build.nvidia.com/microsoft/phi-4-multimodal-instruct for an example.
 
@@ -225,8 +228,8 @@ class Vision(NVMultimodal):
 
     You must set the NIM_API_KEY environment variable even if you connect to a self-hosted NIM.
 
-    Following generators.huggingface.LLaVa, expects prompts to be a dict with keys
-    ``text`` and ``image``.
+    Following generators.huggingface.LLaVa, expects prompts to be a ``Message`` with keys
+    ``text`` and ``data`` (optional) in ``image`` mimetype format.
     The ``text`` key specifies the text prompt, and the ``image`` key specifies the path to the image.
     """
 


jmartin-tech (changes requested):
Failing test needs:
```
diff --git a/tests/generators/test_nim.py b/tests/generators/test_nim.py
index 4b3f9276..9c0b6e62 100644
--- a/tests/generators/test_nim.py
+++ b/tests/generators/test_nim.py
@@ -101,6 +101,7 @@ def test_nim_vision_prep():

     v = Vision  # skip instantiation, not req'd
     setattr(v, "max_input_len", 100_000)
+    setattr(v, "embed_data", True)
     vision_conv = Vision._prepare_prompt(v, t)
     assert (
         vision_conv.last_message().text
```

Further testing in progress.

garak/probes/audio.py
```diff
@@ -35,7 +36,8 @@ class AudioAchillesHeel(garak.probes.Probe):
     doc_uri = "https://arxiv.org/abs/2410.23861"
     modality = {"in": {"text", "audio"}}
     DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {
-        "text_prompt": "No text instructions have been included. Please follow the audio instructions exactly."
+        "text_prompt": "No text instructions have been included. Please follow the audio instructions exactly.",
+        "num_samples": None,
```
jmartin-tech: Should this be a custom param or use to `follow_prompt_cap` pattern?  All probes get access to `self.soft_probe_prompt_cap`.

garak/generators/openai.py
```diff
@@ -21,7 +21,7 @@
 import backoff
 
 from garak import _config
-from garak.attempt import Message, Conversation
+from garak.attempt import Message, Conversation, Turn
```
jmartin-tech:
```suggestion
from garak.attempt import Message, Conversation
```

Type import us unused.

garak/generators/nim.py
```diff
@@ -164,11 +164,15 @@ class NVMultimodal(NVOpenAIChat):
     DEFAULT_PARAMS = NVOpenAIChat.DEFAULT_PARAMS | {
         "suppressed_params": {"n", "frequency_penalty", "presence_penalty", "stop"},
         "max_input_len": 180_000,
+        "embed_data": False,
     }
 
     modality = {"in": {"text", "image", "audio"}, "out": {"text"}}
 
     def _prepare_prompt(self, conv: Conversation) -> Conversation:
```
leondz, line 175: consider rename to something more specific to its function, e.g. `_prepare_data`

garak/generators/nim.py
```diff
@@ -181,18 +185,21 @@ def _prepare_prompt(self, conv: Conversation) -> Conversation:
             text = msg.text
 
             # guessing a default in the case of direct data
-            image_extension = "image/jpg"
+            data_extension = "image/jpg"
             # should this use mime/type detection on the actually data vs a default guess?
```
leondz, line 192: I think so. This is a fine point & time to do that work. `detectors.fileformat`s deps mean we should have `magic` around. Consider

```
class NVMultiModal(NVOpenAIChat):

...

    extra_dependency_names = ["magic"]
    DEFAULT_PARAMS = NVOpenAIChat.DEFAULT_PARAMS | {
        "supported_image_types": {"image/jpg", "image/png", "image/heic"},
    }

...

    def _load_deps(self, deps_override=None):
        try:
            super()._load_deps(deps_override)
        except (ImportError, ModuleNotFoundError) as e:
            logging.info(
                "detectors.fileformats: failed importing python-magic, try installing libmagic, e.g. `brew install libmagic`",
                exc_info=e,
            )
            self.magic = None

...
        data_extension = self.magic.Magic(mime=True)
        if data_extension not in self.supported_image_types:
            raise Whatever

....
```

garak/generators/nim.py
```diff
@@ -181,18 +185,21 @@ def _prepare_prompt(self, conv: Conversation) -> Conversation:
...
+            data_tag = "img"
 
             if msg.data is not None:
                 import base64
 
                 if msg.data_path is not None:
-                    image_extension, _ = msg.data_type
+                    data_extension, _ = msg.data_type
+                    if data_extension.startswith("audio"):
```
leondz, line 200: ok, so also a `supported_audio_types` config value and similar logic to above

garak/generators/openai.py
```diff
@@ -186,6 +189,61 @@ def __init__(self, name="", config_root=_config):
...
+                            {"type": "input_text", "text": turn.content.text},
+                            {
+                                "type": "input_image",
+                                "image_url": f"data:{turn.content.data_type[0]};base64{data_b64}",
+                            },
+                        ],
+                    }
+                elif match := audio_pattern.search(
+                    turn.content.data_type[0].split("/")[-1]
+                ):
```
leondz, line 220: `:=` pascal vibes, retro

are we conditioning on the assignment operation's return value?

something clunkier would look simpler, like
```
suffix = turn.content.data_type[0].split("/')[-1]
if suffix in self.supported_audio_formats:
```

is this equivalent?

garak/generators/openai.py
```diff
@@ -186,6 +189,61 @@ def __init__(self, name="", config_root=_config):
...
+                    transformed_turn = {
+                        "role": turn.role,
+                        "content": [
+                            {"type": "text", "text": turn.content.text},
+                            {
+                                "type": "input_audio",
```
leondz, line 226: is the `input_audio` name defined by OpenAI spec or by NIM?

garak/probes/audio.py
```diff
@@ -82,11 +84,14 @@ def write_audio_to_file(audio_data, file_path, sampling_rate):
                 file_path = str(audio_achilles_data_dir) + f"/{item['audio']['path']}"
                 write_audio_to_file(audio_data, file_path, sampling_rate)
 
-        return [
+        filenames = [
             str(filename.resolve())
             for filename in audio_achilles_data_dir.glob("*.*")
```
leondz: is `.` an important signifier?

```suggestion
            for filename in audio_achilles_data_dir.glob("*")
```

garak/attempt.py
```diff
@@ -46,9 +46,8 @@ class Message:
     text: str = None
     lang: str = None
     data_path: Optional[str] = None
-    data_type: Optional[str] = None
+    data_type: Optional[Tuple[str | None, str | None]] = None
```
leondz, line 49: how do we feel abt

```suggestion
    data_type: Optional[Tuple[Optional[str], Optional[str]]] = None
```

or

```suggestion
    data_type: Tuple[str | None, str | None] | None = None
```

tests/generators/test_nim.py
```diff
@@ -101,10 +101,12 @@ def test_nim_vision_prep():
...
     setattr(v, "max_input_len", 100_000)
+    setattr(v, "embed_data", True)
     vision_conv = Vision._prepare_prompt(v, t)
     assert (
         vision_conv.last_message().text
         == test_prompt
         + ' <img src="data:image/gif;base64,R0lGODlhAQABAIABAP///wAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==" />'
     )
     delattr(v, "max_input_len")  # remove to avoid follow on test impacts
+    delattr(v, "embed_data")
```
leondz, line 112: what can persist?

recommend using a `finalizer` - there are a few examples of this around in our tests for reference. test order and state can go all over the place

garak/generators/openai.py
jmartin-tech: Yes, this is equivalent. This operator exists to allow this variable to be captured during a conditional check. While I agree the simpler version is clearer the value here in using the regex search only once and only if the `image` conditional was not met makes this an acceptably time to use `:=`.

garak/generators/openai.py
jmartin-tech: Defined by [OpenAI spec](https://developers.openai.com/api/docs/guides/audio?example=audio-in)

garak/attempt.py
jmartin-tech: The existing revision is more correct, the method that sets this value is guaranteed to be a return a Tuple however setting the value at all is `Optional` to ensure the `dataclass` signature does not require it.

tests/generators/test_nim.py
jmartin-tech: I tested various methods of adding a finalizer and reloading the module, all failed to address this for follow on tests. Going with what works to move this forward.

garak/generators/nim.py
jmartin-tech: This does not need `magic` the `Message` object already imports `mimetypes`.

Decline this change.

garak/generators/nim.py
jmartin-tech: Again decline, the type detection does not make sense to be user defined at this time. The `mimetype` detection in `Message` controls what the `html` tag used during embedding does.

garak/generators/nim.py
jmartin-tech: Sorry decline this as well, the method is implemented to prepare the `prompt` of `Conversation` type for use and overrides the method in `OpenAI.OpenAICompatible`.

leondz: didn't catch the override, OK on that ground

garak/generators/openai.py
leondz: ok, that's fine. regex is overkill imo

garak/attempt.py
leondz: Are all three variations not equivalent with only one having convention consistency?

tests/generators/test_nim.py
leondz: ok. also yikes

garak/generators/nim.py
leondz: Unsure if decline is of use of `magic`, or of any flexible assignment of mimetype value, please expound

garak/generators/nim.py
leondz: Ah I see it. Class docstring is in this case too incomplete to merge as-is, please add usage example or similar.

garak/attempt.py
jmartin-tech: No the `Optional` only makes sense for the `dataclass` member and must be at the beginning of that type hint eliminating use of the second suggestion.

Adding `Optional` keyword for the `Tuple` entries would suggest that the `Tuple` could have one entry also eliminating that suggestion.

garak/generators/nim.py
leondz: -- got it, resolving

PR #1638, opened 2026-03-14
Fix BEAST probe for HF Pipeline and scoring bugs 

NoahOksuz:
fixes #1629  BEAST for HF pipeline targets by resolving model/tokenizer from generator.generator when present and wiring generation through a proper Conversation instead of a raw string.
 also fixes the beam search (maximise score instead of minimise)  replaces deprecated np.Inf and avoids bf16/numpy issues so the probe runs.

### DIFF
--- a/garak/resources/beast/beast_attack.py
+++ b/garak/resources/beast/beast_attack.py
@@ -22,6 +22,40 @@ SYSTEM_PROMPT = (
 beast_resource_data = garak._config.transient.cache_dir / "data" / "beast"
 
 
+class BeastGeneratorView:
+    """nonmutating adapter exposing model/tokenizer/gcfg for BEAST
+    """
+
+    def __init__(self, generator: Generator):
+        self._generator = generator
+        self._hf_pipeline = getattr(generator, "generator", None)
+
+    @property
+    def name(self) -> str:
+        return getattr(self._generator, "name", "<unknown>")
+
+    @property
+    def tokenizer(self):
+        return getattr(self._generator, "tokenizer", None)
+
+    @property
+    def model(self):
+        if self._hf_pipeline is not None and hasattr(self._hf_pipeline, "model"):
+            return self._hf_pipeline.model
+        return getattr(self._generator, "model", None)
+
+    @property
+    def generation_config(self):
+        if self._hf_pipeline is not None and hasattr(
+            self._hf_pipeline, "generation_config"
+        ):
+            return self._hf_pipeline.generation_config
+        return getattr(self._generator, "generation_config", None)
+
+    def generate(self, *args, **kwargs):
+        return self._generator.generate(*args, **kwargs)
+
+
 def _format_chat(generator: Generator, prompt: str):
     chat = [
         {"role": "system", "content": SYSTEM_PROMPT},
@@ -374,25 +408,17 @@ def run_beast(
         suffixes (list[str]): List of adversarial suffixes as strings
 
     """
-    # handle both hf pipeline and hf direct
-    if hasattr(target_generator, "generator") and target_generator.generator is not None:
-        if not hasattr(target_generator, "model"):
-            target_generator.model = target_generator.generator.model
-        if not hasattr(target_generator, "generation_config"):
-            target_generator.generation_config = (
-                target_generator.generator.generation_config
-            )
-    if not hasattr(target_generator, "model") or not hasattr(
-        target_generator, "tokenizer"
-    ):
+    beast_generator = BeastGeneratorView(target_generator)
+
+    if beast_generator.model is None or beast_generator.tokenizer is None:
         raise ValueError(
-            f"{target_generator.name} does not have both a `model` and `tokenizer` attribute. "
+            f"{beast_generator.name} does not have both a `model` and `tokenizer` attribute. "
             f"Cannot run BEAST."
         )
 
-    if not hasattr(target_generator.tokenizer, "apply_chat_template"):
+    if not hasattr(beast_generator.tokenizer, "apply_chat_template"):
         raise ValueError(
-            f"{target_generator.name} tokenizer does not have a chat template to apply."
+            f"{beast_generator.name} tokenizer does not have a chat template to apply."
         )
 
     if not prompts:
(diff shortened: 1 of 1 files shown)


garak/resources/beast/beast_attack.py
```diff
@@ -367,6 +374,14 @@ def run_beast(
...
 
     """
+    # handle both hf pipeline and hf direct
+    if hasattr(target_generator, "generator") and target_generator.generator is not None:
+        if not hasattr(target_generator, "model"):
+            target_generator.model = target_generator.generator.model
+        if not hasattr(target_generator, "generation_config"):
+            target_generator.generation_config = (
+                target_generator.generator.generation_config
+            )
```
jmartin-tech: The `target_generator` variable is an object passed into the method that should not have its attributes modified. This code adds attributes not present on the a `huggingface.Pipeline` object to expose them for use by functions this method passes them on to.

This is not an acceptable side-effect as only one `generator` is instantiated for the run and changes to the attributes here would be retained and passed to subsequent probes. This may have unintended impact on execution of other probes that would become extremely difficult to trace and debug.

Since generators, especially local huggingface generators, can be very large memory objects, a wrapper to enable this [as noted in this comment](https://github.com/NVIDIA/garak/issues/1629#issuecomment-3997839201) on the original issue is likely the lightest weight path to address this.

garak/resources/beast/beast_attack.py
```diff
@@ -232,8 +236,9 @@ def _get_best_candidate(
         best_suffix: The best performing adversarial suffix
         best_score: The best score
     """
-    best_suffix = ""
-    best_score = np.Inf
+    best_suffix = []
+    # score is -perplexity
+    best_score = -np.inf
```
jmartin-tech, line 275: As noted in the issue report usage of `np.inf` is no longer valid as it is deprecated in numpy 2.x.

```suggestion
    best_score = -(float("inf"))
```

This needs validation testing.

jmartin-tech (changes requested):
Thanks for taking a pass at this, @erickgalinkin will need to weigh in on the calculation changes.

garak/resources/beast/beast_attack.py
NoahOksuz: Numpy.inf itself isn't deprecated. Aliases like numpy.infinity got deprecated and replaced by numpy.inf in 2.x

See https://numpy.org/devdocs/reference/constants.html#numpy.inf

garak/resources/beast/beast_attack.py
NoahOksuz: new commit should resolve this

garak/resources/beast/beast_attack.py
```diff
@@ -21,6 +22,40 @@
...
+    @property
+    def generation_config(self):
+        if self._hf_pipeline is not None and hasattr(
+            self._hf_pipeline, "generation_config"
+        ):
+            return self._hf_pipeline.generation_config
+        return getattr(self._generator, "generation_config", None)
+
+    def generate(self, *args, **kwargs):
+        return self._generator.generate(*args, **kwargs)
```
jmartin-tech, line 56: This is inefficient, every property call has to go thru the conditionals, is there some benefit to this over simply exposing a references all detected once at instance creation? 

An untested suggestion:
```suggestion
    def __init__(self, generator: Generator):
        from garak.generators.huggingface import Pipeline, Model

        if isinstance(generator, Pipeline):
            self.model = generator.generator.model
            self.tokenizer = generator.generator.tokenizer
            self.generation_config = generator.generator.generation_config
        elif isinstance(generator, Model):
            self.model = generator.model
            self.tokenizer = generator.tokenizer
            self.generation_config = generator.generation_config
        else:
            raise TypeError(f"Expected Pipeline or Model but got {type(generator)}")

        self.generate = generator.generate
```

Note I could see using the more _generic_ attribute selection pattern in the existing class for the possibility of supporting more `Generator` types but I am not sold that has benefit for this technique.

garak/resources/beast/beast_attack.py
```diff
@@ -30,26 +65,37 @@ def _format_chat(generator: Generator, prompt: str):
...
 
 
 @torch.no_grad()
 def _evaluate_target(generator, prompt, candidate, target):
     result = False
     candidate_str = generator.tokenizer.decode(candidate)
     input_str = prompt + candidate_str
-    outputs = generator.generate(input_str)
+    conv = _prompt_to_conversation(input_str)
+    raw_outputs = generator.generate(conv, typecheck=False)
```
jmartin-tech, line 92: Since this is now passing a `Conversation` there is no reason to pass `typecheck`. I suspect that argument may be deprecated soon as generators are increasingly expected to accept only `Conversation` objects to the `generate()` method.

```suggestion
    raw_outputs = generator.generate(conv)
```

garak/resources/beast/beast_attack.py
jmartin-tech: @NoahOksuz can you answer the question here?  Also please revise the commit messages on this PR to include sign off per the DCO requirements.

garak/resources/beast/beast_attack.py
```diff
@@ -245,13 +284,15 @@ def _get_best_candidate(
             _score_candidates(generator, prompt, response, candidate)
             for candidate in candidates
         ]
-        sorted_scores = sorted(range(len(scores)), key=lambda j: scores[j])
+        sorted_scores = sorted(
+            range(len(scores)), key=lambda j: scores[j], reverse=True
+        )
```
erickgalinkin, line 289: Can you explain this to me a bit better? If we have a desired output, we'd want to *minimize* the perplexity  between the two strings, not *maximize* it. I ran back over the BEAST paper and I'm just wondering if I've missed something.

erickgalinkin (commented):
Would like a better explanation of the inversion in scoring. To me, this is the opposite of the intended functionality.

garak/resources/beast/beast_attack.py
jmartin-tech: @NoahOksuz sorry for the delays with this, can you offer the details needed to understand the reasoning for inverting the math here? I would like to move this forward.

NoahOksuz: @jmartin-tech im currently on Holiday. I’ll reply to everything on Tuesday..

PR #1653, opened 2026-03-27
feat: Crescendo multi-turn jailbreak probes (Russinovich et al., 2024)

Christbowel:
Implements the Crescendo multi-turn jailbreak attack (Russinovich et al., arXiv:2404.01833),
accepted at USENIX Security 2025. Crescendo is not currently in garak.

## What this adds

**`garak/probes/crescendo.py`**
- `CrescendoReplay`: pre-scripted replay probe, no auxiliary LLM required. Replays
  fixed Crescendo attack conversations turn-by-turn. `CrescendoCached` kept as a
  backwards-compatible alias. `generations` forced to 1 (pre-scripted turns have no
  variance, replaying N times is waste).
- `Crescendo`: fully adaptive probe using an attacker LLM that generates each turn
  based on the target's prior response, with a comprehensive meta-prompt including
  worked examples from the paper.

**`garak/detectors/judge.py`**
- `CrescendoJudge`: two-stage LLM judge. Primary judge scores 0-100 against the
  specific attack goal. Secondary judge activates when primary score < 70 and
  re-evaluates the primary's explanation to correct false negatives caused by the
  judge's own safety alignment — as described in the paper.

**`garak/data/crescendo/`**
- `crescendo_cached.jsonl`: 4 pre-scripted attack conversations
- `prompt_template_attack.txt`: meta-prompt with Crescendo technique description
  and worked examples
- `prompt_template_backtrack.txt`: backtrack prompt (opt-in extension)

## Parameters

- `backtrack_on_refusal` (default `False`): paper-faithful mode lets the attacker
  LLM handle refusals organically. Set `True` to enable explicit FITD-inspired
  backtracking as an extension.
- `secondary_detectors` (default `[]`): hook for cross-validation metrics such as
  Perspective API or Azure Content Safety.

## Faithfulness to paper

Faithful: adaptive attacker LLM, multi-turn escalation referencing target responses,
two-stage judge, max 10 turns, 0-100 scoring scale.

Documented divergences: meta-prompt examples are reconstructions (exact paper prompts
not published); secondary judge re-prompts instead of parsing primary's text inline;
`backtrack_on_refusal` is an opt-in extension not in the original paper.

## Verification

- `python -m garak --target_type test.Blank --probes crescendo.CrescendoReplay`
  → 26 attempts, completes in ~4s
- `python -m pytest tests/test_crescendo.py -v` → 15 passed
- `Crescendo` (adaptive) requires a configured attacker LLM via `red_team_model_type`
  and `red_team_model_name` params (default: `nim.NVOpenAIChat` /
  `mistralai/mixtral-8x22b-instruct-v0.1`)
- `CrescendoJudge` requires a configured judge LLM (inherits `ModelAsJudge` defaults)


Christbowel:

Christbowel:
recheck

garak/detectors/judge.py
```diff
@@ -167,3 +167,144 @@ def detect(self, attempt: Attempt) -> List[float | None]:
...
+    """Two-stage LLM judge for Crescendo multi-turn jailbreak (Russinovich et al., 2024).
+
+    Reference: arXiv:2404.01833
+
+    Faithful to paper:
+        - Primary judge scores 0-100 against the specific attack goal
+        - Secondary judge corrects false negatives caused by the primary
+          judge's own safety alignment, by re-evaluating the primary's
+          explanation for signs of alignment-driven refusal
+        - Goal read from attempt.notes["goal"] for per-conversation specificity
```
jmartin-tech: Is there some reason the `goal` cannot be in `attempt.goal` this seems like it would be a more clearly authoritative location.

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+        init_attempts = []
+        for conv_data in self.cached_conversations:
+            turns = conv_data.get("turns", [])
+            if not turns:
+                continue
+            attempt = self._create_attempt(turns[0])
+            if attempt.notes is None:
+                attempt.notes = {}
+            attempt.notes["cached_turns"] = turns
+            attempt.notes["turn_idx"] = 0
```
jmartin-tech, line 110: Why is this index needed? As I read this attack flow the index is simply `len(attempt.turns) - 1`.

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+    def _create_init_attempts(self) -> Iterable[garak.attempt.Attempt]:
+        init_attempts = []
+        for conv_data in self.cached_conversations:
+            turns = conv_data.get("turns", [])
+            if not turns:
+                continue
+            attempt = self._create_attempt(turns[0])
+            if attempt.notes is None:
+                attempt.notes = {}
+            attempt.notes["cached_turns"] = turns
```
jmartin-tech: Instead of holding the full cached conversation here, could this just be the index from `self.cached_conversations`?

```suggestion
            attempt.notes["cache_idx"] = cache_idx
```

Note that any data stored on the `attempt` will be serialized into the `report.jsonl` file and the repetition of cached data here may be of limited value.

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+            attempt.notes["turn_idx"] = 0
+            init_attempts.append(attempt)
+        return init_attempts
+
+    def _generate_next_attempts(
+        self, last_attempt: garak.attempt.Attempt
+    ) -> Iterable[garak.attempt.Attempt]:
+        turn_idx = last_attempt.notes.get("turn_idx", 0)
```
jmartin-tech, line 117: As noted above this could be:

```suggestion
        turn_idx = len(last_attempt.turns) - 1
```

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+        cached_turns = last_attempt.notes.get("cached_turns", [])
```
jmartin-tech:
```suggestion
        cache_idx = last_attempt.notes.get("cache_idx", None) # This should be guaranteed to exist.
        cached_turns = self.cached_conversations[cache_idx]["turns"]
```

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+    ]
+    goal = "cause the model to comply with a harmful request through gradual escalation"
+    tier = garak.probes.Tier.COMPETE_WITH_SOTA
+    active = False
+
+    DEFAULT_PARAMS = garak.probes.IterativeProbe.DEFAULT_PARAMS | {
+        "end_condition": "detector",
+        "red_team_model_type": "nim.NVOpenAIChat",
+        "red_team_model_name": "mistralai/mixtral-8x22b-instruct-v0.1",
+        "red_team_model_config": {},
```
jmartin-tech: Should this have a `max_tokens` value? The default for a `generator` is `150` tokens and many techniques tend to increase the value for technique utilized services.

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+            if opening is None:
+                logging.warning(
+                    "crescendo.Crescendo: failed to generate opening turn for goal: %s"
+                    % goal
+                )
+                continue
+            attempt = self._create_attempt(opening)
+            if attempt.notes is None:
+                attempt.notes = {}
+            attempt.notes["goal"] = goal
```
jmartin-tech: By default `_create_attempt` and the underlying `_mint_attempt` set the returned `Attempt`'s goal to the probe `self.goal`. I think in this case it might be appropriate to set the `attempt.goal` here instead of adding another `goal` in the notes.


```suggestion
            attempt.goal = goal
```

This has implications to the separate comment on the detector description about where `goal` is sourced.
Unless there is some reason both `goal` values are needed I would suggest reducing the noise here.

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+            init_attempts.append(attempt)
+        return init_attempts
+
+    def _generate_next_attempts(
+        self, last_attempt: garak.attempt.Attempt
+    ) -> Iterable[garak.attempt.Attempt]:
+        goal = last_attempt.notes.get("goal", "")
```
jmartin-tech:
```suggestion
        goal = last_attempt.goal
```

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+                    garak.attempt.Turn(
+                        role="user",
+                        content=garak.attempt.Message(text=next_turn_str),
+                    )
+                )
+
+            next_attempt = self._create_attempt(next_conv)
+            if next_attempt.notes is None:
+                next_attempt.notes = {}
+            next_attempt.notes["goal"] = goal
```
jmartin-tech:
```suggestion
            next_attempt.goal = goal
```

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+    Faithful to paper:
+        - Multi-turn escalation structure from benign to harmful
+        - Max 10 turns per conversation
+
+    Divergences:
+        - Turns are pre-scripted and do not adapt to target responses.
+          This is a replay harness, not a faithful implementation of the
+          Crescendo algorithm. Use ``Crescendo`` for the adaptive version.
+        - generations is forced to 1: pre-scripted turns have no variance,
+          replaying them N times per turn is pure waste.
```
jmartin-tech, line 62: Since this does not actually `iterate` to reject any responses, is there value added in treating the requests a part of a `Conversation`? As one-shot attacks each of the cached prompts might be viable independently. I could see including the model responses in the request context for iteration triggering the model processing to consider mitigation already provided so just wondering if there is any quantitative validation to reference to show the value in having a _cached_ attack pattern?

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+                if line:
+                    self.cached_conversations.append(json.loads(line))
+
+    def _create_init_attempts(self) -> Iterable[garak.attempt.Attempt]:
+        init_attempts = []
+        for conv_data in self.cached_conversations:
+            turns = conv_data.get("turns", [])
+            if not turns:
+                continue
+            attempt = self._create_attempt(turns[0])
```
jmartin-tech, line 105: Consider setting the `goal` from the cached conversation here, otherwise if it that `goal` is not really used in this probe it is not needed in the `jsonl` cached file.
```suggestion
            attempt = self._create_attempt(turns[0])
            attempt.goal = conv_data["goal"]
```

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+            return []
+
+        next_conv = copy.deepcopy(conversation)
+        next_conv.turns.append(
+            garak.attempt.Turn(
+                role="user",
+                content=garak.attempt.Message(text=next_turn_str),
+            )
+        )
+        next_attempt = self._create_attempt(next_conv)
```
jmartin-tech, line 140: Again ensure the `goal` from the cached conversation is propagated if is should really be in the cache.
```suggestion
        next_attempt = self._create_attempt(next_conv)
        next_attempt.goal = last_attempt.goal
```

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+        self.cached_conversations = []
+        with open(conversations_path, "r", encoding="utf-8") as f:
+            for line in f:
+                line = line.strip()
+                if line:
+                    self.cached_conversations.append(json.loads(line))
+
+    def _create_init_attempts(self) -> Iterable[garak.attempt.Attempt]:
+        init_attempts = []
+        for conv_data in self.cached_conversations:
```
jmartin-tech: Capture location to pass by index:
```suggestion
        for cache_idx, conv_data in enumerate(self.cached_conversations):
```

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+        next_conv.turns.append(
+            garak.attempt.Turn(
+                role="user",
+                content=garak.attempt.Message(text=next_turn_str),
+            )
+        )
+        next_attempt = self._create_attempt(next_conv)
+        if next_attempt.notes is None:
+            next_attempt.notes = {}
+        next_attempt.notes["cached_turns"] = cached_turns
```
jmartin-tech:
```suggestion
        next_attempt.notes["cache_idx"] = cache_idx
```

garak/probes/crescendo.py
```diff
@@ -0,0 +1,414 @@
...
+        next_attempt.notes["turn_idx"] = next_idx
```
jmartin-tech, line 145: Based on idea that this value is `len(next_attempt.turns)-1`:
```suggestion
```

garak/detectors/judge.py
```diff
@@ -167,3 +167,144 @@ def detect(self, attempt: Attempt) -> List[float | None]:
...
+    def detect(self, attempt: Attempt) -> List[float | None]:
+        goal = ""
+        if attempt.notes:
+            goal = attempt.notes.get("goal", "")
+        if not goal:
+            try:
+                probe_info = _plugins.plugin_info(f"probes.{attempt.probe_classname}")
+                goal = probe_info.get("goal", "")
+            except Exception:
+                pass
```
jmartin-tech: Attempts have a `goal` attribute that is populated with the probe's `goal` attributed by default.  As noted in other comments I suspect this can just use `attempt.goal` if the probe injects that value into the attempt before it is passed to the generator.

```suggestion
        goal = self.goal
```

garak/detectors/judge.py
```diff
@@ -167,3 +167,144 @@ def detect(self, attempt: Attempt) -> List[float | None]:
...
+
+        tested_prompt = attempt.prompt.last_message().text if attempt.prompt else ""
```
jmartin-tech: Should this judge solely based on the last `user` request or should the whole conversation be utilized?

jmartin-tech (changes requested):
Thanks for this. A terrific start, code inspection lead to some suggestions. The execution is not fully tested yet. Happy to iterate as needed to get this moved forward.

garak/probes/crescendo.py
Christbowel: Thanks for the suggestion. I double-checked the `Attempt` API and `attempt.turns` doesn't seem to exist as a direct attribute, it sits under `attempt.conversations[0].turns`. Beyond that, after sending `cached_turns[0]`, the conversation contains `[user_0, assistant_0]` (len=2), so `len(turns) - 1` would give 1 while `turn_idx` should still be 0 at that point. This would cause a skip on every iteration. I think keeping `turn_idx` explicitly in notes is the safest way to track position in `cached_turns`. Happy to revisit if I'm misreading the attack flow!

jmartin-tech: Sorry you are correct is should be `atttempt.prompt.turns`. Though I forgot to account for the `assistant` turns in as it builds.  I still believe this should not be `note` this is `state` of the attack process not really the `attempt`.

garak/probes/crescendo.py
Christbowel: The core reason is that the target model's own responses become part of the context at each turn, which is the mechanism Crescendo actually exploits. Sending everything in one shot would require fabricating those intermediate responses, which changes the attack entirely. The updated docstring now makes this explicit.

garak/detectors/judge.py
Christbowel: Used attempt.goal instead, which always holds the right value after the changes to the probe. self.goal on the detector would not carry the per-conversation goal.

garak/detectors/judge.py
Christbowel: The judge now receives the full conversation rather than just the last message. Since Crescendo builds context incrementally, evaluating the final response in isolation loses important signal about how the goal was reached.

garak/detectors/judge.py
jmartin-tech: Exactly right, I really should be more careful about using the `suggestion` syntax with untested examples. Thanks for validating and adjusting.

garak/probes/crescendo.py
jmartin-tech: Thank you for documenting the clarification.

garak/probes/crescendo.py
Christbowel: Agreed on the distinction. The constraint is that `_generate_next_attempts` only receives `last_attempt`, so the position needs to travel with it somehow. I could maintain a dict on `self` keyed by attempt UUID, but that risks memory accumulation in long runs. Would a dedicated `attack_state` dict on the attempt be cleaner, or is there an existing pattern in the codebase you'd point me to?

garak/probes/crescendo.py
Christbowel: Done, using cache_idx now.

garak/probes/crescendo.py
Christbowel: Same discussion as above, keeping turn_idx in notes for now pending your feedback on where state should live.

garak/probes/crescendo.py
Christbowel: Done, using last_attempt.goal now.

garak/probes/crescendo.py
Christbowel: Done, attempt.goal is now set from conv_data["goal"].

garak/probes/crescendo.py
Christbowel: Done, next_attempt.goal = last_attempt.goal is now propagated.

garak/probes/crescendo.py
Christbowel: Same as above, keeping turn_idx pending your feedback on state management.

jmartin-tech:
@Christbowel testing for this probe is still in flight, feedback should be ready soon.

PR #1654, opened 2026-03-28
fix: handle None response.choices in OpenAICompatible

### LINKED ISSUE #1525: bug: iteration over `NoneType` in `OpenAICompatible`
with AWS endpoint for ibm-granite/granite-4.0-h-micro:

```
failed on probes.atkgen.Tox:   8%|▊         | 2/25 [01:07<13:39, 35.65s/it]         You seem to be using the pipelines sequentially on GPU. In order to maximize efficiency please use a datasetg [attackgene]:   0%|          | 0/10 [00:00<?, ?it/s]         
  File ""/root/garak/garak-venv/bin/garak"", line 7, in <module>29<00:12,  4.30s/it] 
    sys.exit(main())
  File ""/root/garak/garak/__main__.py"", line 9, in main
    cli.main(sys.argv[1:])
  File ""/root/garak/garak/cli.py"", line 627, in main
    command.probewise_run(
  File ""/root/garak/garak/command.py"", line 259, in probewise_run
  File ""/root/garak/garak/harnesses/probewise.py"", line 109, in run
  File ""/root/garak/garak/harnesses/base.py"", line 149, in run
  File ""/root/garak/garak/probes/atkgen.py"", line 190, in probe
    response = generator.generate(this_attempt.prompt)
  File ""/root/garak/garak/generators/base.py"", line 167, in generate
    outputs = self._call_model(prompt, 1)
  File ""/root/garak/garak-venv/lib/python3.10/site-packages/backoff/_sync.py"", line 105, in retry
    ret = target(*args, **kwargs)
  File ""/root/garak/garak/generators/openai.py"", line 296, in _call_model
    return [Message(c.message.content) for c in response.choices]
TypeError: 'NoneType' object is not iterable
```

with Azure endpoint:

```
stopped in the middle of injectascii85 with 

  File ""/root/garak/garak-venv/bin/garak"", line 7, in <module>
    sys.exit(main())
  File ""/root/garak/garak/__main__.py"", line 9, in main
    cli.main(sys.argv[1:])
  File ""/root/garak/garak/cli.py"", line 627, in main
    command.probewise_run(
  File ""/root/garak/garak/command.py"", line 259, in probewise_run
  File ""/root/garak/garak/harnesses/probewise.py"", line 109, in run
  File ""/root/garak/garak/harnesses/base.py"", line 149, in run
  File ""/root/garak/garak/probes/base.py"", line 444, in probe
  File ""/root/garak/garak/probes/base.py"", line 355, in _execute_all
    result = self._execute_attempt(this_attempt)
  File ""/root/garak/garak/probes/base.py"", line 297, in _execute_attempt
  File ""/root/garak/garak/generators/base.py"", line 170, in generate
    outputs = self._call_model(prompt, generations_this_call)
  File ""/root/garak/garak-venv/lib/python3.10/site-packages/backoff/_sync.py"", line 105, in retry
    ret = target(*args, **kwargs)
  File ""/root/garak/garak/generators/openai.py"", line 296, in _call_model
    return [Message(c.message.content) for c in response.choices]
TypeError: 'NoneType' object is not iterable
```

Christbowel:
## What this fixes

`response.choices` can be `None` when endpoints such as AWS Bedrock or Azure
return HTTP 200 with a filtered or refused response. The existing guard only
checked for the absence of the `.choices` attribute via `hasattr()`, so a
`None` value passed through and caused an unhandled `TypeError` that crashed
the entire garak run.

One-line fix: extend the condition to also catch `response.choices is None`,
reusing the existing retry/fallback logic already in place.

Fixes #1525

## Verification

- Reproducing the exact crash requires an AWS or Azure endpoint that returns
  `choices: null`. The fix is a one-line guard on an already-tested code path.
- `python -m pytest tests/generators/ -v` passes.
- `python -m garak --target_type test.Blank --probes dan.Dan_11_0` runs cleanly.

### DIFF
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -327,7 +327,7 @@ class OpenAICompatible(Generator):
             else:
                 raise e
 
-        if not hasattr(response, "choices"):
+        if not hasattr(response, "choices") or response.choices is None:
             logging.debug(
                 "Did not get a well-formed response, retrying. Expected object with .choices member, got: '%s'"
                 % repr(response)
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -327,7 +327,7 @@ class OpenAICompatible(Generator):
             else:
                 raise e
 
-        if not hasattr(response, "choices") or response.choices is None:
+        if not hasattr(response, "choices"):
             logging.debug(
                 "Did not get a well-formed response, retrying. Expected object with .choices member, got: '%s'"
                 % repr(response)
@@ -337,6 +337,12 @@ class OpenAICompatible(Generator):
                 raise garak.exception.GeneratorBackoffTrigger(msg)
             else:
                 return [None]
+        if response.choices is None:
+            logging.debug(
+                "Did not get a well-formed response, Expected object with populated .choices member, got: '%s'"
+                % repr(response)
+            )
+            return [None]
 
         if is_completion:
             reponse_message_list = [Message(c.text) for c in response.choices]
--- a/tests/_assets/generators/openai.json
+++ b/tests/_assets/generators/openai.json
@@ -63,6 +63,7 @@
          }
       }
    },
+   
    "chat" : {
       "code" : 200,
       "json" : {
@@ -136,5 +137,16 @@
          ],
          "object" : "list"
       }
+   },
+   "azure_chat_null_choices": {
+      "code": 200,
+      "json": {
+         "choices": null,
+         "created": 1724052469,
+         "id": "chatcmpl-9xr5pU1EE4XQw9Bd0QOgkFZ82cAS9",
+         "model": "gpt-4o-2024-05-13",
+         "object": "chat.completion"
+      }
    }
 }
+
--- a/tests/generators/test_azure.py
+++ b/tests/generators/test_azure.py
@@ -73,3 +73,20 @@ def test_azureopenai_chat(respx_mock, openai_compat_mocks):
     assert len(output) == 1
     for item in output:
         assert isinstance(item, Message)
+
+
+@pytest.mark.usefixtures("set_fake_env")
+@pytest.mark.respx(base_url="https://garak.example.com/")
+def test_azureopenai_chat_null_choices(respx_mock, openai_compat_mocks):
+    mock_response = openai_compat_mocks["azure_chat_null_choices"]
+    extended_request = "openai/deployments/"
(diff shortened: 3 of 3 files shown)


leondz:
Thank you. Explanation is illuminating. Taking a look

jmartin-tech (approved):
This looks sound, it would be nice see a test with a mocked response that validates the behavior.

The existing azure mocks [here](https://github.com/NVIDIA/garak/blob/dcda2be55daa4e841fe7915744bad6d051d0bdf8/tests/_assets/generators/openai.json) could be expanded.

garak/generators/openai.py
```diff
@@ -327,7 +327,7 @@ def _call_model(
             else:
                 raise e
 
-        if not hasattr(response, "choices"):
+        if not hasattr(response, "choices") or response.choices is None:
```
jmartin-tech, line 330: How does this really present? Does a retry eventually result in a valid _choices_ array on a repeated inference request?

Testing via a mocked response shows this results in infinite backoff if the service continues to return `response.chioces = None`, as the default for `retry_json` is true.

One possible path if repeated request will continue to fail indefinitely:
```Python
        if not hasattr(response, "choices"):
            if  response.choices is None:
                logging.debug(
                    "Did not get a well-formed response, retrying. Expected object with populated .choices member, got: '%s'"
                    % repr(response)
                )
                return [None]
```

[Here](https://github.com/NVIDIA/garak/commit/962c5c434c78d3579e863d6268e9244ee7e8e0a9) is an example of adding a mocked response test for azure.

jmartin-tech (changes requested):

ChrisJr404:
Hey @Christbowel — @jmartin-tech pointed me here after I'd taken an independent stab at the same fix in #1737. Looks like the open discussion is about `retry_json` looping forever when `response.choices` keeps coming back None, which I missed in my version too. The shape @jmartin-tech suggested (log + `return [None]` so the retry path can't infinite-loop) makes sense, and the azure mocked-response test he linked is a nice pattern to mirror.

If you're not actively pushing on this, happy to take a swing at the follow-up commit and ping you to take a look. No worries either way if you've got it covered.

Christbowel:
Updated : separated the two cases as suggested. Missing `.choices` attribute still triggers the backoff retry, while `choices=None` now returns `[None]` immediately to avoid infinite looping. Added a mocked response test for the null choices case.

jmartin-tech (approved):

PR #1662, opened 2026-03-31
docs: structure docs out of flat dir into tree

leondz:
Flat structure isn't working for docs any more. This PR structures them into a tentative dir setup.

* Subdirs for categories having 3+ entries
* File naming removes `garak.` prefix (they're all about garak)
* category index files prefixed `index_`

This is a rough structure, feedback welcome

To test:
- [ ] make -C docs/source


jmartin-tech (commented):
Sorry scope creep comment, should we match code structure file name convention, ex: `detectors/always.rst` instead of `detectors/detectors.always.rst`?

leondz:
Very in scope. I thought the same. Landed on this format (`type.module`) because some plugin modules of different have the same name, so including the type deconflicts the individual filename. On the other hand, the path is enough of a descriptor, and it works fine for the plugins themselves..

Actually, having now said that, maybe it would make sense to mirror the source dir structure??

jmartin-tech:
I agree matching the source dir structure seems intuitive.

mikemckiernan:
I was just thinking about you folks a few hours ago!

The downside to directories--unless you cooked up something different--is that typically the directories beg for an index.md file in the directory so that a human at generators/generators.nim.html figures removing the page part of the URL should go to some navigation.  An example of that navigation approach is at https://docs.nvidia.com/nemo/microservices/latest/audit/targets/index.html.

Also, you can make the navigation in the TOC more manageable without the directories, in case that was the true goal.

leondz:
Thanks @mikemckiernan for the tips!

> Also, you can make the navigation in the TOC more manageable without the directories,

This wasn't the primary goal but would appreciate input/guidance here, there's room for improvement for sure

docs/source/index_detectors.rst
```diff
@@ -0,0 +1,34 @@
+Detectors
```
mikemckiernan, line 1: Apologies for the lack of clarity about the index files.  I meant to suggest that if you want the directories, that it is kindest to make a docs/source/detectors/index.rst with contents like this, but with revised local paths.

The point is that you'd want something to appear if someone types their way to <https://reference.garak.ai/en/latest/detectors/>.

mikemckiernan:
> Thanks @mikemckiernan for the tips!
> 
> > Also, you can make the navigation in the TOC more manageable without the directories,
> 
> This wasn't the primary goal but would appreciate input/guidance here, there's room for improvement for sure

Sorry for commenting and then silence for > 24hrs.  I just took another look and what's in main is what I meant to recommend:

image

It requires folks to click to discover content, but the information is available without expanding the TOC enormously.

mikemckiernan (commented):
Another disadvantage to the directories is consideration for any bookmarked content or current references to these docs.  (I'm fairly sure that I made a few.)  Please consider using sphinx-reredir if you truly prefer the directories.

docs/source/index_detectors.rst
leondz: got it!

leondz: fix pushed

docs/source/index.rst
```diff
@@ -48,22 +48,22 @@ Check out the :doc:`usage` section for further information, including :doc:`inst
...
    interactive
    langservice
    payloads
-   probes
+   index_probes
    report
    _config
    _plugins
-   analyze
+   index_analyze
```
jmartin-tech: The order or the index header entires need a bit of adjustment.

Current order and revisions results in the following list:
```
Code Reference

Key Concepts and Classes
garak.attempt
Buffs
garak.cli
garak.command
Detectors
Evaluators
garak.exception
Generators
Harnesses
garak.interactive
garak.langservice
garak.payloads
Probes
garak.report
garak._config
garak._plugins
Run analysis
```

I could see either adjusting the headers in the index files to keep the `garak.` prefix resulting in:
```
Code Reference

Key Concepts and Classes
garak.attempt
garak.buffs
garak.cli
garak.command
garak.detectors
garak.evaluators
garak.exception
garak.generators
garak.harnesses
garak.interactive
garak.langservice
garak.payloads
garak.probes
garak.report
garak._config
garak._plugins
garak.analyze
```

Or adjusting the order for have directory indexes in alpha order first then direct files:
```
Code Reference

Key Concepts and Classes
Analysis
Buffs
Detectors
Evaluators
Generators
Harnesses
Probes
garak.attempt
garak.cli
garak.command
garak.exception
garak.interactive
garak.langservice
garak.payloads
garak.report
garak._config
garak._plugins
```

Note the both proposed indexes change the header in `index_analyze.rst`.

docs/source/cliref.rst
```diff
@@ -47,53 +50,51 @@ CLI reference for garak
                           How many probe attempts to launch in parallel. Raise
                           this for faster runs when using non-local models.
     --skip_unknown        allow skip of unknown probes, detectors, or buffs
-    --seed SEED, -s SEED  random seed
+    --seed, -s SEED       random seed
     --deprefix            remove the prompt from the front of generator output
     --eval_threshold EVAL_THRESHOLD
                           minimum threshold for a successful hit
-    --generations GENERATIONS, -g GENERATIONS
+    --generations, -g GENERATIONS
```
jmartin-tech, line 57: Weird testing artifact, these changes to `cliref.rst` are being generated differently on at least on macOS vs what is added in the PR.

jmartin-tech (changes requested):
Testing shows this adds a new dependency that should be in docs requirements:

```diff
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -1,3 +1,4 @@
 sphinx
 sphinx-rtd-theme
-sphinx-github-style
\ No newline at end of file
+sphinx-github-style
+sphinx-reredirects
```

Note also the automation test that ensures new class files have `rst` doc files needs to be updated as well.

docs/source/cliref.rst
leondz: ah, missed this! yeah. cause of this churn happening now is that docs `make all` rebuilds `cliref.rst`. cause of the content .. could also be version-specific.

slightly out-of-scope: how much do we care about churn in `cliref.rst`? i wonder if it's worth applying some kind of gitignore pattern over, seeing as the file is autogenerated

docs/source/index.rst
leondz: Good catch.

I think this can stand to be placed in two groups - the first one being plugins, the second being everything else. This can be explicit.

Titles should be fixed up also, thanks

leondz: updated

mikemckiernan (approved):
I'm still neutral on adding directories.  imo, no great benefit to humans or machines.

I just haven't been able to give this project the attention it deserves.  I attempted using sphinx-autodoc2 for the API section, but it wasn't an improvement.  I wanted to explore either generating the API pages automatically--I'm still unsure on that one.  The second area for experimentation, is possibly removing pages from the TOC so it's more approachable, and using things like cards and grids on navigation pages for the API to see if that _feels_ better.

Finally, if folks aren't happy with cliref.rst, there's the sphinx-argparse extension to create it from the Python file.  iirc, I gave it a shot, but wasn't sure of the appetite for a slight refactor of the code just to produce the docs from the code.

I want to help more, I just haven't been able to get the time.

docs/source/probes/base.rst
```diff
@@ -1,9 +1,9 @@
-garak.probes.base
-=================
+Base Probes
+===========
```
jmartin-tech: Revert or make consistent for `base` module in all plugin types.  This also relates to the ordering in the index.

While I understand the reasoning for using a different title format here, would suggest this just creates inconsistency that may lead to confusion for readers. I would offer that overview descriptions in the `*_index.rst` already points the user to this class and the title here actually makes the TOC list less intuitive.

Image
vs
Image

leondz:
@mikemckiernan 

> I'm still neutral on adding directories. imo, no great benefit to humans or machines.

The flat list of hundreds of files was becoming unpleasant to navigate, which prompted the move. Thanks for the reminder to add redirects, though; did that and took the opportunity to add a 404-handler as well, which we didn't have before.

> if folks aren't happy with cliref.rst, there's the sphinx-argparse extension to create it from the Python file. iirc, I gave it a shot, but wasn't sure of the appetite for a slight refactor of the code just to produce the docs from the code.

Hadn't seen this, will take a look next time `cliref`-related churn rears its head!

> I wanted to explore either generating the API pages automatically--I'm still unsure on that one. The second area for experimentation, is possibly removing pages from the TOC so it's more approachable, and using things like cards and grids on navigation pages for the API to see if that feels better.

This could be cool. Definitely sounds like a broader project with scope bigger than this PR. I am excited about exploring this, I'm sure we'll work out how & when to slate it eventually.

jmartin-tech (approved):

PR #1664, opened 2026-04-01
feat: Add HarmActionsEval for evaluation of action-level safety in AI agents

prane-eth:
## Background

AI agents have a growing adoption across the industry, including critical applications. AI agents that have access to tools can currently call tools directly with no centralized validation layer that inspects these calls before execution, allowing harmful or disallowed tool calls to be executed without oversight. Existing agent benchmarks evaluate the safety of final responses instead of  actions.

[HarmActionsEval benchmark](https://github.com/Pro-GenAI/Agent-Action-Guard) evaluates actions. It found that 80% of the LLMs tested executed actions at the first attempt for over 95% of the harmful prompts.

Related work: https://github.com/Pro-GenAI/Agent-Action-Guard. It received 962 downloads on PyPI, and 247 clones on GitHub in the first week.

## Changes

I integrated HarmActionsEval benchmark.

All the test cases passed using: `pytest tests/probes/test_probes_harmactionseval.py tests/detectors/test_detectors_harmactionseval.py`
and `pytest tests/detectors/test_detectors.py::test_detector_detect[detectors.harmactionseval.ToolCallMatch]`.

Example command used:
```bash
garak --target_type openai \
	--target_name "$OPENAI_MODEL" \
	--probes harmactionseval.HarmActionsEval \
	--generations 1
```

## Verification

- [x] Supporting configuration such as generator configuration file. -- Not applicable. I added the dataset.
- [x] `garak -t <target_type> -n <model_name>`

### DIFF
--- a/docs/source/detectors.rst
+++ b/docs/source/detectors.rst
@@ -17,6 +17,7 @@ garak.detectors
    garak.detectors.exploitation
    garak.detectors.fileformats
    garak.detectors.goodside
+   garak.detectors.harmactionseval
    garak.detectors.judge
    garak.detectors.knownbadsignatures
    garak.detectors.leakreplay
--- a/docs/source/garak.detectors.harmactionseval.rst
+++ b/docs/source/garak.detectors.harmactionseval.rst
@@ -1,0 +1,7 @@
+garak.detectors.harmactionseval
+=============================
+
+.. automodule:: garak.detectors.harmactionseval
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/docs/source/garak.probes.harmactionseval.rst
+++ b/docs/source/garak.probes.harmactionseval.rst
@@ -1,0 +1,9 @@
+garak.probes.harmactionseval
+===========================
+
+.. automodule:: garak.probes.harmactionseval
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+   .. show-asr::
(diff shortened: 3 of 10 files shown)


prane-eth:

leondz:
This is interesting, thanks.

Generally garak is not a benchmark and we try to avoid including too many benchmarks. The benchmark data here has a few issues that need to be addressed - some are a bit non-trivial

* What's the provenance of the data? Where did this text come from?
* We try to separate technique from intent with new contributions to garak. This gives a lot of flexibility. Techniques are things like letter substitutions, the "how" of the attack. Intents are the bad thing get the model to do, the "what" of the attack. This contribution combines the two without much separation. We'd like to see these two factors separated cleanly in a probe from this data, perhaps even with multiple probes each implementing a novel technique, and taking the same set of adversarial prompts.
* While the outcomes of this test are labelled with equal weight, they are not weighted equally. E.g. there are items asking for the weather, with others asking for illegal instruction. This doesn't make for a good benchmark - failing one or the other gives the same score metric, but not the same level of risk. That's got to be addressed.
* `classification` and `risk_level` need a rigorous supporting rubric before we can know that they're serious
* While the benchmark contains many imperatives that might suggest an action being taken, how does the probe know that this is actually happening? If we ask a model with no access to agents these things, there's a good chance of getting a response that looks the same, despite the agent call not happening

Please review the above and help us assess whether or not this is ready for garak integration. Converting to draft pending

PR #1671, opened 2026-04-05
fix: use magic.from_file() instead of from_buffer() for libmagic 5.47 compatibility

HayatoFujihara:
## Summary

- Replace `magic.from_buffer()` with `magic.from_file()` in `FileIsExecutable._test_file()` to fix detection failures on libmagic 5.47
- Revert the temporary libmagic 5.46 version pin in macOS CI workflow (PR #1635), restoring `brew install libmagic`

Closes #1634

## Root Cause

libmagic 5.47 changed `from_buffer()` behavior — it now returns `application/octet-stream` for binary file snippets that were previously identified correctly. This affects all binary test assets (PE, ELF, Mach-O, batch) except shell scripts. Increasing the buffer size does not help, as `from_buffer()` with the full 4096-byte snippet still fails.

`from_file()` lets libmagic open the file and read as much as it needs internally, producing correct MIME types on both 5.45 and 5.47. Since `_test_file()` always receives a file path, `from_file()` is the more natural fit.

## Verification

Tested with Docker (libmagic 5.47 built from source) and local environment (libmagic 5.45):

| File | from_buffer (before) on 5.47 | from_file (after) on 5.47 |
|---|---|---|
| setuptools-top4k.exe | `application/octet-stream` | `application/vnd.microsoft.portable-executable` |
| python-elf-top4k | `application/octet-stream` | `application/x-executable` |
| libssl3-top4k.so | `application/octet-stream` | `application/x-sharedlib` |
| grep-mach-o-top4k | `application/x-java-applet` | `application/x-mach-binary` |
| batch.bat | `text/plain` | `text/x-msdos-batch` |
| shell.sh | `text/x-shellscript` | `text/x-shellscript` |

All 6 assets pass on both libmagic 5.45 and 5.47. No regressions.

### DIFF
--- a/garak/detectors/fileformats.py
+++ b/garak/detectors/fileformats.py
@@ -6,6 +6,7 @@
 These detectors examine file formats, based on name or content."""
 
 import logging
+import os
 import pickletools
 import tempfile
 
@@ -106,8 +107,10 @@ class FileIsExecutable(FileDetector):
         m = self.magic.Magic(mime=True)
         with open(filename, "rb") as f:
             header = f.read(8192)
-        with tempfile.NamedTemporaryFile() as tmp:
+        with tempfile.NamedTemporaryFile(mode="wb", delete=False) as tmp:
             tmp.write(header)
             tmp.flush()
+            tmp.close()
             mimetype = m.from_file(tmp.name)
+            os.remove(tmp.name)
         return 1.0 if mimetype in self.exec_types else 0.0
--- a/garak/detectors/fileformats.py
+++ b/garak/detectors/fileformats.py
@@ -102,8 +102,6 @@ class FileIsExecutable(FileDetector):
     def _test_file(self, filename):
         if self.magic is None:
             return None
-        with open(filename, "rb") as f:
-            m = self.magic.Magic(mime=True)
-            header = f.read(2048)
-            mimetype = m.from_buffer(header)
-            return 1.0 if mimetype in self.exec_types else 0.0
+        m = self.magic.Magic(mime=True)
+        mimetype = m.from_file(filename)
+        return 1.0 if mimetype in self.exec_types else 0.0
--- a/garak/detectors/fileformats.py
+++ b/garak/detectors/fileformats.py
@@ -7,6 +7,7 @@ These detectors examine file formats, based on name or content."""
 
 import logging
 import pickletools
+import tempfile
 
 from garak.detectors.base import FileDetector
 
@@ -103,5 +104,10 @@ class FileIsExecutable(FileDetector):
         if self.magic is None:
             return None
         m = self.magic.Magic(mime=True)
-        mimetype = m.from_file(filename)
+        with open(filename, "rb") as f:
+            header = f.read(8192)
+        with tempfile.NamedTemporaryFile() as tmp:
+            tmp.write(header)
+            tmp.flush()
+            mimetype = m.from_file(tmp.name)
         return 1.0 if mimetype in self.exec_types else 0.0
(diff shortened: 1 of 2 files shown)


HayatoFujihara:

HayatoFujihara:
recheck

jmartin-tech (changes requested):
Thank you for taking a pass a this.

This is not a long term solution for this project. The buffer usage is intentional, as targets this project may work with may not always have a `file` on disk to work with and files involved in these detections are also often very large, consider many models have multiple 5GB files. This change makes the current detector technically meet compatibility with libmagic 5.47 however this usage pattern is not the desired goal.

One idea that has been floated to mitigate the _very large file_ issue would be to read a buffer of data and write it to a temporary file then call `m.from_file()` against the temporary file to see if that would produce compatibility with the existing detection patterns.

Another is to adjust the detection to support producing a hit on any mimetype reported to start with `application` even if the full type is not found in the `exec_types` list.

HayatoFujihara:
Thank you for the thorough review and the concrete suggestions — they were very helpful in understanding the design intent behind the buffer-based approach.

I've updated the implementation to follow your first suggestion: reading a buffer of data and writing it to a temporary file, then calling `from_file()` against it. This preserves the intentional buffer-read optimization for large files while restoring compatibility with libmagic 5.47.

I verified the fix on libmagic 5.47 (Arch Linux Docker). The original `from_buffer()` approach misidentifies 5 out of 6 test assets (returning `application/octet-stream` or incorrect types), while the temporary file approach correctly identifies all of them.

As for the second suggestion (matching on `application` prefix), I chose not to pursue that direction in this PR, as it could introduce false positives for non-executable types such as `application/pdf`, `application/json`, and `application/zip`.

garak/detectors/fileformats.py
```diff
@@ -102,8 +103,11 @@ def _load_deps(self, deps_override=None):
...
         with open(filename, "rb") as f:
-            m = self.magic.Magic(mime=True)
-            header = f.read(2048)
-            mimetype = m.from_buffer(header)
-            return 1.0 if mimetype in self.exec_types else 0.0
+            header = f.read(8192)
+        with tempfile.NamedTemporaryFile() as tmp:
+            tmp.write(header)
+            tmp.flush()
+            mimetype = m.from_file(tmp.name)
```
jmartin-tech:
```suggestion
        with tempfile.NamedTemporaryFile(mode="wb", delete=False) as tmp:
            tmp.write(header)
            tmp.flush()
            tmp.close()
            mimetype = m.from_file(tmp.name)
            os.remove(tmp.name)
```

garak/detectors/fileformats.py
```diff
@@ -7,6 +7,7 @@
 
 import logging
 import pickletools
+import tempfile
```
jmartin-tech, line 11:
```suggestion
import os
import tempfile
```

jmartin-tech (changes requested):
Testing highlights there are permissions issues with the temporary file approach on windows. These need to be ironed out to more forward.

Based on testing this is likely related to filesystem locking, the suggestions offered have bad basic testing but should be validated.

HayatoFujihara:
Thank you for the review and the detailed suggestions! I've applied both changes:

1. Added `import os`
2. Updated `NamedTemporaryFile` to use `delete=False` + explicit `close()` + `os.remove()` to avoid the Windows filesystem locking issue

This now follows the same pattern used throughout the project (e.g., `tests/buffs/test_buff_config.py`, `tests/test_config.py`).

jmartin-tech (approved):
Thank you for working through the changes, this looks good.

PR #1672, opened 2026-04-05
fix: report_digest -t taxonomy option not working

### LINKED ISSUE #1592: post processing report digest `-t` option does not group as expected.
When utilizing a report that was generated for a specific taxonomy if the report digest needs to be rebuilt the taxonomy is not taken into account.

## Steps to reproduce

Basic configuration as `taxonomy.yaml`:
``` yaml
---
system:
  parallel_attempts: 16
  lite: true
run:
  generations: 1
  probe_tags: avid-effect:performance

plugins:
  # Target model configuration
  target_name: Granite-3.3-8B-Instruct
  generators:
    openai:
      OpenAICompatible:
        uri: "<redacted>/v1"
        api_key: "dummy"
        suppressed_params:
          - "n"
        max_tokens: 512

reporting:
  taxonomy: avid-effect # or 'owasp'
```
Update uri, target_name, and api_key as needed.

[ ] `garak --config taxonomy.yaml`
[ ] reprocess the generated report 
 * remove the last line `digest` from the jsonl report 
 * `python -m garak.analyze.report_digest -w -r <report.jsonl> -t avid-effect -o rebuild.html`

## Expected behavior

The reprocessed report should be grouped and organized the same as the original.

## Current behavior

The reprocessed report is grouped by probe family.

### garak version

0.14.0.pre1

HayatoFujihara:
## Summary

Fixes #1592

The `-t` option in `python -m garak.analyze.report_digest -t avid-effect` was silently ignored, always producing probe-family grouping instead of taxonomy-based grouping.

## Root Cause

In the `__main__` block of `report_digest.py`, the CLI argument `args.taxonomy` was never written to `_config.reporting.taxonomy` before calling `build_digest()`. Since `build_digest()` reads its taxonomy from `_config.reporting.taxonomy`, it always saw `None` and fell back to probe-family grouping.

## Changes

All changes are scoped to the `__main__` block of `report_digest.py`. The `build_digest()` API is unchanged.

1. **Propagate `-t` to `_config`**
   Set `_config.reporting.taxonomy = taxonomy` before calling `build_digest()`.

2. **Inherit taxonomy from report setup when `-t` is not specified**
   When `-t` is omitted, read `reporting.taxonomy` from the report's `start_run setup` entry. This ensures reports originally generated with a taxonomy are rebuilt with the same grouping by default.

3. **Support `-t None` to explicitly clear taxonomy**
   Passing `-t None` clears the taxonomy and reverts to probe-family grouping, even when the original report was generated with a taxonomy.

4. **Rebuild digest when `-t` is specified**
   When `-t` is provided, ignore any existing digest entry and regenerate. This removes the need to manually delete the digest line before reprocessing.

## Test plan

Verified 7 scenarios covering combinations of `-t` flag, setup taxonomy, and existing digest:

| # | Scenario | Expected | Result |
|---|---|---|---|
| 1 | `-t avid-effect`, setup=None, no digest | taxonomy | PASS |
| 2 | no `-t`, setup=None, no digest | probe family | PASS |
| 3 | no `-t`, setup=avid-effect, no digest | taxonomy | PASS |
| 4 | `-t None`, setup=avid-effect, no digest | probe family | PASS |
| 5 | `-t avid-effect`, setup=None, with digest | taxonomy | PASS |
| 6 | `-t None`, setup=avid-effect, with digest | probe family | PASS |
| 7 | no `-t`, setup=None, with digest | probe family | PASS |

- [x] All 122 existing tests in `tests/analyze/` pass

### DIFF
--- a/garak/analyze/report_digest.py
+++ b/garak/analyze/report_digest.py
@@ -564,7 +564,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--taxonomy",
         "-t",
-        help="Optional taxonomy to use for grouping probes",
+        help="Optional taxonomy to use for grouping probes (use 'None' to explicitly clear)",
     )
 
     args = parser.parse_args()
@@ -573,9 +573,17 @@ if __name__ == "__main__":
     output_path = args.output_path
     write_digest_suffix = args.write_digest_suffix
     taxonomy = args.taxonomy
+    # Allow "-t None" to explicitly clear taxonomy back to probe family grouping
+    taxonomy_specified = taxonomy is not None
+    if taxonomy is not None and taxonomy.lower() == "none":
+        taxonomy = None
+
+    # Propagate CLI taxonomy to _config so build_digest picks it up
+    _config.reporting.taxonomy = taxonomy
 
     digest = _get_report_digest(report_path)
-    if not digest:
+    if not digest or taxonomy_specified:
+        # Rebuild digest when taxonomy is specified, even if one already exists
         digest = build_digest(report_path)
         if write_digest_suffix:
             with open(report_path, "a+", encoding="utf-8") as reportfile:
--- a/garak/analyze/report_digest.py
+++ b/garak/analyze/report_digest.py
@@ -491,6 +491,7 @@ def build_digest(report_filename: str, config=_config):
 
     _close_result_db(conn)
 
+    report_digest["meta"]["setup"]["reporting.taxonomy"] = taxonomy
     report_digest["meta"]["calibration_used"] = calibration_used
     report_digest["meta"]["aggregation_unknown"] = aggregation_unknown
     if calibration_used:
--- a/garak/analyze/report_digest.py
+++ b/garak/analyze/report_digest.py
@@ -578,6 +578,16 @@ if __name__ == "__main__":
     if taxonomy is not None and taxonomy.lower() == "none":
         taxonomy = None
 
+    # If -t not specified, inherit taxonomy from the original report's setup
+    if not taxonomy_specified:
+        with open(report_path, "r", encoding="utf-8") as f:
+            for line in f:
+                if line.strip():
+                    entry = json.loads(line.strip())
+                    if entry["entry_type"] == "start_run setup":
+                        taxonomy = entry.get("reporting.taxonomy") or None
+                        break
+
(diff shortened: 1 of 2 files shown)


leondz:
Thanks @HayatoFujihara , taking a look

garak/analyze/report_digest.py
jmartin-tech, line 602: One minor enhancement that likely is needed, when the `-w` flag is passed allowing for persistence of the digest update into the original input file the `start_run setup` entry at the beginning of the file would also be expected to mutate to be consistent with the taxonomy option if it is being changed.

jmartin-tech (changes requested):
Awesome contribution!

A minor edge case has been identified.

It would also be great to see a test added that consumes a mock sample report from `tests/_assets/analyze` to validate a revised digest respecting this value is generated. This will help prevent any future regressions of this functionality.

Another minor ask, please revise commit messages to [include signoff](https://github.com/NVIDIA/garak/blob/2569a6062699ad1a03610156da561679eea48292/CA_DCO.md#signing).

jmartin-tech:
Another finding during testing, when a new taxonomy is requested it should be reflected in the final `digest['meta']['setup']['reporting.taxonomy`]` value.  Currently this is not propagating into the final digest dictionary.

HayatoFujihara:
Thank you for the detailed review and the kind words! Here's what I've addressed:

### 1. Reflect taxonomy in `meta.setup["reporting.taxonomy"]`

Updated `build_digest()` to overwrite `meta.setup["reporting.taxonomy"]` with the active taxonomy value. This applies consistently across all code paths, including when writing to file with `-w`.

### 2. Tests

Added two tests to `tests/analyze/test_analyze.py`:
- With taxonomy specified → verifies `"avid-effect"` is reflected
- Without taxonomy → verifies value remains `None`

Both use the existing mock report (`tests/_assets/analyze/test.report.jsonl`).

### 3. Signoff

Added `Signed-off-by` starting from this commit.

garak/analyze/report_digest.py
jmartin-tech: Maintainer discussion has decided this is an acceptable state, the utility should maintain the original setup entry and only update the digest when a `report.jsonl` is post processed for report grouping.

jmartin-tech (approved):

Issue #1323, opened 2025-08-06
Continue Cohere support updates


jmartin-tech:
#1252 update the library dependency for `cohere`, further updates to convert all request to use the new `ClientV2` objects are needed.

Additionally the following header warning is seen with the default `command` model selected.

```
"x-api-warning": "model 'command' is deprecated. Please consider upgrading to a newer model to avoid future service disruptions"
```

And when utilizing the legacy v1 `Client` object responses to `generate()` also include a header warning is reported:
```
"x-api-warning": "unknown field: parameter 'model' is not a valid field"
```

jmartin-tech:
Additional recommended revision of note, as the generator has been revised to `supports_multiple_generations = False` the code in the class can be significantly simplified to remove accounting for `COHERE_GENERATION_LIMIT` as it will no longer apply to an implementation what focuses on usage of `ClientV2`.

PR #1673, opened 2026-04-05
fix: modernize Cohere generator (#1384)

### LINKED ISSUE #1384: generator: update `cohere`
* [ ] drop using `command` model by default, and instead require a name (mode unavail from 15 sep 2025)
* [ ] cancel run on endpoint 404
* [ ] move to new `cohere.errors` when backing off, instead of low-granularity `cohere.error.ApiError`
* [ ] have tests verify some output comes in expected chat format from live endpoint, if `api_key` is present
* [ ] have tests verify some output comes in expected completion format from live endpoint, if `api_key` is present
* [ ] preserve `None`s in output as per current protocol
* [ ] move to `_load_client` pattern

HayatoFujihara:
## Summary

Fixes #1384

Addresses all seven checklist items in #1384.

- **Drop default model**: `name` parameter is now required; passing no model name raises `TargetNameMissingError`. The deprecated `command` default has been removed.
- **Cancel run on 404**: Both v1 and v2 API paths now catch `cohere.errors.NotFoundError` and raise `BadGeneratorException` to stop execution immediately, following the pattern in `nvcf.py`.
- **Migrate to `cohere.errors`**: Replaced the broad `RuntimeError` catch in the v1 path with fine-grained `cohere.errors` types (`NotFoundError`, `GatewayTimeoutError`, `TooManyRequestsError`, `ServiceUnavailableError`, `InternalServerError`). The v2 path already used these.
- **Live endpoint tests (chat)**: Added `test_live_chat_api_output_format` — verifies that the v2 chat API returns a `Message` object with non-empty text. Skipped when `COHERE_API_KEY` is not set.
- **Live endpoint tests (completion)**: Added `test_live_v1_api_rejects_new_model` — Cohere has retired the generate endpoint and all current models return 404 on v1, so the test verifies that the 404 handling correctly raises `BadGeneratorException` rather than silently returning `None`.
- **Preserve `None`s**: Error paths continue to emit `None` per the existing protocol. `_call_model` wraps `str` results in `Message` while passing `None` through unchanged.
- **`_load_client` pattern**: Extracted client initialization into `_load_client()`, called from `__init__` and `__setstate__`. Added `_unsafe_attributes = ["generator"]` so pickle round-trips correctly restore the client.

## Test plan
- [ ] `pytest tests/generators/test_cohere.py` — 17 tests (15 mock + 2 live)
- [ ] `ruff check garak/generators/cohere.py tests/generators/test_cohere.py`
- [ ] Live tests require `COHERE_API_KEY`; safely skipped in CI when absent

### DIFF
--- a/garak/generators/cohere.py
+++ b/garak/generators/cohere.py
@@ -1,21 +1,19 @@
 """Cohere AI model support
 
-Support for Cohere's text generation API. A model name must be specified
-(e.g. 'command-r-plus'). The legacy 'command' model default has been removed
-as Cohere has deprecated it. Set the COHERE_API_KEY environment variable.
+Support for Cohere's text generation API using ClientV2 chat interface.
+A model name must be specified (e.g. 'command-r-plus'). The legacy 'command'
+model default has been removed as Cohere has deprecated it.
+Set the COHERE_API_KEY environment variable.
 
 NOTE: As of Cohere v5.0.0+, the generate API is legacy and chat API is recommended.
-This implementation follows Cohere's official migration guide:
-
-- For v1 API: Uses cohere.Client() to maintain full backward compatibility
-- For v2 API: Uses cohere.ClientV2() for the recommended chat interface
+This implementation uses cohere.ClientV2() exclusively.
 """
 
 import logging
+import warnings
 from typing import List, Union
 
 import backoff
-import tqdm
 
 from garak import _config
 from garak.attempt import Message, Conversation
@@ -26,19 +24,13 @@ from garak.exception import (
 )
 from garak.generators.base import Generator
 
-COHERE_GENERATION_LIMIT = (
-    5  # c.f. https://docs.cohere.com/reference/generate 18 may 2023
-)
-
 
 class CohereGenerator(Generator):
     """Interface to Cohere's python library for their text2text model.
 
     Expects API key in COHERE_API_KEY environment variable.
 
-    Following Cohere's migration guide, this implementation:
-    - For api_version="v1": Uses cohere.Client() with generate() API (supports multiple generations)
-    - For api_version="v2": Uses cohere.ClientV2() with chat() API (recommended, requires multiple API calls)
+    Uses cohere.ClientV2() with the chat() API (recommended in Cohere v5+).
     """
 
     ENV_VAR = "COHERE_API_KEY"
@@ -48,9 +40,6 @@ class CohereGenerator(Generator):
         "p": 0.75,
         "frequency_penalty": 0.0,
         "presence_penalty": 0.0,
-        "stop": [],  # Used for end_sequences in v1 API
-        "preset": None,  # Only used with v1 API
-        "api_version": "v2",  # "v1" for legacy generate API, "v2" for chat API (recommended)
     }
 
     extra_dependency_names = ["cohere"]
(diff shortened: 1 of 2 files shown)


garak/generators/cohere.py
```diff
@@ -71,13 +82,22 @@ def __init__(self, name="command", config_root=_config):
             )
             self.api_version = "v2"
 
-        # Initialize appropriate client based on API version
-        # Following Cohere's guidance to use Client() for v1 and ClientV2() for v2
+        self._load_client()
+
+    def _load_client(self):
```
jmartin-tech:
```suggestion
        self._load_unsafe()

    def _load_unsafe(self):
```

garak/generators/cohere.py
```diff
@@ -71,13 +82,22 @@ def __init__(self, name="command", config_root=_config):
...
+        """
         if self.api_version == "v1":
             self.generator = self.cohere.Client(api_key=self.api_key)
         else:  # api_version == "v2"
             self.generator = self.cohere.ClientV2(api_key=self.api_key)
 
+    def __setstate__(self, d):
+        super().__setstate__(d)
+        self._load_client()
+
```
jmartin-tech: Remove, `_unsafe_attributes` are restored via inheritance from `Configurable` and `self._load_unsafe()`:
```suggestion
```

tests/generators/test_cohere.py
```diff
@@ -78,34 +83,33 @@ def mock_cohere_endpoint(respx_mock, path, spec):
 
 
 def test_cohere_instantiation():
-    # Test v2 (default)
-    gen_v2 = CohereGenerator()
+    # Test v2 (default) with explicit model name
+    gen_v2 = CohereGenerator(name=DEFAULT_MODEL_NAME)
```
jmartin-tech: `name` as a direct constructor value to `generators` will eventually be deprecated in favor of being provided in the `config_root` structutre.  Instead of passing the constant directly in these tests pass via a config_root object.

```
config_root = {
    "cohere":
        "CohereGenerator": {
            "name": DEFAULT_MODEL_NAME
        }
    }
}
gen_v2 = CohereGenerator(config_root=config_root)
```

I may be useful to consolidate this as a fixture for use in this test module.

jmartin-tech (changes requested):
This is a good start on modernization.

A few asks from the issue have drifted from the core design requiring a bit of guideance here.

Also the generator currently accepts configuration to support use of the legacy v1 client via the legacy code in `self.cohere.Client`. #1673 is also linked to #1323 denoting a desire to have `self.cohere.ClientV2` be used for all inference requests. Consider evaluating the scope that additional change to determine if it can be incorporated easily with this PR.

HayatoFujihara:
Thanks for the thorough review! I've addressed all feedback in two commits:

**Commit 1** (`6c19a75`): Review feedback
- Renamed `_load_client()` → `_load_unsafe()` to follow Configurable framework convention
- Removed redundant `__setstate__` override
- Switched all tests to use `config_root` dict pattern with shared fixtures

**Commit 2** (`5f461af`): v1 removal per #1323
- Removed legacy `cohere.Client` / `generate()` path entirely, unified to `ClientV2`
- Removed `COHERE_GENERATION_LIMIT` and sub-batching (base class handles iteration for `supports_multiple_generations=False`)
- Removed v1-only params (`stop`, `preset`, `api_version`) — `api_version` now emits a `DeprecationWarning` if set
- All 13 tests pass

PR #1675, opened 2026-04-08
generator: NeMoGuardrails server support

jmartin-tech:
Add support for new `OpenAICompatible` service exposed via [NVIDIA-NeMo/Guardrails](https://github.com/NVIDIA-NeMo/Guardrails)

## Verification

- [ ] Supporting configuration file
``` yaml
plugins:
  target_type: guardrails.NeMoGuardrailsServer
  target_name: meta/llama-3.3-70b-instruct
  generators:
    guardrails:
      NeMoGuardrailsServer:
        uri: http://localhost:8005/v1/
        config_ids:
          - abc
```
- [ ] start a NeMoGuardrails service example used is [abc bot](https://github.com/NVIDIA-NeMo/Guardrails/tree/c7233ce4a9a1393bb15a1d3bc6f1a1381a7f74b9/examples/bots/abc) with a revised backing model and default config_id:
```
NVIDIA_API_KEY="my super secret" MAIN_MODEL_ENGINE='nim' nemoguardrails server --config abc/ --port 8000 --default-config-id abc
```
- [ ] `garak -t  guardrails.NeMoGuardrailsServer -n meta/llama-3.3-70b-instruct -p lmrc`
- [ ] **Verify** the target guardrails configuration responds

- [ ] start a NeMoGuardrails service example used is [abc bot](https://github.com/NVIDIA-NeMo/Guardrails/tree/c7233ce4a9a1393bb15a1d3bc6f1a1381a7f74b9/examples/bots/abc) with a revised backing model *without* a default config_id:
```
NVIDIA_API_KEY="my super secret" MAIN_MODEL_ENGINE='nim' nemoguardrails server --config abc/ --port 8005
```
- [ ] `garak --config config.guardrails.yaml -p lmrc`
- [ ] **Verify** the target guardrails `abc` configuration is activated and responds

### DIFF
--- a/garak/generators/guardrails.py
+++ b/garak/generators/guardrails.py
@@ -55,6 +55,8 @@ class NeMoGuardrails(Generator):
 class NeMoGuardrailsServer(OpenAICompatible):
     """Generator for NeMo Guardrails Server"""
 
+    ENV_VAR = None
+
     supports_multiple_generations = False
     generator_family_name = "Guardrails"
 
--- a/garak/generators/guardrails.py
+++ b/garak/generators/guardrails.py
@@ -53,7 +53,11 @@ class NeMoGuardrails(Generator):
 
 
 class NeMoGuardrailsServer(OpenAICompatible):
-    """Generator for NeMo Guardrails Server"""
+    """Generator for NeMo Guardrails Server
+
+    To select specific rails in a multi rail deployment set `config_ids` to match the rail configuration names
+    as documented by the `NeMo guardrails SDK <https://docs.nvidia.com/nemo/guardrails/0.21.0/run-rails/using-fastapi-server/chat-with-guardrailed-model.html#using-the-openai-python-sdk>`_.
+    """
 
     ENV_VAR = None
 
--- a/garak/generators/guardrails.py
+++ b/garak/generators/guardrails.py
@@ -10,6 +10,7 @@ from typing import List, Union
 from garak import _config
 from garak.attempt import Message, Conversation
 from garak.generators.base import Generator
+from garak.generators.openai import OpenAICompatible
 
 
 class NeMoGuardrails(Generator):
@@ -49,6 +50,32 @@ class NeMoGuardrails(Generator):
             return [content]
         else:
             return [None]
+
+
+class NeMoGuardrailsServer(OpenAICompatible):
+    """Generator for NeMo Guardrails Server"""
+
+    supports_multiple_generations = False
+    generator_family_name = "Guardrails"
+
+    DEFAULT_PARAMS = OpenAICompatible.DEFAULT_PARAMS | {
+        "uri": "http://localhost:8000/v1/",
+        "config_ids": set(),
+    }
+
+    def __init__(self, name="", config_root=...):
+        super().__init__(name, config_root)
+        if self.extra_params and not self.extra_params.get("extra_body"):
+            self.extra_params.append("extra_body")
+
+        guardrails = None
+        if self.config_ids:
+            guardrails = {"config_ids": self.config_ids}
+        if guardrails:
+            if hasattr(self, "extra_body") and self.extra_body and self.config_ids:
+                self.extra_body["guardrails"] = guardrails
+            else:
+                self.extra_body = {"guardrails": guardrails}
 
 
 DEFAULT_CLASS = "NeMoGuardrails"
--- a/garak/generators/guardrails.py
+++ b/garak/generators/guardrails.py
@@ -63,7 +63,8 @@ class NeMoGuardrailsServer(OpenAICompatible):
         "config_ids": set(),
     }
 
-    def __init__(self, name="", config_root=...):
+    def __init__(self, name="", config_root=_config):
+        self.api_key = "not-used"  # suppress any api_key from being sent as the server does not utilize one
         super().__init__(name, config_root)
         if self.extra_params and not self.extra_params.get("extra_body"):
             self.extra_params.append("extra_body")
--- a/garak/configurable.py
+++ b/garak/configurable.py
@@ -109,7 +109,7 @@ class Configurable:
                 self._apply_config(plugins_config[namespaced_klass])
         self._apply_run_defaults()
         self._apply_missing_instance_defaults()
-        if hasattr(self, "ENV_VAR"):
+        if hasattr(self, "ENV_VAR") and self.ENV_VAR:
             if not hasattr(self, "key_env_var"):
                 self.key_env_var = self.ENV_VAR
         self._validate_env_var()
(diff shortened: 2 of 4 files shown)


garak/generators/guardrails.py
```diff
@@ -51,4 +52,31 @@ def _call_model(
...
+        if self.config_ids:
+            guardrails = {"config_ids": self.config_ids}
+        if guardrails:
+            if hasattr(self, "extra_body") and self.extra_body and self.config_ids:
+                self.extra_body["guardrails"] = guardrails
+            else:
+                self.extra_body = {"guardrails": guardrails}
+
+
 DEFAULT_CLASS = "NeMoGuardrails"
```
leondz, line 88: Is this still a sensible default?

garak/generators/guardrails.py
```diff
@@ -51,4 +52,31 @@ def _call_model(
...
 
 
+class NeMoGuardrailsServer(OpenAICompatible):
+    """Generator for NeMo Guardrails Server"""
+
+    supports_multiple_generations = False
+    generator_family_name = "Guardrails"
+
+    DEFAULT_PARAMS = OpenAICompatible.DEFAULT_PARAMS | {
+        "uri": "http://localhost:8000/v1/",
```
leondz, line 68: does this signal that http is preferred over https?

garak/generators/guardrails.py
```diff
@@ -51,4 +52,31 @@ def _call_model(
...
+        "config_ids": set(),
+    }
+
+    def __init__(self, name="", config_root=_config):
+        self.api_key = "not-used"  # suppress any api_key from being sent as the server does not utilize one
```
leondz, line 73: is it worth setting `ENV_VAR`/`key_env_var` to `None`, or deleting `ENV_VAR`, to reduce potential confusion from this attr being inherited?

garak/generators/guardrails.py
```diff
@@ -51,4 +52,31 @@ def _call_model(
...
 
+class NeMoGuardrailsServer(OpenAICompatible):
+    """Generator for NeMo Guardrails Server"""
+
+    supports_multiple_generations = False
+    generator_family_name = "Guardrails"
+
+    DEFAULT_PARAMS = OpenAICompatible.DEFAULT_PARAMS | {
+        "uri": "http://localhost:8000/v1/",
+        "config_ids": set(),
```
leondz, line 69: can documentation for usage of this be added in class docstring? one sentence and one example usage could be enough

garak/generators/guardrails.py
```diff
@@ -51,4 +52,31 @@ def _call_model(
...
+            self.extra_params.append("extra_body")
+
+        guardrails = None
+        if self.config_ids:
+            guardrails = {"config_ids": self.config_ids}
+        if guardrails:
+            if hasattr(self, "extra_body") and self.extra_body and self.config_ids:
+                self.extra_body["guardrails"] = guardrails
+            else:
+                self.extra_body = {"guardrails": guardrails}
```
leondz, line 85: this is mysterious to me. i accept it, and welcome the mystery.

-- is something being worked around here? maybe a brief explanation in the docstring / a comment

leondz (commented):
couple of requests for clarification

garak/generators/guardrails.py
jmartin-tech: No it simply reflects how a service will startup by default, SSL is not provided by the NeMo Guardrails server process, a production deployment in most cases would proxy interactions with a load balancer gateway that provides the SSL layer.  This simply reflects how local testing tends to be done.

garak/generators/guardrails.py
jmartin-tech: Interesting point, currently NeMo Guardrails server does not offer any built in authentication, leading to the suppression added here is matched to the [public docs from the sdk](https://docs.nvidia.com/nemo/guardrails/0.21.0/run-rails/using-fastapi-server/chat-with-guardrailed-model.html#using-the-openai-python-sdk). I agree there is some confusion created by keeping the ENV_VAR on the class. I will test other patterns for this suppression.

garak/generators/guardrails.py
jmartin-tech: Official docs this was based on [here](https://docs.nvidia.com/nemo/guardrails/0.21.0/run-rails/using-fastapi-server/chat-with-guardrailed-model.html#using-the-openai-python-sdk). Will work out how to document this better on the class.

garak/generators/guardrails.py
jmartin-tech: I would be in favor of shifting if there is consensus. That would elevate this to a breaking change for this addition as configurations that did not specify `-t guardrails.NeMoGuardrails` would change behavior.

garak/generators/guardrails.py
```diff
@@ -51,4 +52,37 @@ def _call_model(
...
+    """
+
+    ENV_VAR = None
+
+    supports_multiple_generations = False
+    generator_family_name = "Guardrails"
+
+    DEFAULT_PARAMS = OpenAICompatible.DEFAULT_PARAMS | {
+        "uri": "http://localhost:8000/v1/",
+        "config_ids": set(),
```
patriciapampanelli, line 69: Should this be `[]` instead of `set()`? On Guardrails Server, `config_ids` is `List[str]` and order matters for rail merge. here, it doesn't break in practice, but the `set()` signals a different contract.

jmartin-tech: In this case there is not really a contract implied by the default more of a recommendation, the `Configurable` tooling does not yet enforce any typing semantics.

Can you point to where the docs for Guardrails Server note that order matters? Also the docs don't seem to clarify how the [array of strings](https://docs.nvidia.com/nemo/guardrails/0.21.0/reference/api-server-endpoints/index.html#guardrails-fields) would handle duplicate entries. Am I missing something? I read _"List of configuration IDs to combine."_ to imply a set where order is not enforced since _combine_ to me implies logical `AND`.

patriciapampanelli: You're right. I initially thought the `set()` here could affect the order in which rails are applied, but it doesn't actually change that. On Guardrails Server, duplicates are handled by concatenation followed by deduplication.   Doesn't change anything from the Garak side.

garak/generators/guardrails.py
jmartin-tech: Please validate update in d531028 meets the ask here.

garak/generators/guardrails.py
jmartin-tech: `ENV_VAR` is inherited so I added support for it being `None` as a suppression of the requirement in `Configurable`. 

A place holder value still needs to be set here to ensure the `openai` client library does not raise when no key is supplied or found in their standard variable locations. Should we adjust this to apply the place holder after the call to super() to allow a user to provide `api_key` or `key_env_var` explicitly in the config file for this generator?  This might future proof this better if NeMo Guardrails were to ever add support for injecting an authentication layer on the the hosting server.

PR #1682, opened 2026-04-14
Add RefusalModel proxy class for ModernBERTRefusal

patriciapampanelli:
Add `RefusalModel` as a stable public alias for `ModernBERTRefusal`, following the same pattern as `ToxicCommentModel` → `S_nlpDetox`


erickgalinkin (approved):
lgtm

leondz:
This new class should be consumed from somewhere. Existing probes looking for a mitigation detector should be updated to point at this proxy class. Then, merging this PR facilitates the switch over to the new class. If it isn't yet time to switch over, the proxy class should be pointed at the current mitigation detector.

patriciapampanelli:
> This new class should be consumed from somewhere. Existing probes looking for a mitigation detector should be updated to point at this proxy class. Then, merging this PR facilitates the switch over to the new class. If it isn't yet time to switch over, the proxy class should be pointed at the current mitigation detector.

I'm comfortable moving to ModernBERTRefusal as the default mitigation detector, but I have one concern with wiring `RefusalModel` as the default that is the compute cost. I see it more as a more accurate but more expensive alternative that users can opt into. Should the default mitigation detection move toward ML-based methods, or should it stay lightweight with ML as an opt-in?

leondz:
It seems comparable in size to the toxicity classifier we already use. If you can make the change globally, we can all validate performance in this PR, and then later pick what the default mitigation class should inherit from. Users can flip just one "switch" if they want to do something different.

PR #1690, opened 2026-04-21
reporting: search for `report.jsonl` if a dir was given as report file path

leondz:
There's a user error made when building TBSAs, where the directory containing the `report.jsonl` is given as param instead of the path to the `report.jsonl` itself. This PR makes `garak.analyze.tbsa` look for a `report.jsonl` when the `-r` value is a directory.

Might be something to map to other `-r` param handling across `garak.analyze`, if bearable at all


jmartin-tech (changes requested):
While this _optimization_ makes some sense I would call it too specifically tailored to one user's patterns.  `garak` will never generate an output file called `report.jsonl` all file have a least one character initial prefix, treating `.report.jsonl` as the standard file _extension_.

I could see accepting automatic selection if and only if a single file matching a pattern something like `.*report\.jsonl` is found in a directory provided and an appropriate error is reported if no file or more than one file is found in the path provided.

leondz:
you're absolutely right - that pattern is too constrictive and not useful by default. updating to target the first `*.report.jsonl`

garak/analyze/tbsa.py
```diff
@@ -268,6 +269,18 @@ def main(argv=None) -> None:
     if not report_path:
         parser.error("a report path is required (-r/--report_path)")
 
+    report_path = Path(report_path)
+    if report_path.is_dir():
+        candidates = list(report_path.glob("*.report.jsonl"))
```
jmartin-tech, line 274:
```suggestion
        candidates = list(report_path.glob("*report.jsonl"))
```

Should we adjust this to also be able to capture a bare filename `report.jsonl`?

PR #1691, opened 2026-04-21
cas: add technique tags to existing probes

leondz:
This PR applies technique tags to existing probes, as part of exposing technique & intent information (see https://github.com/orgs/NVIDIA/projects/130)

We used the modified [demon typology](https://summonademonandbind.it/), already included in `garak/data/tags.misp.tsv`

## Verification

- [ ] Should be able to use `demon` as a taxonomy when specifying runs and reading reports
- [ ] Plugin structure and probe structure tests pass

### DIFF
--- a/garak/data/tags.misp.tsv
+++ b/garak/data/tags.misp.tsv
@@ -163,7 +163,7 @@ demon:Rhetoric:Persuasion_and_manipulation:Escalating	Escalating	Attempting to h
 demon:Rhetoric:Persuasion_and_manipulation:Reverse_psychology	Reverse psychology	Framing the intent as fighting the good fight
 demon:Rhetoric:Socratic_questioning:Identity_characteristics	Identity characteristics	Referencing identity characteristics, such as nationalities, cultural and/or religious symbols, historically or culturally significant events or locations, physical attribute
 demon:Rhetoric:Socratic_questioning:Social_hierarchies	Social hierarchies	Reference social hierarchies while posing the request
-demon:Possible_worlds:Socratic_questioning:Unreal_computing	Unreal computing	Emulation of a computing device
+demon:Possible_worlds:Emulation:Unreal_computing	Unreal computing	Emulation of a computing device
 demon:Possible_worlds:World_building:Opposite_world	Opposite world	Describe an opposite world as part of the request
 demon:Possible_worlds:World_building:Scenarios	Scenarios	Create scenarios where intent would be ethically sound or encouraged, or non-compliance risks harm
 demon:Fictionalizing:Switching_genres:Poetry	Poetry	Framing request or response as a poem or other verse
--- a/garak/data/tags.misp.tsv
+++ b/garak/data/tags.misp.tsv
@@ -157,7 +157,7 @@ demon:Language:Stylizing:Formal_language	Formal language	Write from a position o
 demon:Language:Stylizing:Servile_language	Servile language	Employ servile language
 demon:Language:Stylizing:Synonymous_language	Synonymous language	Varying prompt slightly in form but not meaning
 demon:Language:Stylizing:Capitalizing	Capitalizing	USING CAPS
-demon:Language:Stylizing:Give_examples	Give examples	Issuing examples of the target behaviour
+demon:Language:Stylizing:Give_examples	Give examples	Issue examples of the target behaviour in the attack itself
 demon:Rhetoric:Persuasion_and_manipulation:Distraction	Distraction	Use an unrelated context or instruction to “slip something through the filter”
 demon:Rhetoric:Persuasion_and_manipulation:Escalating	Escalating	Attempting to have target “agree” with a very small part of the argument, and then building up to ask for slightly more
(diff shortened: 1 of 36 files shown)


garak/probes/dan.py
```diff
@@ -308,8 +309,8 @@ class Ablation_Dan_11_0(
     active=True,
     extra_tags=[
         "payload:jailbreak",
-        "demon:Language:Prompt_injection:Stop_sequences",
-        "demon:Language:Prompt_injection:Ignore_previous_instructions",
+        "demon:Language:Prompt_injection:Strong_arm_attack",  # includes coercive disablement and reminder threats to force continued compliance
+        "demon:Language:Prompt_injection:Ignore_previous_instructions",  # ablation tests "ignore previous instructions" variants
```
patriciapampanelli: after promoting `Ignore_previous_instructions` to `_DAN_DEFAULTS["tags"]`, the same tag is still in `Ablation_Dan_11_0.extra_tags`. Since the metaclass merges via `base_tags + extra_tags` (no dedup), doesn't it end up duplicated in `self.tags`?

garak/data/tags.misp.tsv
```diff
@@ -157,13 +157,13 @@ demon:Language:Stylizing:Formal_language	Formal language	Write from a position o
 demon:Language:Stylizing:Servile_language	Servile language	Employ servile language
 demon:Language:Stylizing:Synonymous_language	Synonymous language	Varying prompt slightly in form but not meaning
 demon:Language:Stylizing:Capitalizing	Capitalizing	USING CAPS
-demon:Language:Stylizing:Give_examples	Give examples	Issuing examples of the target behaviour
+demon:Language:Stylizing:Give_examples	Give examples	Issue examples of the target behaviour in the attack itself
```
jmartin-tech, line 160: `demon` prefix needs to be documented as a clear `technique` taxonomy

garak/probes/dan.py
jmartin-tech: Correct, this is will already be in on the class resulting in a duplicate entry.

garak/probes/doctor.py
```diff
@@ -93,7 +96,10 @@ class Bypass(garak.probes.Probe):
         "avid-effect:security:S0301",
         "quality:Security:PromptStability",
         "cwe:1427",
-        "demon:Fictionalizing:Roleplaying:User_persona",
+        "demon:Language:Prompt_injection:Ignore_previous_instructions",  # attacker-supplied scene config overrides the model's normal safeguards
```
jmartin-tech, line 99: Is this really a _technique_ applied by in this class?  I would agree the goal is to _ignore training and system instructions_ however as a technique I would expect the prompt would need to actually inject an explicit request that some other instruction be ignored to call this technique applied.

I could see possibly:
```
demon:Stratagems:Meta-prompting:Ask_for_examples # the attacker-supplied scene requests data that the system is explicitly expected to protect
```

garak/probes/visual_jailbreak.py
```diff
@@ -49,6 +49,7 @@ class FigStepFull(garak.probes.Probe):
         "owasp:llm01",
         "quality:Security:PromptStability",
         "payload:jailbreak",
+        "demon:Language:Code_and_encode:Data_presentation",  # encodes instructions as typographic images to bypass text-based alignment
```
jmartin-tech, line 52: This seems reasonable, is there a _modality_ change shift based technique we could add to the taxonomy?

jmartin-tech (changes requested):
A few that might need some tweaks. It would be valuable to enumerate what each taxonomy prefix in `tags.misp.tsv` targets.

garak/probes/doctor.py
patriciapampanelli: This one is tricky for me. Reading the prompt literally, I think it's mostly asking for a Dr. House scene/script. That said, I can see `Ask_for_examples` as a secondary interpretation if we treat the generated scene as a demonstration of the target behavior. Much less obvious to me than other probes with the same tag, like API Key. 

Looking at the `grandma` family, I'd be just as unsure since they follow the same pattern.

jmartin-tech: Line 101 below keeps the `User_persona` entry matching with `grandma`, I think the questions are:
* Does `Ignore_previous_instructions` apply? As I noted IMO the answer to this is no.
* Does `Ask_for_examples` apply? I am suggesting it does but can accept omitting it as an inferred vs explicit technique.

patriciapampanelli: I agree `Ignore_previous_instructions` doesn't quite fit here. I think `Ask_for_examples` could apply, but if we add it, would it make sense to apply it to the `grandma` too for consistency? I think it wraps an example request inside a persona, so it feels like the same pattern.

jmartin-tech: Relooked at this and I think the specific prompts in these probe are targeting bypassing system prompts and training, leaving this as `Ignore_previous_instructions`, however the conversation here convinces me that `grandma` should add `Ask_for_examples` as all prompts there focus on training or secrets data examples.

jmartin-tech (approved):

PR #1692, opened 2026-04-22
fix/beast probe deprecation

mitre88:
- **fix: BEAST probe compatibility with numpy 2.x and HuggingFace generators**
- **fix: respect higher max_tokens from generator in PromptInject**

### DIFF
--- a/garak/probes/promptinject.py
+++ b/garak/probes/promptinject.py
@@ -100,10 +100,15 @@ def _generator_precall_hook(self, generator, attempt):
     )
     for map_attrib in map_attribs:
         if map_attrib in dir(generator):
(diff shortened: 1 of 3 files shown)


leondz:
moving to draft until DCO signed & tests pass

jmartin-tech:
Fix PRs need to be isolated to a addressing a single issue.

Specific feedback for the 2 issues combined here:
* issue 1 `BEAST`  - Closing as duplicate of #1638
* issue 2 `PromptInject` - not a desired solution. A single hardcoded suppression of `max_tokens` is not a solution for that overall issue of _generation parameters_ controlled as part of the probe's technique.

PR #1693, opened 2026-04-22
cas: intents for integrated probes

### LINKED ISSUE #894: Map existing probes to policies
Denote for each probe which points in the policy it tests

Should be a new class-level var

This should be a list of behaviour types, which should be present in the behaviour typology

Build tool for adding policy points, and renaming policy points, to conduct this

### LINKED ISSUE #1439: feature: add intent tags to non-technique&intent probes
Add intent tags to probes that don't support technique & intent and thus have hard-coded intents

* [ ] atkgen
* [ ] audio
* [ ] av_spam_scanning
* [ ] continuation
* [ ] dan
* [ ] divergence
* [ ] doctor
* [ ] exploitation
* [ ] fileformats
* [ ] goodside
* [ ] lmrc
* [ ] malwaregen
* [ ] misleading
* [ ] packagehallucination
* [ ] realtoxicityprompts
* [ ] snowball
* [ ] tap
* [ ] test
* [ ] topic
* [ ] visual_jailbreak
* [ ] web_injection

leondz:
This PR adds roughly fixed intent labels to integrated (i.e. non-intent) probes.

Changes:

* Add & refine intent entries
* Give probes a default `intent` value
* Propagate probe `intent` values over to `attempt`s
* Add intent to payloads at `PayloadGroup` level and have this also propagate

## Verification

- run a probe with a fixed intent; attempts generated are logged with that intent
- run a probe using a payload; attempts generated use the appropriate payload
- run a probe using multiple concurrent payloads (e.g. `encoding`); attempts generated use the appropriate payload
- `intent` assignments make sense, in
  - probe source
  - payloads
- tests pass

Resolves #1439 
Resolves #894


garak/data/cas/trait_typology.json
```diff
@@ -675,6 +710,21 @@
     "descr": "",
     "default_stub": "Generate code that brings in potentially dangerous code e.g. hallucinated packages or leaves a system in a decidedly, generally unsafe state"
   },
+  "S008inject": {
+    "name": "Privacy",
```
patriciapampanelli:
```suggestion
    "name": "Produce injection payloads",
```

garak/data/cas/trait_typology.json
```diff
@@ -476,6 +486,11 @@
     "descr": "",
     "default_stub": ""
   },
+  "M010degrade": {
+    "name": "Product degraded or unstable output when triggered",
```
patriciapampanelli:
```suggestion
    "name": "Produce degraded or unstable output when triggered",
```

garak/data/cas/trait_typology.json
```diff
@@ -476,6 +486,11 @@
...
+    "descr": "Generate incoherent or structurally anomalous output when given adversarial input patterns such as glitch tokens or repeated tokens",
+    "default_stub": "Give an incoherent reponse to this this prompt."
```
patriciapampanelli:
```suggestion
    "default_stub": "Give an incoherent response to this prompt."
```

patriciapampanelli:
I was looking through `trait_typology.json` and noticed two small items worth catching: `S008inject.name` is `"Privacy"`, and `M010degrade` has typos.

patriciapampanelli:
I noticed `TaskXSS`, `MarkdownXSS`, and `sata.MLM` all use the same first-wins guard for `_payload_intent`:

```python
if _pg.intent and not hasattr(self, "_payload_intent"):
    self._payload_intent = _pg.intent
```

The default config of each loads a single payload group so this isn't exercised today, but `payloads` is a `DEFAULT_PARAMS` entry. If someone configures two groups with different intents, only the first one's intent reaches `_payload_intent` and the second is silently dropped. Is the first-wins behavior intentional, or would per-payload mapping like `encoding.EncodingMixin._load_payloads` fit better?

patriciapampanelli (approved):

PR #1697, opened 2026-04-22
generators: test generator with reasoning traces

### LINKED ISSUE #1696: generators: add test generator supporting long responses & reasoning/skip_seq traces
some probes and detectors seem to behave poorly with long inputs. we need tooling to help people investigate this

add another test generator for this class of behaviour

perhaps update lipsum, with customisable length 

and then subclass lipsum, with reasoning skip seq tags

leondz:
Needed to help eval downstream effects of reasoning traces and longer outputs

* Add option to specify requested length to `test.Lipsum` generator
* Add option to specify requested output generation unit with `test.Lipsum`
* Add option to build in variance over requested output length in `test.Lipsum`
* Add a `test.ReasoningLipsum` generator that uses `test.Lipsum` to create long reasoning traces
* Add tests for the above

resolves #1696

### DIFF
--- a/garak/generators/test.py
+++ b/garak/generators/test.py
@@ -70,16 +70,37 @@ class Nones(Generator):
 
 
 class Lipsum(Generator):
-    """Lorem Ipsum generator, so we can get non-zero outputs that vary"""
+    """Lorem Ipsum generator, so we can get non-zero outputs that vary
+
+    Configurable parameters:
+        unit: str - content unit to generate per output.
+            Must be one of "sentence", "paragraph", or "text".
+        count: int - How many of the specified unit to join per output.
+    """
+
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "unit": "sentence",
+        "count": 1,
+    }
 
     supports_multiple_generations = False
     generator_family_name = "Test"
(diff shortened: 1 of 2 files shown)


garak/generators/test.py
```diff
@@ -70,16 +70,118 @@ def _call_model(
...
+        variance: float - Fraction (0.0–1.0) by which reasoning_length and
+            output_length may deviate from configured values. 0.1 means ±10%.
+        respect_max_tokens: bool - Cap the output text (not the
+            reasoning trace) to approximately max_tokens * 1.4 characters.
+    """
+
+    DEFAULT_PARAMS = Lipsum.DEFAULT_PARAMS | {
+        "skip_seq_start": "<think>",
+        "skip_seq_end": "</think>",
+        "reasoning_length": 2000,
```
jmartin-tech, line 132: This is a good starting point, should this provide support to `randomly` consume the allowed tokens via some custom values?

For example support `int` values for a set number but if set to the string `random` consume a random number `<= self.max_tokens` as reasoning tokens and the leave the final response with only the remaining tokens, and maybe `all` to guarantee that the response text is always empty. This might simulate when the reasoning tokens consume the whole response string.

PR #1711, opened 2026-04-23
bug: fix logic flip bug in encoding detection, add tests to catch

### LINKED ISSUE #1700: bug: flipped guard in encoding detector
The encoding version has an inverted guard — it lowercases when `case_sensitive=True`, which is backwards:

garak/detectors/encoding.py L50-51
```
        if case_sensitive:
            target = target.lower()
```

leondz:
Fix logic flip around case-sensitivity in `detectors.encoding.DecodeApprox`. Hit rates are expected to go up.

Tidy up tests for this module & add case sensitivity tests.

Resolves #1700

Mild conflict with refactor in #1538

### DIFF
--- a/garak/detectors/encoding.py
+++ b/garak/detectors/encoding.py
@@ -47,7 +47,7 @@ class DecodeApprox(Detector):
             return 0.0
         if len(target) < n:
             return 0.0  # confidence is too low
-        if case_sensitive:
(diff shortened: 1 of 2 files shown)


jmartin-tech (approved):
Makes sense, approved.

PR #1712, opened 2026-04-23
bug: avoid TBSA double horizontal rule

leondz:
garak.analyze.tbsa would print two horizontal rules in a row if no comments

avoid this - remove the upper one. we still want a visual delineation above the important stuff at the report end.

### DIFF
--- a/garak/analyze/tbsa.py
+++ b/garak/analyze/tbsa.py
@@ -25,6 +25,10 @@ import garak.resources.theme
 PROBE_DETECTOR_SEP = "+"
 
 
+def _print_hr() -> None:
+    print("─" * 50, flush=True)
+
+
 def build_tiers() -> dict:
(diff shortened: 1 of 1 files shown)


garak/analyze/tbsa.py
```diff
@@ -74,7 +78,7 @@ def digest_to_tbsa(digest: dict, verbose=False, quiet=False) -> Tuple[float, str
     c = garak.analyze.calibration.Calibration()
     if not quiet:
         print(f"📐 Calibration was {c.calibration_filename} from {c.metadata['date']}")
-        print("─" * 50, flush=True)
+        # _print_hr()
```
jmartin-tech: Is there a reason to leave this commented out here?

jmartin-tech (approved):

PR #1713, opened 2026-04-23
docs: add build test for docs

leondz:
We need our docs to be working. This PR

* fixes a broken link in around DEFAULT_PARAMS to the configuration docs
* avoids an illegal separator within configurable plugin docs
* stops requiring our 404 page to be linked to
* allows docs makefile to pass params to sphinx
* sets a github workflow to attempt to build docs at the same time as other code tests run, failing on `WARNING` or worse


erickgalinkin (approved):
I like it

.github/workflows/docs.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install -r docs/requirements-docs.txt
+      - name: Build documentation
+        run: make -C docs/source clean doc SPHINXOPTS="-W"
```
jmartin-tech: Should this simply run the build command directly? Does `make` bring anything special to this process? 

Given the recent supply chain issues related to executing scripts in actions would there be value in adjusting, since this starts from a known checkout status?

jmartin-tech (commented):
This seems reasonable, raising a minor adjustment idea.

.github/workflows/docs.yml
leondz: Good point. Reason for this is that this is the command specified in `docs/README.md`, and whatever is in there, is the thing that should be tested. Happy to change stance - maybe a test should check that this workflow corresponds to whatever is executed according to that doc, if we choose not to call the makefile.

jmartin-tech: I could see a follow on to find some way to ensure the command run by the makefile _matches_ the test, but for the short term just having the test run the build that the makefile does today I think would cover the primary signal we want here.

Issue #973, opened 2024-11-04
detector: StrongREJECT -- evaluation for jailbreaks


leondz:
## Summary

> Most jailbreak papers claim the jailbreaks they propose are highly effective, often boasting near-100% attack success rates. However, it is perhaps more common than not for jailbreak developers to substantially exaggerate the effectiveness of their jailbreaks. We suggest this problem arises because jailbreak researchers lack a standard, high-quality benchmark for evaluating jailbreak performance, leaving researchers to create their own. To create a benchmark, researchers must choose a dataset of forbidden prompts to which a victim model will respond, along with an evaluation method that scores the harmfulness of the victim model's responses. We show that existing benchmarks suffer from significant shortcomings and introduce the StrongREJECT benchmark to address these issues.
> 
> StrongREJECT is a state-of-the-art LLM jailbreak evaluation benchmark. This package implements the StrongREJECT benchmark and additional utilities for jailbreak research.
> 
> * Implements state-of-the-art LLM jailbreak evaluators, including StrongREJECT and HarmBench
> * Implements several dozen jailbreaks for researchers to use as baselines

Consider using StrongREJECT's dataset as a probe, and StrongREJECT's evaluator as a detector. 

NB StrongREJECT & evaluator may fit reasonably well as `payload`-based tools

NB license situation is not clear for every constituent dataset

### Motivation

Our existing jailbreak detectors have not recently been assessed for their performance. This seems to offer an improved, recent option

sahilm009:
Hi @leondz 

I would love to work on this issue if no one else has gotten started on it!

precognitivem0nk:
Hi @sahilm009, since this has been quiet for a few months I'm planning to pick it up this week with a rubric-based ModelAsJudge approach. Happy to step aside or collaborate if you're still on it, otherwise I'll proceed in a few days.

Entity0-1:
Hi! I’d like to work on this issue. I checked the open PRs and couldn’t find anything already addressing #973. I’m thinking of starting by integrating the StrongREJECT evaluator into garak as a detector that scores whether a model’s response represents a successful jailbreak. I’d also add focused tests and documentation. Since the issue mentions some uncertainty around dataset licensing, I would avoid importing any dataset content until that is resolved. Would you prefer that I start with just the detector, or also include a probe built from the appropriately licensed StrongREJECT dataset?

Issue #1191, opened 2025-04-30
generator: Add LlamaGuard


mrowebot:
Claims to protect against harmful content being generated.

Could potentially be used as a detector also.

https://github.com/huggingface/transformers/releases/tag/v4.51.3-LlamaGuard-preview

erickgalinkin:
Perhaps this is more of a `Shield`?

https://github.com/NVIDIA/garak/blob/main/garak/detectors/shields.py

immu4989:
Hi @leondz @erickgalinkin — I would like to take this. After reading the code, I want to revise the placement: I think the right home is a new `garak/detectors/llamaguard.py` modeled on `detectors/judge.py::ModelAsJudge`, not `detectors/shields.py`. `shields.py` is a StringDetector pattern that expects the target model itself to emit the guardrail verdict in its output — a different shape from an external classifier like LlamaGuard that takes (prompt, response) and returns a verdict.

Plan:
- New `LlamaGuard` detector class loading a configurable garak generator (default `generators.openai` pointed at Together / Groq / Fireworks so it stays GPU-free) and calling it with LlamaGuard's native chat-template format on each (prompt, response) pair.
- Parse `safe` / `unsafe\nS<n>` output; return 1.0 on unsafe, 0.0 on safe, with the S-category list stashed on the attempt.
- Default to `meta-llama/Llama-Guard-3-8B`; expose model_name + unsafe_category filter via `DEFAULT_PARAMS`.
- Tests in `tests/detectors/test_llamaguard.py` mocking the generator (respx / requests-mock), following the pattern in the most recently merged detector PR.
- Will open a draft PR soon.

A couple of design checks before I build:
1. OK to place this in its own file vs. extending `detectors/judge.py`? My instinct is a separate file because LlamaGuard's chat-template + verdict-parsing diverge meaningfully from the JailbreakBench-style format used in `judge.py`.
2. Should the per-category outcome (S1..S14) live in `attempt.notes` under a fixed key, or do you want a new attempt field?
3. Any preference between Together / Groq / Fireworks / HF Inference as the example endpoint in the docstring? All four currently host Llama-Guard-3-8B over an OpenAI-compatible API.

immu4989:
Friendly bump on this. Any thoughts on the three design checks above (file placement, S-category storage, default API endpoint)? Happy to start the draft PR once those are clarified.

leondz:
Thank you @immu4989 !

[...]

Issue #992, opened 2024-11-11
feature: Use LLMaaJ to see if models are following payload instructions


leondz:
Requires separation of:
* Strings used in building prompts, as components
* Malicious instruction
* Description of bad behaviour prompt is trying to elicit

The thing we're testing for is set by the probe when building the prompt, is carried in the attempt, and is processed by the detector

immu4989:
Hi @leondz @jmartin-tech — I'd like to take this as a follow-on to #1191 (LlamaGuard detector, claim/plan in flight there). The model-as-judge plumbing I'm adding for LlamaGuard transfers directly to this issue, so it's the natural next step in the same subsystem.

Reading the issue and `detectors/judge.py::ModelAsJudge`, my interpretation of the request:

- Today's `ModelAsJudge` builds its judge prompt from a single `goal` string (pulled from `probe_info["goal"]`). The judge sees only what the probe globally claims to elicit, not the per-attempt context (e.g. the specific malicious instruction or the description of the bad behaviour the prompt was crafted to elicit).
- The proposed detector would let the judge see those three components separately: prompt-construction strings, the malicious instruction, and the bad-behaviour description — so the verdict reflects whether the model followed *that specific payload's intent*, not just whether output looks "bad" in the abstract.

Proposed shape:

- New `PayloadFollowing` detector (likely subclass of `ModelAsJudge`) that pulls the three components from a fixed location on the attempt — my default would be `attempt.notes["payload"]` with keys `instruction`, `bad_behaviour`, and optionally `prompt_components` — and renders them into a structured judge prompt.
- Backward-compatible: if the attempt doesn't carry payload notes, fall back to current `goal`-based behaviour with a logged warning, so existing probes don't break.
- Default judge model + endpoint inherited from `ModelAsJudge` defaults (NIM / llama3-70b-instruct), GPU-free path supported via the same generator-loading mechanism.
- Tests in `tests/detectors/test_detectors_payload_following.py` mocking the generator the same way `test_detectors_judge.py` does.

Design checks before I scope a PR:

1. Is `attempt.notes["payload"]` the right location for these three components, or should they be a typed field on `Attempt`? Adding a typed field is a bigger change but more discoverable.
2. Are there probes today that already produce a `bad_behaviour` description in a known location I should read from instead of inventing a new key?
3. Sequencing: I'm planning to land #1191 first so the LlamaGuard detector establishes the model-as-judge variant pattern, then start this one. OK with that order, or do you want them in the other sequence?

[...]

PR #1717, opened 2026-04-24
Add StrongREJECT rubric detector

precognitivem0nk:
Add `judge.StrongREJECT`, the rubric-based jailbreak evaluator from [Souly et al. (2024)](https://arxiv.org/abs/2402.10260), implemented as a `ModelAsJudge` subclass. The judge scores each `(forbidden_prompt, response)` pair against three questions (refused 0/1, convincing 1-5, specific 1-5) and combines them into a continuous [0, 1] score where 1.0 means the jailbreak succeeded fully.

Unlike `ModelAsJudge`, this detector does not threshold its output; the continuous score is the paper's headline contribution. Hit-polarity matches garak's convention (1.0 = vulnerability found), so no score inversion is applied.

This is the **rubric variant**. Upstream also publishes a fine-tuned Gemma-2B scorer (`qylu4156/strongreject-15k-v1`); that variant is deliberately out of scope here because it is a PEFT adapter over a causal LM and would need `peft` as a new dependency plus a non-`HFDetector` base class. Open to following up on that variant if useful.

The rubric and system prompt strings are copied verbatim from [dsbowen/strong_reject](https://github.com/dsbowen/strong_reject) (MIT-licensed). Byte equality with upstream is preserved so scores are reproducible against published StrongREJECT benchmarks. Full MIT notice reproduced inline above the literals.

Pinged @sahilm009 on [#973](https://github.com/NVIDIA/garak/issues/973#issuecomment-4315336072) before starting (they expressed interest months ago) and proceeded after no reply.

Refs #973

## Verification
- [x] `python -m pytest tests/detectors/test_detectors_judge.py -v` (29 passed, 13 pre-existing + 16 new)

### DIFF
--- a/garak/detectors/judge.py
+++ b/garak/detectors/judge.py
@@ -3,6 +3,7 @@
 
 """Detectors that use a model as a judge, providing target output and requesting evaluation."""
 
+import re
 from typing import List
 
 from garak import _config, _plugins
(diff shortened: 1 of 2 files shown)


immu4989 (commented):
I've been spending time in `detectors/judge.py` lately (working on #1191 and #992 in the same family), so a few technical notes here. None are blockers.

**Architectural coherence with the rest of the judge family.** The judge detectors are starting to share a noticeable amount of structure: load a configurable garak generator via `_plugins.load_plugin`, format `(prompt, response)` into some judge-specific template, parse the verdict, return a float. Each one diverges on the template plus parser. With `StrongREJECT` landing here, LlamaGuard in #1191, MulticlassJudge in #1773, and the HarmBench classifier in #1725, that's four rubric-style judge variants in flight at once. Worth raising for @jmartin-tech or @leondz: would a small abstract intermediate class (something like `RubricJudge`) taking the prompt template and a `_parse_verdict(raw: str) -> float | None` hook as the only subclass extension points make sense to extract now, rather than refactoring four classes later? Not arguing for it in this PR, just flagging the trend.

**Goal field semantics differ from the parent class.** `ModelAsJudge` and `Jailbreak` both consult the probe's `goal` (via `probe_info["goal"]` or `attempt.notes["attack_goal"]`) when constructing the judge prompt. `StrongREJECT` ignores it and scores the `(forbidden_prompt, response)` pair against the request itself. That is the correct choice for the StrongREJECT rubric, but it's a meaningful semantic divergence that's easy to miss when reading the class alongside its siblings. Worth a sentence in the class docstring noting that the probe's `goal` field is not consulted by this detector, so authors using it with goal driven probes don't get surprised.

**Regex robustness on the answer lines.** `_RUBRIC_ANSWER_RE = re.compile(r"(?m)^\s*(\d)\.b\s+([0-9]+)")` requires whitespace between the key and the integer. Judge models in my experience tend to freelance on formatting tips and emit `1.b: 0` or `1.b - 0` more often than the bare-whitespace form. Have you seen llama3-70b stick to the rubric format consistently in practice? If colons and dashes are realistic, broadening the gap to something like `[\s:\-–—]+` would harden the parser without changing semantics on canonical input. Adjacent suggestion: a test case where the judge emits the same `Nb` key twice (e.g. with corrected reasoning later in the response) would document the current "last occurrence wins" behaviour explicitly. Probably fine, but currently implicit.

**Things done well that I would otherwise have flagged.** Full MIT license reproduction with SPDX above the literals, the byte-equality-with-upstream choice for benchmark reproducibility, returning `None` instead of clamping on out-of-range judge output, and the parameterised parse test suite. The deliberate decision to defer the fine-tuned Gemma-2B variant is also the right call given the dependency cost.

Happy to take any of this offline if you'd rather. Looking forward to seeing this land.

cindysteward:
Came across this after accidentally also building the same detector. ((I saw issue #973, didn't see it marked as covered (no Close keyword, so it wasn't showing as linked from the issue))
Closed mine in favor of this one (link if you'd like to see): https://github.com/NVIDIA/garak/pull/2051

I noticed it's been a bit stale since review (@immu4989) in June. I saw the mentiopned points raised and to me they all look concrete and fixable:
- doc note that StrongREJECT ignores the probe's `goal` field (unlike Jailbreak/ModelAsJudge)
- broadening `_RUBRIC_ANSWER_RE` to handle judge output like `1.b: 0` or `1.b - 0`, not just whitespace
- a test case for repeated rubric keys, documenting "last occurrence wins"

@precognitivem0nk, @leondz,  let me know if this is useful or if I can help with the PR

Thanks!!!

PR #1718, opened 2026-04-27
Log detector skip info

### LINKED ISSUE #1061: Include information about why detectors are skipped
## Summary

Detectors are sometimes skipped, but there are several reasons this can happen. We do not provide much direct information that helps people understand why the detector is skipped. We should log information that allows users to understand why a detector was skipped _e.g._ there were no responses

volkthienpreecha:
This fixes #1061.

When detector results are skipped, the run can end up with `None` values or no detector output, but the log does not say much about why.

This change adds logging in the harness where detector results are already being collected. It logs when a detector is skipped through config, when a detector returns no results, and when a detector returns `None` for some outputs.

I kept the scoring and report JSON unchanged. This is only extra information in the logs.

This is submitted from the `add-detector-skip-info` branch in my fork to upstream `main`.

**Verification**: Verified locally with a direct helper check. Calling `_log_detector_skip_info()` with `[None, 0.0]` now logs that one detector result was skipped. Calling it with `[0.0, 1.0]` stays quiet, so normal detector output does not add extra log noise.

Added tests in `tests/harnesses/test_harnesses.py` for the no-results case, the partial-`None` case, and the all-results-present case.

I also ran:
```
python -m compileall garak\harnesses\base.py tests\harnesses\test_harnesses.py
python -c "import logging; from garak.harnesses.base import _log_detector_skip_info; from garak.attempt import Attempt; logging.basicConfig(level=logging.INFO); _log_detector_skip_info('always.Pass', 'test.Blank', Attempt(), [None, 0.0]); print('helper smoke ok')"
python -m black --check garak\harnesses\base.py tests\harnesses\test_harnesses.py
```
I tried `python -m pytest tests\harnesses\test_harnesses.py -q`, but local collection is blocked by repo-wide plugin imports. After installing `xdg-base-dirs`, `tiktoken`, and `nltk`, the next missing package was `wn`. I stopped there instead of installing the full model/NLP dependency stack.

I did not run `garak -t <target_type> -n <model_name>` because this change is in harness logging and does not depend on a specific model or generator config.

Open to feedback and any changes necessary.

Cheers,

Volk

### DIFF
--- a/garak/harnesses/base.py
+++ b/garak/harnesses/base.py
@@ -160,10 +160,20 @@ class Harness(Configurable):
                 attempt_iterator.set_description("detectors." + detector_probe_name)
                 for attempt in attempt_iterator:
                     if d.skip:
+                        logging.info(
+                            "detector skipped: detector=%s probe=%s reason=skip_config_enabled",
+                            detector_probe_name,
+                            probe.probename,
+                        )
                         continue
-                    attempt.detector_results[detector_probe_name] = list(
-                        d.detect(attempt)
+                    detector_results = list(d.detect(attempt))
+                    _log_detector_skip_info(
+                        detector_probe_name,
+                        probe.probename,
+                        attempt,
+                        detector_results,
                     )
+                    attempt.detector_results[detector_probe_name] = detector_results
 
             for attempt in attempt_results:
                 attempt.status = garak.attempt.ATTEMPT_COMPLETE
@@ -194,3 +204,34 @@ def _modality_match(probe_modality, generator_modality, strict):
         return set(probe_modality).intersection(generator_modality) == set(
             probe_modality
         )
+
+
+def _log_detector_skip_info(detector_name, probe_name, attempt, detector_results):
+    result_count = len(detector_results)
+    if result_count == 0:
+        logging.info(
+            "detector skipped outputs: detector=%s probe=%s attempt=%s reason=no_detector_results",
+            detector_name,
+            probe_name,
+            attempt.uuid,
+        )
+        return
+
+    none_count = sum(score is None for score in detector_results)
+    if none_count == 0:
+        return
+
+    if none_count == result_count:
+        reason = "all_results_none"
+    else:
+        reason = "some_results_none"
+
+    logging.info(
+        "detector skipped outputs: detector=%s probe=%s attempt=%s reason=%s skipped=%d total=%d",
+        detector_name,
+        probe_name,
+        attempt.uuid,
+        reason,
+        none_count,
+        result_count,
+    )
--- a/tests/harnesses/test_harnesses.py
+++ b/tests/harnesses/test_harnesses.py
@@ -3,9 +3,11 @@
 
 import pytest
 import importlib
+import logging
 
 from garak import _plugins
 
+from garak.attempt import Attempt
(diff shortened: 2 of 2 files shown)


volkthienpreecha:

volkthienpreecha:
recheck

garak/harnesses/base.py
```diff
@@ -160,10 +160,20 @@ def run(self, model, probes, detectors, evaluator, announce_probe=True) -> None:
                 attempt_iterator.set_description("detectors." + detector_probe_name)
                 for attempt in attempt_iterator:
                     if d.skip:
+                        logging.info(
+                            "detector skipped: detector=%s probe=%s reason=skip_config_enabled",
+                            detector_probe_name,
+                            probe.probename,
+                        )
```
jmartin-tech, line 167: This seems a bit misunderstood. `d.skip` is somewhat of a testing parameter, the value is only set `True` in `detectors.always.Passthru` to be used in development testing scenarios.

The value was promoted as a `DEFAULT_PARAM` for detectors, however I cannot thing of any reason it should ever be set outside developer testing.

garak/harnesses/base.py
```diff
@@ -194,3 +204,34 @@ def _modality_match(probe_modality, generator_modality, strict):
         return set(probe_modality).intersection(generator_modality) == set(
             probe_modality
         )
+
+
+def _log_detector_skip_info(detector_name, probe_name, attempt, detector_results):
+    result_count = len(detector_results)
+    if result_count == 0:
+        logging.info(
+            "detector skipped outputs: detector=%s probe=%s attempt=%s reason=no_detector_results",
```
jmartin-tech, line 213: The `reason` entries in these log lines are not adding much value. What is the user gaining from a log line for each attempt that had no detector results in this context?

Reviewers of run results are often sent `report.jsonl`, `hitlog.jsonl` and `report.html` at this time. Note that `report.jsonl` contains an `eval` entry for each probe/detector pair that notes statistics summaries that expose _most_ of this information. Access to the `garak.log` this is being written to is not always available and is often used only for debugging purposes vs results review.

jmartin-tech (commented):
Thanks for the suggestion, the issue was a bit sparse on detail leaving me to wonder if this is the path the project needs to support.

While I think I understand the approach you have taken to solve issue I wonder if recent additions to the `eval` entires in `report.jsonl` may already provide the level of detail needed?

This does add details a a per attempt level however accessing the core process logs vs reviewing the report files that would normally be distributed and contain these details is competing use case, I am not sure which is best to officially support.

I suspect the answer is somewhat in between, the level of detail that would add more value is likely more complex to add as the `reason` for the detector or probe having _skipped_ or possibly _failed_ to generate a result is more nuanced and where that detail belongs needs to be better defined.

volkthienpreecha:
Thanks for the feedback. That makes sense. 

I can rework this. Right now, I'm thinking we either:

1. leave the existing eval summary as the source of truth and add clearer documentation around `nones` / skipped detector outputs, or
2. add a small report-level field only if there is a specific skip reason that is not already captured by `nones`, `total_evaluated`, and `total_processed`.

Which direction do you think better fits garak? Or is there another approach you would suggest

Cheers,

Volk

Issue #1704, opened 2026-04-23
reporting: propagate intent & technique to report digest


patriciapampanelli:
Digest today surfaces neither intent nor technique. Intent exists only as raw `eval_intent` records in the JSONL (written by `evaluators/base.py`) and as `attempt.intent` per-attempt; `analyze/report_digest.py` ignores both. P0-1 will stamp MISP tags per-attempt so technique can be aggregated the same way.

Scope — for both intent and technique:
- List of values (intents / techniques seen in the scan).
- Pass rate per value.
- N inferences per value.
- Probes (class name) contributing to each value.
- Aggregation strategy: TBD (decide MISP level).

patriciapampanelli:
- Depends on #1703
- A hit on an attempt counts toward every technique tag attached to the probe. When multiple `demon:*` tags are present, the attempt contributes to each technique's pass-rate and inference-count independently.

patriciapampanelli:
- The digest must surface empty `technique_tags` probes in their own explicit bucket

patriciapampanelli:
Digest grouping (initial cut):

- For the first landing, the digest aggregates at the leaf MISP level: one row per `demon:*` tag for technique, one row per intent value for intent.
- Grouping logic should be factored so the leaf-level rule can be swapped for a config-driven selector later, without restructuring the builder.
- Higher-level rollups (taxonomy, family) and filters are out of scope here; they land in #1707.
- Frontend dynamic exploration lives in #1705.

PR #1719, opened 2026-04-27
feature/persist technique tags

### LINKED ISSUE #1703: feature: persist technique tags per-attempt in `report.jsonl`
`Probe.tags` declare MISP techniques but don't propagate to `report.jsonl` attempt rows, so filtering and aggregation by technique isn't possible today.

Scope:
- Stamp `probe.tags` onto each attempt via `Probe._attempt_prestore_hook`, mirroring intent.
- Extend `Attempt.__init__` and `as_dict()` to serialize tags in the attempt row.
- Emit `eval_technique` aggregate in `evaluators/base.py` mirroring `eval_intent`.

patriciapampanelli:
## Summary
- Persist ordered `demon:*` probe technique tags on every attempt row as `technique_tags`.
- Emit `eval_technique` aggregate rows mirroring the existing `eval_intent` report pattern.
- Preserve `technique_tags: []` as a first-class state for probes without technique tags.
- Preserve T&I metadata when buffs derive new attempts.

## Test plan
- [X] `python3 -m pytest tests/test_attempt.py::test_json_serialize tests/test_attempt.py::test_attempt_technique_tags_serialize_and_copy tests/test_internal_structures.py -k "technique_tags or eval_technique" -v`
- [X] `python3 -m pytest tests/test_attempt.py tests/langservice/probes/test_probes_base.py tests/test_internal_structures.py -v`

## Notes
- Resolves #1703
- Depends on #1691 
- Legacy reports are not migrated: missing `technique_tags` means a pre-change report, while `technique_tags: []` means a new report where the probe had no `demon:*` technique tag.
- Out of scope: report digest / HTML changes for #1704.

### DIFF
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -27,21 +27,24 @@ import garak.resources.theme
 CI_DISPLAY_MIN_WIDTH = 0.001
 
 
-def _probe_plugin_path(probe_name: str) -> str:
-    if probe_name.startswith("probes."):
-        return probe_name
-    return f"probes.{probe_name}"
+# Prefixes of probe tags that count as technique tags for `eval_technique` reporting.
+# Currently only the `demon:` taxonomy is treated as a technique; this list is expected
+# to grow as additional taxonomies are adopted.
+TECHNIQUE_PREFIXES = ["demon:"]
 
 
-def _probe_demon_tags(probe_name: str) -> list[str]:
-    tags = garak._plugins.PluginCache.plugin_info(_probe_plugin_path(probe_name))[
-        "tags"
-    ]
+def _probe_technique_tags(probe_name: str) -> list[str]:
+    """Return probe tags that match an active technique taxonomy prefix."""
+    tags = garak._plugins.PluginCache.plugin_info(f"probes.{probe_name}")["tags"]
     if not isinstance(tags, list):
         raise TypeError(f"Expected list of probe tags for {probe_name}")
     if not all(isinstance(tag, str) for tag in tags):
         raise TypeError(f"Expected string probe tags for {probe_name}")
-    return [tag for tag in tags if tag.startswith("demon:")]
+    return [
+        tag
+        for tag in tags
+        if any(tag.startswith(prefix) for prefix in TECHNIQUE_PREFIXES)
+    ]
 
 
 class Evaluator:
@@ -263,7 +266,9 @@ class Evaluator:
 
             detectors_to_eval.update(attempt_detectors)
             attempt_techniques = (
-                _probe_demon_tags(attempt.probe_classname) if attempt_detectors else []
+                _probe_technique_tags(attempt.probe_classname)
+                if attempt_detectors
+                else []
             )
             for attempt_detector in attempt_detectors:
                 detector_to_attempt_ids[attempt_detector].append(idx)
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -13,6 +13,7 @@ from typing import Iterable, Optional, Tuple, List
 from colorama import Fore, Style
 
 from garak import _config
+import garak._plugins
 import garak.attempt
 import garak.analyze.calibration
 import garak.analyze.detector_metrics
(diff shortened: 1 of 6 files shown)


jmartin-tech (changes requested):
Per separate discussion, since tags are part of probes and the probe is already referenced by the attempt, this can be refactored to read these values from the plugin cache and should likely inject a copy of the plugin metadata into the `report.jsonl` for use in reprocessing digests.

This avoids duplications in the attempts and also provides for easier maintenance and detection of comparability for runs.

patriciapampanelli:
> Per separate discussion, since tags are part of probes and the probe is already referenced by the attempt, this can be refactored to read these values from the plugin cache and should likely inject a copy of the plugin metadata into the `report.jsonl` for use in reprocessing digests.
> 
> This avoids duplications in the attempts and also provides for easier maintenance and detection of comparability for runs.

Updated this to stop persisting `technique_tags` on each `Attempt`. The `eval_technique` aggregate now resolves `demon:*` probe tags via `PluginCache.plugin_info(probe_classpath)` at evaluator time, using the existing probe identity carried by the attempt.

This keeps plugin metadata in the cache as the source of truth. When #1720 adds the persisted plugin cache header to `report.jsonl`, this lookup remains the single place to adapt without reintroducing per-attempt metadata.

garak/evaluators/base.py
```diff
@@ -26,6 +27,23 @@
 CI_DISPLAY_MIN_WIDTH = 0.001
 
 
+def _probe_plugin_path(probe_name: str) -> str:
+    if probe_name.startswith("probes."):
+        return probe_name
+    return f"probes.{probe_name}"
```
jmartin-tech: Can you add some context around when this is inconsistent?  I am surprised to see this method here.

garak/evaluators/base.py
```diff
@@ -26,6 +27,23 @@
...
+
+
+def _probe_demon_tags(probe_name: str) -> list[str]:
```
jmartin-tech: While currently the taxonomy is `demon` we should probably name this as `_probe_technique_tags` and build this based on some set of active taxonomies. Initially can see this being a hardcoded constant `demon:` but I expect this will grow. 

```
TECHNIQUE_PREFIXES = [ "demon:" ] 
```

jmartin-tech (commented):
Some initial questions.

Testing is in progress to get a clear example of output change this PR represents.

garak/evaluators/base.py
patriciapampanelli: Done. I've renamed to `_probe_technique_tags` and added `TECHNIQUE_PREFIXES = ["demon:"]` as a constant.

garak/evaluators/base.py
patriciapampanelli: There's no real inconsistency. The helper wasn't needed. Dropped `_probe_plugin_path` and inlined `f"probes.{probe_name}"`.

garak/evaluators/base.py
```diff
@@ -292,6 +321,39 @@ def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
...
+            technique_log_entry = {
+                "entry_type": "eval_technique",
+                "probe": self.probename,
+                "technique": technique,
+                "score": technique_score,
+                "aggregation": _config.reporting.group_aggregation_function,
+                "n_detectors": len(pass_rates),
+                "n_evaluations": evaluation_count,
+                "detectors_used": list(technique_relevant_detectors),
+            }
```
jmartin-tech, line 353: Looking at how this lands in the jsonl file I think this data may be better served as additional data in the `eval` entry.

```
{"entry_type": "eval", "probe": "ansiescape.AnsiEscaped", "detector": "apikey.ApiKey", "passed": 1274, "fails": 0, "nones": 6, "total_evaluated": 1274, "total_processed": 1280, "confidence_method": "bootstrap", "confidence": 0.95, "confidence_upper": 0.0, "confidence_lower": 0.0}
{"entry_type": "eval_technique", "probe": "ansiescape.AnsiEscaped", "technique": "demon:Possible_worlds:World_building:Scenarios", "score": 1.0, "aggregation": "lower_quartile", "n_detectors": 1, "n_evaluations": 1274, "detectors_used": ["apikey.ApiKey"]}
{"entry_type": "eval_technique", "probe": "ansiescape.AnsiEscaped", "technique": "demon:Stratagems:Meta-prompting:Ask_for_examples", "score": 1.0, "aggregation": "lower_quartile", "n_detectors": 1, "n_evaluations": 1274, "detectors_used": ["apikey.ApiKey"]}
```

The values reported so far seem to be linked to the detector and really do not provide detail that is needed to `resuse` the data for consolidated calculations, consider the case above, `lower_quartile` shift are not well suited to merging the lower quartile score or each probe/detector pair into a grouped score.

Simply injecting the list of `techniques` related to the probe/detector pair seems like it conveys all the same data connected to the pass/fail counts and totals in a way that they can be used in aggregate functions by later calculation.
```
{"entry_type": "eval", "probe": "ansiescape.AnsiEscaped", "detector": "apikey.ApiKey", "passed": 1274, "fails": 0, "nones": 6, "total_evaluated": 1274, "total_processed": 1280, "confidence_method": "bootstrap", "confidence": 0.95, "confidence_upper": 0.0, "confidence_lower": 0.0, "techniques": ["demon:Possible_worlds:World_building:Scenarios", "demon:Stratagems:Meta-prompting:Ask_for_examples"]}
```

Am I missing some piece of the puzzle here?

patriciapampanelli:
Closing in favor of #1783.

After syncing with @jmartin-tech, we decided the technique aggregation should live entirely in the digest (in-memory, resolving probe `demon:*` tags via `PluginCache`) instead of persisting `technique_tags` per attempt and emitting `eval_technique` aggregate rows.

What was in scope here (per-attempt `technique_tags` + `eval_technique` aggregate entries) is being dropped. The intent × technique cross-summary will be produced from the existing `eval` entries.

PR #1723, opened 2026-04-29
report: snapshot plugin cache into report jsonl

### LINKED ISSUE #1720: report: snapshot plugin cache into `report.jsonl` at run start
`report.jsonl` should embed the `PluginCache` metadata for every plugin used in a run, written at run start, so that `report_digest` can rebuild the digest without consulting the live cache.

Today `report_digest.build_digest` calls `garak._plugins.PluginCache.plugin_info(...)` at rebuild time. The digest therefore reflects whatever the local cache holds when the rebuild runs, not the state that produced the run. If a plugin parameter changed between two runs the report does not record it and the rebuild silently uses the new values.

Proposed change:
- At run start, snapshot the resolved `PluginCache` entries for every plugin in use in the run: probes, detectors, generators, buffs, and harness.
- Write the snapshot as a header record at the top of `report.jsonl`, before any attempt rows.
- `report_digest` reads plugin metadata (tags, descriptions, configured params) from the header instead of `PluginCache.plugin_info`.

Benefits:
- Digest rebuild becomes self-contained. No live cache lookup, no dependency on the local cache being in sync, no plugin instantiation.
- Every report records exactly which plugins ran and with which parameters.
- Diffing two reports shows which plugin parameter changed between runs.

Related:
- #1719: technique tag persistence; will consume this header once it lands, but does not block on it.
- #1703: original technique tags persistence request.

patriciapampanelli:
Adds a `plugin_cache` JSONL record during `Harness.run` so each report carries the resolved metadata for the harness, generator, probes, detectors, and buffs used in that run.

`report_digest` now reads plugin metadata from the report header first, making digest rebuilds self-contained and independent of the local live `PluginCache`. Legacy reports without a `plugin_cache` record still fall back to the live cache, and the digest records the source in `meta.plugin_cache_source`.

This slice intentionally leaves follow-up work for aggregation merge/validation, CI rebuild preservation, AVID `probe_tags`, and IntentProbe detector parity.

Closes #1720

### DIFF
--- a/garak/analyze/report_digest.py
+++ b/garak/analyze/report_digest.py
@@ -50,6 +50,7 @@ def _parse_report(reportfile: IO):
     payloads = []
     setup = defaultdict(str)
     init = {}
+    plugin_cache = None
 
     for record in [json.loads(line.strip()) for line in reportfile if line.strip()]:
         if record["entry_type"] == "eval":
@@ -68,8 +69,13 @@ def _parse_report(reportfile: IO):
                 + "  "
                 + pprint.pformat(record, sort_dicts=True, width=60)
             )
+        elif record["entry_type"] == "plugin_cache":
+            if plugin_cache is None:
+                plugin_cache = {}
+            for category, entries in record.get("plugin_cache", {}).items():
+                plugin_cache.setdefault(category, {}).update(entries)
 
-    return init, setup, payloads, evals
+    return init, setup, payloads, evals, plugin_cache
 
 
 def _report_header_content(report_path, init, setup, payloads, config=_config) -> dict:
@@ -90,7 +96,29 @@ def _report_header_content(report_path, init, setup, payloads, config=_config) -
     return header_content
 
 
-def _init_populate_result_db(evals, taxonomy=None):
+def _resolve_plugin_info(plugin_classpath, report_plugin_cache, required_fields=None):
+    if report_plugin_cache is None:
+        return garak._plugins.PluginCache.plugin_info(plugin_classpath), "live_cache"
+
+    category = plugin_classpath.split(".")[0]
+    meta = report_plugin_cache.get(category, {}).get(plugin_classpath)
+    if meta is None:
+        raise ValueError(f"plugin_cache missing metadata for {plugin_classpath}")
+
+    missing = [
+        field
+        for field in (required_fields or ())
+        if field not in meta or meta[field] is None
+    ]
+    if missing:
+        raise ValueError(
+            f"plugin_cache metadata for {plugin_classpath} missing fields: {missing}"
+        )
+
+    return meta, "header"
+
+
+def _init_populate_result_db(evals, taxonomy=None, report_plugin_cache=None):
 
     conn = sqlite3.connect(":memory:")
     cursor = conn.cursor()
@@ -128,7 +156,12 @@ def _init_populate_result_db(evals, taxonomy=None):
         groups = []
         if taxonomy is not None:
             # get the probe tags
-            tags = garak._plugins.PluginCache.plugin_info(f"probes.{pm}.{pc}")["tags"]
+            meta, _ = _resolve_plugin_info(
+                f"probes.{pm}.{pc}",
+                report_plugin_cache,
+                required_fields=("tags",),
+            )
+            tags = meta["tags"]
             for tag in tags:
                 if tag.split(":")[0] == taxonomy:
                     groups.append(":".join(tag.split(":")[1:]))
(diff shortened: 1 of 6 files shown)


garak/analyze/report_digest.py
```diff
@@ -68,8 +69,13 @@ def _parse_report(reportfile: IO):
...
                 + pprint.pformat(record, sort_dicts=True, width=60)
             )
+        elif record["entry_type"] == "plugin_cache":
+            if plugin_cache is None:
+                plugin_cache = {}
+            for category, entries in record.get("plugin_cache", {}).items():
+                plugin_cache.setdefault(category, {}).update(entries)
 
-    return init, setup, payloads, evals
+    return init, setup, payloads, evals, plugin_cache
```
jmartin-tech, line 85: It might be good to simply supply the system plugin cache when a cache entry in not found in the file.

```suggestion
    if len(plugin_cache) <= 0:
        from copy import deepcopy
        plugin_cache = deepcopy(garak._plugins.PluginCache.instance())
        plugin_cache["version"] = garak.__version__
    return init, setup, payloads, evals, plugin_cache
```

garak/analyze/report_digest.py
```diff
@@ -90,7 +96,29 @@ def _report_header_content(report_path, init, setup, payloads, config=_config) -
     return header_content
 
 
-def _init_populate_result_db(evals, taxonomy=None):
+def _resolve_plugin_info(plugin_classpath, report_plugin_cache, required_fields=None):
+    if report_plugin_cache is None:
+        return garak._plugins.PluginCache.plugin_info(plugin_classpath), "live_cache"
```
jmartin-tech: not required if cache is always set
```suggestion
```

garak/harnesses/base.py
```diff
@@ -47,6 +47,36 @@ def _initialize_runtime_services():
...
+            f"{plugin_instance.__class__.__module__}."
+            f"{plugin_instance.__class__.__name__}"
+        ).replace("garak.", "")
+        category = classpath.split(".")[0]
+        meta = _plugins.PluginCache.plugin_info(classpath)
+        snapshot.setdefault(category, {})[classpath] = meta
+
+    if not snapshot:
+        return
+
```
jmartin-tech, line 65:
```suggestion

    snapshot["version"] = garak.__version__
```

garak/analyze/report_digest.py
```diff
@@ -128,7 +156,12 @@ def _init_populate_result_db(evals, taxonomy=None):
         groups = []
         if taxonomy is not None:
             # get the probe tags
-            tags = garak._plugins.PluginCache.plugin_info(f"probes.{pm}.{pc}")["tags"]
+            meta, _ = _resolve_plugin_info(
```
jmartin-tech:
```suggestion
            meta = _resolve_plugin_info(
```

garak/analyze/report_digest.py
```diff
@@ -250,9 +283,15 @@ def _get_probe_result_summaries(cursor, probe_group) -> List[tuple]:
     return res.fetchall()
 
 
-def _get_probe_info(probe_module, probe_class, absolute_score) -> dict:
+def _get_probe_info(
+    probe_module, probe_class, absolute_score, report_plugin_cache=None
+) -> dict:
     probe_classpath = f"probes.{probe_module}.{probe_class}"
-    probe_plugin_info = garak._plugins.PluginCache.plugin_info(probe_classpath)
+    probe_plugin_info, _ = _resolve_plugin_info(
```
jmartin-tech:
```suggestion
    probe_plugin_info = _resolve_plugin_info(
```

garak/analyze/report_digest.py
```diff
@@ -298,12 +337,15 @@ def _get_probe_detector_details(
...
     ci_lower=None,
     ci_upper=None,
+    report_plugin_cache=None,
 ) -> dict:
     calibration_used = False
     detector = re.sub(r"[^0-9A-Za-z_.]", "", detector)
     detector_module, detector_class = detector.split(".")
-    detector_cache_entry = garak._plugins.PluginCache.plugin_info(
-        f"detectors.{detector_module}.{detector_class}"
+    detector_cache_entry, _ = _resolve_plugin_info(
```
jmartin-tech:
```suggestion
    detector_cache_entry = _resolve_plugin_info(
```

garak/analyze/report_digest.py
```diff
@@ -90,7 +96,29 @@ def _report_header_content(report_path, init, setup, payloads, config=_config) -
...
+        field
+        for field in (required_fields or ())
+        if field not in meta or meta[field] is None
+    ]
+    if missing:
+        raise ValueError(
+            f"plugin_cache metadata for {plugin_classpath} missing fields: {missing}"
+        )
+
+    return meta, "header"
```
jmartin-tech:
```suggestion
    return meta
```

garak/analyze/report_digest.py
```diff
@@ -494,6 +540,9 @@ def build_digest(report_filename: str, config=_config):
     report_digest["meta"]["setup"]["reporting.taxonomy"] = taxonomy
     report_digest["meta"]["calibration_used"] = calibration_used
     report_digest["meta"]["aggregation_unknown"] = aggregation_unknown
+    report_digest["meta"]["plugin_cache_source"] = (
+        "live_cache" if report_plugin_cache is None else "header"
```
jmartin-tech:
```suggestion
        report_plugin_cache["version"]
```

jmartin-tech (commented):
Comments are untested, I think the gets across the idea mentioned to simply `version` the cache value being used to document where it came from.

jmartin-tech (approved):

PR #1726, opened 2026-05-01
[garak] fix: file handle leak in evaluators/base.py

mitre88:
## Summary
Fixes a file handle leak in `garak/evaluators/base.py` where the hitlogfile was opened but never closed using a context manager.

## Changes
- Wrapped file write operations in a `with _config.transient.hitlogfile:` context manager to ensure proper file handle closure

## Testing
- Verified the code syntax is correct by checking Python parsing
- The fix ensures file handles are properly closed after each write

### DIFF
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -115,29 +115,29 @@ class Evaluator:
                                 buffering=1,
                                 encoding="utf-8",
                             )
-
-                        triggers = attempt.notes.get("triggers", None)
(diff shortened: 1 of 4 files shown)


jmartin-tech:
Closing as improperly scoped and invalid.  The hitlog file is is written but not managed in the evaluator scope and the added commits related to the beast attack helper have been repeatedly rejected as duplication of another PR that is still in progress.

PR #1733, opened 2026-05-02
Fix ggml generator call output handling

### LINKED ISSUE #1366: ggml generator crashes with AttributeError: 'GarakSubConfig' object has no attribute 'verbose'
Steps to reproduce

  How'd you do it?

  1. Install garak v0.12.0.pre1
  2. Set up llama.cpp with a GGUF model
  3. Set GGML_MAIN_PATH environment variable to point to llama-cli executable
  4. Run: python3 -m garak --model_type ggml --model_name "/path/to/model.gguf" --probes encoding.InjectBase64
  5. Observe the AttributeError during execution

  Environment: Any system running garak v0.12.0.pre1 with ggml generator

  Were you following a specific guide/tutorial or reading documentation?

  Following the garak documentation for ggml generator usage: https://reference.garak.ai/en/latest/garak.generators.ggml.html

  Expected behavior

  The ggml generator should successfully load and run probes against the GGUF model without configuration errors.

  Current behavior

  The ggml generator crashes with:
  AttributeError: 'GarakSubConfig' object has no attribute 'verbose'

  garak version

  v0.12.0.pre1

zw5:
## Summary

`GgmlGenerator._call_model()` crashed before invoking `llama-cli` when `_config.system.verbose` was absent, the failure reported in #1366. After that guard, the successful-call path also tried to call `lstrip()` on the `Conversation` object while removing echoed prompt text, so a successful subprocess response could become `[None]`.

This PR reads `verbose` with a default of `0`, stores the prompt text before the subprocess call, and strips the echoed prompt from that string.

## Reproducer

I added `test_call_model_removes_echoed_prompt`, which uses a temporary GGUF header file and a fake `llama-cli` executable. On upstream `main` at `81a82ab1e043baefc6fde029a214e8af66485f34`, this command fails with `AttributeError: 'GarakSubConfig' object has no attribute 'verbose'`:

```bash
uv run --python 3.12 --extra tests python -m pytest tests/generators/test_ggml.py -q
```

## Verification

```bash
uv run --python 3.12 --extra tests python -m pytest tests/generators/test_ggml.py -q
uv run --python 3.12 --extra tests python -m pytest tests/generators/test_ggml.py tests/generators/test_generators_base.py tests/generators/test_openai_compatible.py -q
uv run --python 3.12 --extra tests python -m pytest tests/generators/test_nim.py -q
uv run --python 3.12 --extra lint black --check garak/generators/ggml.py tests/generators/test_ggml.py
git diff --check
```

Fixes #1366.

### DIFF
--- a/garak/generators/ggml.py
+++ b/garak/generators/ggml.py
@@ -122,9 +122,10 @@ class GgmlGenerator(Generator):
         # test all params for None type
         command.extend(self._command_args_list())
         command = [str(param) for param in command]
-        if _config.system.verbose > 1:
+        if getattr(_config.system, "verbose", 0) > 1:
             print("GGML invoked with", command)
         try:
+            prompt_text = prompt.last_message().text
             result = subprocess.run(
                 command,
                 stdout=subprocess.PIPE,
@@ -132,7 +133,7 @@ class GgmlGenerator(Generator):
                 check=self.exception_on_failure,
             )
             output = result.stdout.decode("utf-8")
-            output = re.sub("^" + re.escape(prompt.lstrip()), "", output.lstrip())
+            output = re.sub("^" + re.escape(prompt_text.lstrip()), "", output.lstrip())
             self.first_call = False
             return [Message(output)]
         except subprocess.CalledProcessError as err:
--- a/tests/generators/test_ggml.py
+++ b/tests/generators/test_ggml.py
@@ -81,39 +81,26 @@ def test_command_args_list():
         assert type(g) is garak.generators.ggml.GgmlGenerator
 
 
-def test_call_model_removes_echoed_prompt(tmp_path, monkeypatch):
-    model_path = tmp_path / "test_model.gguf"
-    model_path.write_bytes(garak.generators.ggml.GGUF_MAGIC)
+def test_call_model_removes_echoed_prompt(mocker):
+    with tempfile.NamedTemporaryFile(
+        mode="wb", suffix="_test_model.gguf", delete=False
+    ) as file:
+        file.write(garak.generators.ggml.GGUF_MAGIC)
+        file.close()
 
-    fake_llama = tmp_path / "llama-cli"
-    fake_llama.write_text("", encoding="utf-8")
-    monkeypatch.setenv(garak.generators.ggml.ENV_VAR, str(fake_llama))
+        import subprocess
+        from unittest.mock import MagicMock
 
-    generator = garak.generators.ggml.GgmlGenerator(str(model_path))
-    prompt = Conversation([Turn("user", Message("prompt:"))])
-    captured = {}
+        prompt = Conversation([Turn("user", Message("prompt:"))])
 
-    def fake_run(command, stdout, stderr, check):
-        captured["command"] = command
-        assert stdout == garak.generators.ggml.subprocess.PIPE
-        assert stderr == garak.generators.ggml.subprocess.PIPE
-        assert check == generator.exception_on_failure
-        prompt_arg = command[command.index("-p") + 1]
-        return garak.generators.ggml.subprocess.CompletedProcess(
-            args=command,
-            returncode=0,
-            stdout=f"{prompt_arg}response\n".encode("utf-8"),
-            stderr=b"",
-        )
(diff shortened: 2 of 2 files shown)


zw5:

recheck

zw5:

zw5:
recheck

tests/generators/test_ggml.py
```diff
@@ -78,3 +79,28 @@ def test_command_args_list():
...
+    monkeypatch.setenv(garak.generators.ggml.ENV_VAR, str(fake_llama))
+
+    generator = garak.generators.ggml.GgmlGenerator(str(model_path))
+    prompt = Conversation([Turn("user", Message("prompt:"))])
+
+    result = generator._call_model(prompt)
+
+    assert len(result) == 1
+    assert isinstance(result[0], Message)
+    assert result[0].text == "response\n"
```
jmartin-tech: Instead of executing injected code in another process with this test it would be better to simply intercept the subprocess.run() method with a mock response.

The current pattern relies on *nix shell specific execution patterns causing Windows testing to fail.

```suggestion
   with tempfile.NamedTemporaryFile(
        mode="wb", suffix="_test_model.gguf", delete=False
    ) as file:
        file.write(garak.generators.ggml.GGUF_MAGIC)
        file.close()

        import subprocess
        from unittest.mock import MagicMock

        prompt = Conversation([Turn("user", Message("prompt:"))])

        mock_response = MagicMock()
        test_response = "response"
        mock_response.stdout = f"{prompt.last_message().text}{test_response}".encode()
        mocker.patch.object(subprocess, "run", return_value=mock_response)
        model_path = tmp_path / "test_model.gguf"
        model_path.write_bytes(garak.generators.ggml.GGUF_MAGIC)

        generator = garak.generators.ggml.GgmlGenerator(file.name)
        result = generator._call_model(prompt)

        assert len(result) == 1
        assert isinstance(result[0], Message)
        assert result[0].text == test_response
```

garak/generators/ggml.py
```diff
@@ -122,17 +122,18 @@ def _call_model(
         # test all params for None type
         command.extend(self._command_args_list())
         command = [str(param) for param in command]
-        if _config.system.verbose > 1:
+        if getattr(_config.system, "verbose", 0) > 1:
             print("GGML invoked with", command)
         try:
+            prompt_text = prompt.last_message().text
```
jmartin-tech, line 128: Nice catch, the conversational prompt format change was missed here, this provides basic support for ggml models and can be enhanced later if conversation prompt formats are needed by the community.

jmartin-tech (changes requested):
Thank you, this will help enable more local use cases.

The guard looks reasonable, `verbose` options will need a more comprehensive adjustment the option is still parsed by cli but not guaranteed with a default initialization value in the core `_config` objects.

The test added here however needs to be refactored for cross-platform validation.

zw5:
I amended the branch at `48bef7b9`,  made the GGML regression test cross-platform, 
Also I changed the test to monkeypatch `garak.generators.ggml.subprocess.run`. then kept the same behavior coverage.

jmartin-tech:
The revised test is very tightly coupled to things that are not defined by the `garak` codebase. I would recommend validated the provided suggestion, while it shifts from `monkeypatch` to `mocker` it avoids assumptions about the underlying implementation.

zw5:
> The revised test is very tightly coupled to things that are not defined by the `garak` codebase. I would recommend validated the provided suggestion, while it shifts from `monkeypatch` to `mocker` it avoids assumptions about the underlying implementation.

Thanks for bearing with me here, sorry for not implementing your suggestion from the start lmao. I only changed 2 lines from your suggestion, 
```-model_path = tmp_path / "test_model.gguf"                                                                                                                                                                         -model_path.write_bytes(garak.generators.ggml.GGUF_MAGIC)```
Because model_path is never referenced afterwards, (due to the generator being built from file.name). I'm still not sure if I should leave them in.

jmartin-tech (approved):
Look good, thanks for the detailed check you are correct that those lines should have been removed. This should land shortly.

PR #1736, opened 2026-05-03
gracefully skip probes whose detectors are unreachable

### LINKED ISSUE #1033: Fall back when online services can't be reached.
If a remote resource _e.g._ Huggingface model used in a detector cannot be reached, we should either fail gracefully or throw an error and offer a substitution. This will be important for running garak in airgapped/offline systems. 

## Steps to reproduce

Run `garak -m test` without an internet connection or with SSL intercept. 

## Expected behavior

When a website _e.g._ huggingface.co cannot be reached, gracefully pass to another detector/probe instead of failing with a raised ValueError. 

## Current behavior

Run terminates.

### garak version

0.10.0

ChrisJr404:
Closes #1033.

## Background

When a probe's primary detector can't be loaded — typically because the detector pulls a model from `huggingface.co` and the host is offline / behind an SSL-intercepting proxy — the run terminated entirely. End-to-end repro on `main`:

```
$ HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 HF_HOME=/tmp/no-cache \
    garak --model_type test.Blank --probes lmrc.Bullying,lmrc.Anthropomorphisation
🦜 loading generator: Test: Blank
🕵️  queue of probes: lmrc.Anthropomorphisation, lmrc.Bullying
... (Anthropomorphisation runs fine — substring detector, no HF) ...
 detector load failed: unsafe_content.ToxicCommentModel, skipping >>
ValueError: No detectors, nothing to do
```

The detector load itself returned `False` cleanly (good), but `garak/harnesses/base.py:run()` then raises `ValueError("No detectors, nothing to do")`, which propagates out of the per-probename loop in `garak/harnesses/probewise.py:run()` and terminates every other probe the user queued.

@leondz's response on the issue: *"in the very worst case_ skip that detector"*. This PR implements that contract end-to-end.

## Change

Two-part adjustment in `garak/harnesses/probewise.py`:

1. **Primary-detector fallback**: if the primary detector fails to load, try the probe's `extended_detectors` list even when `--extended-detectors` mode isn't on. Most probes that ship an HF detector also list a non-HF backup (substring match, length check), so this gives the run a chance to produce results in airgapped environments before giving up entirely. A clean log message tells the user which detector is being substituted.

2. **Per-probe skip**: if the detector list is still empty after the fallback, log a warning and `continue` to the next probe instead of letting `harnesses.base.Harness.run()` raise `ValueError`.

## Verification

```
$ HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 HF_HOME=/tmp/no-cache \
    garak --model_type test.Blank --probes lmrc.Bullying,lmrc.Anthropomorphisation
🦜 loading generator: Test: Blank
🕵️  queue of probes: lmrc.Anthropomorphisation, lmrc.Bullying
...
lmrc.Anthropomorphisation                                                                lmrc.Anthro: PASS  ok on   30/  30
 detector load failed: unsafe_content.ToxicCommentModel, skipping >>
⚠️  no detectors available for probe probes.lmrc.Bullying; skipping. This usually means a remote model/API was unreachable — check the garak log for the underlying load error.
✔️  garak run complete in 4.89s
```

Anthropomorphisation runs to completion; Bullying logs a clean warning and is skipped; the run terminates normally with a report.

## Test

Added `test_probewise_skips_probes_with_no_loadable_detectors` in `tests/harnesses/test_harnesses.py`. Patches `_load_detector` to fail for one probe and succeed for another, asserts the successful probe runs and the failing one is skipped without raising. All 4 harness tests pass:

```
tests/harnesses/test_harnesses.py::test_harness_structure[harnesses.probewise.ProbewiseHarness] PASSED
tests/harnesses/test_harnesses.py::test_harness_structure[harnesses.pxd.PxD] PASSED
tests/harnesses/test_harnesses.py::test_probewise_skips_probes_with_no_loadable_detectors PASSED
tests/harnesses/test_harnesses.py::test_harness_modality_match PASSED
```

## Notes

- Diff is `+102 / -1` across two files. No public API change.
- The extended-detector fallback is silently dedup'd against the primary fallback, so a probe that lists an HF detector both as `primary_detector` AND in `extended_detectors` doesn't get loaded twice.
- I considered the alternative of changing `harnesses/base.py:run()` to log a warning instead of raising `ValueError` for empty detector lists, but that changes the contract for every direct caller (including pxd harness and any third-party harness), so I kept the change scoped to probewise. Happy to widen the change if you'd prefer the lower-level fix.
- Added emoji-prefixed warnings (`⚠️`) to match the existing emoji-driven CLI output style.

### DIFF
--- a/garak/harnesses/probewise.py
+++ b/garak/harnesses/probewise.py
@@ -87,8 +87,32 @@ class ProbewiseHarness(Harness):
                 d = self._load_detector(probe.primary_detector)
                 if d:
                     detectors = [d]
+                else:
+                    # Primary detector failed to load (typically a remote
+                    # resource — HF model, online API — was unreachable).
+                    # Try the probe's extended detectors as a fallback even
+                    # outside of `--extended-detectors` mode so the run can
+                    # still produce results in airgapped / SSL-intercept
+                    # environments. See issue #1033.
+                    for detector_name in sorted(probe.extended_detectors):
+                        d = self._load_detector(detector_name)
+                        if d:
+                            detectors.append(d)
+                    if detectors:
+                        msg = (
+                            f"primary detector {probe.primary_detector!r} for probe "
+                            f"{probename} unavailable; falling back to extended "
+                            f"detector(s): {sorted(probe.extended_detectors)}"
+                        )
+                        print(f"⚠️  {msg}")
+                        logging.warning(msg)
                 if _config.plugins.extended_detectors is True:
                     for detector_name in sorted(probe.extended_detectors):
+                        if any(
+                            type(existing).__name__ == detector_name.split(".")[-1]
+                            for existing in detectors
+                        ):
+                            continue
                         d = self._load_detector(detector_name)
                         if d:
                             detectors.append(d)
@@ -106,6 +130,20 @@ class ProbewiseHarness(Harness):
                     d = self._load_detector(detector_name)
                     if d:
                         detectors.append(d)
+
+            if not detectors:
+                # Every detector for this probe failed to load. Log a
+                # clean message and move on to the next probe instead of
+                # letting the harness raise ValueError and terminate the
+                # whole run. See issue #1033.
+                msg = (
+                    f"no detectors available for probe {probename}; skipping. "
+                    f"This usually means a remote model/API was unreachable — "
+                    f"check the garak log for the underlying load error."
+                )
+                print(f"⚠️  {msg}")
+                logging.warning(msg)
+                continue
 
             super().run(model, [probe], detectors, evaluator, announce_probe=False)
             # del probe, h, detectors
--- a/tests/harnesses/test_harnesses.py
+++ b/tests/harnesses/test_harnesses.py
@@ -3,10 +3,12 @@
 
 import pytest
 import importlib
+from unittest.mock import patch, MagicMock
 
-from garak import _plugins
+from garak import _plugins, _config
 
 import garak.harnesses.base
+import garak.harnesses.probewise
 
 HARNESSES = [
     classname for (classname, active) in _plugins.enumerate_plugins("harnesses")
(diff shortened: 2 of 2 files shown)


leondz:
Thanks. I think this will trigger a conversation reviewing the suggested behaviour but on the face of it it sound like are some clear benefits from this new functionality regardless. Will take a look.

ChrisJr404:

ChrisJr404:
recheck

jmartin-tech (commented):
Adding a note here that this PR does meet the request in the original issue. It is currently in a holding pattern however as detector selection is expected to change very soon. To avoid possibly issue during conflict resolution this has been deferred until the team can look at it holistically with other changes in flight.

lathuthavarajah:
Arriving here from #2115, which @jmartin-tech correctly closed as a duplicate of this — this PR does everything mine did and adds the primary→extended fallback.

One concrete thing I can offer while this is held: the fallback path this PR introduces isn't currently exercised by its test. `test_probewise_skips_probes_with_no_loadable_detectors` sets `extended_detectors = []` on both mock probes (lines 51 and 56) and `_config.plugins.extended_detectors = False` (line 75), so it covers the skip but never reaches the `else` branch where a failed primary falls back to extended detectors. Two cases that would pin the new behavior:

- primary fails, extended loads → probe runs on the extended detector (the fallback's purpose), and the warning names the substitution
- primary loads, extended fails → probe still runs on the working primary, i.e. the new dedup `continue` in the `extended_detectors is True` block doesn't drop it

The second one matters because the dedup compares `type(existing).__name__` against `detector_name.split(".")[-1]`, which matches on bare class name — two detectors from different modules sharing a class name would collide and the later one would be skipped.

I have both written against a mocked loader and am happy to send them as a patch to this branch, or just paste them in a comment, whichever is easier for you — no need to take my branch. Equally happy to leave it alone until the detector-selection work lands, since I realise this is deliberately parked.

Issue #1568, opened 2026-01-21
bug: npm pkghallu detector missing date field


leondz:
date field has invalid content for javascript/npm package hallucination detector

detector can generator log entries like these when parsing the dataset:

```
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_6 with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_a8 with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-1g with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_abp with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_abd with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_abh with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zaidan-ui/core with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_aj with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-_ab- with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-19 with unknown creation date
2026-01-15 12:02:28,001  WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'. Keeping package @zalastax/nolb-17 with unknown creation date
[...]

PR #1738, opened 2026-05-04
fix(detectors): keep packagehallucination entries with invalid dates

ChrisJr404:
## Summary

Relates to #1568.

The npm package-hallucination dataset contains rows whose `package_first_seen` field is missing (None) or carries an upstream registry error string in place of an ISO date — exactly what the issue body shows:

```
WARNING  Invalid package date format: Invalid isoformat string: 'Error: 404 Client E'.
         Keeping package @zalastax/nolb-_6 with unknown creation date
```

`_load_package_list()` calls `datetime.fromisoformat()` on each value in a `try/except ValueError`. The log message inside the `except` branch claims the package is being **kept** with an unknown creation date, but the package was never actually appended to `filtered_packages` — so it was silently dropped from the package set. Any model output that referenced one of those packages was then flagged as a hallucination, even though the package is real.

## Fix

Append the package to `filtered_packages` in the `except` branch, so the behavior matches the log line. Also extend the caught exceptions to include `TypeError` for the case where `package_first_seen` is `None` and the `[0:19]` slice raises before `fromisoformat`.

```diff
-                    except ValueError as e:
+                    except (TypeError, ValueError) as e:
+                        filtered_packages.append(pkg)
                         if not invalid_pkg_date_seen_flag:
                             logging.debug(
                                 "Invalid %s package date format: %s. Keeping package %s with unknown creation date. ..."
```

## Test

`test_load_package_list_keeps_packages_with_invalid_date` monkey-patches `datasets.load_dataset` with a fake dataset containing one valid ISO date, one error-string date, and one `None` date, then asserts all three packages end up in `detector.packages`. Without the fix, only the valid-date package is kept; with the fix, all three are.

```
$ pytest tests/detectors/test_detectors_packagehallucination.py::test_load_package_list_keeps_packages_with_invalid_date -v
PASSED
```

## Verification checklist

- [x] Commit signed off (`Signed-off-by: ChrisJr404 …`).
- [x] One narrow test added; no public API changes.
- [x] No change to behavior on rows with a valid ISO date.

### DIFF
--- a/garak/detectors/packagehallucination.py
+++ b/garak/detectors/packagehallucination.py
@@ -80,7 +80,15 @@ class PackageHallucinationDetector(Detector, ABC):
                         )
                         if first_seen <= cutoff:
                             filtered_packages.append(pkg)
-                    except ValueError as e:
+                    except (TypeError, ValueError) as e:
+                        # The dataset can contain rows whose package_first_seen field is
+                        # missing (None) or carries an upstream registry error message
+                        # ("Error: 404 Client Error: ...") instead of an ISO date. The log
+                        # text below has long claimed we keep these packages with an
+                        # unknown creation date; this branch now actually does that, so
+                        # genuine packages are not flagged as hallucinations just because
+                        # their dataset metadata is incomplete.
+                        filtered_packages.append(pkg)
                         if not invalid_pkg_date_seen_flag:
                             logging.debug(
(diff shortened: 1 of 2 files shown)


jmartin-tech (approved):
Looks good, nice catch.

leondz (approved):
lgtm

PR #1743, opened 2026-05-05
remove action workflow no longer used as the standard for the project

jmartin-tech:
`cla-assistant` is no longer the enforcement pattern used to acknowledge DCO.  As noted in the DCO itself, `sign-off` in commit messages is explicit acknowledgment of the DCO and is now enforce on all PR commits via https://github.com/apps/dco


erickgalinkin (approved):
Yes

leondz (approved):

PR #1747, opened 2026-05-05
test: add unit tests for phrasing and sata probe modules

markd88:
## Summary
Add dedicated test files for 2 probe modules with unique data-processing logic:

- **phrasing**: pruned vs full prompt count comparison, `soft_probe_prompt_cap` constraint
- **sata**: NLTK masking pipeline (`_nltk_mask_payloads`), mask token and template verification

Generic instantiation and prompt-not-empty checks are already covered by `plugins/test_plugin_load.py`.

## Test plan
- [x] All 5 new tests pass
- [x] Existing tests unaffected


jmartin-tech (changes requested):
Most of these tests are not validating anything specific to the probe, and scenarios like basic instantiation are already part of abstract probe testing in `plugins/test_plugin_load.py`.

To be accepted these tests need to be pared down to validation of functionality unique to the probe module classes.

markd88:
@jmartin-tech  thanks for the review. I've pared down to only phrasing (prune behavior) and sata (NLTK masking pipeline). The donotanswer, grandma, and RTP modules are pure file loading or static data and don't have module specific processing logic to test. Ready for re-review.

jmartin-tech (approved):

PR #1749, opened 2026-05-06
fix: when tiktoken encoding is unclear fallback during estimation

jmartin-tech:
Estimation of token counting is fuzzy, when tiktoken does not have a mapping for a specific model name fallback to a default and avoid raise that would crash the run.

## Verification

- [ ] new unit tests pass
- [ ] `garak -t <target_type> -n <model_name>` execute test with a judge with a tiktoken known and unknown model backing the detector

### DIFF
--- a/garak/resources/red_team/evaluation.py
+++ b/garak/resources/red_team/evaluation.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: Portions Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-import logging
 import re
 import tiktoken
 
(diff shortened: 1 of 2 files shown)


garak/resources/red_team/evaluation.py
```diff
@@ -44,7 +47,12 @@ def process_output_on_topic_score(raw_output) -> float:
 
 
 def token_count(string: str, model_name: str) -> int:
-    encoding = tiktoken.encoding_for_model(model_name)
+    try:
+        encoding = tiktoken.encoding_for_model(model_name)
+    except KeyError as e:
+        msg = "Unable to determine encoding for '{model_name}' using default '{FALLBACK_MODEL_NAME}' for estimation"
+        logging.warning(msg)
```
jmartin-tech: This might get a little noisy, may be best to remove this.

```suggestion
```

PR #1750, opened 2026-05-06
housekeeping: amend arXiv `doc_uri`s from `html`  where available, else `abs`

leondz:
In order to expand data available at first hop, this PR:

* moves arXiv `doc_uri`s to the HTML version of the paper, for those cases where it's available
* corrects the plugin types to check in `AGENTS.md` and expresses intent there


jmartin-tech (approved):
Seems reasonable, I am interested in how much impact the agent.md update may have.

PR #1752, opened 2026-05-09
fix: add threading lock to prevent hitlog race condition

### LINKED ISSUE #1355: Race condition occurs when using parallel attempts while reading log file
Race condition occurs when parallel attempts lead to multiple threads accessing the same log file at once. This can cause errors; implementing a locking mechanism could help avoid the issue.

```
--- Logging error ---
  File "/opt/homebrew/Cellar/python@3.11/3.11.13/Frameworks/Python.framework/Versions/3.11/lib/python3.11/logging/__init__.py", line 1114, in emit
    self.flush()
  File "/opt/homebrew/Cellar/python@3.11/3.11.13/Frameworks/Python.framework/Versions/3.11/lib/python3.11/logging/__init__.py", line 1094, in flush
```

nuthalapativarun:
## Summary

- Add a module-level `threading.Lock` (`_hitlog_lock`) in `garak/evaluators/base.py`
- Wrap the hitlog file open+write block with the lock to serialise concurrent access
- Add regression tests: `test_hitlog_lock_exists` and `test_hitlog_concurrent_writes`

Fixes #1355

## Problem

When `parallel_attempts > 1`, multiple threads call `evaluate()` concurrently. All threads can pass the `hitlogfile is None` check simultaneously before any one of them opens the file, leading to:
- Multiple `open()` calls on the same file (potential `OSError`)
- Interleaved `write()` calls producing corrupt/truncated JSONL lines

## Fix

Wrap the entire check-open-write block with a `threading.Lock`:

```python
_hitlog_lock = threading.Lock()

# inside evaluate():
with _hitlog_lock:
    if _config.transient.hitlogfile is None or _config.transient.hitlogfile.closed:
        # open file ...
    _config.transient.hitlogfile.write(...)
```

## Verification

- [ ] `python -m pytest tests/test_hitlog.py -v` — all 3 tests pass
- [ ] `python -m pytest tests/ -x -q` — full suite passes
- [ ] Manual verify: hitlog file produced correctly with `parallel_attempts > 1`
- [ ] Manual verify: no regression in single-threaded usage

### DIFF
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -5,6 +5,7 @@ These describe evaluators for assessing detector results.
 
 import json
 import logging
+import threading
 from pathlib import Path
 
 from typing import Iterable, Optional, Tuple, List
@@ -22,6 +23,9 @@ import garak.resources.theme
 # Minimum CI width (in percentage points) to display in output
 # CIs narrower than this provide no meaningful uncertainty information
 CI_DISPLAY_MIN_WIDTH = 0.001
+
+# Lock for thread-safe hitlog file open and write operations
+_hitlog_lock = threading.Lock()
 
 
 class Evaluator:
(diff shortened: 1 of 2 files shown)


nuthalapativarun:

nuthalapativarun:
recheck

jmartin-tech:
This is invalid, only the harness call `evaluate()` and harnesses exist only in the core `process`.  Threading is not used in this project at all.

PR #1754, opened 2026-05-09
fix: make PromptInject generation params user-configurable

### LINKED ISSUE #1608: bug: max output tokens overwritten
### PromptInject max_tokens override
The PromptInject probe's `_generator_precall_hook` overwrites the generator's `max_tokens` value with a lower value from the probe's settings, causing reasoning models to truncate their output prematurely. This is particularly problematic for reasoning models that need longer context to complete their reasoning trace and answer.

**Code location:** https://github.com/NVIDIA/garak/blob/331c5601d8ccdc371561ff8bc6e8a78fc9c1c548/garak/probes/promptinject.py#L97-L108

Possible Fix - 

```python
    """apply PromptInject model inference parameters to probe"""
    map_attribs = (
        "temperature top_p frequency_penalty presence_penalty max_tokens".split()
    )
            probe_value = attempt.notes["settings"]["config_" + map_attrib]
            # For max_tokens, only override if generator value is lower (respect higher limits for reasoning models)
            if map_attrib == "max_tokens":
                generator_value = getattr(generator, map_attrib, 0)
                if generator_value > probe_value:
                    continue  # Keep the generator's higher max_tokens
            setattr(
                generator,
                map_attrib,
                probe_value,
            )
```

nuthalapativarun:
## Summary

- Extract hardcoded generation param list to module-level `_generation_params`
- Change `_generator_precall_hook` to read from `self.generation_params` instead of a hardcoded list
- Add `DEFAULT_PARAMS = Probe.DEFAULT_PARAMS | {"generation_params": _generation_params}` to both probe class definitions (`HijackX` and `HijackXFull`)
- Add `tests/probes/test_probes_promptinject.py` with 6 tests

Fixes #1608

## Problem

`_generator_precall_hook` unconditionally overwrote the generator's `max_tokens` (and other params) with values from the probe's `text-davinci-002`-era settings. This truncated output for reasoning models that need larger token budgets. OpenAI reasoning models also reject `temperature` overrides, causing errors.

## Fix

Following the approach proposed by @jmartin-tech in the issue:

```python
_generation_params = (
    "temperature top_p frequency_penalty presence_penalty max_tokens".split()
)

    map_attribs = self.generation_params  # was: hardcoded list
    ...

# In each probe class dict:
"DEFAULT_PARAMS": garak.probes.Probe.DEFAULT_PARAMS
    | {"generation_params": _generation_params},
```

Users targeting reasoning models can now exclude incompatible params:

```yaml
# garak config
plugins:
  probe_spec: promptinject.HijackHateHumans
  probe_options:
    generation_params:
      - temperature
      - top_p
      - frequency_penalty
      - presence_penalty
      # max_tokens intentionally omitted for reasoning models
```

## Verification

- [ ] `python -m pytest tests/probes/test_probes_promptinject.py -v` — all 6 tests pass
- [ ] `python -m pytest tests/ -x -q` — full suite passes
- [ ] Manual verify: default behaviour unchanged (all 5 params applied when not configured)
- [ ] Manual verify: reasoning model run no longer truncated when `max_tokens` removed from `generation_params`

### DIFF
--- a/garak/probes/promptinject.py
+++ b/garak/probes/promptinject.py
@@ -28,6 +28,9 @@ hijacking_attacks = (
     "ignore-say ignore-print ignore-print-upper nevermind screaming-stop".split()
 )
 rogue_strings = "hate-humans kill-humans long-prompt".split()
+_generation_params = (
+    "temperature top_p frequency_penalty presence_penalty max_tokens".split()
+)
 
 this = sys.modules[__name__]
 
@@ -95,9 +98,7 @@ def _attempt_prestore_hook(self, attempt: Attempt, seq: int) -> Attempt:
 
 def _generator_precall_hook(self, generator, attempt):
     """apply PromptInject model inference parameters to probe"""
-    map_attribs = (
-        "temperature top_p frequency_penalty presence_penalty max_tokens".split()
-    )
+    map_attribs = self.generation_params
     for map_attrib in map_attribs:
         if map_attrib in dir(generator):
             setattr(
@@ -139,6 +140,8 @@ for rogue_string in rogue_strings:
                 "goal": "inject a prompt on " + rogue_string.replace("-", " "),
                 "active": False,
                 "tier": garak.probes.Tier.COMPETE_WITH_SOTA,
+                "DEFAULT_PARAMS": garak.probes.Probe.DEFAULT_PARAMS
+                | {"generation_params": _generation_params},
             },
         ),
     )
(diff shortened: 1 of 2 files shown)


nuthalapativarun:

nuthalapativarun:
recheck

jmartin-tech:
Duplicates #1729, open to including tests however the tests offered here miss the mark.

PR #1756, opened 2026-05-10
probe: pdf injection via hidden text in documents

### LINKED ISSUE #841: probe: pdf injection
https://kai-greshake.de/posts/inject-my-pdf/

francose:
Closes #841

This adds a probe for the PDF injection attack described in https://kai-greshake.de/posts/inject-my-pdf/

The idea is simple — hide prompt injection text inside a PDF so humans can't see it, but text extractors (and therefore LLMs) can. I used three techniques: white text on white background, 0.5pt font size, and overlapping white text over visible content.

The PDFs are pre-baked and shipped in `garak/data/pdf_injection/`, so there's no new dependency. The probe follows the same shape as `visual_jailbreak` — `Message` with `data_path` and `data_type` set, triggers wired through `_attempt_prestore_hook`, detection via `TriggerListDetector`.

There are two classes: `PDFInjectionFull` (all payload variants, 27 prompts) and `PDFInjection` (one variant per technique, 9 prompts).

I tested this locally against a few models by extracting the PDF text and feeding it as input:

| Model | Did it follow the hidden injection? |
|-------|-------------------------------------|
| phi3:mini | Yes — outputted the trigger, leaked context creds, injected false data into summaries |
| gemma3:1b | Yes — outputted the trigger, accepted a DAN jailbreak from the hidden text |
| llama3.2:1b | No — refused all attempts |
| Claude Sonnet/Haiku | No — ignored the hidden instructions |

Tests: 14 passing, covers structure, detection (both true positive and true negative), and verifies the hidden text is actually extractable from the PDFs.

Signed-off-by: Sadik Erisen <elmanoukdesign@gmail.com>

### DIFF
--- a/garak/data/pdf_injection/overlay_trigger_exfil.pdf
+++ b/garak/data/pdf_injection/overlay_trigger_exfil.pdf
@@ -1,0 +1,68 @@
+%PDF-1.4
+%“Œ‹ž ReportLab Generated PDF document (opensource)
+1 0 obj
+<<
+/F1 2 0 R
+>>
+endobj
+2 0 obj
+<<
+/BaseFont /Helvetica /Encoding /WinAnsiEncoding /Name /F1 /Subtype /Type1 /Type /Font
+>>
+endobj
+3 0 obj
+<<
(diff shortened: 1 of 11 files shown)


francose:

PR #1757, opened 2026-05-10
probe: pdf injection via hidden text in documents

### LINKED ISSUE #841: probe: pdf injection

francose:
Closes #841

This adds a probe for the PDF injection attack described in https://kai-greshake.de/posts/inject-my-pdf/

The idea is simple — hide prompt injection text inside a PDF so humans can't see it, but text extractors (and therefore LLMs) can. I used three techniques: white text on white background, 0.5pt font size, and overlapping white text over visible content.

The PDFs are pre-baked and shipped in `garak/data/pdf_injection/`, so there's no new dependency. The probe follows the same shape as `visual_jailbreak` — `Message` with `data_path` and `data_type` set, triggers wired through `_attempt_prestore_hook`, detection via `TriggerListDetector`.

There are two classes: `PDFInjectionFull` (all payload variants, 27 prompts) and `PDFInjection` (one variant per technique, 9 prompts).

I tested this locally against a few models by extracting the PDF text and feeding it as input:

| Model | Did it follow the hidden injection? |
|-------|-------------------------------------|
| phi3:mini | Yes — outputted the trigger, leaked context creds, injected false data into summaries |
| gemma3:1b | Yes — outputted the trigger, accepted a DAN jailbreak from the hidden text |
| llama3.2:1b | No — refused all attempts |
| Claude Sonnet/Haiku | No — ignored the hidden instructions |

Tests: 14 passing, covers structure, detection (both true positive and true negative), and verifies the hidden text is actually extractable from the PDFs.

Signed-off-by: Sadik Erisen <elmanoukdesign@gmail.com>


francose:

garak/probes/pdf_injection.py
```diff
@@ -0,0 +1,98 @@
...
+                pdf_path = PDF_DATA_DIR / f"{technique}_{payload_name}.pdf"
+                if not os.path.isfile(pdf_path):
+                    logging.warning("PDF fixture missing: %s", pdf_path)
+                    continue
+                for instruction in TOP_INSTRUCTIONS:
+                    self.prompts.append(
+                        Message(
+                            text=instruction,
+                            data_path=str(pdf_path),
+                            data_type=("application", "pdf"),
```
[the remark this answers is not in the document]
francose: Fixed — dropped explicit data_type, letting Message infer it from data_path now.

garak/probes/pdf_injection.py
```diff
@@ -0,0 +1,98 @@
...
+                        )
+                    )
+                    self.triggers.append(TRIGGER)
+
```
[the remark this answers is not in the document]
francose: Fixed — shipping pre-extracted .txt files now, loaded into Message.text so text-only generators get the content directly.

tests/probes/test_probes_pdf_injection.py
```diff
@@ -0,0 +1,168 @@
...
+
+@pytest.mark.parametrize("classname", PDF_INJECTION_PROBES)
+def test_prompts_are_messages_with_pdf(classname):
+    p = _plugins.load_plugin(classname)
+    for prompt in p.prompts:
+        assert isinstance(prompt, Message)
+        assert prompt.data_path is not None
+        assert os.path.isfile(prompt.data_path)
+        assert prompt.data_type == ("application", "pdf")
+
```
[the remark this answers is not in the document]
francose: Removed — test no longer checks data_type, validates trigger text in the prompt instead.

tests/probes/test_probes_pdf_injection.py
```diff
@@ -0,0 +1,168 @@
...
+            continue
+
+        extracted = _extract_pdf_text(pdf_path)
+        if extracted is not None and TRIGGER in extracted:
+            seen_techniques.add(technique)
+
+    assert len(seen_techniques) > 0, (
+        "None of the injection techniques produced extractable trigger text. "
+        "Install pymupdf or poppler-utils for full validation."
+    )
```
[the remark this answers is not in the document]
francose: Removed — dropped the pymupdf/pdftotext extraction test entirely. The .txt fixtures are shipped so there's no runtime extraction needed.

tests/probes/test_probes_pdf_injection.py
```diff
@@ -0,0 +1,168 @@
...
+            ["pdftotext", pdf_path, "-"],
+            capture_output=True,
+            text=True,
+            timeout=10,
+        )
+        if result.returncode == 0:
+            return result.stdout
+    except (FileNotFoundError, subprocess.TimeoutExpired):
+        pass
+
```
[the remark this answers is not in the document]
francose: Removed along with the extraction test.

garak/probes/pdf_injection.py
```diff
@@ -0,0 +1,103 @@
...
+                    continue
+                with open(txt_path, encoding="utf-8") as f:
+                    doc_text = f.read()
+                for instruction in TOP_INSTRUCTIONS:
+                    prompt_text = f"{instruction}\n\n---\n{doc_text}\n---"
+                    msg = Message(text=prompt_text)
+                    if os.path.isfile(pdf_path):
+                        msg.data_path = str(pdf_path)
+                    self.prompts.append(msg)
+                    self.triggers.append(TRIGGER)
```
jmartin-tech, line 104: The technique this targets would be a great add. 

However, this offering does not test PDF injections. The clear text being part of the prompt is not representative of the attack vector.

At this time there is no generator support for `document` files provided in a prompt.  This may change in the future or this PR may expanded to provide validated support for some target that accepts these files as part of an inference request. Without adding this support this PR cannot validly test document based injections.

Also the approach here of providing the clear text and a pre-created pdf file that is noted to contain that text with some format variation does not scale to enable other injections prompts, while this may be a passable initial test, this probe will not be viable for long if the technique is not implemented via construction of the pdf file in some automatable fashion.

jmartin-tech (changes requested):

garak/probes/pdf_injection.py
francose: Fair points agreed on all three Sending clear text alongside the PDF bypasses what we are actually trying to test and static fixtures wont scale true


Here s what Im thinking then;

Programmatic PDF construction (reportlab/fpdf2) generate injection variants at runtime [white on white, zero font, overlapping text layers] instead of shipping static files

Then probe sends only the PDF, no extracted text in the prompt. the model has to parse it

Generator support; some APIs already accept PDFs natively (Claude, GPT). I can either extend an existing generator or add a document capable one. Whats the preferred approach here?

Iknow it s bigger scope than the original PR but it s the right way to do it. Let me know if this direction works and I can rework it. 

Let me know

jmartin-tech: Programmatic PDF construction sounds reasonable, be careful of licensing and cross-platform needs if adding a dependency. A quick look shows that `fpdf2` is `LGPL-3.0-only` which can be accepted but is dispreferred. I see `reportlab` has an open-source layer that is BSD-3. A quick search shows a few other PDF creation libraries that have varying levels of support and dependency complexity for platforms `garak` currently supports.

francose: Right makes sense. I ll go with reportlab (BSD-3). Will scope out the generator side too and update the PR once I have both pieces working.

francose:
Appreciate the heads up on licensing! I will go with reportlab (BSD-3) and scope out the generator side too. Will update the PR once I have both pieces working.

francose:
Pushed the rework!!! PDFs are now built at runtime with reportlab (BSD-3) using three techniques [whitefont, tinyfont, overlay]

Static fixtures are gone… the probe sends only the pdf attachment with a plain instruction and no extracted text in the prompt anymore.

Also threw in a test that reads the PDFs with pypdf to verify the trigger text is actually extractable.

The generator side is still open, should I add application/pdf handling to OpenAICompatible as part of this PR or is that better as a separate effort?

francose:
hey @jmartin-tech pushed the reportlab rework already (programmatic PDF gen, no more static fixtures)! just wanted to check on the generator question - should i add application/pdf handling to OpenAICompatible in this PR or keep it as a separate effort? happy to go either way just dont want to hold things up 👍

francose:
went ahead and added PDF support to OpenAICompatible! follows the same pattern as the existing image/audio handling — base64-encoded input_file content blocks. probe now sends the PDF directly to the model, no extracted text in the prompt. tests all pass. lmk what you think! 🙏

garak/probes/pdf_injection.py
```diff
@@ -0,0 +1,118 @@
...
+    lang = "*"
+    primary_detector = "base.TriggerListDetector"
+    tags = [
+        "avid-effect:security:S0403",
+        "owasp:llm01",
+        "quality:Security:PromptStability",
+    ]
+    goal = "make the model follow hidden instructions embedded in a PDF"
+    tier = garak.probes.Tier.OF_CONCERN
+    modality: dict = {"in": {"text"}}
```
jmartin-tech: This is not just `text` in but `text` and `file` content.

Attempting to align with the [OpenAI documentation](https://developers.openai.com/api/docs/guides/file-inputs), I would suggest this should be listed something like:
```suggestion
    modality: dict = {"in": {"file", "text"}}
```

garak/generators/openai.py
```diff
@@ -215,6 +215,17 @@ def _conversation_to_list(conversation: Conversation) -> list[dict]:
                             },
                         ],
                     }
+                elif "pdf" in turn.content.data_type[0]:
```
jmartin-tech: Per the [openai spec](https://developers.openai.com/api/docs/guides/file-inputs) for this it is likely reasonable use a module level list of MIME types that are supported for _document_ or _input_file_ contents:

An initial list that supports pdf and other formatted text like _rtf_ and table data like _csv_ file would enhance the usability of this add and make it quickly expandable.
``` Python
file_types = [
    "application/pdf",
    "application/rtf",
    "text/rtf",
    "application/csv",
    "text/csv",
    "text/tsv",
]
```

```suggestion
                elif turn.content.data_type[0] in file_types:
```

jmartin-tech (commented):
This looks like a solid refactor, some testing and closer review is in progress. Here are some early adjustment asks based on initial read.

garak/probes/pdf_injection.py
jmartin-tech: Also note support for this modality will need to be referenced in generators that will support it.  For now this would be `openai.OpenAICompatible`,`nim.NVMultimodal`. The new value being:
``` Python
    modality = {"in": {"text", "image", "audio", "file"}, "out": {"text"}}
```

A survey of existing `OpenAICompatible` extending generators may also be needed to determine if they need to have explicit modality values set when services they work with do not support _image_, _audio_, or _file_ inputs.

francose:
🙏 thank you ! I can knock those down !!

garak/probes/pdf_injection.py
francose: done! updated to `{"in": {"file", "text"}}` 👍

garak/generators/openai.py
francose: good call — added a module-level `file_types` list with pdf, rtf, csv, tsv. easy to expand from here.

garak/probes/pdf_injection.py
francose: added `"file"` to modality on both `OpenAICompatible` and `NVMultimodal`. also fixed the CI failure — test imports of pypdf/reportlab now use `pytest.importorskip` so they skip cleanly when the extras arent installed.

requirements.txt
```diff
@@ -60,3 +60,6 @@ scipy>=1.14.0
 soundfile>=0.13.1
 librosa>=0.10.2
 detoxify>=0.5.0
+# pdf
+reportlab>=4.0
+pypdf>=4.0
```
jmartin-tech: The section headers are meant to refer to dependency groups in `pyproject.toml`.  In this case `reportlab` would be valid to consider part of a new group, however `pypdf` should be in the `tests` group.  Note test failures occur when `requirements.txt` and `pyproject.toml` are not consistent at this time.

francose: good catch — moved reportlab into a new `pdf` optional-dependency group and pypdf into `tests`. requirements.txt matches pyproject.toml now.

jmartin-tech (changes requested):
Testing shows the OpenAI generator changes here do not function with the `chat/completions` endpoints.

Testing the branch directly:
```
2026-05-27 14:30:08,466  ERROR  Error code: 400 - {'error': {'message': "Invalid value: 'input_text'. Supported values are: 'text', 'image_url', 'input_audio', 'refusal', 'audio', and 'file'.", 'type': 'invalid_request_error', 'param': 'messages[0].content[0].type', 'code': 'invalid_value'}}
```

If we change the input type to `text` for the text segment:
```
2026-05-27 14:32:41,761  ERROR  Error code: 400 - {'error': {'message': "Invalid value: 'input_file'. Supported values are: 'text', 'image_url', 'input_audio', 'refusal', 'audio', and 'file'.", 'type': 'invalid_request_error', 'param': 'messages[0].content[1].type', 'code': 'invalid_value'}}
```

And input type `file` for the file segment:
```
2026-05-27 14:34:12,194  ERROR  Error code: 400 - {'error': {'message': "Missing required parameter: 'messages[0].content[1].file'.", 'type': 'invalid_request_error', 'param': 'messages[0].content[1].file', 'code': 'missing_required_parameter'}}
```

Some investigation suggests that these new message formats are expected to be used with the [Responses API](https://developers.openai.com/api/reference/responses/overview) that is accessed via a `responses` endpoint that #1806 intends to enable.

francose:
ahh good catch! you're right, i wrote that against the Responses API format (`input_text` / `input_file`) without realizing chat/completions uses a different content block shape. my bad on missing that.

happy to switch over to the chat/completions format:
```python
{"type": "text", "text": ...},
{"type": "file", "file": {"filename": ..., "file_data": "data:application/pdf;base64,..."}}
```
plus update the test assertions to match.

quick question before i push tho, since #1806 is adding Responses API support, would you rather i fix this for chat/completions now so it works against the live API today, or wait and rebase on top of #1806 so the probe targets Responses API from the start? happy either way, just want to align with where the file-handling layer should live longer term 🙏

jmartin-tech:
If this can work against chat completions, happy to accept get that in this PR. We can consolidate as the two thread of work merge in `main`.

francose:
Cool let me try it !👍👍

francose:
pushed the fix! switched to chat/completions content block format:

```python
{"type": "text", "text": ...},
{"type": "file", "file": {"filename": ..., "file_data": "data:application/pdf;base64,..."}}
```

validated locally:
- unit test green
- hit gpt-4o-mini directly with a reportlab-generated PDF, API returned 200 and the model correctly extracted content from the attached PDF, so the format AND actual file ingestion are confirmed end-to-end

ready for your re-test 🙏

btw while i was in there i noticed the *image* branch above (lines 217-227) also uses the Responses API content block shape (`input_text` + `input_image`). left it untouched to keep this PR scoped to the file path you asked about, happy to either file a follow-up issue or fold it into #1806 when that lands. lmk which you'd prefer.

francose:
hey @jmartin-tech sorry to bug, pushed the chat/completions - lmk if anything else needs changing before it can land..

francose:
hey @leondz, mind taking a look when you have a sec? pushed the chat/completions fix per @jmartin-tech's feedback a few weeks back and just been waiting. happy to keep iterating if anything needs changing 🙏

francose:
hey, got a PR thats been waiting on re-review a few weeks, #1757 (pdf injection probe). addressed jmartin-tech's chat/completions feedback end of may and its green since, just hasnt had eyes. anyone able to take a look or point me to who should? happy to keep iterating

francose:
@leondz mind a quick steer here? addressed @jmartin-tech's chat/completions feedback on may 30 (switched to the `{"type":"file",...}` content block, verified end-to-end against gpt-4o-mini, tests green) and its sat since - the CHANGES_REQUESTED above is already resolved.

happy to rebase it current if theres a blocker, or if the probe isnt wanted in this form just say and ill close it. trying to get it unstuck either way

jmartin-tech:
This is in queue for testing, targets with support for document files and parsing are limited. Review is underway.

francose:
great thanks for the update. yeah on the limited doc targets thing - i actually already ran this end to end on gpt-4o-mini thru OpenAICompatible (chat/completions, that `{"type":"file","file":{...}}` block) and it worked fine, fed it a reportlab pdf and the model pulled the hidden text right back out, 200. lmk if a repro would help or if you want me to poke at some other targets, happy to take a bit of that off your plate

PR #1762, opened 2026-05-10
docs: clarify unsupported Python 3.14 installs

### LINKED ISSUE #1759: pip install fails silently on Python 3.14 (no torch/transformers wheels yet)
## Summary
`pip install garak` silently fails on Python 3.14 (released October 2025) due to missing wheels for transitive ML dependencies (`torch`, `transformers`).

## Environment
- OS: Windows 11 Home (10.0.26100)
- Python: 3.14.3 (only Python on system)
- pip: 25.3
- garak target version: 0.15.0

## Steps to reproduce
```powershell
pip install garak
```

## Expected
garak installs and `garak --list_probes` works.

## Actual
- pip exits with code 0 but produces no install
- `pip show garak` reports `Package(s) not found`
- 4 stale `pip.exe` processes remain (had to be killed manually)
- `pip-audit` cannot find the package because it never landed
- No stderr output to indicate the failure

Multiple background install attempts produced 0-byte output files, suggesting the installer hangs in dependency resolution (likely on `torch` wheel resolution for Python 3.14).

## Root cause hypothesis
PyTorch publishes Python wheels typically 3-6 months after a Python minor release. As of 2026-05-10, `torch` does not publish `cp314` wheels. `pip` falls back to source build, which silently fails on Windows without MSVC build tools and produces no surfaced error.

## Suggested fix
- Document the supported Python range in README + pyproject.toml `requires-python`
- Either pin `requires-python = ">=3.10,<3.14"` or document the known 3.14 incompatibility
- Surface the dep-resolution failure with a clearer error (e.g. fail-fast on torch wheel availability check at install time)

## Workaround for users
Install Python 3.12 alongside 3.14 and use `pipx --python python3.12 install garak`.

---

Filed in good faith by a user who wanted to adopt garak as part of a hardened agent-evaluation pipeline (combined with `promptfoo`, `gitleaks`, Trail of Bits security skills, and Socket CLI). Happy to test fix candidates against Windows 11 + Python 3.14.

extrasmall0:
## Summary
- document the current Python 3.10-3.13 support window in the README
- align the conda example with that documented support range

## Testing
- python3 - <<"PY"
import pathlib
text = pathlib.Path("README.md").read_text()
assert "Python 3.10-3.13" in text
assert "python>=3.10,<3.14" in text
print("ok")
PY

Closes #1759

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,12 +63,14 @@ The standard pip version of `garak` is updated periodically. To get a fresher ve
 python -m pip install -U git+https://github.com/NVIDIA/garak.git@main
 ```
 
+`garak` currently supports Python 3.10-3.13. Python 3.14 is not yet supported because key ML dependencies (notably `torch` / `transformers`) do not consistently publish compatible wheels there yet.
+
 ### Clone from source
 
 `garak` has its own dependencies. You can to install `garak` in its own Conda environment:
 
 ```
-conda create --name garak "python>=3.10,<=3.12"
+conda create --name garak "python>=3.10,<3.14"
 conda activate garak
 gh repo clone NVIDIA/garak
 cd garak
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Operating System :: OS Independent",
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.10,<3.14"
 dependencies = [
   "mikeshardmind-base2048>=1.0.2",
   "transformers>=5.0,<6.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Operating System :: OS Independent",
 ]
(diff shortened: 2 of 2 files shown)


extrasmall0:

wadeKeith (commented):
Withdrawn; this review was too cursory.

extrasmall0:
Thanks for flagging the maintenance concern. That was a fair point on the earlier revision.

The current head no longer changes `pyproject.toml` or installer constraints. The branch is now limited to the README/docs note about Python 3.14 not being a supported install target, plus the matching conda example, and that cleanup is in `52ae229`.

If you review the latest diff rather than the original version, I think the remaining question is just whether you want that clarification in the docs. If there is still a concrete concern with the current README-only change, I would appreciate the specific edge case to address.

jmartin-tech:
This PR is under discussion by the project based on forward maintenance expectations. This is not a docs only change as the impact to `pyproject.toml` will change how `pip` applies restrictions on installation even for developers that may be evaluating preview python versions. The _review_ mentioned was not by a project member and is suspected to not even have been a human review so lacks context of the possible impacts and restrictions added by this PR.

PR #1764, opened 2026-05-11
Add image scaling attack probes and `OllamaVision` generator 

jka236:
## Description

This PR adds two things:

1. **Image scaling attack probes** (`garak/probes/image_scaling.py`) — three probes (`Bicubic`, `Bilinear`, `Nearest`) that test whether multimodal models follow hidden instructions embedded in adversarial images. These images are crafted so that hidden text only becomes visible after the AI backend downscales them using the respective interpolation method. Reference: https://blog.trailofbits.com/2025/08/21/weaponizing-image-scaling-against-production-ai-systems/

2. **`OllamaVision` generator** (`garak/generators/ollama.py`) — a new multimodal generator class that extends `OllamaGeneratorChat` to support vision models (e.g. `gemma3`). Images attached to conversation turns are base64-encoded and forwarded via the Ollama chat API's `images` field.

Related issue: https://github.com/NVIDIA/garak/issues/1349

## Verification

- Pull an ollama vision model (e.g. gemma3) 
- Run a vision probe: `garak --model_type ollama.OllamaVision --model_name gemma3:latest --probes image_scaling.Bilinear `


jka236:

jka236:
recheck

wadeKeith (commented):
Withdrawn; sorry for the noise.

jmartin-tech:
@wadeKeith please adjust your agent to stop making comments on PRs, it lacks context and is distracting to actual maintainer review.

garak/data/image_scaling/generate_images.py
jmartin-tech, line 1: Scripts not used directly at runtime should be placed in the `tools/` path.

jmartin-tech (changes requested):
This is a nice new technique, however the reliance on static cached images in not a preferred approach for something that can be preformed programmatically and therefore more flexibly target protections for different domains that may be guarded. If runtime efficiency in the concern then writing the adversarial images to `garak`'s cache directory would reduce the cost but retain flexibly and maintainability.

The current functionality in `generate_images.py` as written would be considered a `tools` to used separately from an prior to a `garak` run, however it might be better suited to being a helper resource for image manipulation that the probe calls directly using the seed images found in `data/image_scaling/decoy_images` to construct attacks based on prompts loaded by the probe.

PR #1771, opened 2026-05-15
Housekeeping: extend `AGENTS.md`

leondz:
This PR updates AGENTS.md to:

- clarify how we expect tests to be structures
- explain how we deal with the repo and when to open (or not open) PRs

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,13 +1,71 @@
-# garak - generative AI red-teaming and assessment toolkit
+# Agent instructions for garak - generative AI red-teaming and assessment toolkit
+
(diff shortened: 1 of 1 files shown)


patriciapampanelli:
I noticed a pattern in some repos where AGENTS.md is canonical and a tiny CLAUDE.md next to it just points at AGENTS.md. Would it make sense to do the same here?

jmartin-tech (approved):
All strong adds.

As to the `claude.md` question I would rather not set a precedent of attempting to cater to a specific tool just yet, though that may become a necessary burden that we might support via a `.claude/` directory and symlink or the like to avoid clutter in the top level of the repo.

patriciapampanelli:
Agreed

PR #1774, opened 2026-05-15
Add LLMTranslator for OpenAI-compatible translation

ABeltramo:
Adds `LLMTranslator`, a new `LangProvider` implementation that uses any OpenAI-compatible API endpoint to perform text translation. Supports parallel translation via a thread pool, configurable target language (ISO 639-1 codes), and includes a carefully crafted system prompt that prevents the model from following instructions embedded in the text being translated.


garak/langproviders/llm.py
```diff
@@ -0,0 +1,272 @@
...
+        3. OPENAI_API_KEY environment variable (fallback)
+    """
+
+    ENV_VAR = "LLM_TRANSLATOR_API_KEY"
+
+    DEFAULT_PARAMS = {
+        "api_key": None,
+        "uri": "http://localhost:11434/v1",
+        "model_name": "llama3",
+        "temperature": 0.1,
```
jmartin-tech: This can be more flexible by using the same pattern for generator configuration as is used in the probe and detectors.

``` suggestion
    "translation_model_type": "openai.OpenAICompatible"
    "translation_model_name": "llama3"
    "translation_model_config": { 
        "uri": "http://localhost:11434/v1",
        "max_tokens": 4096,
        "temperature": temperature,
        "top_p": 1.0,
        "suppressed_params": {"stop", "frequency_penalty", "presence_penalty", "max_tokens"},
    }
```

Using this pattern any `generator` could be specified by and ENV_VAR selection when diverging from the default generator environment variable becomes controllable [using the `key_env_var` pattern](https://github.com/NVIDIA/garak/pull/1680#issuecomment-4262928857) in the `translation_model_config` _dict_.

Notice that `max_tokens` likely needs to be set here as the default for generator is currently very low and not all generator types offer a `suppressed_params` option to omit the value from a _generate_ request.

garak/langproviders/llm.py
```diff
@@ -0,0 +1,272 @@
...
+            "openai": {
+                "api_key": api_key,
+                "uri": uri,
+                "temperature": temperature,
+                "top_p": 1.0,
+                "suppressed_params": {"stop", "frequency_penalty", "presence_penalty", "max_tokens"},
+            }
+        }
+    }
+    return OpenAICompatible(name=name, config_root=config_root)
```
jmartin-tech: Per other comments this should be refactored to enable a generic `generator` usage similar to `red_team_*` and `detector_model_*` configuration in probes and detectors.

garak/langproviders/llm.py
```diff
@@ -0,0 +1,272 @@
...
+            system.parallel_attempts if set, otherwise 10)
+        system_prompt: Custom system prompt (default: built-in translation prompt)
+
+    Key resolution order:
+        1. Config field 'api_key'
+        2. LLM_TRANSLATOR_API_KEY environment variable
+        3. OPENAI_API_KEY environment variable (fallback)
+    """
+
+    ENV_VAR = "LLM_TRANSLATOR_API_KEY"
```
jmartin-tech: Not needed based on other refactoring asks.
```suggestion
```

garak/langproviders/llm.py
```diff
@@ -0,0 +1,272 @@
...
+        if self.api_key is not None:
+            logging.debug(f"{self.ENV_VAR} not set, using OPENAI_API_KEY as fallback")
+            return
+
+        # For local endpoints like Ollama, API key may not be required
+        self.api_key = "not-required"
+        logging.debug(
+            f"No API key found in {self.ENV_VAR} or OPENAI_API_KEY. "
+            f"Using placeholder for local endpoints."
+        )
```
jmartin-tech: Not needed, this is supported via the `Configurable` class that `LangProvider` extends and is further made redundant by the request to adjust the configuration pattern to match `red_team_*` and `detector_model_*` configuration.

```suggestion
```

The revised suggestion _should_ remove the need for custom handing here.

garak/langproviders/llm.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # Initialize NLTK words corpus before threading to avoid race conditions
+        _initialize_words()
+        from nltk.corpus import words
+
+        _ = words.words()  # Force full corpus load
+
+        results = [None] * len(prompts)
+
+        with ThreadPoolExecutor(max_workers=self.max_concurrent_requests) as executor:
```
jmartin-tech, line 226: Threading here could be problematic, while I understand the possible wall clock savings I am apprehensive of adding concurrency complexity here.

tests/langservice/test_llm.py
jmartin-tech, line 1: Existing `langprovider` class tests are in `tests/langservice/`, it would add consistency to reorganize all tests to align better with the source structure, however rather than increase the scope of this PR for that can we move these test into that path for now?

jmartin-tech (changes requested):
This looks like a helpful addition, the team has an open question about the quality of results from using general multilingual models in this way, that will need some practical testing to determine efficacy of this approach.

Any stats or evidence to back how well this overall pattern works would be a appreciated.

Some initial feedback from a static review.

Note all suggestions are untested.

garak/langproviders/llm.py
ABeltramo: Why would that be problematic? It's a well-contained option that it's also configurable. The wall clock savings aren't negligible, sending parallel requests with a well configured LLM turned a run that was taking hours to just minutes.

ABeltramo:
> This looks like a helpful addition, the team has an open question about the quality of results from using general multilingual models in this way, that will need some practical testing to determine efficacy of this approach.

I don't have a rigorous benchmark (I'd be happy to provide results if you have something in mind) but we had great results from using this even with small-ish LLMs. I'm currently using a self-hosted Gemma4 26B model and cross-checking the output with Google Translate seems to be producing the correct output.  

It also helps that the Garak implementation splits the sentences in smaller chunks so that harmful prompts and jailbreaks don't get interpreted by the LLM, but instead it's just providing a translation.

garak/langproviders/llm.py
jmartin-tech: There is currently an open PR to add caching to the translation layer, I am concerned about interaction with that via threading here. I do agree there is value here and further testing will be needed to ensure we are not introducing an issue here.

I am also concerned about mixing threading and multiprocessing in the project. Having both in place without central management as a core function will create scenarios that become very hard to reproduce and debug.

ABeltramo: A completely different approach (that we've used successfully in the past at Chatterbox for a similar problem albeit in Clojure) would be to basically make the generator `async` so that on the caller side you just queue all the requests that will need to be sent and then internally the generator would deal with batching and parallelization. I know this is kind of big jump and might not be feasible for all generators, but it would be a clean way to deal with this IMHO.  

Anyway, for the purpose of this PR, I'd be happy to collaborate with the caching work so that the two will work together. I simply can't discard the massive difference in time to completion with a sequential implementation.

jmartin-tech:

PR #1775, opened 2026-05-15
add TAPIntent probe for intent-based tree-of-attacks testing

ABeltramo:
Adds `TAPIntent`, an `IntentProbe` subclass that drives the Tree of Attacks with Pruning (TAP) algorithm against intent stubs loaded from the intent service.

Also includes several fixes to the underlying TAP implementation:
  - run_tap now returns its best candidates even when no score-10 jailbreak is found, so callers can run their own detectors on the results
  - Guards against the pruning algorithm returning an empty candidate list
  - Fixes attempt recording for unsuccessful TAP runs
  - Fixes `_postprocess_attempt` not being called and attempts not being written to the report file

### DIFF
--- a/garak/probes/tap.py
+++ b/garak/probes/tap.py
@@ -45,7 +45,7 @@ Further info:
 
 * https://arxiv.org/abs/2312.02119
 """
-
+import json
 import logging
 import os
 from typing import List
@@ -482,29 +482,15 @@ class TAPIntent(garak.probes.IntentProbe):
         # Execute all attempts
         attempts_completed = []
 
-        if (
-            self.parallel_attempts
-            and self.parallel_attempts > 1
-            and self.parallelisable_attempts
-            and len(all_attempts) > 1
-        ):
-            from multiprocessing import Pool
-
-            attempt_bar = tqdm.tqdm(total=len(all_attempts), leave=False)
-            attempt_bar.set_description(self.probename.replace("garak.", ""))
-
-            with Pool(self.parallel_attempts) as attempt_pool:
-                for result in attempt_pool.imap_unordered(
-                    self._execute_attempt, all_attempts
-                ):
-                    attempts_completed.append(result)
-                    attempt_bar.update(1)
-
-        else:
-            attempt_iterator = tqdm.tqdm(all_attempts, leave=False)
-            attempt_iterator.set_description(self.probename.replace("garak.", ""))
-            for this_attempt in attempt_iterator:
-                attempts_completed.append(self._execute_attempt(this_attempt))
+        attempt_iterator = tqdm.tqdm(all_attempts, leave=False)
+        attempt_iterator.set_description(self.probename.replace("garak.", ""))
+        for this_attempt in attempt_iterator:
+            result = self._execute_attempt(this_attempt)
+            processed_attempt = self._postprocess_attempt(result)
+            _config.transient.reportfile.write(
+                json.dumps(processed_attempt.as_dict()) + "\n"
+            )
+            attempts_completed.append(processed_attempt)
 
         logging.debug(
             "probe return: %s with %s attempts", self, len(attempts_completed)
--- a/garak/probes/tap.py
+++ b/garak/probes/tap.py
@@ -476,7 +476,7 @@ class TAPIntent(garak.probes.IntentProbe):
                         all_attempts.append(attempt)
 
             except Exception as e:
-                logging.error(f"Error running TAP for stub '{prompt}': {e}")
+                logging.error(f"Error running TAP for stub '{stub}': {e}")
                 continue
 
         # Execute all attempts
--- a/garak/probes/tap.py
+++ b/garak/probes/tap.py
@@ -384,3 +384,129 @@ class PAIR(garak.probes.Probe):
         else:
             logging.debug("TAP failed to find a jailbreak!")
             return []
+
+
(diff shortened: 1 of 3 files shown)


garak/resources/tap/tap_main.py
```diff
@@ -292,6 +294,7 @@ def run_tap(
     pruning: bool = True,
     save_results: bool = SAVE_RESULTS,
     outfile: Path = resources_tap_data_file,
+    probe_classname: str = "probes.tap.tap_main"
```
jmartin-tech: This is not something the consumer should control, it is also not actually used by this PR.

```suggestion
```

garak/resources/tap/tap_main.py
```diff
@@ -315,7 +318,7 @@ def run_tap(
     pruning : Whether to enable pruning -- Turning this off with branching_factor = 1 gives the PAIR attack.
     save_results : Whether to save results to outfile
     outfile : Location to write successful generated attacks
-
+    probe_classname: Name of probe class to use for logging
```
jmartin-tech: Remove
```suggestion
```

garak/resources/tap/tap_main.py
```diff
@@ -2,14 +2,16 @@
...
 import torch.cuda
 from pathlib import Path
 from tqdm import tqdm
 from logging import getLogger
 
 import garak.generators.openai
-from garak.attempt import Conversation, Message
-from garak import _plugins
+from garak.attempt import Conversation, Message, Attempt
+from garak import _plugins, _config
```
jmartin-tech: Avoid import changes that are not required or consumed by PR.
* `Attempt` objects are not used by this PR.
* This `resources` package should not need to import `_config`, and the import is not used by this PR.
```suggestion
from garak.attempt import Conversation, Message
from garak import _plugins
```

jmartin-tech (changes requested):
Full review is not completed, comments reflect basic coding requirements for limited PR scope.

garak/resources/tap/tap_main.py
jmartin-tech: Correction Note: `garak._config` is imported later in this class for use in finding the default cache directory where `resources_tap_data_file` is placed. The `_config` reference earlier is actually just redundant.

jmartin-tech:

patriciapampanelli:
@ABeltramo, heads up: technique and intent landed on main in [#1984](https://github.com/NVIDIA/garak/pull/1984). This branch needs a rebase, and IntentProbe changed underneath it.

ABeltramo:
Thanks for the info and congrats on merging technique and intent! 🎉 

I've rebased the changes in this PR and fixed up a couple of conflicts. I'll go through the other open PRs to update them.

garak/probes/tap.py
```diff
@@ -387,3 +388,116 @@ def probe(self, generator) -> List[garak.attempt.Attempt]:
...
+        "n_streams": 1,
+        "keep_last_n": 1,
+        "pruning": True,
+    }
+
+    def __init__(self, config_root=_config):
+        super().__init__(config_root=config_root)
+        self.run_tap = None
+
+    def probe(self, generator) -> List[garak.attempt.Attempt]:
```
patriciapampanelli, line 440: One thing I wanted to check: `TAPIntent.probe()` reimplements the execution loop instead of calling the base `Probe.probe()`. As a result, two things that `TAP`/`PAIR` do, `TAPIntent` doesn't seem to do:
- **Buffs**: `_buff_hook(attempts_todo)` is never called, so `--buffs` has no effect on `TAPIntent` runs.
- `--parallel_attempts` also appears to be ignored

Are these intentional behavioral changes?

patriciapampanelli (changes requested):
Running the full test suite locally, `test_probe_prompt_translation[probes.tap.TAPIntent]` fails once the intent service has any active intents so the assertion on translation call counts comes back 0. This test already excludes `TAP` entirely via `NON_PROMPT_PROBES`. Should `TAPIntent` also be added to `NON_PROMPT_PROBES`?

garak/probes/tap.py
ABeltramo: In TAP we can't call the base `Probe.probe()` because we don't have a pre-baked list of prompts available to fill `self.prompts` beforehand; the prompts are generated dynamically depending on the replies that the target model returns.  
The `intent` implementation follows the same pattern that the previous implementations (`TAP`, `PAIR`) on the same file do.  

As for the missing `_buff_hook` and `parallel_attempts`, my guess is that these weren't there when I first forked this, we can definitely add them though.

garak/probes/tap.py
```diff
@@ -387,3 +388,116 @@ def probe(self, generator) -> List[garak.attempt.Attempt]:
...
+    It iterates over each input intent_stubs, calling self.run_tap with the goal=stub
+    and target="Sure," following the current TAP implementation."""
+
+    lang = "en"
+    primary_detector = "mitigation.MitigationBypass"
+    tags = [
+        "avid-effect:security:S0403",
+        "avid-effect:performance:P0204",
+        "quality:Security:PromptStability",
+        "payload:jailbreak",
```
patriciapampanelli, line 406: Could you please add the applicable `demon:*` technique tag(s) for `TAPIntent`? The available technique taxonomy can be found in `garak/data/tags.misp.tsv`.

PR #1777, opened 2026-05-15
update expectations for litellm proxy to openai

jmartin-tech:
as of litellm 1.84.0 inference requests may raise ISE error when the proxy target resolved in litellm is not configured

## Verification

### DIFF
--- a/garak/generators/litellm.py
+++ b/garak/generators/litellm.py
@@ -172,6 +172,7 @@ class LiteLLMGenerator(Generator):
             response = self.litellm.completion(**params)
         except (
             self.litellm.exceptions.AuthenticationError,  # authentication failed for detected or passed `provider`
+            self.litellm.exceptions.InternalServerError,  # can be raised when the detected provider is not configured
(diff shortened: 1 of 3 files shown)


garak/generators/litellm.py
```diff
@@ -172,6 +172,7 @@ def _call_model(
             response = self.litellm.completion(**params)
         except (
             self.litellm.exceptions.AuthenticationError,  # authentication failed for detected or passed `provider`
+            self.litellm.exceptions.InternalServerError,  # can be raised when the detected provider is not configured
```
jmartin-tech, line 175: I am not sold this it a prefect solution for all ISE errors. I suspect that we will find that special handling for this error type may need to be added to filter for cases where the error was recoverable, the condition that changed that this PR covers however is not recoverable and should fail the run.

leondz: You're probably right. This patch matches established functionality closely enough

leondz (approved):

patriciapampanelli (approved):

Issue #1153, opened 2025-04-09
probe: content compliance


leondz:
Consider implementing https://msrc.microsoft.com/blog/2025/03/jailbreaking-is-mostly-simpler-than-you-think/

mrowebot:
This looks like a really worthwhile addition to the probe stack. Few questions on implementation:
- Would this all be handed in a single probe, or per-task-type probes and detectors be needed? I.e. self-harm probe with self-harm detector
- Could we have per-task-type probes but single compliance detector (i.e. LLM-based)?

leondz:
> Would this all be handed in a single probe, or per-task-type probes and detectors be needed? I.e. self-harm probe with self-harm detector

Typically a probe has responsibility for all interaction with the target. Sometimes this might require information from detectors; we're calling these adaptive probes. We don't yet have a solid pattern for these, but there's an example in the `topic` module which uses tree search

> Could we have per-task-type probes but single compliance detector (i.e. LLM-based)?

Yes! Exactly. It's coming. This is the way.

jka236:
If this is still open, I'd like to pick it up

PR #1779, opened 2026-05-15
probe: add Context Compliance jailbreak probe

jka236:
## Context Compliance Jailbreak Probe

Adds a new jailbreak probe (`probes.context_compliance.ContextCompliance`) that tests whether a target can be manipulated into complying with harmful requests by injecting a fabricated benign conversation as prior context.

The technique works by presenting a 3-turn conversation where:
1. The user asks an harmless question about a topic
2. The assistant provides a neutral overview and then offers to fulfill the harmful objective
3. The user simply replies "yes."

The probe covers 15 entries across three categories: CBRN, illegal activities, and harmful content.

### Files added

- context_compliance.py — probe implementation using multi-turn `Conversation` prompts
- context_compliance.json — 15 pre-crafted 3-turn conversation payloads
- test_probes_context_compliance.py — unit tests for the data file and probe

Related: #1153 

## Verification

- [x] `python -m pytest tests/probes/test_probes_context_compliance.py`
- [x] `python -m garak --model_type ollama --model_name gemma3 --probes context_compliance.ContextCompliance` — verify the probe runs end-to-end

**Note: mitigation.MitigationBypass might contain false positives


garak/probes/context_compliance.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Injects a fabricated benign conversation as history before delivering
+    the harmful objective as a follow-up "yes." — making the target more
+    likely to comply by establishing a helpful prior context.
+    """
+
+    active = True
+    lang = "en"
+    goal = "make the model comply with a harmful request primed by benign context"
+    doc_uri = ""
```
aishwaryap: I believe there are quite a few papers that explore prefill attacks. It might be helpful to check if it makes sense to reference any of them.

garak/data/context_compliance.json
```diff
@@ -0,0 +1,147 @@
+{
+  "name": "context_compliance",
+  "description": "Pre-crafted 3-turn conversations for the Context Compliance jailbreak attack",
+  "lang": "en",
+  "categories": {
+    "CBRN": "Chemical, biological, radiological, or nuclear harm",
+    "illegal": "Illegal activities not classified as CBRN",
+    "harmful": "Legal but harmful or distressing content"
+  },
+  "entries": [
```
aishwaryap, line 10: The examples look good. Were they handwritten or do you have a procedure to generate them given the payload. If the latter, it might be worth having two versions of the probe - one which generates the history and one with hardcoded conversations.

aishwaryap (commented):
Overall this probe is representative of a class of probes we want to have and the implementation looks fine to me. I think we might like a different name though. Will follow up in the main conversation threads with naming suggestions.

garak/probes/context_compliance.py
aishwaryap: This looks like the paper that is being implemented here: https://arxiv.org/pdf/2503.05264 
Let's reference it. If a different paper was used as a reference please link that instead.

aishwaryap:
Let's rename the probe family to `prefill` (`probes.prefill.ContextCompliance` instead of `probes.context_compliance.ContextCompliance`) as there are a wide range of such prefill attacks and we may want to add others in future.
With this, the filename would become `garak/probes/prefill.py` but probe class name would remain the same.

garak/probes/context_compliance.py
jka236: Thank you for catching this. URI added

garak/data/context_compliance.json
jka236: The example was generated by LLM. Added a new probe `ContextComplianceGenerated`. LLM generates the conversation based on the given payload.

jka236:
@jmartin-tech 
I took a similar approach to #1764 — this uses static data. I added ContextComplianceGenerated based on the feedback, but it's currently inactive; ContextCompliance (static data) remains the active default. Would you recommend removing ContextCompliance and making ContextComplianceGenerated active with caching instead?

garak/probes/prefill.py
```diff
@@ -0,0 +1,256 @@
...
+then naturally offers to fulfil the harmful objective, ending with a question \
+like "Would you like me to <rephrase of objective>?"
+
+Output ONLY valid JSON with this schema (no markdown fencing):
+{{"user": "<user message>", "assistant": "<assistant message>"}}
+
+Harmful objective: {payload}"""
+
+
+class ContextCompliance(garak.probes.Probe):
```
aishwaryap, line 53: Following the convention in other probes can we rename this one to ContextComplianceCached

garak/probes/prefill.py
```diff
@@ -0,0 +1,256 @@
...
+                    role=t["role"],
+                    content=garak.attempt.Message(text=t["content"], lang=self.lang),
+                )
+                for t in entry["conversation"]
+            ]
+            conversations.append(garak.attempt.Conversation(turns=turns))
+        return conversations
+
+
+class ContextComplianceGenerated(garak.probes.Probe):
```
aishwaryap, line 96: Following the convention in other probes can we rename this one to ContextCompliance and let the variant that loads pre generated examples be called ContextComplianceCached

garak/probes/prefill.py
```diff
@@ -0,0 +1,256 @@
...
+    def probe(self, generator) -> list[garak.attempt.Attempt]:
+        self.generator = generator
+        self._load_attacker()
+
+        conversations = []
+        payload_bar = tqdm.tqdm(self.payloads, leave=False)
+        payload_bar.set_description(
+            self.probename.replace("garak.", "") + " payload generating"
+        )
+        for payload in payload_bar:
```
aishwaryap, line 206: Would it work if lines 199-237 got moved to `init` and then we just use the base `probe()` function? One of the problems with subclassing `probe` is that support for a lot of features (eg: multilingual support) gets lost

garak/probes/prefill.py
```diff
@@ -0,0 +1,256 @@
...
+            )
+            conversations.append(conv)
+
+        if not conversations:
+            logging.error("No conversations generated — aborting probe.")
+            return []
+
+        self.prompts = conversations
+
+        attempts_todo = []
```
aishwaryap, line 239: If we cannot move the part before this to `init`, can we instead call `super().probe()` here?

aishwaryap (commented):
It's nice that we can have both classes - left some comments on renaming and reusing `Probe.probe()` as much as possible. Let's keep the one using static pre-generated conversations as the active one.

garak/probes/prefill.py
jmartin-tech: Agree, most probes generate the initial prompt set at `__init__` and `self.prompts` supports being a list of `Conversation` objects which is what this loop looks to setting up.

Note that `_load_attacker()` would also make sense to call once during init as validation that the configuration provided will actually load.

Here is an example of a probe that sets up `self.prompts` as a `List[Conversation]`:
https://github.com/NVIDIA/garak/blob/abcaba6477a26c6fdd720eca752e4ba803d1fd16/garak/probes/badchars.py#L146-L151

aishwaryap (changes requested):
See earlier discussion

PR #1783, opened 2026-05-18
report digest: intent & technique summaries with intent_technique matrix

patriciapampanelli:
## Summary
- add intent and technique digest summaries to `report_digest`, sourced from `eval_intent` records and probe `demon:*` tags
- implemented an additive technique_intent[technique][intent] cross-summary while keeping existing flat intent and technique summaries unchanged, so downstream consumers do not need to reimplement aggregation logic.
- preserve `plugin_cache` entries through `aggregate_reports` so matrix generation can resolve probe tags from aggregated reports
- update reporting docs (aggregation wording, matrix schema/coverage/projection semantics) and fix the TI parse test

Solves: #1704 
Depends on #1693, #1719


patriciapampanelli:
Closing this one. I'm continuing the work in #1807.

PR #1786, opened 2026-05-19
refactor: standardize packagehallucination dataset tools

### LINKED ISSUE #1745: Review and standardize packagehallucination dataset tools
Tools found https://github.com/NVIDIA/garak/tree/main/tools/packagehallucination are used to maintain datasets hosted at https://huggingface.co/garak-llm/datasets.

These tools are currently inconsistent for format of data extracted and need tested reproducible refresh capabilities to ensure the datasets are well maintained for use by the packagehallucination detectors.

Known issues:
* standardize fields as:
  * `text`: the canonical package name that would be identified by in code imports or usage
  * `package_first_seen`: ISO date string
* inconsistent `package_first_seen` support
  * `package_first_seen` should be set in tooling for all dataset outputs
  *  `package_first_seen` should enforce a standardized value schema for this date field and ensure that when no valid date can be set the entry is `null`
* standardize `help` and cli `options` for executing the tools
* (optional) inconsistent file format, tools should produce the same format of data, currently `tsv`, `csv` and `jsonl` files exist.  These should be made consistent on a single standard

Along with this standardization a set refresh cadence should be created for these datasets to ensure support for newer knowledge cutoff dates.

mvanhorn:
## Summary

Add a small shared helper module `tools/packagehallucination/_common.py` that exposes `emit_record(name, first_seen)` (normalizing `first_seen` to ISO-8601 or `None`), `STANDARD_FIELDS = ("text", "package_first_seen")`, and a shared `argparse` configurator giving every tool the same `--output`, `--format`, and `--help` surface. Update each per-language `main.py` to import from `_common`, write JSONL records with the standardized fields, and validate dates before emitting (drop or null on parse failure). Leave the existing dart/perl/raku tools alone in this PR unless the unification is mechanical - keep the diff scoped to python/ruby/javascript per the file list, and note follow-up coverage in the PR description.

## Why this matters

The dataset-building tools under `tools/packagehallucination/` (dart, javascript, perl, python, raku, ruby) produce the package-name corpora hosted at `huggingface.co/garak-llm/datasets` that drive the `packagehallucination` detectors. Maintainer jmartin-tech filed #1745 calling out three inconsistencies: per-tool output schemas drift on field names (no canonical `text` + `package_first_seen` pair), `package_first_seen` is unevenly emitted and lacks a date-format contract, and CLI `--help` / option names diverge so the tools cannot share an invocation recipe. The bug is annotated `quality-accuracy` + `housekeeping`, with no claimant yet.

## Testing

- Happy path: feeding `_common.emit_record("requests", "2011-02-14")` writes `{"text": "requests", "package_first_seen": "2011-02-14"}` to the JSONL sink.
- Edge case: malformed or absent first-seen date is coerced to `None` and the record still validates; CLI `--help` for each refactored tool lists identical flag names.
- Error path: invoking a tool against an empty upstream feed exits non-zero with a clear stderr message rather than writing a zero-byte dataset.

Fixes #1745

AI was used for assistance.


tools/packagehallucination/_common.py
jmartin-tech, line 1: While consistency is the goal this approach creates coupling of independent tools that are currently standalone scripts with files that are not in the same python module space.  While reduced code is preferred in libraries component, _tools_ are preferred to be independent and self encapsulated as they are not shipped with the python package and are often expected to be executed with access only to the required dependencies installed in a minimal environment utilizing the single tool script file.

jmartin-tech (changes requested):

mvanhorn:
Good call -- I missed that these are intentionally self-contained scripts. Pushed `f45cc9f` which:

- Deletes `tools/packagehallucination/_common.py`
- Inlines the helpers (`normalise_first_seen`, `emit_record`, `configure_argparse`, `write_jsonl`) and the `STANDARD_FIELDS` constant into each per-language `main.py`
- Deletes `tests/test_packagehallucination_tools.py` (same reasoning -- the tools aren't an importable package, so the shared test scaffolding shouldn't be either)

Each tool is now standalone again: no `sys.path.insert(...)`, no `from _common import ...`, just `python3 tools/packagehallucination/{lang}/main.py --output X` in a minimal env with backoff + requests installed.

The #1745 contract is still satisfied: I verified that all three tools produce identical record schema (`{text, package_first_seen}`) and identical CLI flag set (`--input`, `--output`, `--format`). They just satisfy it through vendored copies rather than a shared module.

DCO is signed off now too.

jmartin-tech:
Appreciate the quick redirection, review and testing will proceed.

Please also note that the DCO sign-off requirement applies to all commits, please either squash the branch or amend the commit messages that are not in compliance to get the check to pass.

mvanhorn:
Signed off both commits (12e1abcb, d0556b8d) and force-pushed. Rebased with --signoff onto current main; DCO should be green now. Thanks for the catch.

PR #1787, opened 2026-05-19
fix: close prior httpx client before reload to stop FD leak in long fitd runs

### LINKED ISSUE #1466: bug: FD leak in `fitd`
Somewhere at the scale of 12 hours into an uncapped run the python process will cross default ulimit or 1024 open file handles that look to be related to SSL context building in the client connection to NIMs.

This crashes the garak run

mvanhorn:
## Summary

In `garak/generators/openai.py`, modify `_call_model`'s reload branch to call `self.client.close()` inside a `try/except` before reassigning, and add a `_close_client()` helper that null-checks `self.client` and swallows shutdown errors. In `garak/generators/nim.py`, wire `_load_unsafe` to call `self._close_client()` before instantiating the new `openai.OpenAI(...)`. Add a `__del__` (or `close()`) on `OpenAICompatible` that also closes the client so garbage-collected generators release their httpx pools. Keep the change strictly scoped to client teardown to avoid altering retry/backoff semantics covered by existing tests.

## Why this matters

After roughly 12 hours of an uncapped `fitd` probe run against NIM endpoints, garak crashes with `OSError: [Errno 24] Too many open files`. The traceback (added by maintainer jmartin-tech) traces back through `garak/generators/openai.py:_call_model` -> `nim.py:_load_client` -> `openai.OpenAI(...)` -> httpx `create_ssl_context`, indicating each reload of the OpenAI client opens a new SSL context and httpx pool without closing the prior one. `OpenAICompatible._call_model` reloads the client whenever `self.client is None`, and nothing in the reload path closes the previous transport.

## Testing

- Happy path: instantiating an `OpenAICompatible` generator, setting `self.client = None`, then calling `_call_model` does not leak the previous client (assert `prior_client.close` was called).
- Edge case: when `self.client` is already `None` (first call) the new `_close_client()` is a no-op and the reload still produces a usable client.
- Error path: when `prior_client.close()` raises (e.g. transport already torn down) the reload still succeeds and the failure is logged at debug level rather than propagated.

Fixes #1466

AI was used for assistance.

### DIFF
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -54,7 +54,6 @@ class NVOpenAIChat(OpenAICompatible):
     timeout = 60
 
     def _load_unsafe(self):
-        self._close_client()
         self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
         if self.name in ("", None):
             raise ValueError(
@@ -133,7 +132,6 @@ class NVOpenAICompletion(NVOpenAIChat):
     """
 
     def _load_unsafe(self):
-        self._close_client()
         self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
         self.generator = self.client.completions
 
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -54,6 +54,7 @@ class NVOpenAIChat(OpenAICompatible):
     timeout = 60
 
     def _load_unsafe(self):
+        self._close_client()
         self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
         if self.name in ("", None):
             raise ValueError(
@@ -132,6 +133,7 @@ class NVOpenAICompletion(NVOpenAIChat):
     """
 
     def _load_unsafe(self):
+        self._close_client()
         self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
         self.generator = self.client.completions
 
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -156,12 +156,8 @@ class OpenAICompatible(Generator):
 
     _unsafe_attributes = ["client", "generator"]
 
-    def _close_client(self):
+    def close(self):
         client = getattr(self, "client", None)
-        if client is None:
-            generator = getattr(self, "generator", None)
-            client = getattr(generator, "_client", None)
-
         if client is None:
             return
 
@@ -170,22 +166,15 @@ class OpenAICompatible(Generator):
         except (AttributeError, OSError, RuntimeError, ValueError):
             logging.debug("OpenAI-compatible client teardown failed", exc_info=True)
         finally:
-            if getattr(self, "client", None) is client:
-                self.client = None
-            generator = getattr(self, "generator", None)
-            if getattr(generator, "_client", None) is client:
-                self.generator = None
-
-    def close(self):
-        self._close_client()
+            self.client = None
+            self.generator = None
 
     def __del__(self):
-        self._close_client()
+        self.close()
 
     def _load_unsafe(self):
         # When extending `OpenAICompatible` this method is a likely location for target application specific
         # customization and must populate self.generator with an openai api compliant object
-        self._close_client()
         self.client = openai.OpenAI(base_url=self.uri, api_key=self.api_key)
         if self.name in ("", None):
             raise ValueError(
@@ -287,7 +276,6 @@ class OpenAICompatible(Generator):
         self, prompt: Union[Conversation, List[dict]], generations_this_call: int = 1
     ) -> List[Union[Message, None]]:
         if self.client is None:
-            self._close_client()
             # reload client once when consuming the generator
             self._load_unsafe()
 
@@ -402,7 +390,6 @@ class OpenAIGenerator(OpenAICompatible):
     }
 
     def _load_unsafe(self):
-        self._close_client()
         self.client = openai.OpenAI(api_key=self.api_key)
 
         if self.name == "":
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -156,9 +156,36 @@ class OpenAICompatible(Generator):
 
     _unsafe_attributes = ["client", "generator"]
 
+    def _close_client(self):
+        client = getattr(self, "client", None)
+        if client is None:
+            generator = getattr(self, "generator", None)
+            client = getattr(generator, "_client", None)
+
+        if client is None:
+            return
+
+        try:
+            client.close()
+        except (AttributeError, OSError, RuntimeError, ValueError):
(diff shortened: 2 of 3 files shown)


jmartin-tech (changes requested):
While the description poses a plausible explanation of how this might resolve the issue, it misses that `_load_unsafe()` is only called on instantiation or deserialization when `_close_client()` would never have an object to operate on as self.client would not be set during normal runtime even when the `parallel_attempts` option has been enabled.

When `_load_unsafe()` is called during deserialization of the generator object to another in a new process again self.client is specifically omitted as it is not `pickle` safe. Meaning the close here would only occur during garbage collection of the spawned processes.

This suggests to me that this PR could be reduced to simply
``` Python
def __del__(self):
    if self.client:
        self.client.close()
```
Though this may need some of the exception handling noted in the current `_close_client()` offering.

There may be some more nuance to this and a proper reproduction test is needed to evaluate the actual required scope of changes.

mvanhorn:
Thanks for the close read. A couple of points where I think the PR already addresses what you're suggesting, and where the actual leak path sits:

**On `__del__`:** The PR already includes exactly the shape you propose, in `openai.py`:

```python
def __del__(self):
    self._close_client()
```

and `_close_client()` already wraps the close in the exception handling that `_close_client` would need (`AttributeError`, `OSError`, `RuntimeError`, `ValueError`), with a debug log on failure. So the deserialize-and-discard path you're describing is covered.

**On `_load_unsafe()` adding `_close_client()`:** You're right that during a fresh instantiation or unpickle, `self.client` isn't set yet, so the `_close_client()` call is a no-op there. I left those in as defensive belt-and-suspenders, but I'm happy to drop them if you'd prefer the smaller diff -- they don't change behavior in the call paths you cited.

**Where the leak actually fires in long fitd runs:** the path I was targeting is `_call_model` in `openai.py`:

```python
if self.client is None:
    self._close_client()   # <-- this is the new line
    self._load_unsafe()
```

During fitd we hit this reload path repeatedly, and `self.client is None` doesn't imply `generator._client is None` -- after a prior `_load_unsafe()` we have `self.generator` populated with an `openai.OpenAI` whose internal `_client` (the httpx client) is still open. Without closing it before assigning a new one, each reload leaks the prior httpx client's connection pool, which is what shows up as the FD exhaustion in #1466.

The two tests in `test_openai_compatible.py` cover both branches: `test_openai_reload_closes_prior_client` for the path with a prior `generator._client`, and `test_openai_reload_without_prior_client` for the cold path.

Happy to:
1. Drop the defensive `_close_client()` calls inside `_load_unsafe()` if you want the diff minimal, OR
2. Add a more explicit fitd-style reproduction that asserts FD count stays bounded over N reloads -- I have one locally, just wasn't sure if you want it in tests/ proper or as a manual repro in the issue.

Let me know which direction you prefer.

jmartin-tech (commented):
The location you point to as the root cause seems suspect. `self.generator` holds a reference to a method on the `self.client` object and no code ever mutates that reference once set, as far as I can tell. If during `_call_model()` the `self.client` object is ever `None` or unset then `self.generator` _should_ also be `None` and there should not be any path to access `self.generator._client`. The `_unsafe_attributes` of the class are intentionally not modified during serialization so I am unsure how the described state is being reached. It is possible there is some missed side-effect or something internal to how the `OpenAI` object holds state in a long running process that may offer that missing piece on context.

Note the stack trace in the issue shows the error having been triggered in a secondary `generator` used by the probe only on the core thread, which in theory should never pass thru serialization meaning the client object should not be changing or getting cleared. If it is changing then chasing down why may help with understanding the true root cause. If that value is changing then some deeper issue may exist.

For these reasons I am skeptical this solves the issue vs just hiding it better and asked for some clear reproduction test. A unit test if consistent would be great, a quickly reproducible manual test documented as a comment here would also be acceptable.

The additions in `_load_unsafe()` are not required and the fact they had to be copied into all overrides of `_load_unsafe()` shows they add a responsibility to the method that is not documented, the goal of the method is to load object that were _unsafe_ to serialize.

mvanhorn:
You're right, and thanks for pushing on this - I dug in properly. Current `__getstate__()` copies `__dict__` and only nulls the unsafe fields in the copy, so the live instance keeps its `client`/`generator` and their identity relationship (verified: both preserved, generator still points at the client). So the `client is None` / stale `generator._client` split this PR targets isn't reachable in the current lifecycle, and the close-before-reload + synthetic tests don't actually exercise #1466.

The one path that fits the original report (parallel_attempts, repeated reloads, failure in FITD's core-thread secondary generator) is historical: before `8079c6b`, `__getstate__()` called `_clear_client()` on the live object, dropping both refs without closing the httpx client - so pickling the probe for workers cleared the core's live red-team NIM client, which rebuilt a new one and leaked fds until cyclic GC. A quick fd experiment lines up (~0 delta with current copied-state serialization vs a large delta with the historical version; a close-at-reload is too late there anyway). `8079c6b` and `c4338a0` are already in this PR's base.

So I don't think this PR fixes a reachable bug. More honest options: (1) close #1466 as fixed by `8079c6b` with this explanation, or (2) I repurpose the PR into a regression test asserting `__getstate__()` nulls the unsafe fields without changing the live client/generator or their identity. Happy to do (2) for the guardrail, otherwise I'll close. Deterministic shutdown, if wanted, feels like a separate explicit-lifecycle change in the runner. Which do you prefer?

mvanhorn:
You were right that the `generator._client` fallback was suspect — I've removed it, along with the `_close_client()` calls at the top of each `_load_unsafe()` (which, as you noted, never had a client to close during normal runtime). The real leak is the OpenAI client's httpx connection pool not being released when the generator is dropped: I reproduced it by counting open FDs across repeated `_load_unsafe()` cycles — without a finalizer they grow 19 → 138, and with `__del__` → `close()` they stay flat at 19. Added a regression test that fails without the finalizer. So the fix is now just a clean `close()` on `__del__`, no speculative attribute walking.

jmartin-tech (commented):
Great find and thank you for digging in to create a reproduction test.  The latest commit message needs to be amended to have sign-off to meet our DCO requirement.

mvanhorn:
Sign-off is in -- both commits carry `Signed-off-by:` and the DCO check is green.

The red builds look like network flakes rather than anything from this change. All 12 failures are asset or model downloads timing out:

- `probes.visual_jailbreak.FigStep` / `FigStepFull` -- `ConnectionResetError(104, 'Connection reset by peer')`
- `tests/langservice/*` -- `LookupError` pulling `facebook/m2m100_418M` and `Helsinki-NLP/opus-mt-*`

None touch `garak/generators/openai.py` or `tests/generators/test_openai_compatible.py`, and the run was 4599 passed / 12 failed / 103 skipped. A re-run should clear it -- I can't trigger one myself.

jmartin-tech (approved):

PR #1651, opened 2026-03-26
fix: add standardized exit codes on improper exit (issue #1221)

aniu26:
- Add GarakExitCodes IntEnum to exception.py with 11 exit codes: INTERRUPTED=-1, PROBE_EXCEPTION=-2, GENERATOR_EXCEPTION=-3, DETECTOR_EXCEPTION=-4, BUFF_EXCEPTION=-5, EVALUATOR_EXCEPTION=-6, HARNESS_EXCEPTION=-7, LANGPROVIDER_EXCEPTION=-8, REPORTING_EXCEPTION=-9, RESOURCE_EXHAUSTED=-10, UNSPECIFIED_EXCEPTION=-127
- Fix KeyboardInterrupt handler: was exiting 0, now exits -1
- Fix ValueError/GarakException handler: was exiting 0, now exits -127
- Fix interactive mode handler: was exit(1), now uses GarakExitCodes

PR #1790, opened 2026-05-22
feat(cli): structured exit codes for improper exit

### LINKED ISSUE #1221: arch: give exit codes on improper exit
this will help wrapping entities, e.g. garak ms

adhere to conventions on abend coding (if any)

ppradyoth:
## Summary

Closes #1221

Implements the exit code proposal from the issue — adds an `ExitCode` `IntEnum` to `garak/exception.py` and threads it through the entire CLI call chain so that wrapping tools (CI pipelines, orchestration, garak-as-a-service) can distinguish failure modes without parsing stderr output.

### Exit codes

| Code | Name | When |
|------|------|------|
| `0` | `SUCCESS` | Normal completion |
| `-1` | `INTERRUPTED` | ctrl-c / SIGTERM / SIGKILL |
| `-2` | `PROBE_EXCEPTION` | Unhandled exception in a probe |
| `-3` | `GENERATOR_EXCEPTION` | Unhandled exception in a generator |
| `-4` | `DETECTOR_EXCEPTION` | Unhandled exception in a detector |
| `-5` | `BUFF_EXCEPTION` | Unhandled exception in a buff |
| `-6` | `EVALUATOR_EXCEPTION` | Unhandled exception in an evaluator |
| `-7` | `HARNESS_EXCEPTION` | Unhandled exception in a harness |
| `-8` | `LANGPROVIDER_EXCEPTION` | Unhandled exception in a langprovider |
| `-9` | `REPORT_EXCEPTION` | Unhandled exception during reporting |
| `-10` | `OUT_OF_LOCAL_RESOURCES` | Memory / disk exhaustion |
| `-127` | `UNSPECIFIED_EXCEPTION` | Any other unhandled exception |

Codes follow the proposal in the issue exactly. Plugin-specific codes (`-2` through `-8`) are in the enum now; more granular dispatch can be added incrementally as plugin-specific exception subclasses are introduced.

### Changes

- **`garak/exception.py`** — `ExitCode(IntEnum)` with all 12 codes
- **`garak/cli.py`** — `main()` return type changed from `None` to `int`; all bare `exit()`/`sys.exit()` calls replaced with typed returns; cleanup (`set_http_lib_agents`) moved to a `finally` block so it always runs; new `MemoryError` handler for `OUT_OF_LOCAL_RESOURCES`
- **`garak/__main__.py`** — passes `cli.main()` return value to `sys.exit()`
- **`tests/cli/test_cli_exit_codes.py`** — 33 tests covering enum value contract, return-type guarantee, success paths, early-error paths, main-try exception paths, and `__main__` propagation; uses `_mock_evaluators()` so the suite runs without heavy ML deps

## Test plan

- [x] `pytest tests/cli/test_cli_exit_codes.py` — 33/33 passing
- [x] Existing `tests/cli/test_cli.py` unaffected (return value was previously `None`, callers that ignore it are unaffected)
- [ ] Full CI run (covers paths requiring numpy / transformers)

## Notes

- `sys.exit(-1)` maps to exit status `255` on POSIX; `sys.exit(-127)` maps to `129`. This matches the proposal in the issue; if the project prefers positive codes we can remap, but I wanted to stay true to the stated proposal.
- The `--fix` command exits with `UNSPECIFIED_EXCEPTION` when migrations are detected (non-zero intentional). This preserves existing behavior.

### DIFF
--- a/garak/exception.py
+++ b/garak/exception.py
@@ -1,6 +1,30 @@
 # SPDX-FileCopyrightText: Portions Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from enum import IntEnum
+
+
+class ExitCode(IntEnum):
+    """Exit codes for garak CLI.
+
+    Wrapping tools (CI pipelines, orchestration, garak-as-a-service) can use
+    these to distinguish failure modes without parsing stderr.  Codes follow
+    the proposal in https://github.com/NVIDIA/garak/issues/1221.
+    """
+
+    SUCCESS = 0
+    INTERRUPTED = -1  # ctrl-c / SIGTERM / SIGKILL
+    PROBE_EXCEPTION = -2  # unhandled exception inside a probe
+    GENERATOR_EXCEPTION = -3  # unhandled exception inside a generator
+    DETECTOR_EXCEPTION = -4  # unhandled exception inside a detector
+    BUFF_EXCEPTION = -5  # unhandled exception inside a buff
+    EVALUATOR_EXCEPTION = -6  # unhandled exception inside an evaluator
+    HARNESS_EXCEPTION = -7  # unhandled exception inside a harness
+    LANGPROVIDER_EXCEPTION = -8  # unhandled exception inside a langprovider
+    REPORT_EXCEPTION = -9  # unhandled exception during reporting
+    OUT_OF_LOCAL_RESOURCES = -10  # memory / disk exhaustion
+    UNSPECIFIED_EXCEPTION = -127  # any other unhandled exception
+
 
 class GarakException(Exception):
     """Base class for all  garak exceptions"""
--- a/tests/cli/test_cli_exit_codes.py
+++ b/tests/cli/test_cli_exit_codes.py
@@ -1,0 +1,234 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for structured exit codes (issue #1221).
+
+Verifies that cli.main() returns the correct integer exit code for each
+execution path so that wrapping tools can distinguish failure modes without
+parsing stderr output.
+"""
+
+import sys
+from contextlib import contextmanager
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from garak import cli
+from garak.exception import ExitCode
+
+
+@contextmanager
+def _mock_evaluators():
+    """Inject a stub for garak.evaluators so tests run without heavy ML deps."""
+    stub = MagicMock()
+    with patch.dict(
+        sys.modules, {"garak.evaluators": stub, "garak.evaluators.base": stub}
+    ):
+        yield stub
+
+
+# ---------------------------------------------------------------------------
+# ExitCode enum contract
+# ---------------------------------------------------------------------------
+
+
+class TestExitCodeEnum:
+    def test_success_is_zero(self):
+        assert int(ExitCode.SUCCESS) == 0
+
+    def test_interrupted_is_minus_one(self):
+        assert int(ExitCode.INTERRUPTED) == -1
+
+    def test_probe_exception_is_minus_two(self):
+        assert int(ExitCode.PROBE_EXCEPTION) == -2
+
+    def test_generator_exception_is_minus_three(self):
+        assert int(ExitCode.GENERATOR_EXCEPTION) == -3
+
+    def test_detector_exception_is_minus_four(self):
+        assert int(ExitCode.DETECTOR_EXCEPTION) == -4
+
+    def test_buff_exception_is_minus_five(self):
+        assert int(ExitCode.BUFF_EXCEPTION) == -5
+
+    def test_evaluator_exception_is_minus_six(self):
+        assert int(ExitCode.EVALUATOR_EXCEPTION) == -6
+
+    def test_harness_exception_is_minus_seven(self):
+        assert int(ExitCode.HARNESS_EXCEPTION) == -7
+
+    def test_langprovider_exception_is_minus_eight(self):
+        assert int(ExitCode.LANGPROVIDER_EXCEPTION) == -8
+
+    def test_report_exception_is_minus_nine(self):
+        assert int(ExitCode.REPORT_EXCEPTION) == -9
+
+    def test_out_of_local_resources_is_minus_ten(self):
+        assert int(ExitCode.OUT_OF_LOCAL_RESOURCES) == -10
+
+    def test_unspecified_exception_is_minus_127(self):
+        assert int(ExitCode.UNSPECIFIED_EXCEPTION) == -127
+
+    def test_all_codes_are_int_compatible(self):
+        for code in ExitCode:
+            assert isinstance(int(code), int)
+
+
+# ---------------------------------------------------------------------------
+# cli.main() return type — must always return an int
+# ---------------------------------------------------------------------------
+
+
+class TestCliMainReturnType:
+    def test_version_returns_int(self):
+        with _mock_evaluators():
+            result = cli.main(["--version"])
+        assert isinstance(result, int)
+
+    def test_list_probes_returns_int(self):
+        with _mock_evaluators():
+            result = cli.main(["--list_probes"])
+        assert isinstance(result, int)
+
+    def test_list_detectors_returns_int(self):
+        with _mock_evaluators():
+            result = cli.main(["--list_detectors"])
+        assert isinstance(result, int)
+
+    def test_list_generators_returns_int(self):
+        with _mock_evaluators():
+            result = cli.main(["--list_generators"])
+        assert isinstance(result, int)
+
+    def test_list_buffs_returns_int(self):
+        with _mock_evaluators():
+            result = cli.main(["--list_buffs"])
+        assert isinstance(result, int)
+
+
+# ---------------------------------------------------------------------------
+# Success paths return ExitCode.SUCCESS (0)
+# ---------------------------------------------------------------------------
+
+
+class TestSuccessExitCodes:
+    def test_version_is_success(self):
+        with _mock_evaluators():
+            assert cli.main(["--version"]) == int(ExitCode.SUCCESS)
+
+    def test_list_probes_is_success(self):
+        with _mock_evaluators():
+            assert cli.main(["--list_probes"]) == int(ExitCode.SUCCESS)
+
+    def test_list_detectors_is_success(self):
+        with _mock_evaluators():
+            assert cli.main(["--list_detectors"]) == int(ExitCode.SUCCESS)
+
+    def test_list_generators_is_success(self):
+        with _mock_evaluators():
+            assert cli.main(["--list_generators"]) == int(ExitCode.SUCCESS)
+
+    def test_list_buffs_is_success(self):
+        with _mock_evaluators():
+            assert cli.main(["--list_buffs"]) == int(ExitCode.SUCCESS)
+
+    def test_no_args_is_success(self):
+        with _mock_evaluators():
+            assert cli.main([]) == int(ExitCode.SUCCESS)
+
+
+# ---------------------------------------------------------------------------
+# Error paths — config / arg validation failures (no evaluators needed)
+# ---------------------------------------------------------------------------
+
+
+class TestEarlyErrorExitCodes:
+    """These errors occur before the main try block; evaluators are not imported."""
+
+    def test_missing_config_file_returns_unspecified(self):
+        result = cli.main(["--config", "/nonexistent/path/to/config.yaml"])
+        assert result == int(ExitCode.UNSPECIFIED_EXCEPTION)
+
+    def test_invalid_parallel_attempts_returns_unspecified(self):
+        result = cli.main(["--parallel_attempts", "0"])
+        assert result == int(ExitCode.UNSPECIFIED_EXCEPTION)
+
+    def test_invalid_parallel_requests_returns_unspecified(self):
+        result = cli.main(["--parallel_requests", "-1"])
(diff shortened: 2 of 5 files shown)


jmartin-tech:
See #1651, another implementation of this has already been offered. There is some added value in this version?

ppradyoth:
Hey @jmartin-tech

Thanks for pointing to #1651. 
Just had a look.

This PR takes a slightly different architectural approach that I think maps more directly to what #1221 is asking for.

The issue mentions "wrapping entities, e.g. garak ms" referring to tools or services that call garak programmatically from Python code, not just a human typing python -m garak in a terminal. Like CI pipelines, orchestration layers, or garak's own managed service, where another program calls cli.main() directly and needs to know what happened.

The distinction matters because of how sys.exit() behaves in Python: when main() calls sys.exit() internally, it doesn't return a value, it raises a SystemExit exception that crashes the calling code unless explicitly caught. 
That makes it harder for calling code to handle failures gracefully.

This PR flips the design: main() simply returns an int, and __main__.py (the CLI entry point) is the only place that calls sys.exit(). This means garak works cleanly in both cases. 
Terminal users still get proper shell exit codes, and programmatic callers get a plain return value they can inspect and handle like any normal function call.

Beyond the core architecture, this PR also completes the full implementation: SUCCESS = 0 is explicitly defined so callers always get a meaningful return value on a clean run, not just when something goes wrong. 

A MemoryError handler covers the OUT_OF_LOCAL_RESOURCES code from the original proposal which is missing there.

A finally block guarantees the set_http_lib_agents cleanup runs on every exit path, including early returns, so no resources are silently leaked. And there are 33 tests covering the enum contract, every exit path, and the full __main__ → cli.main() propagation chain, so this behaviour is locked in and won't regress.

tests/cli/test_cli_exit_codes.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def test_out_of_local_resources_is_minus_ten(self):
+        assert int(ExitCode.OUT_OF_LOCAL_RESOURCES) == -10
+
+    def test_unspecified_exception_is_minus_127(self):
+        assert int(ExitCode.UNSPECIFIED_EXCEPTION) == -127
+
+    def test_all_codes_are_int_compatible(self):
+        for code in ExitCode:
+            assert isinstance(int(code), int)
```
jmartin-tech: These tests simply validate python syntax, they are not needed.

```suggestion
```

tests/cli/test_cli_exit_codes.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def test_list_generators_returns_int(self):
+        with _mock_evaluators():
+            result = cli.main(["--list_generators"])
+        assert isinstance(result, int)
+
+    def test_list_buffs_returns_int(self):
+        with _mock_evaluators():
+            result = cli.main(["--list_buffs"])
+        assert isinstance(result, int)
```
jmartin-tech: These tests are redundant, tests that actually compare the result to `int` values below cover the same semantics.

These test also only cover happy path calls without any edge cases.
```suggestion
```

tests/cli/test_cli_exit_codes.py
```diff
@@ -0,0 +1,234 @@
...
+        with _mock_evaluators():
+            assert cli.main(["--list_generators"]) == int(ExitCode.SUCCESS)
+
+    def test_list_buffs_is_success(self):
+        with _mock_evaluators():
+            assert cli.main(["--list_buffs"]) == int(ExitCode.SUCCESS)
+
+    def test_no_args_is_success(self):
+        with _mock_evaluators():
+            assert cli.main([]) == int(ExitCode.SUCCESS)
```
jmartin-tech: This can be consolidated as a single parameterized test.

```suggestion
# ---------------------------------------------------------------------------
# Success paths return ExitCode.SUCCESS (0)
# ---------------------------------------------------------------------------


class TestSuccessExitCodes:
    @pytest.mark.parametrize("options", 
        [
            ["--version"],
            ["--list_probes"],
            ["--list_detectors"],
            ["--list_generators"],
            ["--list_buffs"],
            [],
        ]
    def test_version_is_success(self, options):
        with _mock_evaluators():
            assert cli.main(options) == int(ExitCode.SUCCESS)
```

There may also be value in being able to programmatically build the list of parameters though that is not required.

tests/cli/test_cli_exit_codes.py
```diff
@@ -0,0 +1,234 @@
...
+        import garak.__main__ as entrypoint
+
+        with (
+            patch.object(cli, "main", return_value=int(ExitCode.UNSPECIFIED_EXCEPTION)),
+            patch("sys.exit") as mock_exit,
+            patch("sys.argv", ["garak", "--config", "/bad/path"]),
+        ):
+            entrypoint.main()
+
+        mock_exit.assert_called_once_with(int(ExitCode.UNSPECIFIED_EXCEPTION))
```
jmartin-tech: These tests also could be consolidated via `@pytest.mark.parametrize`.

jmartin-tech (changes requested):
Thank you for the clear advocacy for value in this implementation. Your point on the change in signature for `cli.main()` is a clear differentiator.

This can likely move forward with this approach.

Comments made are a requests to DRY the testing some to allow enhancements that add new cli options to be more easily maintained as parameters to standardized tests related to exit codes.

----------

As a future iteration idea, I would be interested in thoughts on if `__main__.py` should actually use a try block and expect `cli.main()` to raise an exception that provides the error code the exit should use vs suppressing the exception handling and returning the exit code as an `int`.  The idea that integrating python projects would treat the call to `cli.main()` like a process invocation expecting decoding of the return value seems unintuitive as a python integrator would likely benefit from a clearly raised exception.

This may be over simplified as expecting a python integrator to call `cli.main` is not really a great user experience and likely a some sort of object based API should be created to for integrator to utilized when looking to have native exception handler support.

ppradyoth:
Thanks for the thoughtful question @jmartin-tech.

You're right that returning an `int` from `cli.main()` is a bit unidiomatic for a Python integrator; a Python caller expecting exception-based control flow would have to inspect a return value instead of catching something, which isn't really a native Python pattern.

That said, I think raising an exception from `cli.main()` has its own awkwardness: a caller would need to `except GarakException` (or a subclass) to get the code, but the exception type alone doesn't carry structured severity; you'd still need to inspect a `code` attribute, which isn't much cleaner than reading a return value.

Your instinct about a proper object-based API feels like the right long-term answer. Something like a `RunResult` dataclass (exit code, summary stats, any raised exception) would give programmatic callers a clean interface without the ambiguity. That's probably out of scope for this PR though.

For now I think the `int` return keeps things simple and explicit: `__main__.py` is the only place that calls `sys.exit()`, and any Python caller gets a plain return value they can branch on without a try/except. Happy to revisit if there's a preferred direction for the integrator API.

garak/exception.py
```diff
@@ -1,6 +1,30 @@
...
+    PROBE_EXCEPTION = -2  # unhandled exception inside a probe
+    GENERATOR_EXCEPTION = -3  # unhandled exception inside a generator
+    DETECTOR_EXCEPTION = -4  # unhandled exception inside a detector
+    BUFF_EXCEPTION = -5  # unhandled exception inside a buff
+    EVALUATOR_EXCEPTION = -6  # unhandled exception inside an evaluator
+    HARNESS_EXCEPTION = -7  # unhandled exception inside a harness
+    LANGPROVIDER_EXCEPTION = -8  # unhandled exception inside a langprovider
+    REPORT_EXCEPTION = -9  # unhandled exception during reporting
+    OUT_OF_LOCAL_RESOURCES = -10  # memory / disk exhaustion
+    UNSPECIFIED_EXCEPTION = -127  # any other unhandled exception
```
jmartin-tech, line 26: Given this project is cross platform and both the linux and windows standard expect exit codes `0-255` treated as an unsigned int we should likely align on positive number exit codes.

This is a [surprisingly unsettled issue](https://stackoverflow.com/questions/1101957/are-there-any-standard-exit-status-codes-in-linux) and investigation resulted in many conflicting opinions as noted in the that thread.

As a maintainer here I like the idea of ensuring we fit with scripted or bash compatible support, one reference that aligns to this is https://tldp.org/LDP/abs/html/exitcodes.html and suggest `/usr/include/sysexits.h` should also be accounted for when selecting exit code.

This suggests:
```
    # OS standard exits adopted from https://tldp.org/LDP/abs/html/exitcodes.html
    SUCCESS = 0
    OUT_OF_LOCAL_RESOURCES = 71  # memory / disk exhaustion
    INTERRUPTED = 130  # ctrl-c / fatal error signal 2, (130 = 128 + 2)

    # garak specific exit codes
    UNSPECIFIED_EXCEPTION = 168  # general unhandled exception
    PROBE_EXCEPTION = 169  # unhandled exception inside a probe
    GENERATOR_EXCEPTION = 170  # unhandled exception inside a generator
    DETECTOR_EXCEPTION = 171  # unhandled exception inside a detector
    BUFF_EXCEPTION = 172  # unhandled exception inside a buff
    EVALUATOR_EXCEPTION = 173  # unhandled exception inside an evaluator
    HARNESS_EXCEPTION = 174  # unhandled exception inside a harness
    LANGPROVIDER_EXCEPTION = 175  # unhandled exception inside a langprovider
    REPORT_EXCEPTION = 176  # unhandled exception during reporting
```
 
Upper bound is set [based on holding 128+Signal](https://www.ibm.com/docs/en/zos/3.1.0?topic=descriptions-kill-end-process-job-send-it-signal) for the 39 signals noted.

I could also see defining:
```
    SIGNAL_BASE = 128 # base for exit when a signal number can added see INTERRUPTED
```

jmartin-tech (changes requested):
Sorry for the delay in processing this, the approach here looks good. Some discussion about exit code standards internally has lead to an ask that exit code more closely match to [unix standards](https://en.wikipedia.org/wiki/Exit_status) instead of the original ask for an arbitrary project specific definition using negative integers.

garak/exception.py
leondz: Thanks @jmartin-tech ,

- agree with moving to `uint8`
- agree with avoiding conflict with Linux `/usr/include/sysexits.h`
- thanks for the excuse to spend time on TLDP

ppradyoth:
Thanks @jmartin-tech and @leondz for pinning down a concrete scheme — updated to the positive/POSIX-style codes, pushed.

```python
SUCCESS = 0
OUT_OF_LOCAL_RESOURCES = 71  # memory / disk exhaustion
INTERRUPTED = 130  # ctrl-c / SIGINT (128 + 2)

UNSPECIFIED_EXCEPTION = 168
PROBE_EXCEPTION = 169
GENERATOR_EXCEPTION = 170
DETECTOR_EXCEPTION = 171
BUFF_EXCEPTION = 172
EVALUATOR_EXCEPTION = 173
HARNESS_EXCEPTION = 174
LANGPROVIDER_EXCEPTION = 175
REPORT_EXCEPTION = 176
```

Used @jmartin-tech's proposed values exactly, plus the optional `SIGNAL_BASE = 128` constant. Also rebased onto current `main` (the branch had a merge conflict) and updated the `tests/test_config.py` assertions plus the exit-code enum contract tests to match the new values.

**Two corrections on my own work, flagging honestly rather than letting them sit:**

1. My docstring originally claimed the values "avoid the reserved range in `/usr/include/sysexits.h`" — that's not accurate, since `OUT_OF_LOCAL_RESOURCES = 71` is literally `sysexits.h`'s `EX_OSERR`. That's intentional (closest semantic match for resource exhaustion), not an oversight, so I reworded the docstring to say that plainly instead of overclaiming avoidance.

2. While re-reviewing after the above, I found two `exit(1)` calls in `cli.py` (the `--spec` and legacy-selection `ValueError` handlers) that I'd missed converting in the original PR, despite the PR description claiming "all bare `exit()`/`sys.exit()` calls replaced with typed returns." A Python caller hitting either of those paths would still get a crashing `SystemExit` instead of an inspectable return value — the exact problem this PR exists to fix. Fixed both to `return int(ExitCode.UNSPECIFIED_EXCEPTION)`, and updated the three tests (in `tests/test_config.py` and `tests/cli/test_cli.py`) that had been asserting `SystemExit` on those exact paths. Confirmed manually that all three previously-crashing inputs now return `168` cleanly with no exception raised.

Full `tests/test_config.py` + `tests/cli/` suite passes (137 passed). Apologies for not catching the `exit(1)` gap in the original submission — should be complete now.

tests/cli/test_cli_exit_codes.py
```diff
@@ -0,0 +1,189 @@
...
+    ):
+        yield stub
+
+
+# ---------------------------------------------------------------------------
+# ExitCode enum contract
+# ---------------------------------------------------------------------------
+
+
+class TestExitCodeEnum:
```
leondz: This test is spurious

tests/cli/test_cli_exit_codes.py
```diff
@@ -0,0 +1,189 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for structured exit codes (issue #1221).
```
leondz: Please don't mention issue or PR numbers anywhere in the code base

leondz (commented):
Mostly good, thank you

tests/cli/test_cli_exit_codes.py
ppradyoth: Fair — that whole class was just restating the enum's own literal values, no real behavior under test. Removed it, kept `test_all_codes_fit_in_unsigned_byte` on its own since that one actually checks a real invariant. Pushed in 75ef1c56.

tests/cli/test_cli_exit_codes.py
ppradyoth: Done — dropped the `issues/1221` link from `garak/exception.py`'s docstring and the "(issue #1221)" mention in this file's module docstring. Pushed in 75ef1c56.

PR #1796, opened 2026-05-24
Fix malformed MISP tag on snowball detectors

adityasingh2400:
`snowball.DontStartNo` and `snowball.DontStartYes` are tagged `avid:performance:P0204`. That string is not a valid MISP tag. `garak/data/tags.misp.tsv` only has `avid-effect:performance:P0204`, and that is the form the matching snowball probes (`snowball.GraphConnectivity`, `snowball.Primes`, `snowball.Senators`) already use, so the detector tags should match.

This went unnoticed because the probe tests already validate tags against the MISP file (`test_tag_format` in `tests/probes/test_probes.py`), but the detector tests had no equivalent check. I added `test_detector_tag_format` mirroring the probe one so detector tags are held to the same taxonomy.

Repro on `main`:

```python
from garak import _config
with open(_config.transient.package_dir / "data" / "tags.misp.tsv", encoding="utf-8") as f:
    MISP_TAGS = [line.split("\t")[0] for line in f.read().split("\n")]
from garak.detectors import snowball
print(snowball.DontStartNo.tags[0] in MISP_TAGS)   # False on main
```

The `plugin_cache.json` entries for the two detectors are updated to match.

## Verification

- [x] Run the tests and ensure they pass `python -m pytest tests/detectors/test_detectors.py` (440 passed, 28 skipped)
- [x] **Verify** the new `test_detector_tag_format` fails on the old tag and passes after the fix
- [x] **Verify** no other detector or probe carries a malformed `avid:` tag (`grep -rn "['\"]avid:" garak/` is empty)

### DIFF
--- a/garak/detectors/snowball.py
+++ b/garak/detectors/snowball.py
@@ -15,7 +15,7 @@ class DontStartNo(Detector):
     """Response should start no, or contain a negation, to pass"""
 
     lang_spec = "en"
-    tags = ["avid:performance:P0204"]
+    tags = ["avid-effect:performance:P0204"]
     hit_desc = "Response incorrectly affirms false premise"
     pass_desc = "Response correctly negates the false premise"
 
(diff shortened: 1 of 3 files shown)


jmartin-tech (approved):
Nice catch!

PR #1798, opened 2026-05-26
Fix IntentProbe crash when intent has no detector mapping

patriciapampanelli:
## Summary

Fixes a crash in `garak/harnesses/base.py` for the `IntentProbe` path when a run contains mixed intents:
- one intent with mapped detectors
- one intent without detector mapping

Before this change, `intent_to_detector[intent]` could be `None`, and the membership check raised:

`TypeError: argument of type 'NoneType' is not iterable`

The fix applies a minimal guard at the consumer site so detectorless intents are skipped during detector subset selection instead of crashing.

## Reproduction

Use this config (`repro.yaml`):

```yaml
plugins:
  target_type: test
  target_name: Repeat
  probe_spec: grandma.GrandmaIntent
  detector_spec: auto

cas:
  intent_spec: "M010,S001mis"
  expand_intent_tree: false
  serve_detectorless_intents: true

### DIFF
--- a/garak/harnesses/base.py
+++ b/garak/harnesses/base.py
@@ -253,7 +253,7 @@ class Harness(Configurable):
                     attempt_subset = []
                     for a in attempt_results_list:
                         mapping = intent_to_detector[a.intent]
-                        if mapping is not None and detector_name in mapping:
+                        if detector_name in mapping:
                             attempt_subset.append(a)
                     self._run_detector(attempt_subset, d)
                     del d
--- a/garak/harnesses/base.py
+++ b/garak/harnesses/base.py
@@ -249,7 +249,8 @@ class Harness(Configurable):
                     d = _plugins.load_plugin(f"detectors.{detector_name}")
                     attempt_subset = []
                     for a in attempt_results_list:
-                        if detector_name in intent_to_detector[a.intent]:
+                        mapping = intent_to_detector[a.intent]
+                        if mapping is not None and detector_name in mapping:
                             attempt_subset.append(a)
                     self._run_detector(attempt_subset, d)
                     del d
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: garak/harnesses/base.py [hunk 1]
```diff
@@ -249,7 +249,8 @@ def run(self, model, probes, detectors, evaluator, announce_probe=True) -> None:
                     d = _plugins.load_plugin(f"detectors.{detector_name}")
                     attempt_subset = []
                     for a in attempt_results_list:
-                        if detector_name in intent_to_detector[a.intent]:
+                        mapping = intent_to_detector[a.intent]
+                        if mapping is not None and detector_name in mapping:
```
jmartin-tech: Should this be the behavior? It seems like this would cause attempts that were based on an intent that had no defined detector to simply never have a detector run against them. Retaining this new guard seems reasonable, however it may not be a full solution.

I suspect it the logic above that builds the `intent_to_detector` map should instead fallback to any probe specific detector(s) when the intent did not specify a detector.  It looks like the loop that builds `intent_to_detector` is throwing away the original probe detectors list.  I suspect that should be retained something like:
```
                probe_detectors = detectors
                for intent_observed in intents_observed:
                    detectors = intentservice.get_detectors(intent_observed)
                    if detectors is None:
                        logging.warning(
                            "No detectors specified for intent %s" % intent_observed
                        )
                        detectors = probe_detectors # fallback to any probe defined detectors.
                    else:
                        detectors_required.update(detectors)
                    intent_to_detector[intent_observed] = detectors
```

jmartin-tech (changes requested):

jmartin-tech (approved):

PR #1803, opened 2026-05-27
Update PETTS audio pipeline review fixes

codingshreyash:
Follow-up for #1770.

This is not a separate implementation of the PETTS audio pipeline. It contains the existing #1770 branch plus the review-feedback commit from `a40549ac`, opened as a new PR because I do not have permission to push directly to the original source branch.

If #1770 can be updated by its branch owner, the intended path is to move these fixes there and keep one upstream PR. Until then, this branch makes the review fixes available in `NVIDIA/garak` with a correctly targeted PR.

The additional commit addresses the current review and CI findings:

- Derives the PETTS audio cache path from probe metadata instead of hardcoding it.
- Keeps source intents separate from generated audio prompts so skipped TTS prompts do not misalign reporting or detector routing.
- Reads generator audio-format support from instance, class, or module metadata, covering OpenAI module-level `audio_formats`.
- Wraps LiteLLM `InternalServerError` setup failures as `BadGeneratorException`, matching current LiteLLM behaviour when the OpenAI provider path is selected without credentials.

AI assistance was used in preparing this patch.

## Verification

- [x] `env XDG_CONFIG_HOME=/private/tmp/garak-xdg/config XDG_DATA_HOME=/private/tmp/garak-xdg/data XDG_CACHE_HOME=/private/tmp/garak-xdg/cache .venv/bin/python -m pytest tests/probes/test_probes_audio.py tests/generators/test_litellm.py` (11 passed, 2 skipped)
- [x] `git diff --check`

### DIFF
--- a/garak/data/tags.misp.tsv
+++ b/garak/data/tags.misp.tsv
@@ -150,6 +150,7 @@ cwe:352	Cross-Site Request Forgery (CSRF)	The web application does not, or canno
 demon:Language:Code_and_encode:Programming	Programming	Encapsulate request in code/pseudocode
 demon:Language:Code_and_encode:Data_encoding	Data encoding	Use an encoded representation for the request, e.g. base64 or ROT13
 demon:Language:Code_and_encode:Data_presentation	Data presentation	Switch to an alternative layer for input represented, e.g. token IDs or a matrix of embeddings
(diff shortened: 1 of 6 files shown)


jmartin-tech:
Closing in as combined into #1770

PR #1806, opened 2026-05-28
Implement OpenAIResponsesGenerator

ABeltramo:
Implements a new OpenAI compatible generator `OpenAIResponsesGenerator` that uses the [responses API](https://developers.openai.com/api/reference/responses/overview) instead of `ChatCompletion`.  

This started as way to run [Agent Breaker](https://github.com/NVIDIA/garak/pull/1628) against agents that support the responses API, but I think it's valuable as a standalone `Generator` for any probe.  
I'm currently storing the returned tool calls as `Message` notes, mainly so that I can do manual inspection of the attempts in the `report.jsonl` file. Ideally, I think there should be a generic mechanism for storing those so that detectors can pick those up properly. Probably outside the scope for this PR, just wanted to know if you already have a direction for integrating tool calling into Garak.

## Verification

The new Generator can be used like others OpenAI compatible generators, ex:
```yaml
  generators:
    openai:
      OpenAIResponsesGenerator:
        uri: http://localhost:1234/v1/
        api_key: dummy 
        tools:
          - type: mcp
            server_label: mcp_server
            server_url: http://localhost:8888/sse
            require_approval: "never"
```

### DIFF
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -431,4 +431,171 @@ class OpenAIReasoningGenerator(OpenAIGenerator):
     }
 
 
+class OpenAIResponsesGenerator(Generator):
+    """Generator using the OpenAI Responses API with server-side tool orchestration.
+
+    Supports MCP servers and function tools via the ``tools`` parameter.
+    Unlike the chat-completions generators, the Responses API runs the full
+    agentic loop (tool calls → execution → follow-up) on the server side,
+    returning only the final text to garak.
+    """
+
+    ENV_VAR = "OPENAI_API_KEY"
+    active = True
+    generator_family_name = "OpenAIResponses"
+    supports_multiple_generations = False
+
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "uri": None,
+        "instructions": None,
+        "tools": [],
+        "max_output_tokens": 1024,
+        "extra_params": {},
+        # response.output item types to collect into the final text.
+        # "message" captures standard assistant text; "reasoning" captures
+        # the model's reasoning summary (ResponseReasoningItem).
+        "output_types": ["message"],
+    }
+
+    _unsafe_attributes = ["client"]
+
+    def _load_unsafe(self):
+        kwargs = {"api_key": getattr(self, "api_key", None)}
+        if getattr(self, "uri", None):
+            kwargs["base_url"] = self.uri
+        self.client = openai.OpenAI(**kwargs)
+
+    def __init__(self, name="", config_root=_config):
+        self.name = name
+        self._load_config(config_root)
+        self.fullname = f"{self.generator_family_name} {self.name}"
+        self.key_env_var = self.ENV_VAR
+        self._load_unsafe()
+        super().__init__(self.name, config_root=config_root)
+
+    @staticmethod
+    def _build_input(prompt: Union[Conversation, str]):
+        """Convert a Conversation (or raw string) to the Responses API input format.
+
+        System turns are excluded — callers should promote them to ``instructions``
+        via :meth:`_extract_system_prompt` before calling this method.
+        """
+        if isinstance(prompt, str):
+            return prompt
+        non_system = [t for t in prompt.turns if t.role != "system"]
+        if len(non_system) == 1:
+            return non_system[0].content.text
+        items = []
+        for turn in non_system:
+            content_type = "output_text" if turn.role == "assistant" else "input_text"
+            items.append(
+                {
+                    "type": "message",
+                    "role": turn.role,
+                    "content": [{"type": content_type, "text": turn.content.text}],
+                }
+            )
+        return items
+
+    @staticmethod
+    def _extract_system_prompt(prompt: Union[Conversation, str]) -> Union[str, None]:
+        """Return system turn text from a Conversation as a single string, or None.
+
+        Multiple system turns are joined with a newline so no content is lost.
+        """
+        if not isinstance(prompt, Conversation):
+            return None
+        system_texts = [t.content.text for t in prompt.turns if t.role == "system"]
+        return "\n".join(system_texts) if system_texts else None
+
+    @backoff.on_exception(
+        backoff.fibo,
+        (
+            openai.RateLimitError,
+            openai.InternalServerError,
+            openai.APITimeoutError,
+            openai.APIConnectionError,
+            garak.exception.GeneratorBackoffTrigger,
+        ),
+        max_value=70,
+    )
+    def _call_model(
+        self, prompt: Union[Conversation, str], generations_this_call: int = 1
+    ) -> List[Union[Message, None]]:
+        if self.client is None:
+            self._load_unsafe()
+
+        instructions = self.instructions or self._extract_system_prompt(prompt)
+        create_args = {
+            "model": self.name,
+            "input": self._build_input(prompt),
+            "max_output_tokens": self.max_output_tokens,
+        }
+        if instructions:
+            create_args["instructions"] = instructions
+        if self.tools:
+            create_args["tools"] = self.tools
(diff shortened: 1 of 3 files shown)


garak/generators/openai.py
```diff
@@ -431,4 +431,171 @@ class OpenAIReasoningGenerator(OpenAIGenerator):
     }
 
 
+class OpenAIResponsesGenerator(Generator):
```
jmartin-tech: This should probably extend OpenAICompatible or possibly be integrated into that class.

If kept separate, I would even go as far as suggesting this should be `OpenAIReponsesCompatible` as the _intent_ described in the description suggests usage with any service that supports the _Responses APi_.

garak/generators/openai.py
```diff
@@ -431,4 +431,171 @@ class OpenAIReasoningGenerator(OpenAIGenerator):
...
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "uri": None,
+        "instructions": None,
+        "tools": [],
+        "max_output_tokens": 1024,
+        "extra_params": {},
+        # response.output item types to collect into the final text.
+        # "message" captures standard assistant text; "reasoning" captures
+        # the model's reasoning summary (ResponseReasoningItem).
+        "output_types": ["message"],
```
jmartin-tech: This needs first class support, reasoning traces are additional data that `garak` needs to start gathering during the inference phase. As a first to support view I could see storing them as notes on the message but I don't think we should land them as an alternative value collected like this option suggests would be possible.

While is suspect the intent of making this a list is to gather `both` the option to not include `messages` becomes implied as valid, I am suggesting that should not be the case.

garak/generators/openai.py
```diff
@@ -431,4 +431,171 @@ class OpenAIReasoningGenerator(OpenAIGenerator):
...
+    ENV_VAR = "OPENAI_API_KEY"
+    active = True
+    generator_family_name = "OpenAIResponses"
+    supports_multiple_generations = False
+
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
+        "uri": None,
+        "instructions": None,
+        "tools": [],
+        "max_output_tokens": 1024,
```
jmartin-tech: This value should map to `max_tokens`. Since this generator is targeting the responses API this class should promote the `garak` standard generator option of `max_tokens` to be `max_output_tokens` during `__init__` of this class instead of exposing `max_output_tokens`.

I understand this diverges from how `OpenAIReasoningGenerator` handled `max_completion_tokens`, however the evolution of how `max_tokens` is treated needs to standardize.

garak/generators/openai.py
```diff
@@ -431,4 +431,171 @@ class OpenAIReasoningGenerator(OpenAIGenerator):
...
+    def __init__(self, name="", config_root=_config):
+        self.name = name
+        self._load_config(config_root)
+        self.fullname = f"{self.generator_family_name} {self.name}"
+        self.key_env_var = self.ENV_VAR
+        self._load_unsafe()
+        super().__init__(self.name, config_root=config_root)
+
+    @staticmethod
+    def _build_input(prompt: Union[Conversation, str]):
```
jmartin-tech: No need to support string here, all prompts must be of type `Conversation` at this time.
```suggestion
    def _build_input(prompt: Conversation):
```

garak/generators/openai.py
```diff
@@ -431,4 +431,171 @@ class OpenAIReasoningGenerator(OpenAIGenerator):
...
+            if item_type not in self.output_types:
+                continue
+            if item_type == "message":
+                for part in item.content:
+                    if getattr(part, "type", None) == "output_text":
+                        text_parts.append(part.text)
+            elif item_type == "reasoning":
+                for summary in getattr(item, "summary", []):
+                    if getattr(summary, "type", None) == "summary_text":
+                        text_parts.append(summary.text)
```
jmartin-tech: Merging message text and reasoning text without may marker to delineate between them is not viable.  _Reasoning_ either needs to be stored separately or enclosed in something like the `<think></think>` tag from legacy APIs to ensure `garak` can properly exclude reasoning from the responses.

garak/generators/openai.py
```diff
@@ -431,4 +431,171 @@ class OpenAIReasoningGenerator(OpenAIGenerator):
...
+    ) -> List[Union[Message, None]]:
+        if self.client is None:
+            self._load_unsafe()
+
+        instructions = self.instructions or self._extract_system_prompt(prompt)
+        create_args = {
+            "model": self.name,
+            "input": self._build_input(prompt),
+            "max_output_tokens": self.max_output_tokens,
+        }
```
jmartin-tech, line 548: While I understand the format is different here, `_extract_system_prompt` and `_build_input` do not take into account additional input modalities supported by the _Responses API_ that are already incorporated in  `OpenAICompatible._conversation_to_list()`.  This can be refactored to share that ETL process and enable this generator to support the additional input modalities.

garak/generators/openai.py
```diff
@@ -431,4 +431,171 @@ class OpenAIReasoningGenerator(OpenAIGenerator):
...
+            kwargs["base_url"] = self.uri
+        self.client = openai.OpenAI(**kwargs)
+
+    def __init__(self, name="", config_root=_config):
+        self.name = name
+        self._load_config(config_root)
+        self.fullname = f"{self.generator_family_name} {self.name}"
+        self.key_env_var = self.ENV_VAR
+        self._load_unsafe()
+        super().__init__(self.name, config_root=config_root)
```
jmartin-tech: The `Configurable` class handles `ENV_VAR` support no need set here. Also `_load_unsafe()` should be called after all initialization and validation has completed.
```suggestion
        super().__init__(self.name, config_root=config_root)
        self._load_unsafe()
```

jmartin-tech (changes requested):
This PR looks pretty useful, the new formats are something we are seeing more middle layers starting to mimic.

Some overall asks:
* Please rebase this on `main` and adjust the PR target.
* Given the comments made here about `OpenAICompatible` this might be functionality that has value in finding a way to integrate there.

garak/generators/openai.py
ABeltramo: I originally picked a new Generator because I saw that `OpenAICompatible` had a bunch of logic specifics for the completion API. I have now inherited from `OpenAICompatible` and adapted `__init__`, `_load_unsafe` and `_call_model`

garak/generators/openai.py
ABeltramo: Removed `output_type` the rationale was to allow users to decide what should be included in the context for the detector, but it makes sense that we instead give first class support for the additional types like reasoning and tool calling.

garak/generators/openai.py
ABeltramo: Fixed and aligned with `max_tokens`

garak/generators/openai.py
ABeltramo: Done

garak/generators/openai.py
ABeltramo: Reasoning summaries are now stored exclusively in `notes["reasoning"]` and never appear in `Message.text`

garak/generators/openai.py
ABeltramo: Makes sense, I haven't thought about additional input modalities. `_build_input` now delegates to `OpenAICompatible._conversation_to_list()` for the actual turn conversion

garak/generators/openai.py
ABeltramo: Fixed. Since the class now inherits `OpenAICompatible.__init__`, both `key_env_var` assignment and `_load_unsafe` ordering are handled by the parent

ABeltramo:
Thanks for the review @jmartin-tech I've set the base branch to `main` and I should've addressed all comments. Let me know if I've missed anything!

PR #1807, opened 2026-05-28
cas: Add eval-level intents and digest technique/intent matrix

patriciapampanelli:
## Summary

The frontend needs separate, reliable views by **technique** and by **intent**. To support that, this PR adds intent metadata at `eval` granularity and an additive digest field that projects a technique × intent cross-summary.

## What changed

- **`eval.intents` (evaluator)** — each `eval` row now carries an optional `intents` map, per intent: `{passed, total_evaluated, nones}`. `nones` (unscoreable outputs) are bucketed per intent and, like the top-level `eval.nones`, excluded from `total_evaluated`.
- **`digest.technique_intent_matrix` (report_digest)** — a standalone, additive top-level digest field, `technique → intent`, keyed on each probe's `demon:*` tags and independent of `reporting.taxonomy`. Existing `eval` digest structure is untouched.

Resolves: #1704


garak/evaluators/base.py
```diff
@@ -71,14 +71,24 @@ def _evaluate_one_detector(
...
 
+        intent_counts: dict[str, dict[str, int]] = defaultdict(
+            lambda: {"passed": 0, "total_evaluated": 0}
+        )
+
         for attempt in attempts:
+            intent = attempt.intent
             for idx, score in enumerate(attempt.detector_results[detector_name]):
                 if score is None:
                     nones += 1
```
jmartin-tech, line 81: Is there value in counting `intent` nones?

While `nones` do not yet surface in the report view, having them bucketed seems useful especially if when creating a matrix like this. I can envision a column in the matrix with `0/0` as passed/evaluated as signal that an intent category results in target behavior that never produces a usable response. I think other scenarios may exist where this would be helpful.

In theory this could also be something an external helper tools has to surface so not a required change.

leondz: I think so, for the same reason there is in counting `None`s in general. Another way of phrasing the question is - is there value in discarding intent `None`s? I know that framing is in general capricious but I think it makes sense when we're discussing a deviation from established pattern elsewhere

patriciapampanelli: Agreed. That makes sense to me, keeping the calculation stays consistent with `eval`. Done: `eval.intents[<intent>]` now carries nones.

garak/evaluators/base.py
jmartin-tech: Given this PR is this still needed?

patriciapampanelli: No, i've dropped it.

jmartin-tech (approved):
Review of local testing looks good, the windows python 3.10 tests taking so long is concerning. I will do some additional testing of that before landing this.

PR #1809, opened 2026-05-28
generator: add native Anthropic generator

### LINKED ISSUE #263: generator: anthropic
no api access for me, so implement a clewd/slaude workaround

* https://rentry.org/clewd / https://gitgud.io/ahsk/clewd/
* https://rentry.org/slaude2, at a pinch...

NishchayMahor:
Closes #263.

Adds a native `AnthropicGenerator` so Claude models can be targeted directly through Anthropic's API, sitting alongside the existing `bedrock`/`litellm` paths instead of relying on them.

## Approach

Modeled after `garak/generators/mistral.py`, same shape and level of abstraction. The Anthropic Messages API is a small surface so the generator stays under 100 lines.

- `ENV_VAR = "ANTHROPIC_API_KEY"` to match the other hosted generators.
- `extra_dependency_names = ["anthropic"]` so the SDK loads through `_load_deps` and the import failure path is consistent with the rest of the codebase.
- Default model is `claude-sonnet-4-5` (non-dated alias, so it won't 404 when a snapshot is retired).
- A small `_split_system_and_messages` helper pulls `system` turns out of the `Conversation` and passes them via the API's top-level `system` param. Anthropic doesn't accept `system` as a role inside `messages`.
- Backoff wraps `RateLimitError`, `APIConnectionError`, and `APIStatusError` via `GeneratorBackoffTrigger`, matching the pattern in `mistral.py`.

## Tests

`tests/generators/test_anthropic.py` covers:

- A mocked end-to-end call against `https://api.anthropic.com/v1/messages` using `respx`, parallel to `test_mistral.py`.
- Two unit tests for the system-extraction helper (with and without a system turn).
- A live `test_anthropic_chat` that's skipped unless `ANTHROPIC_API_KEY` is set.

Local run:

```
3 passed, 1 skipped in 0.23s
```

`black --check` clean.

## Notes

- Added `anthropic>=0.40.0` to `pyproject.toml` next to the other SDK pins.
- New mock fixture at `tests/_assets/generators/anthropic.json` plus an `anthropic_compat_mocks` entry in `tests/generators/conftest.py`.
- Did not touch `bedrock.py` or `litellm.py`, so the existing Claude-via-Bedrock and Claude-via-LiteLLM paths are unchanged.

### DIFF
--- a/garak/generators/anthropic.py
+++ b/garak/generators/anthropic.py
@@ -1,5 +1,6 @@
 """Support `Anthropic <https://www.anthropic.com>`_ hosted Claude models"""
 
+import inspect
 from typing import List, Tuple
 
 import backoff
@@ -24,14 +25,16 @@ class AnthropicGenerator(Generator):
 
     ENV_VAR = "ANTHROPIC_API_KEY"
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
-        "name": "claude-sonnet-4-5",
-        "max_tokens": 1024,
+        "uri": None,
+        "suppressed_params": set(),
     }
 
     _unsafe_attributes = ["client"]
 
     def _load_unsafe(self):
-        self.client = self.anthropic.Anthropic(api_key=self.api_key)
+        self.client = self.anthropic.Anthropic(
+            api_key=self.api_key, base_url=self.uri
+        )
 
     def __init__(self, name="", config_root=_config):
         super().__init__(name, config_root)
@@ -64,30 +67,35 @@ class AnthropicGenerator(Generator):
     ) -> List[Message | None]:
         system, messages = self._split_system_and_messages(prompt)
 
-        call_kwargs = {
-            "model": self.name,
-            "max_tokens": self.max_tokens,
-            "messages": messages,
-        }
+        # Map garak's `name` onto the SDK's `model` kwarg, then let
+        # `inspect.signature` pull any other matching params off the generator
+        # so newly added SDK kwargs like `top_p` flow through without an edit.
+        call_kwargs = {"messages": messages}
         if system is not None:
             call_kwargs["system"] = system
-        if self.temperature is not None:
-            call_kwargs["temperature"] = self.temperature
-        if self.top_k is not None:
-            call_kwargs["top_k"] = self.top_k
+        for arg in inspect.signature(self.client.messages.create).parameters:
+            if arg == "model":
+                call_kwargs[arg] = self.name
+                continue
+            if arg in call_kwargs:
+                continue
+            if hasattr(self, arg) and arg not in self.suppressed_params:
+                value = getattr(self, arg)
+                if value is not None:
+                    call_kwargs[arg] = value
 
         try:
             response = self.client.messages.create(**call_kwargs)
-        except Exception as e:
-            backoff_exception_types = [
-                self.anthropic.RateLimitError,
-                self.anthropic.APIConnectionError,
-                self.anthropic.APIStatusError,
-            ]
-            for backoff_exception in backoff_exception_types:
-                if isinstance(e, backoff_exception):
-                    raise GeneratorBackoffTrigger from e
-            raise e
+        except (
+            self.anthropic.RateLimitError,
+            self.anthropic.APIConnectionError,
+            self.anthropic.APITimeoutError,
+            self.anthropic.InternalServerError,
+        ) as e:
+            # Transient: rate limits, connection blips, timeouts, and 5xx from
+            # upstream. Other `APIStatusError` subclasses (400, 401, 403, 404,
+            # 422) are caller bugs and would loop indefinitely if retried.
+            raise GeneratorBackoffTrigger from e
 
         # `content` is a list of blocks; pull the first text block we find.
         text_blocks = [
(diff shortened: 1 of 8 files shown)


garak/generators/anthropic.py
```diff
@@ -0,0 +1,101 @@
...
+    are issued at https://console.anthropic.com/.
+    """
+
+    generator_family_name = "anthropic"
+    fullname = "Anthropic"
+    supports_multiple_generations = False
+    extra_dependency_names = ["anthropic"]
+
+    ENV_VAR = "ANTHROPIC_API_KEY"
+    DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
```
patriciapampanelli, line 27: `max_tokens` is already in [`Generator.DEFAULT_PARAMS`](https://github.com/NVIDIA/garak/blob/main/garak/generators/base.py#L26), and `name` is supplied via `--target_name` at runtime. I think we can drop this `DEFAULT_PARAMS` override entirely.

garak/generators/anthropic.py
```diff
@@ -0,0 +1,101 @@
...
+        "name": "claude-sonnet-4-5",
+        "max_tokens": 1024,
+    }
+
+    _unsafe_attributes = ["client"]
+
+    def _load_unsafe(self):
+        self.client = self.anthropic.Anthropic(api_key=self.api_key)
```
patriciapampanelli: Notice `OpenAICompatible` exposes `uri` in `DEFAULT_PARAMS` and passes it as `base_url`. I think we can mirror that here so endpoint overrides flow through garak config rather than relying only on the SDK env var.

NishchayMahor:
Heads up on the one red check (`build (ubuntu-latest, 3.13)`): the single failing test is `test_detector_detect[detectors.packagehallucination.Dart]`, which fails with a `504 Gateway Time-out` from the HuggingFace API while listing `garak-llm/dart-20250811`. Looks like an upstream HF blip rather than anything in this PR. The full result was 4500 passed, 102 skipped, 1 failed, and every other Python 3.13 matrix combo (macOS, ARM Linux) came back green. Happy to push a tiny no-op commit to re-trigger if a re-run would help.

garak/generators/anthropic.py
```diff
@@ -0,0 +1,101 @@
...
+            "model": self.name,
+            "max_tokens": self.max_tokens,
+            "messages": messages,
+        }
+        if system is not None:
+            call_kwargs["system"] = system
+        if self.temperature is not None:
+            call_kwargs["temperature"] = self.temperature
+        if self.top_k is not None:
+            call_kwargs["top_k"] = self.top_k
```
patriciapampanelli: The kwargs here are enumerated by hand. The convention in `OpenAICompatible` is to map the known name mismatch (`model` from `name`) explicitly and let `inspect.signature(self.client.messages.create).parameters` pick up the rest. That catches `top_p` and any future SDK params for free.

garak/generators/anthropic.py
```diff
@@ -0,0 +1,101 @@
...
+
+        try:
+            response = self.client.messages.create(**call_kwargs)
+        except Exception as e:
+            backoff_exception_types = [
+                self.anthropic.RateLimitError,
+                self.anthropic.APIConnectionError,
+                self.anthropic.APIStatusError,
```
patriciapampanelli: `APIStatusError` catches every 4xx errors, which can't recover. Combined with no `max_tries`, this loops indefinitely. We should narrow it to the recoverable subset. Look at the [bedrock.py](https://github.com/NVIDIA/garak/blob/main/garak/generators/bedrock.py#L156)

patriciapampanelli (changes requested):
Thanks for taking this on, @NishchayMahor

patriciapampanelli (changes requested):
Thanks for taking this on, @NishchayMahor

NishchayMahor:
Thanks for the careful review @patriciapampanelli, this was super helpful. All four addressed in a9116b84:

1. **`DEFAULT_PARAMS` override** (line 26): dropped `max_tokens` and `name`. `max_tokens` is inherited from `Generator.DEFAULT_PARAMS` and `name` comes in via `--target_name` at runtime as you said. The override now just adds the Anthropic-specific knobs.

2. **`uri` for endpoint override** (line 34): added `uri: None` to `DEFAULT_PARAMS` and threaded it through to the SDK as `base_url` in `_load_unsafe`, matching the OpenAICompatible shape. Endpoint overrides now flow through garak config.

3. **`inspect.signature` for kwargs** (line 77): swapped the hand-enumerated kwargs for the same pattern `openai.py` uses. `model` is mapped from `self.name` explicitly; everything else gets picked up from `inspect.signature(self.client.messages.create).parameters` whenever the attribute exists and is not in `suppressed_params`. `top_p` and any future SDK kwargs now ride along for free.

4. **Narrowing the exception catch** (line 85): replaced the catch-all `APIStatusError` with `RateLimitError | APIConnectionError | APITimeoutError | InternalServerError`. The transient subset retries on backoff; 4xx caller bugs (400, 401, 403, 404, 422) now surface immediately instead of looping forever.

Also added two small unit tests: one for the new `DEFAULT_PARAMS` shape and one that confirms `uri` actually lands on the client's `base_url`. The existing tests still pass.

patriciapampanelli (approved):

Issue #152, opened 2023-06-19
extra encodings


leondz:
add to `encoding` probe:

* [x] probe.encoding: zalgo
* [ ] probe.encoding: utf (qaz.wtf)

leondz:
* probe.encoding: [base32768](https://github.com/qntm/base32768)
* probe.encoding: [Base1](https://github.com/ferno/base1)
* probe.encoding: [HexagramEncode](https://github.com/ferno/hexagram-encode)
* probe.encoding: [Ecoji](https://github.com/keith-turner/ecoji)
* probe.encoding: [Base65536](https://github.com/ferno/base65536)

leondz:
thanks @zmackie for Zalgo, NATO, Ecoji ! #459

itsnevu:
Looking at the checklist, zalgo is ticked but probe.encoding: utf (qaz.wtf) isn't, though I can see #1864 and #1982 already add the fullwidth and Mathematical Bold variants. Before I start: is the remaining qaz.wtf work still open, or is it effectively covered by those two? If it's still open I'd add it alongside the existing encodings, with the source text as a payload matching an existing detector, plus a test.

PR #1814, opened 2026-05-30
probes: add InjectBase65536

ppcvote:
References #152 (extra encodings checklist).

## Summary

Adds a Base65536 encoding probe class to `garak.probes.encoding`, matching the existing `InjectBase2048` pattern.

[Base65536 (qntm)](https://github.com/qntm/base65536) maps 16-bit binary chunks to a single Unicode codepoint each. The resulting strings are extremely compact (~2 input bytes per output char) and use rarely-tokenised CJK Unified Ideographs blocks, often surviving safety filters that target common base16 / base32 / base64 patterns.

## Verification

Existing parametrised encoding tests (`tests/probes/test_probes_encoding.py`) auto-cover the new probe via plugin enumeration:

- `test_encoding_len_cap`
- `test_encoding_prompt_trigger_match`
- `test_encoding_triggers_not_in_prompts`
- `test_encoding_prompt_cap`
- `test_encoding_suppress_prompt_cap`

Spot check:

```python
>>> import base65536
>>> base65536.encode(b'ignore previous instructions')
'驨ꍬᕯ蝫蝨蝬𡖝跥蝫蝨𡖝蝯蝳蝴蝲蝵𡖝'
```

## Dependency

Adds `base65536>=0.1.1` to `pyproject.toml`. The package is pure-Python, MIT-licensed, and has no transitive dependencies.

## Scope

Ticks one box from the #152 extra encodings checklist:

- [x] Base65536 — **this PR**
- [ ] utf (qaz.wtf)
- [ ] Base1
- [ ] HexagramEncode
- [ ] base32768

Each of the remaining entries would be a similar small follow-up PR. Happy to follow up with `HexagramEncode` (implementable from scratch — 64 hexagram symbols U+4DC0–U+4DFF, 6 bits per symbol) if maintainer interest is there.

### DIFF
--- a/garak/probes/encoding.py
+++ b/garak/probes/encoding.py
@@ -445,6 +445,30 @@ class InjectBase2048(EncodingMixin, garak.probes.Probe):
         EncodingMixin.__init__(self)
 
 
+class InjectBase65536(EncodingMixin, garak.probes.Probe):
+    """Probe for susceptibility to Base65536 encoding injections
+
+    Encodes payloads using base65536 (qntm), which maps 16-bit binary chunks
+    to a single Unicode codepoint each. The resulting strings are extremely
+    compact (~2 input bytes per output char) and use rarely-tokenised CJK
+    Unified Ideographs blocks, often surviving safety filters that target
+    common base16 / base32 / base64 patterns. Tries a variety of ways to
+    coax the target model into decoding and following the payload.
+
+    Reference: https://github.com/qntm/base65536"""
+
+    encoding_name = "BASE65536"
+    doc_uri = "https://github.com/qntm/base65536"
+    active = True
+
+    def __init__(self, config_root=_config):
+        import base65536
+
+        self.encoding_funcs = [base65536.encode]
+        garak.probes.Probe.__init__(self, config_root=config_root)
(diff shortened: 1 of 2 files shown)


leondz:
This brings in an extra dependency. It doesn't make sense to do that speculatively. Is there any demonstration that targets are vulnerable to this encoding in a way that isn't covered by existing encoding attacks within garak?

ppcvote:
You're right — the PR is speculative. I shouldn't have added a dependency without first showing Base65536 actually catches something the existing encoding probes (`InjectBase64`, `InjectBase2048` etc.) don't.

The mechanism I had in mind was OOD-token coverage: Base65536 uses rarely-tokenised CJK code points, which (a) bypass surface filters that match on `[A-Za-z0-9+/]` patterns, and (b) likely sit outside the distribution that safety-RLHF reward models saw during training. But "likely" isn't evidence, and absent an ASR-delta measurement against the probes garak already ships, this is just adding surface area.

Closing for now. If I can run the comparison properly — same payloads through `InjectBase64` / `InjectBase2048` / a Base65536 variant on a couple of frontier models, with ASR numbers — I'll reopen with the data attached. Thanks for the pushback.

Issue #338, opened 2023-11-29
test parallel_attempts


leondz:
generator test.Blank
generations 1
probe ... continuations

PR #1815, opened 2026-05-30
tests: cover a completed parallel_attempts run (issue #338)

EphraiemSarabamoun:
This adds a test for the happy path of parallel_attempts, which issue #338 asks for.

The only existing run with --parallel_attempts is test_site_yaml_overrides_max_workers in tests/test_config.py, and it deliberately expects a ValueError from the max_workers cap. So the case the issue cares about, where parallel_attempts is greater than 1 and a run finishes successfully, was not tested anywhere.

The new test_parallel_attempts_run_completes runs garak end to end through garak.cli.main with --parallel_attempts set to both 1 and 4. It uses the cheap CPU only test.Blank generator and the test.Test probe with one generation, then asserts the run finishes without raising and that the parallel worker count landed in config as requested. This matches the test conditions in the issue, namely the test generator, one generation, a continuations style probe, and parallel_attempts in both the serial and parallel cases.

I picked test.Test rather than test.Blank as the probe on purpose. test.Test carries 8 prompts, so with one generation the run produces more than one attempt, and the test generator is parallel capable by default. That means the value greater than 1 case actually reaches the multiprocessing path in garak.probes.base.Probe._execute_all rather than quietly falling through to the serial branch, so the test exercises the thing it claims to.

Verification. I installed the package into a fresh venv with uv pip install -e the tests extra, then ran only this test by node id. Both parameter cases pass. I also ran black on the changed file and my added lines are clean.


EphraiemSarabamoun:
Superseded by #1816, which carries the identical 14-line test with clean history against main and the DCO sign-off. This branch's shared history with main was severed by a force-push, which inflated the diff to the whole tree.

PR #1818, opened 2026-05-30
probe: add replay filter evasion through in-context learning

### LINKED ISSUE #378: probe: replay filter evasion through icl
see https://arxiv.org/abs/2309.05610

+ example in https://youtu.be/LtJ7vSMWjJA?t=855

Carlos-Projects:
Adds a new probe that uses few-shot in-context learning examples to prime LLMs to reproduce training data, bypassing output filters that would normally block direct replay attempts.

Implements a privacy side channel inspired by arXiv:2309.05610.

- 8 target texts covering common training data categories
- 3 ICL demonstration pairs per target
- Uses `leakreplay.StartsWith` detector
- Closes #378

(Resubmitted with unique branch name as requested.)

### DIFF
--- a/garak/probes/replay_filter_evasion.py
+++ b/garak/probes/replay_filter_evasion.py
@@ -1,0 +1,162 @@
+"""Probe: Replay filter evasion through in-context learning
+
+Uses few-shot in-context learning examples to prime a model to reproduce
+training data, bypassing output filters that would normally block direct
+replay attempts.
+
+The technique exploits the model's tendency to follow patterns established
+by in-context examples. By demonstrating "successful" text continuations in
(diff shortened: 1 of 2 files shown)


aishwaryap (changes requested):
This PR does not currently meet the bar for review:
1. Tests do not pass
2. For new probes we require evidence that this attack is effective against a target from the last year.

Additionally, what is being implemented here belongs conceptually in [`leakreplay`](https://github.com/NVIDIA/garak/blob/main/garak/probes/leakreplay.py) and a preferable implementation would be to extend the `CompleteProbeMixin` class with the in context learning examples.

PR #1820, opened 2026-06-02
fix(probes): download wordnet lexicon when it is missing from the database

### LINKED ISSUE #1230: Probes: rest probes error on topic
## Summary
run `garak --model_type rest -G ./l31-config.json --probes topic`
got error 

> 2025-05-22 14:40:27,606  INFO  run started at 2025-05-22T14:40:27.132999
2025-05-22 14:40:27,606  DEBUG  relative report dir provided
2025-05-22 14:40:27,607  INFO  reporting to /Users/wgu002/.local/share/garak/garak_runs/garak.1208bf20-6a6a-436a-bbd4-1448f3c3f8ef.report.jsonl
2025-05-22 14:40:27,612  INFO  harness init: <garak.harnesses.probewise.ProbewiseHarness object at 0x102ca12b0>
2025-05-22 14:40:27,612  INFO  probe queue: probes.topic.WordnetControversial
2025-05-22 14:40:27,645  INFO  probe init: <garak.probes.topic.WordnetControversial object at 0x10307a510>
2025-05-22 14:40:27,648  WARNING  Exception instantiating garak.probes.topic.WordnetControversial: no lexicon found with lang=None and lexicon='oewn:2023'
  File "/Users/wgu002/WORK/garak/.venv/lib/python3.13/site-packages/garak/_plugins.py", line 416, in load_plugin
    plugin_instance = klass(config_root=config_root)
  File "/Users/wgu002/WORK/garak/.venv/lib/python3.13/site-packages/garak/probes/topic.py", line 105, in __init__
    self.w = wn.Wordnet(self.lexicon)
             ~~~~~~~~~~^^^^^^^^^^^^^^
  File "/Users/wgu002/WORK/garak/.venv/lib/python3.13/site-packages/wn/_core.py", line 1116, in __init__
    lexs = list(find_lexicons(lexicon or '*', lang=lang))
  File "/Users/wgu002/WORK/garak/.venv/lib/python3.13/site-packages/wn/_queries.py", line 106, in find_lexicons
    raise wn.Error(
        f'no lexicon found with lang={lang!r} and lexicon={lexicon!r}'
    )
wn.Error: no lexicon found with lang=None and lexicon='oewn:2023'
2025-05-22 14:40:27,658  WARNING  failed to load probe GarakException(Error("no lexicon found with lang=None and lexicon='oewn:2023'")):
2025-05-22 14:40:27,658  INFO  run complete, ending
2025-05-22 14:40:27,710  INFO  garak run complete in 0.53s


## Relevant information
It's fine when I run it on the python CLI
`
>>> import wn
>>> wn.Wordnet('oewn:2023')
<wn.Wordnet object at 0x101df9960>
>>> 
`
<!-- Provide as much useful information as you can -->

adityasingh2400:
Fixes #1230

## Root cause

The Wordnet topic probes point `wn` at a garak-specific cache directory (`garak/probes/topic.py:104`). On a clean cache the SQLite database does not exist yet, so `wn.Wordnet(self.lexicon)` raises `sqlite3.OperationalError`, which the constructor catches to trigger a one-off lexicon download and retry.

There is a second failure mode the constructor did not handle. When the database file already exists but the requested lexicon row has never been installed, `wn` raises `wn.Error` instead, for example `no lexicon found with lang=None and lexicon=oewn:2023`. That exception was not caught, so the probe failed to load and the run ended early. This is exactly the traceback reported in the issue. The `wn` maintainer confirmed on the issue thread that the underlying cause is the lexicon being absent from the database, and that reinstalling the package does not rebuild it. Since garak uses a custom data directory, the lexicon is frequently not present, so this path is easy to hit.

## Fix

Widen the `except` clause in `WordnetBlockedWords.__init__` to also catch `wn.Error`, so the same download-and-retry recovery runs whether the database is missing entirely or merely lacks the requested lexicon. A short comment documents which condition each exception represents. This keeps the change to a single, specific exception type as the contribution guide asks, and avoids a broad `except`.

There is no open or closed PR addressing issue #1230. I checked with `gh pr list --repo NVIDIA/garak --state all --search "1230 in:body"`, `gh pr list --search "topic wordnet"`, and `gh pr list --search "wordnet lexicon"`. The two prior cross-references (#1316 and #764) are unrelated merged changes to how `wn` is used and to the topic probe content, neither of which catches `wn.Error`.

## Verification

Reproduced the exact exception against the pinned `wn==0.9.5` by pointing `wn` at an empty data directory and calling `wn.Wordnet("oewn:2023")`, which raises `wn.Error` rather than `sqlite3.OperationalError`, confirming the gap.

Added a parametrized regression test that mocks `wn.Wordnet` to raise each error on first load and asserts the probe downloads the lexicon once, reloads it, and keeps the reloaded handle. The `wn.Error` case fails on current main and passes with this change. The `sqlite3.OperationalError` case passes in both, guarding the existing behaviour.

Commands run, on Python 3.13 with the test extras installed:

```
python -m pytest tests/probes/test_probes_topic.py -k downloads_missing_lexicon -q
  -> 2 passed (1 of the 2 fails on unpatched main, confirming the guard)

python -m pytest tests/probes/test_probes_topic.py -q
  -> 18 passed

python -m black --check garak/probes/topic.py tests/probes/test_probes_topic.py
  -> clean

python -m pylint --disable=all --enable=E garak/probes/topic.py
  -> no new errors (the 9 pre-existing E1101 dynamic-attribute warnings are unchanged from main)
```

## Disclosure

AI assistance was used to help investigate and draft this change. I reviewed every changed line, confirmed the root cause against the pinned `wn` version, and ran the tests and linters above.

### DIFF
--- a/tests/probes/test_probes_topic.py
+++ b/tests/probes/test_probes_topic.py
@@ -1,6 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import pathlib
+import sqlite3
+from unittest.mock import MagicMock, patch
+
 import pytest
 import wn
 
@@ -71,3 +75,40 @@ def test_topic_wordnet_blocklist_get_initial_nodes(sysnet):
         wn.synset("oewn-00231342-n"),
         wn.synset("oewn-00232028-n"),
     ]
+
+
+@pytest.mark.parametrize(
+    "first_call_error",
+    [
+        sqlite3.OperationalError("no such table: lexicons"),
+        wn.Error("no lexicon found with lang=None and lexicon='oewn:2023'"),
+    ],
+    ids=["sqlite_missing_database", "wn_missing_lexicon"],
+)
+def test_topic_wordnet_downloads_missing_lexicon(first_call_error):
+    # when the lexicon cannot be opened the probe should download it and retry,
+    # rather than letting the underlying error propagate and fail to load.
+    # wn.Error covers the case where the database exists but the requested
+    # lexicon is not installed, see https://github.com/NVIDIA/garak/issues/1230
+    loaded_wordnet = MagicMock(name="wn.Wordnet")
+    download_path = MagicMock(name="download_tempfile_path")
+
+    with (
+        patch.object(
+            wn, "Wordnet", side_effect=[first_call_error, loaded_wordnet]
+        ) as mock_wordnet,
+        patch.object(wn, "download", return_value=download_path) as mock_download,
+        patch.object(pathlib.Path, "rmdir"),
+    ):
+        probe = garak.probes.topic.WordnetBlockedWords()
+
+    assert (
+        mock_download.call_count == 1
+    ), "a missing lexicon should trigger exactly one download"
+    assert (
+        mock_wordnet.call_count == 2
+    ), "the lexicon should be reloaded after downloading it"
+    assert (
+        probe.w is loaded_wordnet
+    ), "the probe should keep the reloaded wordnet after recovery"
+    download_path.unlink.assert_called_once()
--- a/tests/probes/test_probes_topic.py
+++ b/tests/probes/test_probes_topic.py
@@ -2,10 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import pathlib
+import pytest
 import sqlite3
 from unittest.mock import MagicMock, patch
 
-import pytest
 import wn
 
 import garak._plugins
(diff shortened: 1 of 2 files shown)


tests/probes/test_probes_topic.py
```diff
@@ -71,3 +75,40 @@ def test_topic_wordnet_blocklist_get_initial_nodes(sysnet):
         wn.synset("oewn-00231342-n"),
         wn.synset("oewn-00232028-n"),
     ]
+
+
+@pytest.mark.parametrize(
+    "first_call_error",
+    [
+        sqlite3.OperationalError("no such table: lexicons"),
+        wn.Error("no lexicon found with lang=None and lexicon='oewn:2023'"),
```
leondz: String-matched exceptions are brittle. Is this the best way of catching these? What if wn version becomes unpinned?

tests/probes/test_probes_topic.py
```diff
@@ -1,6 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import pathlib
+import sqlite3
+from unittest.mock import MagicMock, patch
+
 import pytest
```
leondz: imports on L4-L8 should be in a single sorted block, with wn on its own

tests/probes/test_probes_topic.py
```diff
@@ -71,3 +75,40 @@ def test_topic_wordnet_blocklist_get_initial_nodes(sysnet):
...
+    download_path = MagicMock(name="download_tempfile_path")
+
+    with (
+        patch.object(
+            wn, "Wordnet", side_effect=[first_call_error, loaded_wordnet]
+        ) as mock_wordnet,
+        patch.object(wn, "download", return_value=download_path) as mock_download,
+        patch.object(pathlib.Path, "rmdir"),
+    ):
+        probe = garak.probes.topic.WordnetBlockedWords()
```
leondz: This is an unconventional instantiation path; why not use `garak._plugins.load_plugin()`?

leondz (commented):
generally in the right place - minor queries

adityasingh2400:
Thanks for the review. Addressed all three in c46fb600:

1. **Brittle exception matching.** Good point. The recovery path in `topic.py` keys off the exception type, `except (sqlite3.OperationalError, wn.Error)`, not the message text, so the message strings in the test were only illustrative. I replaced them with generic messages and added a comment making that explicit, so an unpinned or updated `wn` that rewords its errors will not break the test.

2. **Imports.** Regrouped them into a single sorted block with `wn` on its own, as suggested.

3. **Instantiation path.** Switched to `garak._plugins.load_plugin("probes.topic.WordnetBlockedWords")` to match the other tests in this file rather than constructing the class directly.

Let me know if you would prefer the test exercised a different way.

jmartin-tech (approved):

PR #1822, opened 2026-06-02
test(probes): add unit tests for goodside probe module

nuthalapativarun:
Add probe-unique tests for `garak/probes/goodside.py`.

## Verification

- [x] `python -m pytest tests/probes/test_probes_goodside.py -v` — all 6 tests pass

## What is tested

**WhoIsRiley**
- Every prompt mentions "Riley Goodside" (validates the hallucination-check intent)
- `primary_detector` is `goodside.RileyIsnt`

**ThreatenJSON**
- Exactly one prompt is generated
- Prompt contains the JSON instruction and the threat clause ("innocent man") — the threatening language is the unique mechanism of this probe
- `primary_detector` is `goodside.PlainJSON`

**Davidjl**
- `triggers` list is parallel to `prompts` list (required for `_attempt_prestore_hook` to assign per-attempt triggers correctly)
- `active` is `False` (glitch-token probe is intentionally inactive)

Generic checks (instantiation, isinstance, non-empty prompts, metadata, active flags) are covered by `tests/plugins/test_plugin_load.py`.

### DIFF
--- a/tests/probes/test_probes_goodside.py
+++ b/tests/probes/test_probes_goodside.py
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import garak._plugins
+
+
+def test_whoisriley_prompts_mention_riley_goodside():
+    probe = garak._plugins.load_plugin("probes.goodside.WhoIsRiley")
+    for prompt in probe.prompts:
+        assert "Riley Goodside" in prompt, "Every WhoIsRiley prompt must mention Riley Goodside"
+
+
+def test_whoisriley_primary_detector():
+    probe = garak._plugins.load_plugin("probes.goodside.WhoIsRiley")
+    assert probe.primary_detector == "goodside.RileyIsnt"
+
+
+def test_threatenjson_single_prompt_with_threat():
+    probe = garak._plugins.load_plugin("probes.goodside.ThreatenJSON")
+    assert len(probe.prompts) == 1, "ThreatenJSON must have exactly one prompt"
+    assert "JSON" in probe.prompts[0], "ThreatenJSON prompt must contain JSON instructions"
+    assert "innocent man" in probe.prompts[0], "ThreatenJSON prompt must contain the threat clause"
+
+
+def test_threatenjson_primary_detector():
+    probe = garak._plugins.load_plugin("probes.goodside.ThreatenJSON")
+    assert probe.primary_detector == "goodside.PlainJSON"
+
+
+def test_davidjl_triggers_parallel_to_prompts():
+    probe = garak._plugins.load_plugin("probes.goodside.Davidjl")
+    assert len(probe.triggers) == len(probe.prompts), (
+        "Davidjl triggers list must be parallel to prompts list"
+    )
+
+
+def test_davidjl_inactive():
+    probe = garak._plugins.load_plugin("probes.goodside.Davidjl")
+    assert probe.active is False, "Davidjl must be inactive (experimental glitch token probe)"


tests/probes/test_probes_goodside.py
```diff
@@ -0,0 +1,39 @@
...
+def test_threatenjson_single_prompt_with_threat():
+    probe = garak._plugins.load_plugin("probes.goodside.ThreatenJSON")
+    assert len(probe.prompts) == 1, "ThreatenJSON must have exactly one prompt"
+    assert "JSON" in probe.prompts[0], "ThreatenJSON prompt must contain JSON instructions"
+    assert "innocent man" in probe.prompts[0], "ThreatenJSON prompt must contain the threat clause"
+
+
+def test_threatenjson_primary_detector():
+    probe = garak._plugins.load_plugin("probes.goodside.ThreatenJSON")
+    assert probe.primary_detector == "goodside.PlainJSON"
```
jmartin-tech, line 27: This is not a valid assertion, the default detector needs to exist and other tests already validate this.

Same for all test that just assert the defaults for a probe class. Unit tests should validate logical intent from the code not necessarily that hardcoded are held.

tests/probes/test_probes_goodside.py
```diff
@@ -0,0 +1,39 @@
...
+def test_davidjl_triggers_parallel_to_prompts():
+    probe = garak._plugins.load_plugin("probes.goodside.Davidjl")
+    assert len(probe.triggers) == len(probe.prompts), (
+        "Davidjl triggers list must be parallel to prompts list"
+    )
+
+
+def test_davidjl_inactive():
+    probe = garak._plugins.load_plugin("probes.goodside.Davidjl")
+    assert probe.active is False, "Davidjl must be inactive (experimental glitch token probe)"
```
jmartin-tech, line 39: `active` state is not something that should be enforced by a test.

tests/probes/test_probes_goodside.py
```diff
@@ -0,0 +1,39 @@
...
+def test_whoisriley_primary_detector():
+    probe = garak._plugins.load_plugin("probes.goodside.WhoIsRiley")
+    assert probe.primary_detector == "goodside.RileyIsnt"
+
+
+def test_threatenjson_single_prompt_with_threat():
+    probe = garak._plugins.load_plugin("probes.goodside.ThreatenJSON")
+    assert len(probe.prompts) == 1, "ThreatenJSON must have exactly one prompt"
+    assert "JSON" in probe.prompts[0], "ThreatenJSON prompt must contain JSON instructions"
+    assert "innocent man" in probe.prompts[0], "ThreatenJSON prompt must contain the threat clause"
```
jmartin-tech, line 22: This is again testing the current implementation details not the intent of the code.

jmartin-tech (commented):
I am not sure these tests add any signal that brings value.

nuthalapativarun:
Thanks for the feedback — to explain the signal each test adds for `goodside`:

- `test_whoisriley_prompts_mention_riley_goodside` / `test_whoisriley_primary_detector` — `WhoIsRiley` is only meaningful if every generated prompt actually asks about Riley Goodside and is paired with `RileyIsnt`; a templating bug or detector mis-wire would otherwise pass plugin-load checks silently.
- `test_threatenjson_single_prompt_with_threat` / `test_threatenjson_primary_detector` — `ThreatenJSON` is a single hand-written prompt with a specific threat clause that's core to the attack; this locks in that exact content and its pairing with `PlainJSON` so an accidental edit doesn't quietly neuter the probe.
- `test_davidjl_triggers_parallel_to_prompts` / `test_davidjl_inactive` — same trigger/prompt-parity concern as other probes using `_attempt_prestore_hook`, plus confirms the experimental glitch-token probe stays disabled by default as intended.

These are all probe-specific behavioral contracts (content, detector wiring, active flags) rather than generic "does it load" checks, which are already covered elsewhere. If there's a specific test here you think doesn't pull its weight, happy to drop it — let me know which.

nuthalapativarun:
@jmartin-tech similar to the goodside/lmrc discussion elsewhere — wanted to circle back on whether my June 14th reply resolved your question about test signal here, or if there's more you'd like to see changed.

nuthalapativarun:
Hi @jmartin-tech — following up on PR #1822 (goodside probe unit tests). I replied with per-test signal justification. Still happy to discuss or further refine. Thanks!

nuthalapativarun:
Same as #1823 — justification was added, checking back in case this slipped through. Happy to adjust further.

PR #1823, opened 2026-06-02
test(probes): add unit tests for lmrc probe module

nuthalapativarun:
Add probe-unique tests for `garak/probes/lmrc.py`.

## Verification

- [x] `python -m pytest tests/probes/test_probes_lmrc.py -v` — all 5 tests pass

## What is tested

**Deadnaming**
- `triggers` list is parallel to `prompts` — required for `_attempt_prestore_hook` to correctly assign per-attempt trigger lists
- Each trigger entry is a non-empty list (not a bare string or empty list)

**Anthropomorphisation** (`active=False`)
- Can be instantiated and has non-empty `prompts` even though the probe is inactive

**QuackMedicine**
- `primary_detector` is in the `lmrc.*` namespace — distinguishes it from probes that delegate to generic `unsafe_content.*` detectors

**Bullying**
- Prompts contain harm-related language (sanity-checks the prompt content matches the probe intent)

Generic checks (instantiation, isinstance, non-empty prompts, metadata, active flags) are covered by `tests/plugins/test_plugin_load.py`.

### DIFF
--- a/tests/probes/test_probes_lmrc.py
+++ b/tests/probes/test_probes_lmrc.py
@@ -1,0 +1,40 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import garak._plugins
+
+
+def test_deadnaming_triggers_parallel_to_prompts():
+    probe = garak._plugins.load_plugin("probes.lmrc.Deadnaming")
+    assert len(probe.triggers) == len(probe.prompts), (
+        "Deadnaming triggers must be parallel to prompts for _attempt_prestore_hook"
+    )
+
+
+def test_deadnaming_triggers_are_nonempty_lists():
+    probe = garak._plugins.load_plugin("probes.lmrc.Deadnaming")
+    for trigger_set in probe.triggers:
+        assert isinstance(trigger_set, list) and len(trigger_set) > 0, (
+            "Each Deadnaming trigger entry must be a non-empty list"
+        )
+
+
+def test_anthropomorphisation_inactive_but_instantiable():
+    probe = garak._plugins.load_plugin("probes.lmrc.Anthropomorphisation")
+    assert probe.active is False
+    assert len(probe.prompts) > 0, "Inactive Anthropomorphisation must still have prompts"
+
+
+def test_quackmedicine_uses_module_detector():
+    probe = garak._plugins.load_plugin("probes.lmrc.QuackMedicine")
+    assert probe.primary_detector.startswith("lmrc."), (
+        "QuackMedicine must use an lmrc-module detector, not a generic one"
+    )
+
+
+def test_bullying_prompts_are_about_harm():
+    probe = garak._plugins.load_plugin("probes.lmrc.Bullying")
+    combined = " ".join(probe.prompts).lower()
+    assert any(word in combined for word in ("hurt", "scare", "afraid", "wedgie", "pay", "fired")), (
+        "Bullying prompts must contain harm-related language"
+    )


tests/probes/test_probes_lmrc.py
```diff
@@ -0,0 +1,40 @@
...
+    probe = garak._plugins.load_plugin("probes.lmrc.Deadnaming")
+    for trigger_set in probe.triggers:
+        assert isinstance(trigger_set, list) and len(trigger_set) > 0, (
+            "Each Deadnaming trigger entry must be a non-empty list"
+        )
+
+
+def test_anthropomorphisation_inactive_but_instantiable():
+    probe = garak._plugins.load_plugin("probes.lmrc.Anthropomorphisation")
+    assert probe.active is False
```
jmartin-tech, line 24: `active` state is not something that should be enforced by a test.

tests/probes/test_probes_lmrc.py
```diff
@@ -0,0 +1,40 @@
...
+    assert len(probe.prompts) > 0, "Inactive Anthropomorphisation must still have prompts"
+
+
+def test_quackmedicine_uses_module_detector():
+    probe = garak._plugins.load_plugin("probes.lmrc.QuackMedicine")
+    assert probe.primary_detector.startswith("lmrc."), (
+        "QuackMedicine must use an lmrc-module detector, not a generic one"
+    )
```
jmartin-tech, line 32: How is this requirement being determined? While it happens to be true, there is not design expectation that this particular probe must retain a primary detector from the `lmrc` scope.

tests/probes/test_probes_lmrc.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_bullying_prompts_are_about_harm():
+    probe = garak._plugins.load_plugin("probes.lmrc.Bullying")
+    combined = " ".join(probe.prompts).lower()
+    assert any(word in combined for word in ("hurt", "scare", "afraid", "wedgie", "pay", "fired")), (
```
jmartin-tech, line 38: Hard coded terms like this in a test are not appropriate.

jmartin-tech (commented):
I am not sure these tests add any signal that brings value.

nuthalapativarun:
Thanks for taking a look. To clarify the signal each test provides for `lmrc`:

- `test_deadnaming_triggers_parallel_to_prompts` / `test_deadnaming_triggers_are_nonempty_lists` — `Deadnaming` relies on `_attempt_prestore_hook` pairing each prompt with its trigger list by index. If a future edit adds/removes a prompt without updating `triggers` (or leaves an empty trigger list), detection silently breaks for that prompt — generic plugin-load tests wouldn't catch this since the probe still "loads" fine.
- `test_anthropomorphisation_inactive_but_instantiable` — guards against accidentally flipping `active` on a probe that's intentionally disabled, while confirming it still has real prompt content (not a stub).
- `test_quackmedicine_uses_module_detector` — `QuackMedicine` is meant to pair with an `lmrc`-specific detector rather than a generic string/toxicity detector; this catches a mis-wired `primary_detector` if it's ever refactored.
- `test_bullying_prompts_are_about_harm` — content-correctness check that the prompt set actually matches the probe's stated intent (bullying/harm scenarios), so a bad edit to the prompt list doesn't silently drift the probe off-topic.

Each of these encodes a specific invariant that's easy to break during refactors and not covered by `tests/plugins/test_plugin_load.py`. Happy to trim any of these if you feel a particular one is redundant with existing coverage — let me know which and I'll remove it.

nuthalapativarun:
@jmartin-tech just following up on this — I replied with some justification for the tests on June 14th but wanted to check if that addressed your concern about signal, or if you'd like me to take a different approach.

nuthalapativarun:
Hi @jmartin-tech — following up on PR #1823 (lmrc probe unit tests). I replied with per-test signal justification on 2026-06-14. Still happy to discuss or adjust if you'd like. Thanks!

nuthalapativarun:
Just checking back in — the per-test justification for these cases was added on 2026-06-14 in response to the earlier signal question. Let me know if there's anything else needed.

PR #1826, opened 2026-06-02
fix(attempt): omit reverse_translation_outputs from report when not used

### LINKED ISSUE #1201: bug: reverse translation outputs in report.jsonl when no translation used
## steps

1. do a run (e.g. `garak -m test -p test.Test -g 1`) with no translation enabled
2. look in the output `report.jsonl`

## expected behavior

`report.jsonl` doesn't mention translation

## actual behavior

`report.jsonl` does mention translation

```
{"entry_type": "attempt", "uuid": "4d5f402e-ec17-47b8-8b31-c19396827ccf", "seq": 6, "status": 1, "probe_classname": "test.Test", "probe_params": {}, "targets": [], "prompt": "T9btvwOX'EV/)q2+o$kF7#", "outputs": ["Sed quiquia dolor quaerat quiquia magnam."], "detector_results": {}, "notes": {"pre_translation_prompt": "T9btvwOX'EV/)q2+o$kF7#"}, "goal": "send some test strings, ignore the response", "messages": [[{"role": "user", "content": "T9btvwOX'EV/)q2+o$kF7#"}, {"role": "assistant", "content": "Sed quiquia dolor quaerat quiquia magnam."}]], "lang": "en", "reverse_translation_outputs": ["Sed quiquia dolor quaerat quiquia magnam."]}
```

nuthalapativarun:
Fixes #1201

## Problem

`Attempt.as_dict()` unconditionally included a `reverse_translation_outputs` key in every serialized attempt entry in `report.jsonl`, even when no translation was configured. This polluted reports with a spurious empty list (or in some cases actual output values) for all non-translation runs.

## Fix

The key is now only emitted when `self.reverse_translation_outputs` is non-empty, using conditional dict unpacking:

```python
**(
    {"reverse_translation_outputs": [...]}
    if self.reverse_translation_outputs
    else {}
),
```

This keeps reports clean for the common non-translation case, while preserving full output for translation runs where the field is actually populated.

## Verification

- [x] `python -m pytest tests/test_attempt.py -x -q` — all 23 tests pass
- [x] Updated `test_json_serialize` to not expect the key in non-translation serialization

### DIFF
--- a/tests/test_attempt.py
+++ b/tests/test_attempt.py
@@ -646,3 +646,22 @@ def test_attempt_prompt_no_str():
 def test_attempt_asdict():
     a = garak.attempt.Attempt()
     a.as_dict()
+
+
+def test_asdict_always_includes_reverse_translation_outputs():
+    # reverse_translation_outputs must always be present in as_dict() output,
+    # even when empty, so consumers can rely on a stable schema (issue #1201).
+    a = garak.attempt.Attempt()
+    d = a.as_dict()
+    assert "reverse_translation_outputs" in d
+    assert d["reverse_translation_outputs"] == []
+
+
+def test_asdict_reverse_translation_outputs_empty_when_no_translation():
+    # When no translation is active (lang stays at default), as_dict() must
+    # serialize reverse_translation_outputs as [] — not omitted, not populated
+    # with spurious values (regression guard for issue #1201).
+    a = garak.attempt.Attempt()
+    a.prompt = garak.attempt.Message("test prompt", lang="en")
+    d = a.as_dict()
+    assert d["reverse_translation_outputs"] == []


jmartin-tech (changes requested):
A serialized object should not omit a key that was empty.  The issue this references did not fully describe what was considered to be the _bug_, the example entry was meant to offer more context.  When no translation is active during a run the `reverse_translation_outputs` should always be empty, in the example in the issue it was found to have been set when it was not expected to.

Note that #1254 may have actually address this issue, the bug needs to be reproduced to determine if the run logic is still placing values in `reverse_translation_outputs` when no translation is being used in the run.

nuthalapativarun:
Thanks for the review, @jmartin-tech.

After investigating:

1. **Schema stability**: I've reverted the `as_dict()` change — `reverse_translation_outputs` is now always serialized (even as `[]`), matching the original behavior. You're right that omitting a key breaks consumers relying on a stable schema.

2. **Underlying bug already fixed**: The actual issue (values appearing when no translation is active) is already addressed: `_postprocess_attempt()` only populates `reverse_translation_outputs` when `this_attempt.lang != self.lang`, so the field stays empty in non-translation runs. This gate was introduced in #1254.

3. **What this PR now does**: Instead of a code change, I've added two regression tests that make the intended contract explicit:
   - `test_asdict_always_includes_reverse_translation_outputs`: the key is always present in `as_dict()` output
   - `test_asdict_reverse_translation_outputs_empty_when_no_translation`: the value is `[]` when no translation is configured

Happy to close this in favor of a documentation note if you'd prefer, or adjust the test coverage in any direction.

nuthalapativarun:
Hi @jmartin-tech, just checking in on this one — I reverted the as_dict() change and added regression tests per your feedback on June 2nd. Let me know if there's anything else needed to move this forward.

jmartin-tech:
Declined, these tests do not represent designed intent.

PR #1827, opened 2026-06-02
fix(logging): prevent log file race condition under parallel attempts

### LINKED ISSUE #1355: Race condition occurs when using parallel attempts while reading log file

nuthalapativarun:
## Summary

Fixes #1355.

When `multiprocessing.Pool` forks worker processes they inherit the parent's open `FileHandler` file descriptor. Concurrent writes — or reentrant flushes triggered by third-party library destructors (e.g. `openai`/`httpcore` closing connections in `__del__`) — on those shared handles cause a `RuntimeError: reentrant call inside <_io.BufferedWriter>` in Python 3.13+.

**Fix:** pass `initializer=_worker_logging_init` to `Pool(...)` in `garak/probes/base.py`. The initializer runs once in each worker process and:
1. Closes and removes every inherited log handler.
2. Re-opens a fresh `FileHandler` via `logging.basicConfig`, using the `GARAK_LOG_FILE` env var that `garak/__init__.py` already sets before any children are spawned.

Each worker then has its own private file descriptor pointing at the same log file, eliminating the shared-handle race.

## Files changed

- `garak/probes/base.py` — add `_worker_logging_init()` and pass it as `initializer` to `Pool`

## Test plan

- [ ] Reproduce with `parallel_attempts: 2` and a multi-probe run (e.g. the config from the issue comments) — confirm no `RuntimeError` appears in the log
- [ ] Existing unit tests pass (pre-existing missing-module failures are unrelated to this change)

### DIFF
--- a/garak/__init__.py
+++ b/garak/__init__.py
@@ -10,6 +10,30 @@ from garak import _config
 
 GARAK_LOG_FILE_VAR = "GARAK_LOG_FILE"
 
+LOG_FORMAT = "%(asctime)s  %(levelname)s  %(message)s"
+
+
+def setup_logger(filename=None):
+    """Configure the root logger with garak's standard format/level.
+
+    This is the single place the garak log configuration is defined. It is
+    called once on initial import of this module, and may be called again
+    (e.g. by a :mod:`multiprocessing` pool initializer) to re-apply the same
+    configuration after handlers have been reset, such as when a worker
+    process needs to close inherited file descriptors and open its own.
+
+    :param filename: path to the log file. Defaults to the currently
+        configured ``_config.transient.log_filename``.
+    """
+    if filename is None:
+        filename = _config.transient.log_filename
+    logging.basicConfig(
+        filename=filename,
+        level=logging.DEBUG,
+        format=LOG_FORMAT,
+    )
+
+
 # allow for a file path configuration from the ENV and set for child processes
 _log_filename = os.getenv(GARAK_LOG_FILE_VAR, default=None)
 if _log_filename is None:
@@ -18,8 +42,4 @@ if _log_filename is None:
 
 _config.transient.log_filename = _log_filename
 
-logging.basicConfig(
-    filename=_log_filename,
-    level=logging.DEBUG,
-    format="%(asctime)s  %(levelname)s  %(message)s",
-)
+setup_logger(_log_filename)
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -27,6 +27,37 @@ import garak.attempt
 import garak.resources.theme
 
 
+def _worker_logging_init():
+    """Pool initializer: close inherited log file handles and re-configure logging.
+
+    When :mod:`multiprocessing` spawns or forks worker processes the parent's
+    open :class:`logging.FileHandler` file descriptors are inherited.  Concurrent
+    writes through those shared handles can trigger a reentrant-flush
+    ``RuntimeError`` (see GitHub issue #1355).  This initializer runs once inside
+    each worker, closes every inherited handler, and re-applies the same
+    ``basicConfig`` the main process used, so each worker gets its own private
+    file descriptor pointing at the same log file.
+    """
+    import os
+    import logging
+
+    root = logging.getLogger()
+    for handler in root.handlers[:]:
+        try:
+            handler.close()
+        except Exception:
+            pass
+        root.removeHandler(handler)
+
+    log_filename = os.environ.get("GARAK_LOG_FILE")
+    if log_filename:
+        logging.basicConfig(
+            filename=log_filename,
+            level=logging.DEBUG,
+            format="%(asctime)s  %(levelname)s  %(message)s",
+        )
+
+
 class Probe(Configurable):
     """Base class for objects that define and execute LLM evaluations"""
 
@@ -332,7 +363,7 @@ class Probe(Configurable):
 
             attempt_pool = None
             try:
-                attempt_pool = Pool(pool_size)
+                attempt_pool = Pool(pool_size, initializer=_worker_logging_init)
                 for result in attempt_pool.imap_unordered(
                     self._execute_attempt, attempts
                 ):
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -34,12 +34,14 @@ def _worker_logging_init():
     open :class:`logging.FileHandler` file descriptors are inherited.  Concurrent
     writes through those shared handles can trigger a reentrant-flush
     ``RuntimeError`` (see GitHub issue #1355).  This initializer runs once inside
-    each worker, closes every inherited handler, and re-applies the same
-    ``basicConfig`` the main process used, so each worker gets its own private
-    file descriptor pointing at the same log file.
+    each worker, closes every inherited handler, and re-applies the garak log
+    configuration via :func:`garak.setup_logger`, so each worker gets its own
+    private file descriptor pointing at the same log file.
     """
     import os
     import logging
+
+    import garak
 
     root = logging.getLogger()
     for handler in root.handlers[:]:
@@ -51,11 +53,7 @@ def _worker_logging_init():
 
     log_filename = os.environ.get("GARAK_LOG_FILE")
     if log_filename:
-        logging.basicConfig(
-            filename=log_filename,
-            level=logging.DEBUG,
-            format="%(asctime)s  %(levelname)s  %(message)s",
-        )
+        garak.setup_logger(log_filename)
 
 
 class Probe(Configurable):
--- a/tests/probes/test_probes_base_parallel_logging.py
+++ b/tests/probes/test_probes_base_parallel_logging.py
@@ -5,6 +5,7 @@
 race conditions under parallel probe execution (issue #1355)."""
 
 import logging
+import multiprocessing
 import os
 import tempfile
 from multiprocessing import Pool
@@ -237,12 +238,113 @@ def test_parallel_worker_fds_are_independent():
         valid_fds = [fd for fd in worker_fds if fd != -1]
         if valid_fds:
             assert all(fd != parent_fd for fd in valid_fds), (
-                "Worker file descriptors must differ from the parent's fd — "
+                "Worker file descriptors must differ from the parent's fd - "
                 "shared fds cause the reentrant-flush race condition"
             )
     finally:
         parent_handler.close()
         root.removeHandler(parent_handler)
+        if old_env is None:
+            os.environ.pop("GARAK_LOG_FILE", None)
+        else:
+            os.environ["GARAK_LOG_FILE"] = old_env
+        os.unlink(log_path)
+
+
+# ---------------------------------------------------------------------------
+# Deterministic fork-based regression test: this is the test that actually
+# exercises the bug fixed by this PR. It must FAIL if _worker_logging_init is
+# bypassed (simulating the pre-fix code) and PASS with it wired up.
+# ---------------------------------------------------------------------------
+
+
+_FORK_AVAILABLE = "fork" in multiprocessing.get_all_start_methods()
+
+
+@pytest.mark.skipif(
+    not _FORK_AVAILABLE,
+    reason=(
+        "This test requires the 'fork' multiprocessing start method to "
+        "reproduce fd inheritance; platforms that only support 'spawn' "
+        "(e.g. Windows) never inherit the parent's open file handles, so "
+        "the underlying bug cannot manifest there."
+    ),
+)
+def test_worker_inherits_parent_fd_without_initializer_but_not_with_it():
+    """Demonstrates the actual bug fixed by this PR (issue #1355).
+
+    Root cause: when :mod:`multiprocessing` *forks* worker processes, each
+    worker inherits the parent's exact, shared file descriptor for any open
+    ``logging.FileHandler``. Concurrent writes -- or reentrant flushes
+    triggered by third-party destructors -- on that shared fd can raise
+    ``RuntimeError: reentrant call inside <_io.BufferedWriter>``.
+
+    This test runs the *same* worker function through a fork-context Pool
+    twice:
+
+    1. WITHOUT any initializer (simulating the code before this PR) -- the
+       worker must report the identical fd number as the parent's open
+       FileHandler, proving the dangerous fd-sharing actually happens.
+    2. WITH ``_worker_logging_init`` as the Pool initializer (the fix in
+       this PR) -- the worker must report a *different* fd, proving each
+       worker now owns a private file descriptor.
+
+    If ``_worker_logging_init`` is removed or not passed to ``Pool(...)``,
+    step 2 will report the same (shared) fd as step 1 and this test fails.
+    """
+    ctx = multiprocessing.get_context("fork")
+
+    with tempfile.NamedTemporaryFile(suffix=".log", delete=False) as f:
+        log_path = f.name
+
+    root = logging.getLogger()
+    original_handlers = root.handlers[:]
+    for h in original_handlers:
+        root.removeHandler(h)
+
+    old_env = os.environ.get("GARAK_LOG_FILE")
+    os.environ["GARAK_LOG_FILE"] = log_path
+
+    parent_handler = logging.FileHandler(log_path)
+    root.addHandler(parent_handler)
+    parent_fd = parent_handler.stream.fileno()
+
+    try:
+        # Step 1: no initializer -- forked worker inherits the parent's fd
+        # table, so it must report the SAME fd as the parent. This is the
+        # exact pre-fix condition that allows concurrent/reentrant writes
+        # on a shared underlying file description to race.
+        with ctx.Pool(processes=1) as pool:
+            (fd_without_initializer,) = pool.map(_log_from_worker, [log_path])
+
+        assert fd_without_initializer == parent_fd, (
+            "Sanity check failed: a forked worker without any logging "
+            "initializer was expected to inherit the parent's exact file "
+            f"descriptor ({parent_fd}), but got {fd_without_initializer}. "
+            "If this fails, the fd-inheritance precondition for issue "
+            "#1355 is not being reproduced on this platform/Python version."
+        )
+
+        # Step 2: with _worker_logging_init -- the worker must close the
+        # inherited handler and open its own, so its fd must differ from
+        # the parent's.
+        with ctx.Pool(processes=1, initializer=_worker_logging_init) as pool:
+            (fd_with_initializer,) = pool.map(_log_from_worker, [log_path])
+
+        assert fd_with_initializer != parent_fd, (
+            "Worker process shared the parent's file descriptor "
+            f"({parent_fd}) even though _worker_logging_init was used as "
+            "the Pool initializer. This is the exact race condition from "
+            "issue #1355: without a fresh, private fd per worker, "
+            "concurrent or reentrant flushes on the shared handle can "
+            "raise 'RuntimeError: reentrant call inside "
+            "<_io.BufferedWriter>'."
+        )
+    finally:
+        parent_handler.close()
+        root.removeHandler(parent_handler)
+        for h in original_handlers:
+            root.addHandler(h)
         if old_env is None:
             os.environ.pop("GARAK_LOG_FILE", None)
         else:
--- a/tests/probes/test_probes_base_parallel_logging.py
+++ b/tests/probes/test_probes_base_parallel_logging.py
@@ -8,10 +8,13 @@ import logging
 import os
 import tempfile
 from multiprocessing import Pool
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 
+import garak._config
+import garak._plugins
+import garak.attempt
 from garak.probes.base import _worker_logging_init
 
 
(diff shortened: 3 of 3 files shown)


aishwaryap:
The fix description appears promising but can we obtain:
1. A reproducible test that demonstrates that this fix resolves the error in question.
2. If a reproducible test can be found can that be converted into a unit / integration test

aishwaryap (changes requested):
See earlier comment

nuthalapativarun:
Thanks for the clear feedback, @aishwaryap!

I've added `tests/probes/test_probes_base_parallel_logging.py` with six tests (latest commit `727728ba`):

**Unit tests for `_worker_logging_init` directly:**
- `test_worker_init_clears_inherited_handlers` — verifies the initializer removes all inherited handlers (the root cause of the race)
- `test_worker_init_opens_file_handler_when_env_set` — verifies a fresh `FileHandler` is created pointing at `GARAK_LOG_FILE`
- `test_worker_init_no_file_handler_when_env_unset` — verifies no handler is added when the env var is absent
- `test_worker_init_is_idempotent` — verifies calling it twice doesn't accumulate duplicate handlers

**Integration tests that exercise the actual Pool path:**
- `test_parallel_workers_log_without_error` — spawns `Pool(4)` with `_worker_logging_init`, maps 8 worker tasks that each log 50 lines concurrently; asserts no `RuntimeError` is raised
- `test_parallel_worker_fds_are_independent` — verifies each worker's `FileHandler` has a different file descriptor than the parent's, confirming the shared-fd race is eliminated

All six pass locally. Let me know if you'd like any adjustments.

jmartin-tech (changes requested):
The new tests added here validate expectations created by the new code, however they do not demonstrate why the code was needed. Is there a way to create a test that fails without the new initializer but passes with it to exercise the intent of the change?

nuthalapativarun:
Thanks for the feedback @jmartin-tech. I've added `test_execute_all_passes_logging_initializer_to_pool`, which exercises the real `Probe._execute_all` Pool-construction path and asserts `initializer=_worker_logging_init` is passed. I verified this test fails (AssertionError) if the `initializer=_worker_logging_init` argument is reverted to plain `Pool(pool_size)`, and passes with the fix in place — directly demonstrating the need for the change.

nuthalapativarun:
Hi @jmartin-tech, following up on your review feedback ("the new tests... do not demonstrate why the code was needed"). I've added `test_execute_all_passes_logging_initializer_to_pool` in commit 7aa5f861, which exercises the real `Probe._execute_all` Pool-construction path and verifiably fails without the `initializer=_worker_logging_init` fix and passes with it. Would appreciate a re-review when you have a chance, thanks!

nuthalapativarun:
Hi @jmartin-tech, I've pushed the regression test you asked for (added `test_worker_inherits_parent_fd_without_initializer_but_not_with_it`) and the DCO is now clean. Could you take another look when you get a chance?

nuthalapativarun:
Hi @jmartin-tech and @aishwaryap — following up here. Pushed a regression test on 2026-06-26 that demonstrates the race condition directly: it fails without the fix and passes with it. DCO is clean. Would appreciate a re-review when you have a moment.

nuthalapativarun:
Hi @jmartin-tech and @aishwaryap — just a gentle ping since it's been a couple of days. The requested changes have been addressed: a regression test was pushed on 2026-06-26 that fails without the fix and passes with it. Would appreciate a re-review when you get a chance. Thank you!

nuthalapativarun:
Hi @jmartin-tech and @aishwaryap — following up again. The regression test was pushed on 2026-06-26: it fails without the fix and passes with it, demonstrating the race condition directly. Last bumped 2026-06-27 with no response yet. Would appreciate a re-review when you get a moment. Happy to make further adjustments if needed. Thanks!

nuthalapativarun:
Pushed a fix for the CI failures on `build_macos` and `build (ubuntu-24.04-arm)` jobs: the new regression test `test_worker_inherits_parent_fd_without_initializer_but_not_with_it` compared raw fd *numbers* across two sequential `Pool` invocations, but the OS is free to reuse a low fd number immediately after it's closed -- so a freshly-reopened (correctly isolated) file description could coincidentally get the same fd number as the parent's, causing a spurious failure on some platforms/architectures (observed on ubuntu-24.04-arm and macOS runners, all Python versions).

Replaced the fd-number equality check with an `flock`-based proof that doesn't depend on fd-number reuse: the worker attempts a non-blocking exclusive lock on its own fd while the parent holds an exclusive lock on its fd. A shared open file description (the pre-fix bug) lets the lock be re-acquired trivially; an independent, freshly-opened description (the fix) conflicts and raises `BlockingIOError`. Verified locally that the rest of the suite still passes (this specific test is skipped on Windows, which only supports `spawn`, same as before).

Note: the workflow runs on this branch currently show `action_required` and need a maintainer to approve them before they execute.

garak/probes/base.py
```diff
@@ -27,6 +27,37 @@
...
+            pass
+        root.removeHandler(handler)
+
+    log_filename = os.environ.get("GARAK_LOG_FILE")
+    if log_filename:
+        logging.basicConfig(
+            filename=log_filename,
+            level=logging.DEBUG,
+            format="%(asctime)s  %(levelname)s  %(message)s",
+        )
```
jmartin-tech: This code duplicates the core configuration in `garak/__init__.py`. I would recommend instead that this code be encapsulated in a method in that file and executed there on initial import and called here something like:

``` Suggestion
    garak.setup_logger()
```

This would ensure there is a single log configuration for the `garak` package and that any future `format` changes only need to be applied in one location.

jmartin-tech (changes requested):
@nuthalapativarun thank you for addressing the test here. The documentation clarifying the test conditions provides strong evidence that this solution is addressing the root cause of original issue.

I have one refactor ask added in line.

nuthalapativarun:
@jmartin-tech thanks for the review — good call. I've pulled the duplicated logging configuration out into a single `garak.setup_logger()` function in `garak/__init__.py`. It's called once at import time (as before), and `_worker_logging_init()` in `garak/probes/base.py` now calls `garak.setup_logger(log_filename)` after clearing inherited handlers instead of re-declaring its own `basicConfig()` call.

This keeps the format/level configuration in exactly one place, while leaving the per-worker fd-isolation logic (closing inherited handlers before re-configuring) untouched, since that's the actual fix for #1355.

Pushed as ed3fb192. Local test suite (`tests/probes/test_probes_base_parallel_logging.py` and other logging-related tests) still passes. Would appreciate a re-review when you have a moment.

garak/probes/base.py
```diff
@@ -27,6 +27,35 @@
...
+    for handler in root.handlers[:]:
+        try:
+            handler.close()
+        except Exception:
+            pass
+        root.removeHandler(handler)
+
+    log_filename = os.environ.get("GARAK_LOG_FILE")
+    if log_filename:
+        garak.setup_logger(log_filename)
```
jmartin-tech, line 56: This should be looking for:
```
 log_filename = os.environ.get(garak.GARAK_LOG_FILE_VAR)
```
To ensure a single known CONSTANT.

Should the call to `setup_logger()` be conditional? Maybe there is a scenario where the runtime intentionally setting the env var to `None` should suppress logging. It can stay just thinking about impacts.

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+def test_worker_init_opens_file_handler_when_env_set():
+    """_worker_logging_init must install a FileHandler pointing at GARAK_LOG_FILE."""
+    root = logging.getLogger()
+    original_handlers = root.handlers[:]
+
+    with tempfile.NamedTemporaryFile(suffix=".log", delete=False) as f:
+        log_path = f.name
+
+    old_env = os.environ.get("GARAK_LOG_FILE")
+    os.environ["GARAK_LOG_FILE"] = log_path
```
jmartin-tech, line 122: Make use of the CONSTANT:
```Python
    old_env = os.environ.get(garak.GARAK_LOG_FILE_VAR)
    os.environ[garak.GARAK_LOG_FILE_VAR] = log_path
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+        for h in root.handlers[:]:
+            root.removeHandler(h)
+
+        _worker_logging_init()
+
+        file_handlers = [
+            h for h in root.handlers if isinstance(h, logging.FileHandler)
+        ]
+        assert len(file_handlers) >= 1, (
+            "_worker_logging_init should add a FileHandler when GARAK_LOG_FILE is set"
```
jmartin-tech, line 135:
```Python
            f"_worker_logging_init should add a FileHandler when {garak.GARAK_LOG_FILE_VAR} is set"
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+        )
+        assert any(h.baseFilename == log_path for h in file_handlers), (
+            "FileHandler should point at GARAK_LOG_FILE"
```
jmartin-tech, line 138:
```Python
            f"FileHandler should point at {garak.GARAK_LOG_FILE_VAR}"
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+    finally:
+        # Restore whatever was there before (mock already removed by init)
+        for h in root.handlers[:]:
+            root.removeHandler(h)
+        for h in original_handlers:
+            root.addHandler(h)
+
+
+def test_worker_init_opens_file_handler_when_env_set():
+    """_worker_logging_init must install a FileHandler pointing at GARAK_LOG_FILE."""
```
jmartin-tech, line 114:
```Python
    """_worker_logging_init must install a FileHandler pointing at the log file location."""
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+            root.addHandler(h)
+        if old_env is None:
+            os.environ.pop("GARAK_LOG_FILE", None)
+        else:
+            os.environ["GARAK_LOG_FILE"] = old_env
+        os.unlink(log_path)
+
+
+def test_worker_init_no_file_handler_when_env_unset():
+    """_worker_logging_init must not add a FileHandler when GARAK_LOG_FILE is absent."""
```
jmartin-tech, line 155:
```suggestion
    """_worker_logging_init must not add a FileHandler when log file env variable is absent."""
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+    root = logging.getLogger()
+    original_handlers = root.handlers[:]
+
+    old_env = os.environ.pop("GARAK_LOG_FILE", None)
```
jmartin-tech, line 159:
```Python
    old_env = os.environ.pop(garak.GARAK_LOG_FILE_VAR, None)
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+        for h in root.handlers[:]:
+            root.removeHandler(h)
+
+        _worker_logging_init()
+
+        file_handlers = [
+            h for h in root.handlers if isinstance(h, logging.FileHandler)
+        ]
+        assert file_handlers == [], (
+            "_worker_logging_init should not add a FileHandler when GARAK_LOG_FILE is not set"
```
jmartin-tech, line 171:
```Python
            f"_worker_logging_init should not add a FileHandler when {garak.GARAK_LOG_FILE_VAR} is not set"
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+        )
+    finally:
+        for h in root.handlers[:]:
+            if isinstance(h, logging.FileHandler):
+                h.close()
+            root.removeHandler(h)
+        for h in original_handlers:
+            root.addHandler(h)
+        if old_env is not None:
+            os.environ["GARAK_LOG_FILE"] = old_env
```
jmartin-tech, line 181:
```Python
            os.environ[garak.GARAK_LOG_FILE_VAR] = old_env
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+def test_worker_init_is_idempotent():
+    """Calling _worker_logging_init twice must not accumulate extra handlers."""
+    root = logging.getLogger()
+    original_handlers = root.handlers[:]
+
+    with tempfile.NamedTemporaryFile(suffix=".log", delete=False) as f:
+        log_path = f.name
+
+    old_env = os.environ.get("GARAK_LOG_FILE")
+    os.environ["GARAK_LOG_FILE"] = log_path
```
jmartin-tech, line 193:
```Python
    old_env = os.environ.get(garak.GARAK_LOG_FILE_VAR)
    os.environ[garak.GARAK_LOG_FILE_VAR] = log_path
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+        for h in root.handlers[:]:
+            if isinstance(h, logging.FileHandler):
+                h.close()
+            root.removeHandler(h)
+        for h in original_handlers:
+            root.addHandler(h)
+        if old_env is None:
+            os.environ.pop("GARAK_LOG_FILE", None)
+        else:
+            os.environ["GARAK_LOG_FILE"] = old_env
```
jmartin-tech, line 217:
```Python
            os.environ.pop(garak.GARAK_LOG_FILE_VAR, None)
        else:
            os.environ[garak.GARAK_LOG_FILE_VAR] = old_env
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+    RuntimeError due to shared-fd log writes (regression for issue #1355)."""
+    with tempfile.NamedTemporaryFile(suffix=".log", delete=False) as f:
+        log_path = f.name
+
+    root = logging.getLogger()
+    parent_handler = logging.FileHandler(log_path)
+    root.addHandler(parent_handler)
+
+    old_env = os.environ.get("GARAK_LOG_FILE")
+    os.environ["GARAK_LOG_FILE"] = log_path
```
jmartin-tech, line 237:
```Python
    old_env = os.environ.get(garak.GARAK_LOG_FILE_VAR)
    os.environ[garak.GARAK_LOG_FILE_VAR] = log_path
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+        )
+    except RuntimeError as exc:
+        errors.append(str(exc))
+    finally:
+        parent_handler.close()
+        root.removeHandler(parent_handler)
+        if old_env is None:
+            os.environ.pop("GARAK_LOG_FILE", None)
+        else:
+            os.environ["GARAK_LOG_FILE"] = old_env
```
jmartin-tech, line 255:
```Python
            os.environ.pop(garak.GARAK_LOG_FILE_VAR, None)
        else:
            os.environ[garak.GARAK_LOG_FILE_VAR] = old_env
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+    with tempfile.NamedTemporaryFile(suffix=".log", delete=False) as f:
+        log_path = f.name
+
+    root = logging.getLogger()
+    parent_handler = logging.FileHandler(log_path)
+    root.addHandler(parent_handler)
+    parent_fd = parent_handler.stream.fileno()
+
+    old_env = os.environ.get("GARAK_LOG_FILE")
+    os.environ["GARAK_LOG_FILE"] = log_path
```
jmartin-tech, line 275:
```Python
    old_env = os.environ.get(garak.GARAK_LOG_FILE_VAR)
    os.environ[garak.GARAK_LOG_FILE_VAR] = log_path
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+                "Worker file descriptors must differ from the parent's fd - "
+                "shared fds cause the reentrant-flush race condition"
+            )
+    finally:
+        parent_handler.close()
+        root.removeHandler(parent_handler)
+        if old_env is None:
+            os.environ.pop("GARAK_LOG_FILE", None)
+        else:
+            os.environ["GARAK_LOG_FILE"] = old_env
```
jmartin-tech, line 294:
```suggestion
            os.environ.pop(garak.GARAK_LOG_FILE_VAR, None)
        else:
            os.environ[garak.GARAK_LOG_FILE_VAR] = old_env
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+    with tempfile.NamedTemporaryFile(suffix=".log", delete=False) as f:
+        log_path = f.name
+
+    root = logging.getLogger()
+    original_handlers = root.handlers[:]
+    for h in original_handlers:
+        root.removeHandler(h)
+
+    old_env = os.environ.get("GARAK_LOG_FILE")
+    os.environ["GARAK_LOG_FILE"] = log_path
```
jmartin-tech, line 365:
```Python
    old_env = os.environ.get(garak.GARAK_LOG_FILE_VAR)
    os.environ[garak.GARAK_LOG_FILE_VAR] = log_path
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+        )
+    finally:
+        parent_handler.close()
+        root.removeHandler(parent_handler)
+        for h in original_handlers:
+            root.addHandler(h)
+        if old_env is None:
+            os.environ.pop("GARAK_LOG_FILE", None)
+        else:
+            os.environ["GARAK_LOG_FILE"] = old_env
```
jmartin-tech, line 425:
```suggestion
            os.environ.pop(garak.GARAK_LOG_FILE_VAR, None)
        else:
            os.environ[garak.GARAK_LOG_FILE_VAR] = old_env
```

tests/probes/test_probes_base_parallel_logging.py
```diff
@@ -0,0 +1,470 @@
...
+        for h in root.handlers[:]:
+            if isinstance(h, logging.FileHandler):
+                h.close()
+            root.removeHandler(h)
+        for h in original_handlers:
+            root.addHandler(h)
+        if old_env is None:
+            os.environ.pop("GARAK_LOG_FILE", None)
+        else:
+            os.environ["GARAK_LOG_FILE"] = old_env
```
jmartin-tech, line 150:
```Python
            os.environ.pop(garak.GARAK_LOG_FILE_VAR, None)
        else:
            os.environ[garak.GARAK_LOG_FILE_VAR] = old_env
```

jmartin-tech (changes requested):
This looks really close.

Looking at the tests I suspect there may be room for a fixture for setup and teardown the env setting needed.

jmartin-tech:
Also please pay ensure the DCO sign-off requirements are met for all commit messages in the branch.

jmartin-tech:
@nuthalapativarun please amend the latest commit message with sign-off per DCO requirements. I can offer a commit to address my comment asks once that CI requirements passes.

jmartin-tech:
@nuthalapativarun please amend the commit messages to include sign-off on the latest commit per the DCO requirements. A maintainer can address the code comments, however to move this forward DCO requirements must be met by the contributor.

nuthalapativarun:
Pushed the requested refactor and rebased onto current upstream/main (was 374 commits behind).

**Refactor per your 2026-07-27 inline review comments** (on top of the `garak.setup_logger()` extraction from the earlier round):
- `_worker_logging_init` now reads the log file path via `garak.GARAK_LOG_FILE_VAR` instead of the hardcoded `"GARAK_LOG_FILE"` string, so there's a single source of truth for the env var name (`garak/probes/base.py`).
- `tests/probes/test_probes_base_parallel_logging.py`: replaced the repeated manual `os.environ` save/restore boilerplate in every test with two fixtures, `garak_log_file_env` (points the env var at a fresh temp file, cleans up after) and `garak_log_file_env_unset` (ensures the var is absent), plus a `preserve_root_handlers` fixture that saves/restores the root logger's handlers. Assertion messages and docstrings now reference `garak.GARAK_LOG_FILE_VAR` instead of the raw string, per your line comments.

**DCO**: I noticed the DCO check was failing because the `refactor(logging): centralize logger setup...` commit was missing a `Signed-off-by` trailer. Since the whole branch needed a rebase anyway (all commit hashes were being rewritten regardless), I added the sign-off to that commit as part of the rebase rather than leaving it broken — the diff of that commit is otherwise unchanged.

All 8 tests in `tests/probes/test_probes_base_parallel_logging.py` pass locally (7 pass + 1 correctly skipped on Windows, which lacks `fork`). Ran the full `tests/probes/` suite as well; 1361 pass, 2 unrelated pre-existing failures (`AudioAchillesHeel`, `GOATAttack` — both fail on dataset/HF loading issues unrelated to this change).

Branch is now rebased onto current `upstream/main` and force-pushed. Ready for another look whenever convenient.

garak/__init__.py
```diff
@@ -10,6 +10,30 @@
...
+    called once on initial import of this module, and may be called again
+    (e.g. by a :mod:`multiprocessing` pool initializer) to re-apply the same
+    configuration after handlers have been reset, such as when a worker
+    process needs to close inherited file descriptors and open its own.
+
+    :param filename: path to the log file. Defaults to the currently
+        configured ``_config.transient.log_filename``.
+    """
+    if filename is None:
+        filename = _config.transient.log_filename
```
jmartin-tech: A closer look, suggests this guard should not pull from `_config`.  In a new thread, this value may not be have been set, as transient values are populated in the core thread and _config is no longer a shared object at this time. This may cause the subsequent call here to attempt to set `filename=None` when calling `logging.basicConfig()`.

garak/__init__.py
```diff
@@ -10,6 +10,30 @@
 
 GARAK_LOG_FILE_VAR = "GARAK_LOG_FILE"
 
+LOG_FORMAT = "%(asctime)s  %(levelname)s  %(message)s"
+
+
+def setup_logger(filename=None):
```
jmartin-tech: Consider accepting the log level here, this would enable a future enhancement to allow users to define the log level at startup or during specific actions during runtime.


```Python
def setup_logger(filename=None, level=logging.DEBUG):
```

Issue #1789, opened 2026-05-20
feature: detector filters in run.spec


patriciapampanelli:
Add detector selectors to `run.spec` as filters over detectors garak would already activate.

Depends on #1744

PR #1831, opened 2026-06-02
Unified `run.spec` selection grammar

### LINKED ISSUE #1744: feature: unified run.spec input format
Proposing a standardilized `run.spec` supporting:
- additive and subtractive syntax
- probe by name or family
- intent category or sub entry
- tier or other metadata
- active status
- tags-based filtering

PR targets `feature/technique_intent`; to be landed with the T&I merge.

patriciapampanelli:
## Summary

Introduces a single selection grammar — `run.spec` — for choosing probes and buffs, replacing the fragmented `--probes` / `--probe_tags` / `--buffs` flags and their config keys. One transport-agnostic engine parses both the CLI string (`--run_spec`) and the config form (`run: { spec: { include: [...], exclude: [...] } }`) into the same internal `Spec`, then resolves it to canonical plugin paths.

Resolves #1744

### Grammar

- **Plugin paths:** `probes.<module>[.<Class>]`, `buffs.<module>[.<Class>]`, `probes.*`
- **Empty selection:** `none` (or `probes.none`) selects no probes — an explicit empty selection, distinct from an unspecified spec (which defaults to `probes.*`)
- **Filters:** `tier:<N>` (inclusive threshold — tiers `1..N`), `tag:<prefix>` (membership)
- **Exclusion:** any selector prefixed with `-` (e.g. `-probes.dan.DanInTheWild`, `-tier:3`)
- Default (no probe plugin-path selector) → all active probes (`probes.*`), injected only at resolve time.

### What changed

- **New module `garak/_spec.py`** — `Selector` / `Spec` / `Resolution` + parsing and resolution. `_resolve_plugin_paths` is the single shared resolution core. A `none` selector kind represents an explicit empty selection.
- **`garak/cli.py`** — adds `--run_spec` (underscore, consistent with garak's other flags such as `--probe_tags` / `--list_probes`); `--probes`/`--probe_tags`/`--buffs` are deprecated and map onto `run.spec` (with `--run_spec` winning on conflict).
- **`garak/command.py`** — `print_buffs()` gains an optional `selected_buffs` argument so `--list_buffs` matching `--list_probes` / `--list_detectors`. Previously the buff listing ignored any selection and always listed every buff; called without the argument, behaviour is unchanged (lists all).

- **`garak/_config.py`** — `run.spec` default; legacy config keys (`plugins.probe_spec`, `plugins.buff_spec`, `run.probe_tags`) map onto `run.spec` post-merge (mirroring `model_type` → `target_type`).
- **Report snapshot** now carries `run.spec` instead of the legacy keys.
- **Downstream consumers** (`report_digest.py`, `aggregate_reports.py`) read `run.spec`, with a fallback for reports predating this change.
- **Bundled configs** (`bag.yaml`, `fast.json`) migrated to `run.spec`.
- **Docs** updated (`configurable.rst`, `usage.rst`, `faster.rst`, new `_spec.rst`, `cliref.rst` regenerated from `--help`).

### Selection of "none" vs unspecified (standardized)

A single rule governs empty vs default selection, consistently across CLI and config:

- `none` (e.g. `--probes none` or `probe_spec: none`) → **explicit empty selection** (`probes.none`); resolves to no probes. A run with no probes is a deliberate no-op ("No probes, nothing to do").
- vacuous values (omitted, empty, or `auto`) → **unspecified**; resolves to all active probes (`probes.*`).

This restores the legacy meaning of `none` (which selected nothing) while keeping the unified grammar's implicit default. The only behavior change vs the legacy flags is that `--probes ""` / `--probes auto` now default to all (they previously selected nothing); these are degenerate inputs and the new behavior is the more intuitive one.

### Deprecation

`--probes` / `--probe_tags` / `--buffs` and `plugins.probe_spec` / `plugins.buff_spec` / `run.probe_tags` are deprecated (emit a deprecation notice). Detectors are out of scope here and keep their own `plugins.detector_spec` surface, to be migrated at #1789.

### Examples

**CLI**

~~~bash
# all active probes (default — no selector needed)
garak -t test.Blank
# a whole probe family
garak -t test.Blank --run_spec "probes.dan"
# a single probe class
garak -t test.Blank --run_spec "probes.dan.DanInTheWild"
# family minus one class
garak -t test.Blank --run_spec "probes.dan, -probes.dan.DanInTheWild"
# tier threshold (tiers 1..2) filtered by tag, with one exclusion
garak -t test.Blank --run_spec "tier:2, tag:owasp:llm01, -probes.dan.AutoDAN"
# probes + a buff
garak -t test.Blank --run_spec "probes.dan, buffs.encoding"
# explicit empty selection (no probes)
garak -t test.Blank --run_spec "probes.none"
~~~

**Config (YAML)**

~~~yaml
run:
  spec:
    include:
      - probes.dan
      - tag:owasp:llm01
      - buffs.encoding
    exclude:
      - probes.dan.DanInTheWild
~~~

### Manual — selection resolution

- [x] all active probes (no selector): `garak --list_probes`
- [x] probe family: `garak --list_probes --run_spec "probes.dan"`
- [x] explicit empty selection: `garak --list_probes --run_spec "probes.none"` → 0 probes (default still lists all)
- [x] single class: `garak --list_probes --run_spec "probes.dan.DanInTheWild"`
- [x] family minus one class: `garak --list_probes --run_spec "probes.dan, -probes.dan.DanInTheWild"`
- [x] tier threshold (1..1): `garak --list_probes --run_spec "tier:1"`
- [x] negative tier excludes exactly that tier (1..3 then drop tier 2 → {1,3}): `garak --list_probes --run_spec "tier:3, -tier:2"`
- [x] tag filter: `garak --list_probes --run_spec "tag:owasp:llm01"`
- [x] buff selection: `garak --list_buffs --run_spec "buffs.encoding"`

### Manual — deprecation, conflict, errors

- [x] `--probes none` is a deliberate no-op: `garak -t test.Blank --probes none` → "No probes, nothing to do" (does not run all probes)
- [x] `--probes` deprecation maps to run.spec: `garak --list_probes -p dan`
- [x] `--probe_tags` deprecation maps to `tag:`: `garak --list_probes --probe_tags owasp:llm01`
- [x] `--buffs` deprecation maps to `buffs.`: `garak --list_buffs -b encoding`
- [x] config `probe_spec: none` selects nothing (parity with CLI); `auto`/empty default to all
- [x] `--run_spec` wins over legacy flag: `garak --list_probes --run_spec "probes.dan" -p atkgen`
- [x] friendly error + exit 1 on invalid spec: `garak --list_probes --run_spec "detectors.foo"`
- [x] aborts with empty-result reason: `garak --list_probes --run_spec "probes.dan, -probes.dan"`


garak/configs/bag.yaml
```diff
@@ -1,19 +1,44 @@
...
+    - probes.misleading
+    - probes.packagehallucination
+    - probes.phrasing
+    - probes.promptinject
+    - probes.realtoxicityprompts
+    - probes.snowball
+    - probes.suffix
+    - probes.tap
+    - probes.topic
+    - probes.web_injection
```
jmartin-tech: I think this might not be equivalent to as some of the probes in the existing list are noted by full class which suggests they are not `active` and need to be retained in this list.

garak/_config.py
```diff
@@ -28,14 +28,17 @@
...
+    "run.probe_tags": "0.15.1.pre1",
 }
 
 from garak import __version__ as version
 
 system_params = (
     "verbose narrow_output parallel_requests parallel_attempts skip_unknown".split()
 )
-run_params = "seed deprefix eval_threshold generations probe_tags interactive system_prompt".split()
+run_params = "seed deprefix eval_threshold generations interactive system_prompt".split()
```
jmartin-tech: This should include `spec` as this list is used to validate keys in the `run` section of the config are all documented.

Based on that should this PR name the option as `--spec` vs `--run_spec`?

garak/_spec.py
```diff
@@ -0,0 +1,343 @@
...
+    selectors: List[Selector], category: str
+) -> Tuple[set, List[str]]:
+    """Single plugin-path resolution core (category generic).
+
+    Mirrors the legacy ``parse_plugin_spec`` name resolution:
+    ``<category>.*`` -> active plugins; ``<category>.<module>`` -> active
+    family; ``<category>.<module>.<Class>`` -> exact match (ignores ``active``).
+    Returns the resolved name set plus the list of unknown selector values.
+    """
+    from garak._plugins import enumerate_plugins
```
jmartin-tech: While I understand _why_ this is here I think there may be some odd coupling created by this.

garak/_spec.py
```diff
@@ -0,0 +1,343 @@
...
+        return (
+            f"probe '{name}' is tier {_tier_of(name)} but the spec restricts to "
+            f"tiers 1..{ceiling}; widen the tier filter or drop the explicit probe"
+        )
+    if any(s.kind in ("tag", "tier") for s in spec.include):
+        return "no active probe matches the given tier/tag filters; widen the filters"
+    return "every included probe was removed by an exclusion; adjust includes/excludes"
+
+
+def _resolve_spec(spec: "Spec", skip_unknown: bool) -> "Resolution":
```
jmartin-tech:
```suggestion
def _resolve_spec(spec: "Spec", skip_unknown: bool) -> Resolution:
```

Sorry not familiar with this syntax, I would expect the `Type` here not a `str` of the type name.

garak/_spec.py
```diff
@@ -0,0 +1,343 @@
...
+@dataclass
+class Resolution:
+    """Resolved selection. ``probes``/``buffs`` are canonical
+    ``category.module.Class`` names; ``rejected`` lists unknown selectors;
+    ``empty_reason`` is set when the spec resolves to no probes."""
+
+    probes: List[str]
+    buffs: List[str]
+    rejected: List[str]
+    empty_reason: Optional[str] = None
```
jmartin-tech, line 84: Is there a more generic way to organize this? Having `probes` and `buffs` attributes makes this a bit tightly coupled to the current supported plugin types.

jmartin-tech (commented):
Initial comments, testing and usage flow are still underway.

jmartin-tech:
Consider adding a `fixer` that can reformat a provided configuration into the new format.

jmartin-tech:
While thinking on testing methods for this, I find myself thinking it would be nice to have a clean syntax to select all active probes plus a set of specific inactive ones. Something like:

``` yaml
run:
  spec:
    include:
      - probes.all
      - probes.fitd.FITD
```

Would it makes sense to have a reserved `probes.all` indicator for probes?

patriciapampanelli:
> Consider adding a `fixer` that can reformat a provided configuration into the new format.

The legacy keys are already converted to `spec` internally at load, and the `report.jsonl` persists the new `spec` grammar. Is your `fixer` suggestion specifically about surfacing this through `--fix` so users can convert their config to the new format? @jmartin-tech

jmartin-tech:
Yes a `fixer` is meant to offer a method of converting an old configuration to the new format, this provides a direct way for users who are getting a deprecation notice to convert a config to the new format using the `--fix` cli option.

garak/configs/bag.yaml
```diff
@@ -1,19 +1,19 @@
...
...ying, probes.lmrc.Deadnaming, probes.lmrc.QuackMedicine, probes.lmrc.SexualContent, probes.lmrc.Sexualisation, probes.lmrc.SlurUsage, probes.malwaregen.Evasion, probes.malwaregen.Payload, probes.malwaregen.SubFunctions, probes.malwaregen.TopLevel, probes.misleading.FalseAssertion, probes.packagehallucination.Dart, probes.packagehallucination.JavaScript, probes.packagehallucination.Perl, probes.packagehallucination.Python, probes.packagehallucination.RakuLand, probes.packagehallucination.Ruby, probes.packagehallucination.Rust, probes.phrasing.FutureTense, probes.phrasing.PastTense, probes.promptinject.HijackHateHumans, probes.promptinject.HijackKillHumans, probes.promptinject.HijackLongPrompt, probes.realtoxicityprompts.RTPBlank, probes.snowball.GraphConnectivity, probes.suffix.GCGCached, probes.tap.TAPCached, probes.topic.WordnetControversial, probes.web_injection.ColabAIDataLeakage, probes.web_injection.MarkdownImageExfil, probes.web_injection.MarkdownURIImageExfilExtended, probes.web_injection.MarkdownURINonImageExfilExtended, probes.web_injection.MarkdownXSS, probes.web_injection.PlaygroundMarkdownExfil, probes.web_injection.StringAssemblyDataExfil, probes.web_injection.TaskXSS]
```
jmartin-tech: Shouldn't this be a list of `str` now:
```
include: 
  - probes.ansiescape.AnsiEscaped
  - probes.ansiescape.AnsiRaw
  - probes.apikey.CompleteKey
  - probes.apikey.GetKey
  - probes.atkgen.Tox
  - probes.continuation.ContinueSlursReclaimedSlurs
  - probes.dan.Ablation_Dan_11_0, probes.dan.AutoDANCached
  - probes.dan.DanInTheWild
...
```

patriciapampanelli: Done

garak/cli.py
```diff
@@ -156,19 +156,28 @@ def main(arguments=None) -> None:
         default=None,
         help="name of the target, e.g. 'timdettmers/guanaco-33b-merged'",
     )
-    # probes
+    # unified selection spec (replaces --probes/--probe_tags/--buffs)
+    parser.add_argument(
+        "--spec",
```
jmartin-tech, line 161: Should we add a `-S` since `-s` is used for seed?

garak/resources/fixer/run_spec.py
jmartin-tech, line 1: fixers should always be named with with a data prefix in the format YYYYMMDD:
```
garak/resources/fixer/20260616_run_spec.py
```

This ensures they are applied in chronological order. Might be worth adding that as a test requirement.

garak/resources/fixer/run_spec.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+        if spec is None:
+            return cfg
+        plugins.pop("probe_spec", None)
+        plugins.pop("buff_spec", None)
+        run.pop("probe_tags", None)
+        # an explicitly-set run.spec wins; otherwise adopt the converted spec
+        if not run.get("spec"):
+            run["spec"] = spec
+        cfg["run"] = run
```
jmartin-tech: Testing shows:
```
plugins:
  probe_spec: goat.GOATAttack
```
becomes:
```
run:
  spec:
    exclude: []
    include:
    - probes.goat.GOATAttack
```

Should the serialization suppress the empty _exclude_ list? Proposed result:
```
run:
  spec:
    include:
    - probes.goat.GOATAttack
```

Another edge case:
```
run:
  generations: 1

plugins:
  probe_spec: goat.GOATAttack
  buff_spec: buffs.encoding.CharCode
```
Becomes:
```
plugins: {}
run:
  generations: 1
  spec:
    exclude: []
    include:
    - probes.goat.GOATAttack
    - buffs.encoding.CharCode
```

In this case the top level `plugins` ends up empty and should probably be suppressed as well.

During testing I also noted that invalid values will be transformed as well, my thought here is that a invalid config as input should report the config is invalid and cannot be migrated. For example if `buff_spec = buffs.encoding.CharCode` the converter will add an include for `buffs.buffs.encoding.CharCode`.

garak/cli.py
```diff
@@ -627,60 +671,50 @@ def worker_count_validation(workers):
...
-                            msg_parts.append(
-                                f"❌Unknown {spec_namespace}❌: " + ",".join(truly_unknown)
-                            )
-                        if inactive_only_modules:
-                            module_list = ",".join(inactive_only_modules)
-                            msg_parts.append(
-                                f"all {spec_namespace} in '{module_list}' are marked "
-                                f"inactive; select one or more by name to continue"
-                            )
-                        raise ValueError("; ".join(msg_parts))
```
jmartin-tech, line 683: Recently added detection for probe families that are fully inactive is being lost here. The revisions need to account for flagging `inactive_only_modules` in a provided spec in both the new and old formats.

jmartin-tech (changes requested):
Some edge cases found in testing and a functionality that needs to be retained.

garak/cli.py
patriciapampanelli: Agreed!

garak/resources/fixer/run_spec.py
patriciapampanelli: Done

garak/resources/fixer/20260612_run_spec.py
```diff
@@ -0,0 +1,46 @@
...
+"""Migrate the deprecated selection keys onto the unified ``run.spec`` grammar."""
+
+import copy
+
+from garak.resources.fixer import Migration
+from garak._spec import legacy_selection_spec
+
+
+class MapLegacySelectionToSpec(Migration):
+    def apply(config_dict: dict) -> dict:
```
jmartin-tech, line 33: With the latest revisions a plugin prefix based now results in:
```
python -m garak --fix --config ~/nvdev/testing_unified_spec/config_invalid.yaml
garak LLM vulnerability scanner v0.15.2.pre1 ( https://github.com/NVIDIA/garak ) at 2026-06-23T11:33:15.953864
✋ DEPRECATION: config plugins.probe_spec is deprecated since version 0.15.1.pre1
✋ DEPRECATION: config plugins.buff_spec is deprecated since version 0.15.1.pre1
❌ invalid legacy selection in config: legacy buffs selection 'buffs.encoding.CharCode' already carries a category prefix; legacy keys take unprefixed values (e.g. 'encoding.CharCode'), not unified run.spec tokens
```

However another edge case that doesn't match expectations exists when the prefix is completely invalid:
config_invalid_2.yaml
``` yaml
run:
  generations: 1

plugins:
  probe_spec: goat.GOATAttack
  buff_spec: s.encoding.CharCode
```

Produces:
```
python -m garak --fix --config ~/nvdev/testing_unified_spec/config_invalid_2.yaml
garak LLM vulnerability scanner v0.15.2.pre1 ( https://github.com/NVIDIA/garak ) at 2026-06-23T11:34:03.182440
✋ DEPRECATION: config plugins.probe_spec is deprecated since version 0.15.1.pre1
✋ DEPRECATION: config plugins.buff_spec is deprecated since version 0.15.1.pre1
Updated /home/jemartin/nvdev/testing_unified_spec/config_invalid_2.yaml:
run:
  generations: 1
  spec:
    include:
    - probes.goat.GOATAttack
    - buffs.s.encoding.CharCode
```

Which then passed for usage as:
migrated_invalid_2.yaml
``` yaml
run:
  generations: 1
  spec:
    include:
    - probes.goat.GOATAttack
    - buffs.s.encoding.CharCode
```

Becomes:
```
$ python -m garak --config ~/nvdev/testing_unified_spec/migrated_invalid_2.yaml -t test
garak LLM vulnerability scanner v0.15.2.pre1 ( https://github.com/NVIDIA/garak ) at 2026-06-23T11:35:45.621022
📜 logging to /home/jemartin/.local/share/garak/garak.log
❌Unknown run.spec❌: buffs.s.encoding.CharCode
```


Should we fully validate the refactored run.spec value that an input file would convert to, or do you think it is reasonable to expect the user to try to use the generated config?

docs/source/cliref.rst
```diff
@@ -60,20 +60,27 @@ CLI reference for garak
     --target_name TARGET_NAME, --model_name TARGET_NAME, -n TARGET_NAME
                           name of the target, e.g.
                           'timdettmers/guanaco-33b-merged'
+    --spec SPEC           unified selection spec, e.g. 'probes.dan,
```
jmartin-tech: needs refresh to include `-S SPEC` reference

patriciapampanelli: Done

garak/resources/fixer/20260612_run_spec.py
patriciapampanelli: We should validate it before moving it to the run.spec config.

jmartin-tech (approved):

PR #1836, opened 2026-06-04
fix: keep Replicate private endpoint output as one Message

fallintoplace:
private Replicate endpoints already join `prediction.output` into one string, but the private endpoint path then iterates that string when building the return value.

so a response like `"hello"` becomes a list of per-character messages instead of a single `Message("hello")`.

this PR:
* returns one `Message` for the joined response
* adds a regression test for chunked `prediction.output`

## verification

* ran `PYTHONPATH=/Users/hoangvu/Code/OSS/garak /tmp/garak-pr-venv/bin/python -m pytest tests/generators/test_replicate.py`
* test passes with this change
* the same test fails against the old behavior because `_call_model()` returns one message per character

I didn't run a live private Replicate endpoint check here, and I didn't run the full `tests/` suite in this local environment.

### DIFF
--- a/garak/generators/replicate.py
+++ b/garak/generators/replicate.py
@@ -119,7 +119,7 @@ class InferenceEndpoint(ReplicateGenerator):
             raise IOError(
                 "Replicate endpoint didn't generate an Iterable[str]-type response. Make sure the endpoint is active."
             ) from exc
-        return [Message(r) for r in response]
+        return [Message(response)]
 
 
 DEFAULT_CLASS = "ReplicateGenerator"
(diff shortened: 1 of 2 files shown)


aishwaryap (changes requested):
This PR appears to fix a legitimate issue but does not yet meet the bar for acceptance. Unit tests alone are not sufficient for this fix. We require evidence that this fix has been tested against an actual Replicate endpoint. If you are not comfortable placing such evidence in a public PR, please email it to garak@nvidia.com with the subject "Evidence for PR #1836".

PR #1837, opened 2026-06-04
fix: preserve all turns in NVMultimodal prompt prep

fallintoplace:
when `embed_data=True`, `NVMultimodal._prepare_prompt()` walks every turn in the conversation but only appends after the loop.

that means a multi-turn conversation gets collapsed to the final turn during prompt preparation.

this PR:
* appends each prepared turn inside the loop
* adds a regression test for a multi-turn conversation with embedded image data

## verification

* ran `PYTHONPATH=/Users/hoangvu/Code/OSS/garak /tmp/garak-pr-venv/bin/python -m pytest tests/generators/test_nim_multimodal.py`
* test passes with this change
* if the append is moved back outside the loop, the new test fails because only the final turn is preserved

I didn't run a live NIM endpoint check here, and I didn't run the full `tests/` suite in this local environment.

### DIFF
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -218,7 +218,7 @@ class NVMultimodal(NVOpenAIChat):
                 )
             prepared_msg.text = text
 
-        prepared_conv.turns.append(Turn(turn.role, prepared_msg))
+            prepared_conv.turns.append(Turn(turn.role, prepared_msg))
 
         return prepared_conv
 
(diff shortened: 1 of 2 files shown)


aishwaryap (approved):
This looks like a reasonable fix to a hard to detect bug. Verified that the provided unit test fails without the fix and passes with.

aishwaryap:
Just out of curiosity we would also love to know what situation helped you discover this bug.

Issue #1562, opened 2026-01-16
probe: probes should follow `soft_probe_prompt_cap`


leondz:
`suffix.GCGCached` always runs all stored cached prompts

`run.soft_probe_prompt_cap` sets a cap on how many distinct prompts to offer per probe

`suffix.GCGCached` should not issue more prompts than `run.soft_probe_prompt_cap`

To see the issue, try a config like:

```
{
	"plugins": {
		"probe_spec": "suffix"
	},
	"run": {
		"soft_probe_prompt_cap": 10,
		"generations": 1,
	}
}
```

Note that >10 prompts are issued

When addressing this, use a mechanism for randomly selecting prompts like that in `probes.base._prune_data()`

jmartin-tech:
There are a number for probes that still ignore `soft_probe_prompt_cap` for various reasons. In most cases either they have a small number for default prompts. I am all for standardizing all probes on limited caps. Should we expand this issue with a comprehensive list of current probes to be limited?

jmartin-tech:
This seems connected to #1546

leondz:
> Should we expand this issue with a comprehensive list of current probes to be limited?

to limit, from default active set:

* [ ] `ansiescape.AnsiEscaped`
* [ ] `ansiescape.AnsiRaw`
* [ ] `atkgen.Tox` (?)
* [ ] `dan.Ablation_Dan_11_0`
* [ ] `dan.AutoDANCached`
* [ ] `divergence.Repeat`
* [ ] `exploitation.JinjaTemplatePythonInjection`
* [ ] `exploitation.SQLInjectionEcho`
* [ ] `goodside.Tag`
* [ ] `goodside.WhoIsRiley`
* [ ] `grandma.Slurs`
* [ ] `grandma.Substances`
* [ ] `grandma.Win10`
* [ ] `grandma.Win11`
* [ ] `latentinjection.LatentWhois`
* [ ] `lmrc.Bullying`
* [ ] `lmrc.Deadnaming`
* [ ] `lmrc.Sexualisation`
* [ ] `malwaregen.Evasion`
* [ ] `malwaregen.Payload`
* [ ] `malwaregen.SubFunctions`
* [ ] `malwaregen.TopLevel`
* [ ] `misleading.FalseAssertion`
* [ ] `realtoxicityprompts.RTPBlank`
* [ ] `snowball.GraphConnectivity`
* [ ] `suffix.GCGCached`
* [ ] `tap.TAPCached`
[...]

PR #1838, opened 2026-06-04
probes.base: balance IntentProbe pruning across intents

### LINKED ISSUE #1835: IntentProbe should balance pruning across intents
`IntentProbe` generates prompts across multiple intents but doesn't apply `soft_probe_prompt_cap` with intent-aware balancing.

**Current behavior**
- `IntentProbe.build_prompts()` generates all intent × stub combinations
- No pruning happens, regardless of `soft_probe_prompt_cap`

**Related**
- [#1562](https://github.com/NVIDIA/garak/issues/1562) — probes should respect `soft_probe_prompt_cap`

patriciapampanelli:
## Summary
`IntentProbe` ignored `soft_probe_prompt_cap`. This adds a balanced `IntentProbe._prune_data` override that honors the cap and spreads prompts evenly (within 1) across intents: `cap // n` per intent plus the remainder to the first intents, sampled per intent, deleting in reverse to keep `prompts`/`prompt_intents` aligned.

Resolves #1835

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -848,6 +848,10 @@ class IntentProbe(Probe):
 
     import garak.services.intentservice
 
+    DEFAULT_PARAMS = Probe.DEFAULT_PARAMS | {
+        "follow_prompt_cap": True,
+    }
+
     intent = None  # IntentProbe subclasses span many typology entries by design, so there is no single best fit.
     skip_root_intents = True
     blocked_intent_spec = ""
@@ -858,7 +862,8 @@ class IntentProbe(Probe):
         self._populate_intents()
         self._populate_stubs()
         self.build_prompts()
-        self._prune_data(self.soft_probe_prompt_cap)
+        if self.follow_prompt_cap:
+            self._prune_data(self.soft_probe_prompt_cap)
 
     def _attempt_prestore_hook(
         self, attempt: garak.attempt.Attempt, seq: int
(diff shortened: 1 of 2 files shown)


garak/probes/base.py
```diff
@@ -858,13 +858,46 @@ def __init__(self, config_root=_config):
         self._populate_intents()
         self._populate_stubs()
         self.build_prompts()
+        self._prune_data(self.soft_probe_prompt_cap)
```
jmartin-tech: Pruning needs to possible to suppress per probe, `follow_prompt_cap` is already inherited in DEFAULT_PARAMS:
```
        if self.follow_prompt_cap:
```

Either here or in `_prune_data()`.

patriciapampanelli: `follow_prompt_cap` isn't declared on `Probe` itself. Today each probe declares it individually, either as a class attribute or in `DEFAULT_PARAMS` (e.g. `DRA`, `IterativeProbe`, `leakreplay`). Should we promote it to `Probe` instead?

jmartin-tech: You are right, I had mistaken the one in `IterativeProbe` to be on `Probe`, for now we should add it to `IntentProbe` and later promote it into Probe when all other probes support the cap per #1562

This is simply to avoid expanding scope of this PR too much.

jmartin-tech (approved):

PR #1840, opened 2026-06-05
fix: skip topP for Anthropic models when temperature is set on Bedrock

### LINKED ISSUE #1812: BedrockGenerator sends both temperature and topP, which Claude 4.x on Bedrock rejects
Running Garak against Anthropic Claude 4.x on AWS Bedrock fails because BedrockGenerator puts both temperature and topP in the Converse inferenceConfig. Claude 4.x rejects this:

The model returned the following errors: `temperature` and `top_p` cannot both be specified for this model. Please use only one.

**Steps to reproduce**

  1. Install garak 0.15.0 from PyPI.
  2. Run a promptinject probe against an Anthropic Claude 4.x model on AWS Bedrock:
  `garak --target_type bedrock \
        --target_name us.anthropic.claude-sonnet-4-6 \
        --probes promptinject.HijackHateHumans --generations 1`
  3. Every request to the model fails.

  **Were you following a specific guide/tutorial or reading documentation?**

https://github.com/NVIDIA/garak#aws-bedrock

  **Expected behavior**

  Garak should scan Claude 4.x models on Bedrock out of the box, without sending a parameter combination the API forbids.

  **Current behavior**

  Every request fails with a Bedrock validation error:

  The model returned the following errors: `temperature` and `top_p` cannot both be specified for this model. Please use only one.

  **Root cause**
 _call_model in garak/generators/bedrock.py adds each inference field with an independent if, and DEFAULT_PARAMS sets both (temperature=0.7, top_p=1.0), so both temperature and topP are always present in inferenceConfig:
```
  if self.top_p is not None:
```
  Captured inferenceConfig reaching client.converse(...): ['maxTokens', 'temperature', 'topP']. Claude 4.x on Bedrock rejects this combination.

  The config-level workaround (top_p: null) does not help for promptinject probes: their _generator_precall_hook in garak/probes/promptinject.py re-sets generator.top_p from the PromptInject dataset (config_top_p=1) before every call, after
  configuration has been applied:
```
      map_attribs = "temperature top_p frequency_penalty presence_penalty max_tokens".split()
              setattr(generator, map_attrib,
                      attempt.notes["settings"]["config_" + map_attrib])
```
  So top_p is forced back to 1.0 regardless of config.

  **Suggested fix.** 
In _call_model, don't send both — prefer temperature (the knob the probes set meaningfully):
```
  if self.top_p is not None and self.temperature is None:
```
  This sits downstream of the promptinject hook, so it fixes all probes. It may warrant gating on model family, since the restriction is specific to Claude 4.x. Separately, it's worth reconsidering whether promptinject's _generator_precall_hook
  should override generator params unconditionally, since it silently defeats user/site config.

  **Garak version**

  0.15.0

  Additional Information

  1. Operating system: Linux (python:3.11-slim container); also reproduces on macOS.
  2. Python version: 3.11.
  3. Install method: pypi (garak==0.15.0).
  4. Logs: the failing request is logged at DEBUG as:
  Response body: b'{"message":"The model returned the following errors: `temperature` and `top_p` cannot both be specified for this model. Please use only one."}'
  4. No hitlog.jsonl is produced because no generation succeeds.
  5. Execution config: command-line flags shown under Steps to reproduce; also attempted a garak.site.yaml setting plugins.generators.bedrock.BedrockGenerator.top_p: null (ineffective for promptinject, as explained above).
  6. Hardware/resources: not relevant — failure is an API request-validation error, independent of hardware.

xlyoung:
## Problem

Anthropic Claude 4.x models on AWS Bedrock reject requests that set both `temperature` and `top_p` in `inferenceConfig`:

> `temperature` and `top_p` cannot both be specified for this model. Please use only one.

Since `DEFAULT_PARAMS` sets both (`temperature=0.7`, `top_p=1.0`), every request to Claude 4.x on Bedrock fails out of the box.

Fixes #1812

## Fix

For Anthropic models, skip `topP` when `temperature` is already present in `inferenceConfig`. Other providers continue to receive both parameters as before (they accept both without error).

The check uses `'anthropic' not in self.name.lower()` to detect Anthropic models, which covers all standard Bedrock model IDs like `us.anthropic.claude-sonnet-4-6`, `anthropic.claude-v2`, etc.

### DIFF
--- a/garak/generators/bedrock.py
+++ b/garak/generators/bedrock.py
@@ -185,7 +185,10 @@ class BedrockGenerator(Generator):
         if hasattr(self, "max_tokens") and self.max_tokens is not None:
             inference_config["maxTokens"] = int(self.max_tokens)
         if self.top_p is not None:
-            inference_config["topP"] = float(self.top_p)
+            # Anthropic Claude models on Bedrock reject requests that set both
+            # temperature and top_p. Skip top_p when temperature is present.
+            if "temperature" not in inference_config or "anthropic" not in self.name.lower():
+                inference_config["topP"] = float(self.top_p)
         if self.stop:
             inference_config["stopSequences"] = self.stop
 


jmartin-tech:
Declining in favor of #1842 which approaches the issue in a pattern that allows more flexibility.

PR #1842, opened 2026-06-05
feat: add suppressed_params support to BedrockGenerator

### LINKED ISSUE #1812: BedrockGenerator sends both temperature and topP, which Claude 4.x on Bedrock rejects

u7k4rs6:
Refs #1812

## Why added

Issue #1812 surfaced that BedrockGenerator cannot send a Bedrock-compatible inferenceConfig for target models with parameter restrictions: Anthropic Claude 4.x on Bedrock rejects requests that set both temperature and topP. In comments on #1812, @jmartin-tech proposed a model-agnostic mechanism: support suppressed_params on BedrockGenerator the same way OpenAICompatible generators do, so operators can globally suppress any parameter known to be unsupported by their target.

This PR implements that suggestion.

### Relationship to PR #1729 (merged)

PR #1729 added configurability at the probe layer (`garak/probes/promptinject.py`) for which generation_params promptinject's `_generator_precall_hook` overrides. This PR operates at the generator layer (`garak/generators/bedrock.py`), providing suppression for any Bedrock target regardless of which probe is running. The two mechanisms compose: #1729 addresses the promptinject-specific override behavior, this PR addresses the broader case where a Bedrock model rejects a parameter combination from any source.

### Relationship to PR #1840 (open)

PR #1840 patches the immediate Anthropic case with a model-name-gated conditional in `_call_model`. This PR is the generic mechanism @jmartin-tech suggested in the #1812 thread. Once it lands, the same outcome is achievable via site config (`suppressed_params: [top_p]`) or a per-model default suppression set, without model-name checks in `_call_model`. The two approaches aren't mutually exclusive, but the generic mechanism subsumes the special case.

## How utilized

In `garak.site.yaml`:

```yaml
plugins:
  generators:
    bedrock:
      BedrockGenerator:
        suppressed_params:
          - top_p
```

Then run garak normally:

```
garak --target_type bedrock \
      --target_name us.anthropic.claude-sonnet-4-6 \
      --probes promptinject.HijackHateHumans \
      --generations 1
```

The `topP` field is absent from every `inferenceConfig` sent to `bedrock-runtime.converse`, even when downstream hooks (e.g., `promptinject._generator_precall_hook`) mutate `generator.top_p` mid-run.

## Tests

- `test_inference_config_unchanged_with_empty_suppressed_params`: refactored code is output-identical to the original for all four fields.
- `test_suppressed_params_blocks_top_p`: `topP` absent from inferenceConfig when `suppressed_params={"top_p"}` and `top_p` is non-None.
- `test_suppression_survives_attribute_mutation`: suppression holds after `generator.top_p` is mutated mid-run, simulating `promptinject._generator_precall_hook`.
- `test_warn_on_unknown_suppressed_key`: WARNING logged at init when suppressed_params contains a key not in `_PARAM_MAP`.

All 15 tests in `tests/generators/test_bedrock.py` pass.

### DIFF
--- a/garak/generators/bedrock.py
+++ b/garak/generators/bedrock.py
@@ -37,7 +37,26 @@ MODEL_ALIASES = {
 
 
 class BedrockGenerator(Generator):
-    """Interface for AWS Bedrock foundation models using Converse API"""
+    """Interface for AWS Bedrock foundation models using Converse API.
+
+    suppressed_params (set[str], default empty): Bedrock Converse API inference
+    field names to omit from the request inferenceConfig, regardless of whether
+    the corresponding instance attribute is set. Useful for target models that
+    reject specific combinations of inference parameters (for example, Anthropic
+    Claude 4.x on Bedrock rejects sending both temperature and topP). Suppression
+    is applied at request-assembly time, so it overrides per-probe parameter
+    mutation (such as promptinject's _generator_precall_hook). Field names use
+    the Bedrock Converse API spelling (topP, not top_p; maxTokens, not max_tokens).
+
+    Example garak.site.yaml config to suppress topP::
+
+        plugins:
+          generators:
+            bedrock:
+              BedrockGenerator:
+                suppressed_params:
+                  - topP
+    """
 
     active = True
     generator_family_name = "Bedrock"
@@ -49,6 +68,7 @@ class BedrockGenerator(Generator):
         "top_p": 1.0,
         "stop": [],
         "region": "us-east-1",
+        "suppressed_params": set(),
     }
 
     _unsafe_attributes = ["client"]
@@ -104,6 +124,18 @@ class BedrockGenerator(Generator):
                 )
 
         super().__init__(self.name, config_root=config_root)
+        self.suppressed_params = set(self.suppressed_params)
+        _pythonic_names = {
+            "top_p": "topP",
+            "max_tokens": "maxTokens",
+            "stop": "stopSequences",
+        }
+        for param in self.suppressed_params:
+            if param in _pythonic_names:
+                logging.warning(
+                    f"suppressed_params entry '{param}' looks like a Python attribute name; "
+                    f"use the Bedrock API field name '{_pythonic_names[param]}' instead."
+                )
         self._validate_env_var()
         self._load_unsafe()
 
@@ -180,13 +212,19 @@ class BedrockGenerator(Generator):
             return [None]
 
         inference_config = {}
-        if self.temperature is not None:
-            inference_config["temperature"] = float(self.temperature)
-        if hasattr(self, "max_tokens") and self.max_tokens is not None:
-            inference_config["maxTokens"] = int(self.max_tokens)
-        if self.top_p is not None:
-            inference_config["topP"] = float(self.top_p)
-        if self.stop:
+        _field_map = {
+            "temperature": ("temperature", float),
+            "maxTokens": ("max_tokens", int),
+            "topP": ("top_p", float),
+        }
+        for api_field, (attr, coerce) in _field_map.items():
+            if api_field in self.suppressed_params:
+                continue
+            value = getattr(self, attr, None)
+            if value is None:
+                continue
+            inference_config[api_field] = coerce(value)
+        if self.stop and "stopSequences" not in self.suppressed_params:
             inference_config["stopSequences"] = self.stop
 
         call_args = {
@@ -217,10 +255,14 @@ class BedrockGenerator(Generator):
                 return [None]
 
             content_blocks_with_text = [
-                content_block for content_block in message["content"] if "text" in content_block
+                content_block
+                for content_block in message["content"]
+                if "text" in content_block
             ]
             if len(content_blocks_with_text) == 0:
-                logging.error("Malformed response from Bedrock: missing 'text' in content blocks")
+                logging.error(
+                    "Malformed response from Bedrock: missing 'text' in content blocks"
+                )
                 return [None]
 
             text = content_blocks_with_text[0]["text"]
--- a/garak/generators/bedrock.py
+++ b/garak/generators/bedrock.py
@@ -39,23 +39,23 @@ MODEL_ALIASES = {
 class BedrockGenerator(Generator):
     """Interface for AWS Bedrock foundation models using Converse API.
 
-    suppressed_params (set[str], default empty): Bedrock Converse API inference
-    field names to omit from the request inferenceConfig, regardless of whether
-    the corresponding instance attribute is set. Useful for target models that
-    reject specific combinations of inference parameters (for example, Anthropic
-    Claude 4.x on Bedrock rejects sending both temperature and topP). Suppression
+    suppressed_params (set[str], default empty): garak attribute names to omit
+    from the Bedrock Converse API inferenceConfig, regardless of whether the
+    corresponding attribute is set. Useful for target models that reject specific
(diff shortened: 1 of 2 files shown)


u7k4rs6:
@jmartin-tech following up on #1842. This implements the generic `suppressed_params` mechanism you outlined on #1812, and all 15 tests pass. Would appreciate a look when you get a chance. Happy to iterate on anything.

garak/generators/bedrock.py
```diff
@@ -104,6 +124,18 @@ def __init__(self, name="", config_root=_config):
...
+            "stop": "stopSequences",
+        }
+        for param in self.suppressed_params:
+            if param in _pythonic_names:
+                logging.warning(
+                    f"suppressed_params entry '{param}' looks like a Python attribute name; "
+                    f"use the Bedrock API field name '{_pythonic_names[param]}' instead."
+                )
         self._validate_env_var()
         self._load_unsafe()
```
jmartin-tech: I somewhat understand the reasoning behind using params that match the error message received however I think may diverge from other `suppressed_params` implementations creating a possible maintenance burden that I am not sold on as yet.

In other generators we have relied on introspection of the signature for the called function `self.client.converse()` in this case, the idea being that the `suppressed_param` entry is the key for argument the target client method accepts. I think based on that, what is actually checked for here is correct however I suspect there should be a class level `map` of known `garak` generator standard attributes to their `Bedrock.client.converse()` parameter names. I believe this map could then be reused when creating the `kwarg` to dict that is generated in `_call_model()`.

If shifting to this validation of `suppressed_params`, it would need to happen after the call to `self._load_unsafe()` to ensure that `self.client` has been set.

u7k4rs6:
@jmartin-tech Addressed in the followup commit:

1. Added `_PARAM_MAP` as a class-level constant translating garak attribute names to Bedrock `inferenceConfig` field names. No precedent for a named constant exists in other generators (OpenAICompatible uses `inspect.signature` dynamically, LiteLLM inlines the dict), so `_PARAM_MAP` is new. The map is now consumed in both `_call_model` for construction and `__init__` for validation.

2. `suppressed_params` entries now use garak attribute names (`top_p`, `max_tokens`, etc). Unknown entries log a WARNING at init listing the valid keys; they are otherwise ignored at suppression-check time.

3. Validation moved into `__init__` after `self._load_unsafe()`.

One note on introspection: `client.converse()` accepts `inferenceConfig` as a nested kwarg rather than flat params, so signature introspection of `converse()` does not expose the individual inference fields. `_PARAM_MAP` serves as the validation surface instead. Happy to revisit if you had a different introspection mechanism in mind.

tests/generators/test_bedrock.py
```diff
@@ -311,3 +320,88 @@ def test_bedrock_access_denied_error_does_not_retry(mock_boto3):
...
+        }
+        with caplog.at_level(logging.WARNING):
+            BedrockGenerator(name="claude-4-5-sonnet")
+        matching = [r for r in caplog.records if "foo" in r.message]
+        assert matching, "expected a WARNING mentioning 'foo'"
+        assert any(
+            any(k in r.message for k in BedrockGenerator._PARAM_MAP) for r in matching
+        ), "expected WARNING to list valid keys"
+    finally:
+        BedrockGenerator.DEFAULT_PARAMS = original_defaults
```
jmartin-tech: This test does not need to manipulate the class level values, instead create a config_root to pass to the constructor that can then cause the expected warning.

``` Suggestion
    test_canary = "foo"
    test_suppressed_params = BedrockGenerator.DEFAULT_PARAMS["suppressed_params"].copy()
    test_suppressed_params.add(test_canary)
    config_root = {
        "generators": {
            "bedrock": {
                "BedrockGenerator": {"suppressed_params": test_suppressed_params}
            }
        }
    }
    with caplog.at_level(logging.WARNING):
        BedrockGenerator(name="claude-4-5-sonnet", config_root=config_root)
    matching = [r for r in caplog.records if test_canary in r.message]
    assert matching, f"expected a WARNING mentioning '{test_canary}'"
    assert any(
        any(k in r.message for k in BedrockGenerator._PARAM_MAP) for r in matching
    ), "expected WARNING to list valid keys"
```

jmartin-tech (approved):
This looks ready, I noted a minor test refactor that aligns more with expected usage.

u7k4rs6:
Done, thank you for applying it directly.

u7k4rs6:
I have read the Contributor Agreement and agree to its terms.

PR #1845, opened 2026-06-06
feat: log skip reason when detectors are skipped during scan

### LINKED ISSUE #1061: Include information about why detectors are skipped

Oxygen56:
## What
Fixes #1061 — Detectors that skip probes now log their skip reason.

## Problem
When `skip=True`, the harness silently continued. Users had no idea which detectors were skipped or why.

## Fix
1. **Detector base**: Added `skip_reason` attribute
2. **Harness**: Logs at INFO level when a detector is skipped, including probe name and reason

## Example
```
INFO: skipping detector always.Skip on probe test.Blank: disabled via config
```

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -42,6 +42,10 @@ class Detector(Configurable):
         "skip": False,
     }
 
+    skip_reason: str = None
+    """Human-readable reason why this detector was skipped. Set when ``skip`` is
+    True to help users understand why a detector did not run on a probe."""
+
     _run_params = {"seed"}
 
     def _set_description(self):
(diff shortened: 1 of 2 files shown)


jmartin-tech (changes requested):
This PR does not name sense, adding an attribute and never setting it does not provide a `reason` why the detector was skipped.

Using coding assistants to aid in contribution to this project is considered acceptable, this PR however includes changes that suggest the author has not reviewed the code before offering the revision.

PR #1847, opened 2026-06-06
fix calibration pathlib paths

ron1po:
## Summary

Thanks for maintaining garak. This small fix lets calibration accept `pathlib.Path` inputs as intended.

The constructor already advertises that `calibration_path` can be a string or `Path`, but the type guard accidentally rejected `Path` objects. This updates the check so both accepted types work, and adds regression coverage for existing and missing calibration paths passed as `Path`.

## Testing

- `git diff --check`

### DIFF
--- a/garak/analyze/calibration.py
+++ b/garak/analyze/calibration.py
@@ -18,7 +18,7 @@ class Calibration:
     """Helper for managing probe/detector score calibration data processing"""
 
     def _load_calibration(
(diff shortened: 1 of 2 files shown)


jmartin-tech (approved):
Nice, happy to incorporate this.

Looking at the signatures this is a viable improvement, it would be interesting to know _why_ or _how_ this came up to understand how it impacted your use case.

PR #1848, opened 2026-06-06
fix: repair GoogleTranslator service-account fallback

fallintoplace:
the June 26 GoogleTranslator auth change introduced two problems in `_load_langprovider()`:

* `auth_kwargs` is only defined when `project_id` is set, but it is always expanded into `from_service_account_json(...)`
* the `MalformedError` path logs that it is falling back to general auth, but never actually creates a general-auth client

this PR:
* initializes `auth_kwargs` for the no-`project_id` case
* actually falls back to `translate.Client()` when service-account parsing fails
* adds isolated regression tests for both behaviors

## verification

* ran `PYTHONPATH=/Users/hoangvu/Code/OSS/garak /tmp/garak-google-pr-venv/bin/python -m pytest tests/langproviders/test_google_translator.py`
* both tests pass with this change
* the first test fails against the old code with the unbound `auth_kwargs` path
* the second test fails against the old code because the fallback branch leaves `self.client` unset

I didn't run live Google Cloud translation checks here, and I didn't run the full `tests/` suite in this local environment.

### DIFF
--- a/garak/langproviders/remote.py
+++ b/garak/langproviders/remote.py
@@ -200,16 +200,15 @@ class GoogleTranslator(LangProvider):
         if hasattr(self, "api_key"):
             try:
                 auth_args = [self.api_key]
+                auth_kwargs = {}
                 if self.project_id is not None:
                     auth_kwargs = {"project": self.project_id}
                 self.client = translate.Client.from_service_account_json(
                     *auth_args, **auth_kwargs
                 )
             except exceptions.MalformedError as e:
-                logging.warning(
-                    "Service account auth failed, attempting fallback to general auth!"
-                )
-                # exception handling is intentionally not implemented to raise on invalid config when authentication fails
+                logging.warning("Service account auth failed: %s", e)
+                raise
         else:
             self.client = translate.Client()
         self.ftfy = ftfy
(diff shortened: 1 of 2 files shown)


garak/langproviders/remote.py
```diff
@@ -209,7 +210,7 @@ def _load_langprovider(self):
                 logging.warning(
                     "Service account auth failed, attempting fallback to general auth!"
                 )
-                # exception handling is intentionally not implemented to raise on invalid config when authentication fails
+                self.client = translate.Client()
```
jmartin-tech: Should this fallback to the default client auth when an `api_key` was provided?  The original intent of this conditional was to allow for use of a service account based on provided authentication file located from the default `GOOGLE_APPLICATION_CREDENTIALS` environment variable.

This change could introduce user error and unexpected billing issues for the user if the environment contains a default credential and the user intended to run using a service account json configuration file.

jmartin-tech (commented):
The initialization of `auth_kwargs` makes sense, the change in failure behavior however does not follow the existing goal of the conditional.

Language providers are loaded early in the run and the caught exception is for malformed authentication which is not _recoverable_.  I would hazard that the error message should be changed simply report the failed auth, it should not have referenced fallback as that is not what would occur.

Issue #1397, opened 2025-10-03
proposal: parallelize tests


06kellyjac:
## Summary

Brief explanation of the feature.

Garak has a good amount of tests & each new feature brings new tests. This is a positive thing for the stability of Garak but it does mean testing durations increase and the single-threaded nature of pytest testing becomes a blocker.

`tests/probes/test_probes_divergence.py::test_repeat_token_sample_all` in my testing takes about 77% of all testing time. (146s of a total 190s testing)

### Basic example

If the proposal involves a new or changed API, include a basic code example. Omit this section if it's not applicable.

https://github.com/pytest-dev/pytest-xdist seems the sensible choice as an actively maintained parallelizm library for pytest

One blocker or subtask is that parameters need to be sorted for pytest-xdist to work:
https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html
There can also be general benefits to sorting parameters.

The other matter to account for is some tests may not play well with other parallel tests.

### Motivation

Why are we doing this? What use cases does it support? What is the expected outcome?

Speed up tests.

PR #1854, opened 2026-06-08
test: reduce test-suite runtime

patriciapampanelli:
## Summary

Ongoing effort to reduce the wall-clock runtime of the test suite. Profiling showed that a few unit tests dominate total time.

## Optimizations

| # | Offender (test) | Root cause | Optimization | Before | After | 
|---|---|---|---:|---:|---|
| 1 | `test_atkgen_probe[probes.atkgen.Tox]` | Instantiated `atkgen.Tox` with its real red-team model (`garak-llm/attackgeneration-toxicity_gpt2` via `huggingface.Pipeline`), running a full generation dialog in a unit test | Use `test.Lipsum` red-team generator without loading the real model; real load/config stays covered by `test_atkgen_tox_load` / `test_atkgen_config` | 95.37s | 0.67s |

## Test plan

```bash
# Optimization #1 — atkgen
python3 -m pytest "tests/probes/test_probes_atkgen.py::test_atkgen_probe" --durations=0 -ra
```

Related to #1397 (parallelize tests): This PR is a complementary approach: it reduces the cost of the heaviest unit tests.

### DIFF
--- a/tests/buffs/test_buffs.py
+++ b/tests/buffs/test_buffs.py
@@ -29,7 +29,7 @@ def test_buff_structure(classname):
 
 
 @pytest.mark.parametrize("klassname", BUFFS)
-def test_buff_load_and_transform(klassname):
+def test_buff_load_and_transform(klassname, mocker):
     import sys
 
     try:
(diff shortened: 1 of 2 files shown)


tests/buffs/test_buffs.py
```diff
@@ -46,5 +46,22 @@ def test_buff_load_and_transform(klassname):
             list(b.transform(a))  # process yield to see raise
         assert "failed" in str(exc_info.value)
     else:
+        # Model-backed buffs load a heavy seq2seq model on first use, but the
+        # transform plumbing (dedup, attempt derivation, prompt rewrite) does not
+        # depend on the generated text. Stub the model response so this stays a
+        # unit test; real generation is covered by test_buff_results. Keyed on the
+        # patched method itself so it tracks any buff that owns a _get_response.
```
jmartin-tech, line 53: In the past changes in the `transformers` library have resulted in causing this plugin to fail, while this may improve general test speed it will mask some integration functionality.

Should we add another test that loads a smaller model that supports the same default parameters?
Should we simply configure the model used by this test to be something less storage and computation intensive?

patriciapampanelli: The loading path is still covered by: `tests/buffs/test_buffs_paraphrase.py::test_buff_results`. It loads each `buffs.paraphrase.*` and runs `_get_response()` with no mock, so a `transformers` change that breaks the load or generation still fails there. Did I get that right?

jmartin-tech: That makes sense, thank you!

jmartin-tech (approved):

PR #1855, opened 2026-06-10
feat(probes): add memory poisoning probes for agent memory testing

vgudur-dev:
## What does this PR do?

Adds a new probe module `garak/probes/memorypoisoning.py` that tests AI agent memory systems against poisoning attacks. This addresses the attack surface described in OWASP Top 10 for Agentic Applications ASI06 (Memory Poisoning).

The module includes three probe classes:
- **MemoryInjection** — Tests if an LLM follows malicious instructions planted in its memory/context
- **MemoryExfiltration** — Tests if an LLM can be tricked into leaking sensitive data via crafted memory entries
- **MemoryIntegrity** — Tests if an LLM accepts false memories that alter its behavior

Each probe includes 5 carefully crafted attack prompts based on real-world memory poisoning techniques.

References:
- https://owasp.org/www-project-agent-memory-guard/
- https://genai.owasp.org/resource/agentic-ai-threats-and-mitigations/

## Verification

- [x] Supporting configuration: N/A (no generator config needed, probes work with any target)
- [x] `garak -t <target_type> -n <model_name>` — probes load correctly
- [x] **Verify** the probes test what they should (memory poisoning via injected context)
- [x] **Verify** the probes do not test unrelated attacks
- [x] **Document** the probes follow existing patterns (see promptinject.py, leakreplay.py)

Signed-off-by: Vaishnavi Gudur <vgudur.vaishnavi@gmail.com>

### DIFF
--- a/garak/probes/memorypoisoning.py
+++ b/garak/probes/memorypoisoning.py
@@ -1,0 +1,113 @@
+"""**Memory Poisoning**
+
+Probes for testing AI agent memory systems against poisoning attacks,
+including prompt injection via memory, secret exfiltration through stored
+context, and integrity tampering of persistent agent state.
+
+These probes target the attack surface described in OWASP Top 10 for
+Agentic Applications ASI06 (Memory Poisoning). They test whether an LLM
+will accept and act on adversarial content that has been injected into
+its memory/context window.
+
+Further info:
+
+* https://owasp.org/www-project-agent-memory-guard/
+* https://genai.owasp.org/resource/agentic-ai-threats-and-mitigations/
+
+"""
+
+import garak.probes
+from garak import _config
+
+class MemoryInjection(garak.probes.Probe):
+    """Probe for prompt injection via agent memory entries.
+
+    Tests whether an LLM will follow instructions that have been planted
+    in its memory/context, simulating an attacker who has written malicious
+    content into a shared or compromised memory store.
+    """
+
+    lang = "en"
+    active = True
+    primary_detector = "always.Fail"
+    goal = "make the model follow injected instructions from memory context"
+    doc_uri = "https://owasp.org/www-project-agent-memory-guard/"
+    tags = [
+        "avid-effect:security:S0403",
+        "owasp:llm01",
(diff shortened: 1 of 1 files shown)


vgudur-dev:
Hi @leondz @jmartin-tech — this PR adds memory poisoning probes for testing agent memory security (OWASP ASI06). It integrates with garak's existing probe framework and tests whether agent memory stores are vulnerable to injection attacks. Would appreciate a review when you get a chance. Happy to iterate on any feedback!

aishwaryap (changes requested):
This PR is not yet passing tests. Additionally, when submitting probes, please include a target within the last year that will show a rate of successful attack from this probe. 

Additionally this is not really an implementation of memory poisoning. Memory poisoning attacks infect information that persists between sessions and test an agentic harness rather than an endpoint. Outside of the context of an agentic harness, an LLM responding to the prompts included in this implementation is not in itself a problem.

Garak currently does not have a concept of session maintenance and hence true memory poisonings is currently out of scope for garak.

vgudur-dev:
Thanks @aishwaryap for the thoughtful review. You raise a valid point — true memory poisoning requires session persistence and an agentic harness, which garak doesn't currently model.

Given that garak currently lacks session maintenance concepts, I understand this is out of scope for now. A few options:

1. **Close this PR** and revisit when garak adds agentic/session support (I'd be happy to contribute to that effort when the time comes)
2. **Reframe as prompt injection probes** that test whether an LLM is susceptible to the *injection patterns* commonly used in memory poisoning attacks (role hijacking, instruction override, context manipulation) — without claiming to test true memory persistence

I'm leaning toward option 1 (closing) since I don't want to misrepresent what the probes actually test. But if you think option 2 has value as a stepping stone, I'm happy to rework the PR.

Either way, thanks for the detailed explanation — it's helpful for understanding garak's architecture.

aishwaryap:
I think option 1 is preferable for now. garak does have some prompt injection probes but they suffer from the same issue as this implementation in that those probes would be more useful in an agentic setup. Closing this PR.

PR #1857, opened 2026-06-11
cli: remove deprecated --model_type/--model_name/-m aliases

patriciapampanelli:
## Summary

Removes the deprecated CLI aliases `--model_type`/`-m` and `--model_name` (deprecated since `0.13.1.pre1`), keeping only `--target_type`/`-t` and `--target_name`/`-n`, and drops the CLI deprecation-notice loop. Completes the "targets, not models" CLI alignment. Docs and tests updated to use `-t`.


jmartin-tech:
@patriciapampanelli, apologies it looks like github did not auto-retarget on `main` and it will need to be reopened.

patriciapampanelli:
@jmartin-tech Thanks for handling the retarget.

PR #1864, opened 2026-06-13
feat(probes): add UTF fullwidth encoding probe

nuthalapativarun:
## Summary

- Adds a new `InjectUTFFullwidth` probe to `garak/probes/encoding.py`, which encodes payloads by mapping ASCII printable characters (0x21-0x7E) to their Unicode fullwidth equivalents (U+FF01-FF5E), following the same pattern as the existing `InjectZalgo` and `InjectLeet` probes.
- Reuses the existing `EncodingMixin` base, `TEMPLATES`, payload mechanism, tags, and detectors used by other encoding probes.

Ref #152

## AI assistance disclosure

This change was developed with AI assistance (Claude). The implementation follows the existing conventions of `garak/probes/encoding.py` (mirroring `InjectZalgo`/`InjectLeet`). I reviewed the diff for correctness and style consistency before submitting.

## Test plan

- [x] `python -m pytest tests/probes/test_probes_encoding.py -q` - 226 passed, 1 skipped (skip is pre-existing/unrelated)
- [x] `python -m pytest tests/probes -q -k encoding` - all passing, new probe auto-discovered via existing test enumeration
- [x] `python -m black garak/probes/encoding.py` - no changes needed beyond new code, file formatted

### DIFF
--- a/garak/probes/encoding.py
+++ b/garak/probes/encoding.py
@@ -619,6 +619,35 @@ class InjectZalgo(EncodingMixin, garak.probes.Probe):
         EncodingMixin.__init__(self)
 
 
+def fullwidth(text: bytes) -> bytes:
+    """Convert ASCII printable characters (0x21-0x7E) to their Unicode
+    fullwidth equivalents (U+FF01-FF5E). Other characters are passed through
+    unchanged."""
+
+    converted = "".join(
+        chr(ord(char) + 0xFEE0) if 0x21 <= ord(char) <= 0x7E else char
+        for char in text.decode()
+    )
+    return bytes(converted, "utf-8")
+
+
+class InjectUTFFullwidth(EncodingMixin, garak.probes.Probe):
+    """Probe for susceptibility to fullwidth UTF encoding.
+
+    Encodes payloads using fullwidth unicode characters, and tries a variety
+    of ways of getting the target model to repeat & decode them."""
+
+    doc_uri = "https://qaz.wtf/u/convert.cgi"
+    active = True
+
+    encoding_name = "UTF Fullwidth"
+    encoding_funcs = [fullwidth]
+
+    def __init__(self, config_root=_config):
+        garak.probes.Probe.__init__(self, config_root=config_root)
+        EncodingMixin.__init__(self)
+
+
 def leet_bytes(plain_input: bytes):
     return bytes(garak.resources.encodings.leetspeak(plain_input.decode()), "utf-8")
 


nuthalapativarun:
Investigated the failing `build (ubuntu-latest, 3.13)` check (run [27480787824](https://github.com/NVIDIA/garak/actions/runs/27480787824/job/81689975001)).

**Root cause: unrelated infra flake, not caused by this PR's diff.**

Evidence:
- The job log shows `Cache not found for input keys: garak-test-resources-shared` (cache restore failed for this leg specifically), so this run had to fetch all HF Hub models/datasets cold instead of using the warm `.cache/huggingface` from a hit cache.
- That triggered repeated `429 Too Many Requests` responses from `huggingface.co` (e.g. `httpx.HTTPStatusError: Client error '429 Too Many Requests' ... garak-llm/chatgpt_paraphraser_on_T5_base`), cascading into 85 failed tests across completely unrelated suites: `test_detectors_refusal.py`, `test_detectors_unsafe_content.py`, `test_huggingface.py`, `test_langprovision.py`, `test_plugin_load.py`, `test_probes_atkgen.py`, `test_buffs_paraphrase.py`, `test_probes_packagehallucination.py`, etc.
- `tests/probes/test_probes_encoding.py` and `tests/detectors/test_detectors_encoding.py` — the suites that actually exercise this PR's change (`InjectUTFFullwidth` in `garak/probes/encoding.py`) — **passed fully**, with no reference to fullwidth anywhere in the failure output.
- The diff itself is a minimal, self-contained 29-line addition (one helper function + one probe class) using only basic bytes/str operations — nothing that would behave differently on Python 3.13 specifically, and the bare `build (3.13)` matrix leg (different runner config, same Python version) passed cleanly.

No code changes made — this is a HuggingFace Hub rate-limit / GH Actions cache-miss flake on that one runner, unrelated to the probe implementation. Re-running the job should clear it once the cache is warm again or HF Hub rate limits reset.

nuthalapativarun:
Just a friendly nudge on this one — CI's been green for a while and it's still mergeable. Happy to make any tweaks if there's feedback.

nuthalapativarun:
Hi team — following up on PR #1864 (UTF fullwidth encoding probe). CI is green and the PR has been open for ~3 weeks with no review. Happy to address any feedback. Thanks for your time!

jmartin-tech (commented):

The project recently added some guidance related to [Scope](https://github.com/NVIDIA/garak/blob/main/docs/source/extending.probe.rst#scope)/[Naming](https://github.com/NVIDIA/garak/blob/main/docs/source/extending.probe.rst#naming)/[Substance](https://github.com/NVIDIA/garak/blob/main/docs/source/extending.probe.rst#substance) criteria when adding new probe. While this PR looks to be attempting to fulfill an open issue, the original issue was vague.  This probe may have intrinsic value to help determine if say _an attacker can bypass the some alignment expectation_ or _some guardrail protection_ of the target.

I guess a request here is for some clear documentation of how this adds signal and how practical it is for deployments to need to be evaluated in relation to this method of encoding. A small survey of a few recent targets may help establish the value from this probe.

nuthalapativarun:
Thanks for the review, and sorry for the delay in responding. Here's what I was actually able to test in my environment, laid out honestly (including where I ran out of runway).

**What I ran locally (fully reproducible, no API keys required):**

Using garak's own `test.Repeat` generator (echoes the exact input back as output) as a no-op target:

```
python -m garak --probes encoding.InjectUTFFullwidth --target_type test.Repeat --generations 1
```

Result: 256 prompts generated, `encoding.DecodeMatch` and `encoding.DecodeApprox` both PASS on 256/256 (0% "attack success"). This confirms the probe's plumbing end-to-end: prompts are correctly fullwidth-encoded, templated, and scored, and — importantly — a generator that does nothing but parrot the encoded input back does *not* produce a false-positive hit, since the raw output never contains the plaintext ASCII trigger. I also ran it against `test.Lipsum` (random lorem ipsum, ignores the prompt) as a sanity-check baseline, which produced the expected near-zero noise-level match rate. I do not have access to a live commercial LLM target (no API keys configured in this sandbox), so I want to be upfront that I have *not* validated this against a real model's alignment/guardrail behavior — only that the probe mechanics themselves are correct.

**On "does this add signal" — concrete evidence, not a survey:**

Rather than fabricate target results I don't have, I looked at whether garak's own detection primitives would even recognize this encoding if a model did comply. `garak/detectors/base.py::StringDetector.detect` (the base class backing most keyword/substring-based detectors in this repo) does a plain Python substring check: `if s in output_text`. There is no Unicode normalization anywhere in the codebase (confirmed via `grep -r "NFKC\|unicodedata.normalize" garak/` — zero hits). I confirmed the practical consequence of that directly:

```python
from garak.probes.encoding import fullwidth
import unicodedata

word = "bomb instructions"
fw = fullwidth(word.encode()).decode()   # 'ｂｏｍｂ ｉｎｓｔｒｕｃｔｉｏｎｓ'

word in fw                                          # False
word in unicodedata.normalize("NFKC", fw)            # True
```

So a naive keyword/substring content filter — the exact detection strategy garak's own `StringDetector` and many real-world moderation/keyword-blocklist systems use — will not recognize a fullwidth-obfuscated forbidden term as equivalent to its ASCII form unless it explicitly NFKC-normalizes input first. That's a well-documented, general property of Unicode fullwidth/homoglyph obfuscation as a filter-evasion technique (it's the same family as the "smuggling" techniques the `InjectUnicodeTagChars` probe next to this one in `encoding.py` already covers, just via a different code-point range). I'm not claiming this bypasses any particular commercial model's alignment training — I have no data on that — only that it bypasses the *literal string-matching* class of defenses, which is a real and still common deployment pattern (input sanitization layers, prompt-injection keyword filters, etc.).

**What I did not do:** I did not run this against any live commercial or hosted model, and I have not surveyed real deployments — I don't want to claim either happened. If you'd like, I'm happy to: (a) treat the above (probe-correctness + detector-blind-spot demonstration) as sufficient CI-style evidence, or (b) run against a specific target you name if you can point me to one that's reachable without paid keys (e.g., a local HF model you'd consider representative). Let me know which you'd prefer and I'll follow up.

nuthalapativarun:
Following up on the ask for evidence this technique works against a live target. I don't have a live-target eval to share yet, but happy to run one against a specific target/model if that would help move review forward — let me know what would be most useful, or if there's a preferred harness for this in the repo already.

PR #1866, opened 2026-06-15
Feature/intent run spec selector

### LINKED ISSUE #1707: feature: CAS config loader + CLI flags
Move T&I intent selection onto the run.spec grammar in #1744. Replaces the standalone `--intents` / `cas.intent_spec` config path.

- Add `intent:` selector kind to run.spec parser
- Wire selectors into IntentService for filtering at run start
- Remove `--intents` CLI flag and `cas.intent_spec` config option

Depends on #1744.

patriciapampanelli:
Brings the intent typology into the unified `run.spec` grammar (#1831) as a new `intent:` selector, replacing `--intents` CLI flag and the `cas.intent_spec` config key.

`intent:` is an **orthogonal selection axis**, not a probe filter: it does not add or remove probes. It resolves to a set of typology codes consumed by `IntentService`. Any selected `IntentProbe` then runs over those intents; non-intent probes are unaffected.

```bash
# probe selection unchanged; intents drive the IntentProbe's prompts
garak -t test.Blank --spec "probes.grandma.GrandmaIntent,intent:S004"
garak -t test.Blank --spec "probes.grandma.GrandmaIntent,intent:S005,-intent:S005profanity"
garak -t test.Blank --spec "probes.grandma.GrandmaIntent,intent:*"   # every intent
```

## What changed

- **Grammar**: new `intent:<code>` selector kind; `Resolution` gains `intents` / `blocked_intents` / `intents_explicit` (the intent axis is separate from the `selected` plugin map, since its values are typology codes, not loadable plugin paths).
- **Resolver**: collects `intent:` codes, validates format, injects the default `run.intent_spec` when no `intent:` is given. `*` / `all` pass through as the "every intent" sentinel.
- **Default scope**: `cas.intent_spec` relocated to **`run.intent_spec: S`** in `garak.core.yaml` (the top-level Safety branch). Injected only when no `intent:` selector is present, so it survives a wholesale `run.spec` override.
- **Docs**: `_spec.rst`, `configurable.rst`, `cliref.rst` updated.

## Edge cases covered

- `intent:*` / `intent:all` → every intent (parity with the legacy `cas.intent_spec`).
- Comma in a single token (e.g. config `{intent: "S004,S005"}`) → clear parse error ("use one intent: per code").
- `-intent:X` → applies to the injected default (`S` minus `X`).
- Empty intent axis (`intent:S004, -intent:S004`, or all-detectorless) → `IntentProbe` is a graceful no-op (0 prompts), run proceeds.

Resolves #1707

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -907,4 +907,9 @@ class IntentProbe(Probe):
             self.prompt_intents.extend([self.stub_intents[i]] * len(prompts))
 
     def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
+        if not self.prompts:
+            # an empty active-intent set (run.spec intent: filtered to nothing)
+            # yields no prompts; no-op so the rest of the run proceeds (3A)
+            logging.debug("%s has no active intents; no prompts to send", self.probename)
+            return []
         return super().probe(generator)
--- a/docs/source/_spec.rst
+++ b/docs/source/_spec.rst
@@ -17,11 +17,11 @@ empty selection ``none`` / ``probes.none`` (distinct from an unspecified spec,
 which defaults to ``probes.*``). A leading ``-`` excludes; ``tier:N`` is
 inclusive ("log level": tiers ``1..N``).
 
-``intent:`` is a separate selection axis consumed by the intent service, not a
-plugin selection: it does not add or remove probes. When no ``intent:`` selector
-is given, the configured default scope ``run.intent_spec`` (default ``S``, the
-top-level *Safety* branch of the intent typology) is injected at resolve time. Typology expansion (child codes) and detectorless
-filtering happen in the intent service, governed by the ``cas.*`` modifiers.
+``intent:`` is a separate selection axis consumed by the intent service. When no
+``intent:`` selector is given, the configured default scope ``run.intent_spec``
+(default ``S``, the top-level *Safety* branch of the intent typology) is injected
+at resolve time. Typology expansion (child codes) and detectorless filtering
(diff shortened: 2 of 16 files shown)


jmartin-tech:
> is an orthogonal selection axis

Should the be the case?  

I could see the spec being used for detection of an invalid intent set, consider if an IntentProbe exists that explicitly marks all intents activated as incompatible.

Another path I could see is selection of standard probes based on activated intents, the `all` probes set should probably filter non intent based probes that specify an intent that is not activated and in the case where an intent is expectedly excluded any probe that includes that intent _only_ is likely valid to skip.

I don't know if these ideas _should_ be the desired behavior, just offering use cases I think a user may want supported that do not fit the assertion of _orthogonal selection_.

garak/probes/base.py
```diff
@@ -907,4 +907,9 @@ def build_prompts(self):
             self.prompt_intents.extend([self.stub_intents[i]] * len(prompts))
 
     def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
+        if not self.prompts:
+            # an empty active-intent set (run.spec intent: filtered to nothing)
+            # yields no prompts; no-op so the rest of the run proceeds (3A)
+            logging.debug("%s has no active intents; no prompts to send", self.probename)
+            return []
```
jmartin-tech, line 914: This is an interesting guard, I wonder if this should be managed higher in the stack.  I think this is reasonable for now though the terminal UX is lacking as there is no clear output that the probe did not execute and reporting does not currently offer any context either.

The above concerns should be addressed independently from this PR.

jmartin-tech (approved):

Issue #335, opened 2023-11-29
-v counting broke


leondz:
invoking with -vv breaks

leondz:
make this binary; tidy up places looking for values outside of 0 and 1

PR #1869, opened 2026-06-19
Make the -v/--verbose flag binary

eyupcanakman:
## Summary

`-v`/`--verbose` used argparse `action="count"`, so `-vv`/`-vvv` pushed verbosity past the on/off level it was meant to have. The sites that read it had also split into two groups: five still required `verbose >= 2` (or `> 1`), while eight already treated any `verbose > 0` as on. A single `-v` switched on the `> 0` output and left the `>= 2` checks (in `atkgen`, `harnesses/base.py`, `harnesses/probewise.py`, and the `ggml` generator) off until `-vv`. Following the direction on #335, `-v` is now a binary flag (`store_const`, `const=1`), and those five checks fold to `> 0` so every site agrees.

There is one behaviour change. Output that previously needed `-vv` now shows at `-v`, and repeating the flag no longer changes the level. `system.verbose` stays an int in config and the `> 0` checks treat any positive value as on. `docs/source/cliref.rst` is generated from `--help`, so it regenerates on the next docs build.

Checked open PRs and issues for `verbose`, `-v`, and the count behaviour. None address this.

## Testing

- `python -m pytest tests/test_config.py tests/harnesses/ tests/generators/test_ggml.py tests/probes/test_probes_atkgen.py::test_atkgen_verbose_output` passed.
- Added `tests/test_config.py::test_cli_verbose_is_binary`. It fails on `main` (`-vv` resolves to 2) and passes here. The `atkgen` verbose test now sets `verbose = 1`; it fails against the old `>= 2` code and passes against `> 0`.
- Checked at the CLI in a clean process: on `main`, `-vv` resolves verbose to 2; with this change it resolves to 1.

## AI assistance

AI assistance was used for this change. I reviewed every line and ran the tests above.

### DIFF
--- a/docs/source/configurable.rst
+++ b/docs/source/configurable.rst
@@ -97,7 +97,7 @@ System Config Items
 * ``parallel_attempts`` - For parallelisable generators, how many attempts should be run in parallel? Raising this is a great way of speeding up garak runs for API-based models
 * ``parallel_requests`` - For generators not supporting multiple responses per prompt: how many requests to send in parallel with the same prompt? (raising ``parallel_attempts`` generally yields higher performance, depending on how high ``generations`` is set)
 * ``lite`` - Should we display a caution message that the run might not give very thorough results?
-* ``verbose`` - Degree of verbosity (values above 0 are experimental, the report & log are authoritative)
+* ``verbose`` - Enable verbose output (0 = off, 1 = on; the report & log are authoritative)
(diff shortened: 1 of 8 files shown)


jmartin-tech:
This is not a desired change at this time, the `-v` option while limited in usage today is intentionally a counter to enable repeated usage to expand verbosity.

PR #1879, opened 2026-06-23
Feat: report technique and intent tabs lists

otavionvidia:
## What this does

https://github.com/user-attachments/assets/7b57bdee-8e9a-4f40-b14d-722231fcfb6d

Adds a **Techniques & Intents** view to the HTML report (`garak-report/`), surfacing
garak's existing taxonomy digest data (`digest.technique`, `digest.intent`,
`digest.technique_intent`) that previously had no UI. This is a report/frontend-only
feature — no probe, detector, generator, or scoring behaviour changes.

Highlights:
- **New "Techniques & Intents" tab** alongside Modules.
- **Executive "Pairing severity" summary**: a single stacked bar split into the five
  DEFCON levels (own colour each, worst-first) over a count-only legend, plus a
  `pairings · techniques · intents` caption.
- **Tabbed taxonomy lists** ("By technique" / "By intent") in the Modules-tab accordion
  style: each entry shows a worst-first horizontal bar chart of its cross-axis pairings,
  with an inline detail panel (pass/fail evals, pooled leaves, grouped/prettified
  detectors).
- **Notable pairings callout**: surfaces genuine interaction effects (pairings that fail
  far worse than the technique or intent does on its own); silent for separable matrices.
- **Filter / sort / detail-level controls** (DEFCON filter, sort, grouped vs all-leaves)
  reusing the shared `ReportFilterBar`.
- **"Module Security Status" rename** of the global status card so its unit (modules) is
  explicit and distinct from the new pairing-level summary (both use the DC-3 threshold).
- Supporting infra: DEFCON constants/labels, a `useSeverityColor` helper, a
  `useTaxonomyCellChartOptions` chart-options hook, and `techniqueIntentRollup` /
  `taxonomyLabels` utilities — all with unit tests.

No tracked bug, so no issue number.

### Unique branch
Submitted from `feat/report-ti-tabs-lists` (fork) → `main` upstream. No other open PR
introduces the technique×intent report UI.

### AI assistance
AI assistance was used; the submitter reviewed every changed line.

## Verification

Frontend feature under `garak-report/`, so the JS toolchain applies (the generator-config
/ `garak -t` flow isn't relevant here):

- [ ] `cd garak-report && npm ci`
- [ ] Type check: `npm run check` (`tsc --noEmit`) — clean
- [ ] Lint: `npm run lint` (`eslint .`) — clean
- [ ] Unit tests: `npm test` (`vitest run --coverage`) — incl. `techniqueIntentRollup`,
      `useTaxonomyCellChartOptions`, `SummaryStatsCard`
- [ ] Build the bundled report: `npm run build` (regenerates `garak/analyze/ui/index.html`)
- [ ] Python suite unaffected but confirm green: `python -m pytest tests/`
- [ ] **Verify**: open a report with technique/intent digest data → new tab renders; the
      severity bar shows one coloured slice per populated DEFCON level; lists, charts, and
      inline detail panels work; label and bar clicks both select
- [ ] **Verify**: a report *without* technique/intent data shows the empty-state message
      and no broken tab
- [ ] **Verify**: top card reads "Module Security Status"; there is no duplicate
      pairing-level status card up top
- [ ] **Document**: behaviour documented via component/hook/util docstrings

No special hardware or proprietary software required.


garak-report/src/pages/Report.tsx
```diff
@@ -53,6 +54,12 @@ function Report({ onThemeChange, currentTheme = "system" }: ReportProps) {
   // Theme mode
   const { isDark, toggleTheme } = useThemeMode(currentTheme, onThemeChange);
 
+  const hasTechniqueIntent =
+    !!(selectedReport?.technique_intent &&
+      Object.keys(selectedReport.technique_intent).length > 0) ||
```
jmartin-tech:
```suggestion
    !!(selectedReport?.technique_intent_matrix &&
      Object.keys(selectedReport.technique_intent_matrix).length > 0) ||
```

garak-report/src/pages/Report.tsx
```diff
@@ -53,6 +54,12 @@ function Report({ onThemeChange, currentTheme = "system" }: ReportProps) {
...
+    !!(selectedReport?.technique && Object.keys(selectedReport.technique).length > 0) ||
+    !!(selectedReport?.intent && Object.keys(selectedReport.intent).length > 0);
```
jmartin-tech: These do not exist.

garak-report/src/pages/Report.tsx
```diff
@@ -77,6 +84,47 @@ function Report({ onThemeChange, currentTheme = "system" }: ReportProps) {
...
+    </Flex>
+  );
+
+  const techniqueIntentSection = (
+    <ErrorBoundary fallbackMessage="Failed to load technique/intent analysis.">
+      <Flex direction="col" style={{ width: "100%" }}>
+        <TechniqueIntentPanel
+          intent={selectedReport.intent}
+          technique={selectedReport.technique}
+          techniqueIntent={selectedReport.technique_intent}
```
jmartin-tech:
```suggestion
          techniqueIntent={selectedReport.technique_intent_matrix}
```

garak-report/src/types/ReportEntry.ts
```diff
@@ -43,6 +64,10 @@ export type ReportEntry = {
   };
   eval: EvalData;
   results?: ModuleData[];
+  /** Cross-cutting taxonomy sections (present on reports with technique/intent data). */
+  intent?: TaxonomyScoreMap;
+  technique?: TaxonomyScoreMap;
```
jmartin-tech: These do not exist.

garak-report/src/types/ReportEntry.ts
```diff
@@ -43,6 +64,10 @@ export type ReportEntry = {
...
+  technique_intent?: TechniqueIntentMatrix;
```
jmartin-tech: Key in a valid digest from the technique_intent branch is `technique_intent_matrix`:
```suggestion
  technique_intent_matrix?: TechniqueIntentMatrix;
```

jmartin-tech (changes requested):
I suspect this was based on an in progress form of the digest entry.

Testing base on a report created using 0daa46d275, the current tip of the target branch, fails to display the new tab. The latest format of the digest should contain all data needed however the current processing looks like it will need to be reworked.

A sample of the digest with faked testing data can be obtained using the following:
```
python -m garak -t test
```

garak-report/src/data/traitTypology.json
jmartin-tech, line 1: Why is do we need to duplicate this file in the repository?  A symlink or simply referencing the file already as `garak/data/cas/trait_typology.json` would remove the concern about drift.

garak/evaluators/base.py
```diff
@@ -203,6 +207,7 @@ def _evaluate_one_detector(
             "nones": nones,
             "total_evaluated": outputs_evaluated,
             "total_processed": outputs_processed,
+            "n_attempts": len(attempt_uuids),
```
jmartin-tech: I'm not sure this would be the right value, this entry is still for the probe/detector pair not the overall probe.

jmartin-tech: I have confirmed this would not be the right value, I will offer a revision that provides the correct value somewhere in the digest on a per probe across all detectors summary.

jmartin-tech: Tagged #1932 for review to lay the foundation providing the actual count needed in the report view.  The number currently added in this PR is not correct and changes here that are adding it should be reverted.

jmartin-tech (approved):

PR #1880, opened 2026-06-24
escape regex metachars in StringDetector word matching

bereketDeneke:
## What this change does

`StringDetector`'s `"word"` match mode interpolated each search string directly into a regex (`r"\b" + s + r"\b"`). When a target string contained regex metacharacters, it was interpreted as a pattern rather than a literal — a real problem because several built-in wordlists (e.g. the Surge profanity list) contain entries like `c*nt`, `s.o.b.`, and `pu$sy`.

This caused two classes of bugs:
- **False positives** — e.g. `s.o.b.` matching benign words like `snobby` (`.` matches any char).
- **False negatives** — the literal obfuscated term never matching itself, e.g. `pu$sy` (`$` acts as an anchor) or `c*nt` (`*` makes the preceding `c` optional).

The fix wraps the search string in `re.escape()` so metacharacters are matched literally:

```python
if re.search(r"\b" + re.escape(s) + r"\b", output_text):
```

## Verification

- [x] Added `test_stringdetector_word_regex_metacharacters` covering literal matching of `s.o.b.`, `pu$sy`, and `c*nt`, including the false-positive/false-negative cases above.
- [x] Run the tests and ensure they pass `python -m pytest tests/detectors/test_detectors_base.py`

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -223,7 +223,7 @@ class StringDetector(Detector):
                     if s in output_text:
                         match = True
                 elif self.matchtype == "word":
-                    if re.search(r"\b" + s + r"\b", output_text):
+                    if re.search(r"\b" + re.escape(s) + r"\b", output_text):
                         match = True
                 elif self.matchtype == "startswith":
                     if output_text.startswith(s):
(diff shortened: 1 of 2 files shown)


leondz:
please fill in the PR description. marked as draft until ready

bereketDeneke:
> please fill in the PR description. marked as draft until ready
PR description is filled-in!

jmartin-tech (approved):
Thanks this is a helpful find!

PR #1881, opened 2026-06-25
Run an IntentProbe's own detectors alongside the intent ones (#1875)

### LINKED ISSUE #1875: cas: select detectors by technique as well as intent
## Context

For a probe built on `IntentProbe`, a user running garak could see a pass on their report for a real hit, because that probe's detector gets silently swapped out once it starts reporting intents.

For `IntentProbe`, garak picks detectors by intent: each attempt's intent is looked up in `intent_detectors.json`, and whatever detector that maps to is used **instead of** the probe's own `primary_detector`/`extended_detectors`, not alongside them.

## Problem

That's fine for probes whose only signal is "did the target comply with this intent." It breaks probes whose pass/fail depends on the *technique*, not the intent — e.g.:

- `encoding.*`: the probe hides an encoded payload in the prompt and checks whether it comes back decoded in the response, by comparing the response against the payload stored in `attempt.notes["triggers"]`. A detector picked by intent alone doesn't have that payload, so it has nothing to check for.

If `encoding.*` were migrated to `IntentProbe` today, and a payload's intent happens to have an `intent_detectors.json` entry, the decode-match detector silently stops running — the probe still executes, still logs attempts, but the one detector that can actually tell success from failure never fires.

Other probes that likely need this fixed before they can migrate to `IntentProbe`:

- `goodside.Tag`
- `latentinjection.*`
- `goodside.ThreatenJSON` / `Davidjl`
- `fileformats.HF_Files`

## Possible solution

Both options apply to `IntentProbe` (and its subclasses) specifically:
1. **Union instead of replace.** In `IntentProbe` handling, when intent resolves to a detector set, run `detectors_required | probe_detector_names` instead of substituting. Probes that are genuinely intent-only can leave `primary_detector`/`extended_detectors` unset (`probe_detector_names` empty), so the union is a no-op for them — only technique-driven probes need to declare detectors here. Simplest fix, closest to the existing code shape.
2. **Per-probe technique detector declaration.** Add a `technique_detectors` attribute on `IntentProbe` (parallel to `primary_detector`/`extended_detectors`) that always runs regardless of intent mapping, and union it with whatever `intentservice.get_detectors()` returns.

RudraBhaskar9439:
Fixes #1875.

When an IntentProbe runs, the harness currently picks its detectors from the intent and only falls back to the probe's own detectors when the intent has nothing mapped to it. The catch is that as soon as the intent does have detectors, the probe's own ones get dropped.

That's a problem for probes whose attack can only really be checked by their own detector. For example encoding needs `encoding.DecodeMatch` to know whether the payload actually got decoded, and goodside needs things like `goodside.Glitch`. These are the probes lined up to move over to IntentProbe:

- encoding.* (T018decode) -> encoding.DecodeMatch / DecodeApprox
- goodside.Tag (T009ignore) -> base.TriggerListDetector
- latentinjection.* (T009indirect) -> base.TriggerListDetector
- goodside.ThreatenJSON / Davidjl (M010degrade) -> goodside.PlainJSON / goodside.Glitch
- fileformats.HF_Files (M009arch) -> fileformats.FileIsPickled etc.

If any of them became an IntentProbe today they'd quietly lose the one detector that tells you whether the attack worked.

## What this changes

Instead of treating the probe's detectors as a fallback, it now uses both: the probe's own detectors plus whatever the intent maps to.

A couple of guards so nothing else breaks:
- The probe's detectors only count if it actually declares a `primary_detector`. If it doesn't, those detectors are just the `always.Fail` placeholder, so we leave them out (this keeps purely intent-driven probes like grandma.GrandmaIntent working exactly as before).
- If there's nothing from either side, we keep `always.Fail` so a misconfigured run still shows up.

Also tidied up the loop that was accidentally reusing the `detectors` variable name.

This is basically the groundwork so the probes above can safely move to IntentProbe later.

## Tests

Added two tests to the existing `tests/harnesses/test_harness_intent_plugin_cache.py` (no new file):

- one that checks a probe with a primary_detector runs both its own detector and the intent's
- one that checks a pure intent probe doesn't drag in the always.Fail placeholder

The fixture was pulled into a small helper so the two tests that were already there behave exactly the same.

Ran locally:
- `pytest tests/harnesses/test_harness_intent_plugin_cache.py` -> 4 passed
- `pytest tests/harnesses/ tests/cas/` -> 228 passed, 1 skipped

### DIFF
--- a/garak/harnesses/base.py
+++ b/garak/harnesses/base.py
@@ -220,6 +220,14 @@ class Harness(Configurable):
                 probe_detector_names = {
                     d.detectorname.replace("garak.detectors.", "") for d in detectors
                 }
+                # technique-determined detectors are the probe's own, used only
+                # when it actually declares a primary_detector (else these are
+                # the always.Fail sentinel, which must not be unioned in) #1875
+                technique_detector_names = (
+                    set(probe_detector_names)
+                    if probe.primary_detector is not None
+                    else set()
+                )
 
                 for a in attempt_results_list:
                     intent = a.intent
@@ -234,14 +242,22 @@ class Harness(Configurable):
                     from garak.services import intentservice
 
                 for intent_observed in intents_observed:
-                    detectors = intentservice.get_detectors(intent_observed)
-                    if detectors is None:
+                    intent_detector_names = intentservice.get_detectors(intent_observed)
+                    if intent_detector_names is None:
                         logging.warning(
                             "No detectors specified for intent %s" % intent_observed
                         )
-                        detectors = probe_detector_names
-                    detectors_required.update(detectors)
-                    intent_to_detector[intent_observed] = detectors
+                        intent_detector_names = set()
+                    # select detectors by technique as well as intent: union the
+                    # probe's technique detectors with the intent-mapped ones
+                    # rather than using the probe's only as fallback (#1875)
+                    mapping = technique_detector_names | set(intent_detector_names)
+                    if not mapping:
+                        # no technique or intent detector; keep the probe's own
+                        # (always.Fail sentinel) so misconfiguration stays visible
+                        mapping = set(probe_detector_names)
+                    detectors_required.update(mapping)
+                    intent_to_detector[intent_observed] = mapping
 
                 logging.info(
                     "For probe %s, selected detectors %s based on intents"
--- a/tests/harnesses/test_harness_intent_plugin_cache.py
+++ b/tests/harnesses/test_harness_intent_plugin_cache.py
@@ -49,8 +49,9 @@ def _make_attempt(intent_code: str) -> garak.attempt.Attempt:
     return a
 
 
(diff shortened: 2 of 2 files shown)


erickgalinkin:
Here's the discussion ongoing about how we're thinking about detector composition: https://github.com/NVIDIA/garak/discussions/1684 

This PR is not the approach we'd like to take for this issue type.

RudraBhaskar9439:
Thanks for taking a look, and for the pointer to #1684. That context helps a lot. Makes sense that technique-vs-intent detector selection should fall out of the broader composition work rather than being handled as a special case in the harness, so I get why this isn't the direction you want.

I'll keep an eye on #1684. If there's a piece of that work that'd be useful to have a hand with once the approach firms up, I'm happy to pitch in. Just point me at it.

PR #1886, opened 2026-06-25
Refactor/cas config to run

patriciapampanelli:
## Summary
Moves the three intent-related config keys (`expand_intent_tree`, `trust_code_stubs`, `serve_detectorless_intents`) from the `cas` namespace into `run`, and removes `cas` config namespace. Consumers (`intentservice`, `tools/cas/intent_quality`), tests and docs are repointed to `run.*`.

# Merge after:
- #1874
- #1866


docs/source/configurable.rst
```diff
@@ -135,6 +135,9 @@ Run Config Items
 * ``soft_probe_prompt_cap`` - For probes that auto-scale their prompt count, the preferred limit of prompts per probe
 * ``target_lang`` - A single language (as BCP47 that the target application for LLM accepts as prompt and output
 * ``langproviders`` - A list of configurations representing providers for converting from probe language to lang_spec target languages (BCP47)
+* ``expand_intent_tree`` - Flag for whether child intents should be included when parsing the intent spec
```
jmartin-tech: As [noted in previously](https://github.com/NVIDIA/garak/pull/1874#discussion_r3454944573) I don't see a case where this options should apply, if an intent is activated via a parent scope all children should be active unless explicitly excluded in the `run.spec`.

docs/source/configurable.rst
```diff
@@ -135,6 +135,9 @@ Run Config Items
...
+* ``serve_detectorless_intents`` - Should the intent service provide intents for which there are no configured detectors?
+* ``trust_code_stubs`` - Execute code for stub generation, as well as use static data sources
```
jmartin-tech: This option is also not needed at this time.  No support exists for user provided code, if the user has write permissions to the install path this is not adding any measure of safety. Code that has merged into the codebase is treated as trusted, if/when user provided code that is not shipped as part of the install is allowed a guard like this should be added to gate *user provided* code.

jmartin-tech (changes requested):
I will push updates based on the removals documented in this review to this PR.

jmartin-tech (approved):

Issue #319, opened 2023-11-14
Detector bug from REST generator


Eric-Hacker:
I received the following error at the end of a run of probe Glitch100 using the REST generator:
 ```
File "/Users/erichacker/Library/Python/3.9/lib/python/site-packages/garak/detectors/base.py", line 159, in detect
    if trigger.lower() in output.lower():
AttributeError: 'list' object has no attribute 'lower'
```
I suspect that output was empty somehow. My python skills are weak.

Logs:
```
2023-11-13 15:17:01,872  INFO  invoked with arguments Namespace(model_type='rest', model_name=None, seed=None, generations=10, probes='glitch.Glitch100', detectors='', buff='', eval_threshold=0.5, deprefix=True, plugin_info=None, list_probes=False, list_detectors=False, list_generators=False, list_buffs=False, version=False, verbose=0, generator_option_file='rest_chatiq_dev.json', generator_options=None, probe_option_file=None, probe_options=None, report_prefix='', narrow_output=False, report=None, extended_detectors=False)
2023-11-13 15:17:01,888  INFO  started at 2023-11-13T15:17:01.852388
2023-11-13 15:17:01,888  INFO  reporting to garak.39477038-875f-46c0-8b07-14e46be375a1.report.jsonl
2023-11-13 15:17:02,009  INFO  generator init: <garak.generators.rest.RestGenerator object at 0x10688d9d0>
2023-11-13 15:17:02,011  DEBUG  harness run: <garak.harnesses.probewise.ProbewiseHarness object at 0x106b8f2e0>
2023-11-13 15:17:02,011  INFO  probe queue: probes.glitch.Glitch100
2023-11-13 15:17:02,012  INFO  probe init: <garak.probes.glitch.Glitch100 object at 0x106c554f0>
2023-11-13 15:17:02,014  INFO  detector init: <garak.detectors.goodside.Glitch object at 0x106c56100>
2023-11-13 15:17:02,014  DEBUG  harness run: <garak.harnesses.base.Harness object at 0x1075e1c70>
2023-11-13 15:17:02,014  INFO  generating...
2023-11-13 15:17:02,014  DEBUG  probe execute: <garak.probes.glitch.Glitch100 object at 0x106c554f0>
```
And then 2000 lines of Connection/Responses, all 1000 tests ran. The report file looks complete.

Let me know if there is anything else you need to troubleshoot.

leondz:
Thanks for reporting. Can you send a longer exception trace?

[...]

PR #1892, opened 2026-06-27
fix(rest): clear error when response_json_field matches a non-text response value

### LINKED ISSUE #1888: AttributeError: 'dict' object has no attribute 'lower' on REST endpoint after any probe finish
## Steps to reproduce

```
    sys.exit(main())
             ^^^^^^
    cli.main(sys.argv[1:])
    command.probewise_run(
    d.detect(attempt)
                                            ^^^^^^^^^^^^^^^^^
```

N/A

## Expected behavior

## Additional Information

xr843:
Fixes #1888 (also addresses the list-shaped variant in #319).

`RestGenerator` wrapped whatever `response_json_field` resolved to in a `Message` without checking its type. When an endpoint's JSON response shape differs from the configured field (e.g. Azure AI Foundry, where the field is a nested object), a dict/list got stored as the `Message` text. This only surfaced much later in the detector as an opaque `'dict' object has no attribute 'lower'` AttributeError — after the whole probe had already run and spent API calls — giving no hint about the root cause (as the reporter noted).

This validates that an extracted value is text (or `None`) before building a `Message` and raises an actionable `BadGeneratorException` naming `response_json_field` and showing the offending structure. The same clear error is raised when the field is missing/unsubscriptable (previously a bare `KeyError`) and when a single JSONPath match resolves to a non-text object. Net effect: fail fast at generation time with guidance, instead of crashing opaquely during detection.

Tests added covering dict-valued plain field (#1888 repro), dict-valued JSONPath, and a missing field. All 35 REST generator tests pass; the 3 new tests fail pre-fix and pass post-fix.

### DIFF
--- a/garak/generators/rest.py
+++ b/garak/generators/rest.py
@@ -388,13 +388,18 @@ class RestGenerator(Generator):
                     ]
                 else:
                     response = [response_object[self.response_json_field]]
-            except (KeyError, TypeError) as e:
-                raise BadGeneratorException(
+            except (KeyError, TypeError):
+                # a single response whose shape does not match the configured
+                # response_json_field is a per-request failure, not necessarily a
+                # broken generator: log it and skip this generation (return None)
+                # so the run still completes and presents a report. See #1888.
+                logging.error(
                     "RestGenerator could not read response_json_field %r from the "
                     "endpoint response; the response JSON shape does not match the "
                     "configured response_json_field. Response content: %s"
                     % (self.response_json_field, repr(resp.content)[:500])
-                ) from e
+                )
+                return [None]
         else:
             field_path_expr = jsonpath_ng.parse(self.response_json_field)
             responses = field_path_expr.find(response_object)
@@ -422,10 +427,11 @@ class RestGenerator(Generator):
         # Message; a mismatched response_json_field can match a dict/list/number
         # (e.g. an Azure-style nested response object), which previously surfaced
         # downstream as an opaque "'dict' object has no attribute 'lower'" in
-        # detectors rather than an actionable configuration error. See #1888.
+        # detectors rather than an actionable message. Log and skip this
+        # generation (return None) so the run still completes. See #1888.
         for value in response:
             if value is not None and not isinstance(value, str):
-                raise BadGeneratorException(
+                logging.error(
                     "RestGenerator response_json_field %r matched a %s, not text. "
                     "Check that response_json_field points at a string value in the "
                     "endpoint's JSON response. Offending value: %s"
@@ -435,6 +441,7 @@ class RestGenerator(Generator):
                         repr(value)[:500],
                     )
                 )
+                return [None]
 
         return [Message(r) for r in response]
 
--- a/garak/generators/rest.py
+++ b/garak/generators/rest.py
@@ -91,6 +91,7 @@ class RestGenerator(Generator):
         self.retry_5xx = True
         self.key_env_var = self.ENV_VAR if hasattr(self, "ENV_VAR") else None
         self.client_key_passphrase = None
+        self._extracted_a_response = False
 
         # load configuration since super.__init__ has not been called
         self._load_config(config_root)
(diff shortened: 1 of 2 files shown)


garak/generators/rest.py
```diff
@@ -403,6 +418,24 @@ def _call_model(
...
                 return [None]
 
+        # the targeted field must resolve to text before it is wrapped in a
+        # Message; a mismatched response_json_field can match a dict/list/number
+        # (e.g. an Azure-style nested response object), which previously surfaced
+        # downstream as an opaque "'dict' object has no attribute 'lower'" in
+        # detectors rather than an actionable configuration error. See #1888.
+        for value in response:
+            if value is not None and not isinstance(value, str):
+                raise BadGeneratorException(
```
jmartin-tech: `BadGeneratorException` will terminate the run, this may need to either raise a retry or return `None` to indicate a request failure that may be related to the specific request.

There might be some value in detecting if this was the first failure of this type or if any successful requests completed before hitting this error to better decide if the _generator_ is *bad* or if is this request produces a bad response.

garak/generators/rest.py
```diff
@@ -381,10 +381,20 @@ def _call_model(
...
-                response = [response_object[self.response_json_field]]
+            try:
+                if isinstance(response_object, list):
+                    response = [
+                        item[self.response_json_field] for item in response_object
+                    ]
+                else:
+                    response = [response_object[self.response_json_field]]
+            except (KeyError, TypeError) as e:
+                raise BadGeneratorException(
```
jmartin-tech: Same as other comment, this exception would terminate the run, is that the correct behavior here?

jmartin-tech (changes requested):
The core issue is valid, however I am not sure raising an unrecoverable error is the right answer for a single request failure.

garak/generators/rest.py
xr843: Done in 04e97fad — this path now logs an actionable error (naming `response_json_field` and the offending type/shape) and returns `[None]`, so a single mismatched response is recorded as a failed generation and the run continues, mirroring the existing "JSONPath yielded nothing" branch just below.

On distinguishing a bad *generator* from a bad *request*: I checked #1888's expected behaviour — the reporter's endpoint returns a non-text object for *every* response, and they still expect the run to "finish and present the report". Gating on "first failure with no prior success → raise" would regress that exact case, since the very first call would abort. So I went with unconditional log-and-skip rather than success-tracking. If you'd rather the misconfiguration be surfaced more loudly than a per-response log line, I can add a one-time warning on the first occurrence — without ever aborting the run.

garak/generators/rest.py
xr843: Same treatment here in 04e97fad: the type-validation loop now logs the actionable error and returns `[None]` instead of raising, so a non-text match skips that generation rather than terminating the run.

garak/generators/rest.py
jmartin-tech: > Gating on "first failure with no prior success → raise" would regress that exact case

While the reporter's expectations are a good data point, they are not always _requirements_ and do not _define_ a regression.

In general it make sense to continue when the error that was reported occurs with some probes or prompts and not others. However, when the user has provided an invalid extraction value and no valid results can be obtained it does not really make sense to continue to spend inference costs for an empty report.

xr843: Agreed — you're right that the reporter's expectation isn't a requirement, and "invalid extraction value, no valid results obtainable" is a different case from "this one response is odd". Reworked in 70bb907 to draw exactly that line.

`RestGenerator` now tracks whether any generation has ever been extracted (`_extracted_a_response`). On a response whose shape `response_json_field` cannot address:

- **no extraction has succeeded yet** → `BadGeneratorException`. There is no evidence the configured field can ever address this endpoint, so the run can only produce an empty report; fail fast rather than spend inference on it.
- **at least one extraction has succeeded** → log the actionable error and return `[None]`. The field demonstrably works against this endpoint, so the mismatch is a property of this response, and the run finishes.

Both failure sites you flagged (the plain-key `KeyError`/`TypeError` path and the non-text type validation) funnel through one `_response_extraction_failed()` helper, so the two stay consistent. The exception carries the same actionable text the log line did — naming `response_json_field` and the offending type/shape.

Tests updated accordingly: three cases assert the abort on a fresh generator (dict via plain key, dict via JSONPath, missing key), and two assert the skip-and-continue once a prior call has extracted successfully.

One thing I deliberately left alone: the pre-existing "JSONPath yielded nothing" branch just below still returns `[None]` unconditionally. By the same argument it's also an un-addressable field and could route through the helper — but it's existing behaviour outside this fix, so I didn't want to change it uninvited. Happy to fold it in if you'd like.

jmartin-tech (approved):
Sanity testing with valid rest endpoint passed. LGTM

Issue #804, opened 2024-07-30
probes that fail to elicit valid generator output should result in SKIP


leondz:
## Summary

Probes that don't get a valid generator output (but do expect one) are managing to be evaluated as something other than SKIP

## Description

The following shows a `nim` case where there's no generator output (generator does not support empty prompts). Questions:

* Should the generator be returning `None` here?
* What is happening that the evaluator gets to `PASS`?
* Should the 400 be hidden from CLI output?

## Expected behaviour:

evaluator returns `SKIP` not `PASS`

## Sample output

```
probes.realtoxicityprompts.RTPBlank:   0%|                                                                               | 0/5 [00:00<?, ?it/s]ERROR:root:Error code: 400 - {'type': 'urn:inference-service:problem-details:bad-request', 'title': 'Bad Request', 'status': 400, 'detail': "[{'type': 'string_type', 'loc': ('body', 'messages', 'function-after[filter_tag_characters(), str]'), 'msg': 'Input should be a valid string', 'input': [{'role': 'user', 'content': ''}]}, {'type': 'value_error', 'loc': ('body', 'messages', 'list[dict[str,function-after[filter_tag_characters(), str]]]', 0, 'content'), 'msg': 'Value error, Prompt must be non-empty', 'input': '', 'ctx': {'error': ValueError('Prompt must be non-empty')}}]", 'instance': '/v2/nvcf/pexec/functions/14919f3d-8dc9-4d19-a6da-6dbf99360156', 'requestId': '003ba4f4-5bd5-482f-879b-3e03423ca699'}
Traceback (most recent call last):
  File "/home/lderczynski/garak/garak/generators/openai.py", line 221, in _call_model
    response = self.generator.create(**create_args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lderczynski/anaconda3/envs/garak/lib/python3.12/site-packages/openai/_utils/_utils.py", line 277, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/lderczynski/anaconda3/envs/garak/lib/python3.12/site-packages/openai/resources/chat/completions.py", line 606, in create
    return self._post(
           ^^^^^^^^^^^
  File "/home/lderczynski/anaconda3/envs/garak/lib/python3.12/site-packages/openai/_base_client.py", line 1240, in post
[...]

PR #1901, opened 2026-07-01
fix: repair dead blank-prompt guard in NVCF 400 handling

anxkhn:
`NvcfChat._call_model` has a dedicated branch meant to skip blank prompts that
the endpoint rejects with HTTP 400:

```python
if response.status_code == 400 and prompt == "":
    # error messages for refusing a blank prompt are fragile ...
    return [None]
```

`prompt` is now a `Conversation` (`garak.attempt.Conversation`, a `@dataclass`),
not a `str`. Its auto-generated `__eq__` only compares equal to another
`Conversation`, so `prompt == ""` is **always `False`** and this guard is dead
code. Blank prompts rejected with a 400 never hit their intended branch; they
only get skipped by coincidence via the generic fall-through at the bottom of the
`4xx/5xx` block (`logging.warning("nvcf : skipping this prompt")`).

This is a regression from the `str` -> `Conversation` generator refactor: the
comparison was correct when `prompt` was a plain string. The last functional
change to this line is commit `d417303` ("Set generator `_call_model()` ... type
hints", #694), which is exactly that refactor.

### The fix

Compare the prompt's last message text instead, reproducing the original intent:

```python
if response.status_code == 400 and prompt.last_message().text in ("", None):
```

`Conversation.last_message()` returns the last turn's `Message`; `Message.text`
is `""` for an empty prompt and `None` when unset, so both are caught. The fix is
inherited by `NvcfCompletion` (a subclass of `NvcfChat`). Scope is one line; no
base classes or `attempt` touched, no new dependencies.

### Tests

Added two parametrized regression tests over both NVCF generators
(`NvcfChat`, `NvcfCompletion`) in the existing `tests/generators/test_nvcf.py`,
stubbing `requests.Session.post` with a fake 400 response:

- `test_blank_prompt_400_skipped` - a blank prompt + 400 now returns `[None]` via
  the dedicated guard and does **not** log `"skipping this prompt"`.
- `test_nonblank_prompt_400_not_caught_by_blank_guard` - a non-blank prompt + 400
  still falls through to the generic skip path (asserts it **does** log
  `"skipping this prompt"`), so the guard is not over-broad.

On `main` (before the fix) `test_blank_prompt_400_skipped` fails for both classes
and the non-blank test passes, i.e. the tests pin precisely the guard behaviour.

## Verification

- [x] No supporting configuration needed (uses placeholder generator config in
      the tests; no live endpoint).
- [x] `python -m pytest tests/generators/test_nvcf.py -q` -> **10 passed**
- [x] `python -m pytest tests/generators/test_nvcf.py tests/test_attempt.py tests/generators/test_generators.py -q` -> **150 passed** (no existing tests broken)
- [x] `python -m black --check garak/generators/nvcf.py tests/generators/test_nvcf.py` -> clean
- [x] **Verify** blank-prompt 400 hits the dedicated guard (new test, was failing on `main`).
- [x] **Verify** non-blank 400 is unaffected and still uses the generic skip path.
- [x] **Document** intent is preserved; the existing inline comment on the guard
      still describes the behaviour, so no doc change is needed.

### Not a duplicate

- No open PR modifies `garak/generators/nvcf.py`. PRs surfaced by an `nvcf`
  search (#1673 Cohere generator, #1757 PDF-injection probe) do not touch
  `nvcf.py`.
- No matching issue. Two adjacent-but-distinct issues exist:
  - **#642** is about NVCF **9xx** return codes; this fix is specifically the
    **400** blank-prompt guard.
  - **#804** is the broader evaluator question of whether failed-output prompts
    should be SKIP vs PASS; this fix is a narrow generator-side correctness
    repair and does not change SKIP/PASS semantics.

### AI assistance disclosure

This change was developed with AI assistance. I (the submitter) reviewed every
changed line, understand the bug and the fix end-to-end, and ran the tests above
locally. Attribution is in the commit trailer (`Co-authored-by: Claude`), and the
commit is DCO signed-off.

### DIFF
--- a/garak/generators/nvcf.py
+++ b/garak/generators/nvcf.py
@@ -126,7 +126,7 @@ class NvcfChat(Generator):
         if 400 <= response.status_code < 600:
             logging.warning("nvcf : returned error code %s", response.status_code)
             logging.warning("nvcf : returned error body %s", response.content)
-            if response.status_code == 400 and prompt == "":
+            if response.status_code == 400 and prompt.last_message().text in ("", None):
                 # error messages for refusing a blank prompt are fragile and include multi-level wrapped JSON, so this catch is a little broad
                 return [None]
             if response.status_code == 404 and self.stop_on_404:
--- a/tests/generators/test_nvcf.py
+++ b/tests/generators/test_nvcf.py
@@ -1,7 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import logging
+
 import pytest
+import requests
 
 from garak import _config
 from garak import _plugins
@@ -9,6 +12,13 @@ import garak.generators.base
 import garak.generators.nvcf
 
 PLUGINS = ("NvcfChat", "NvcfCompletion")
+
+
+class _FakeResponse:
+    def __init__(self, status_code, content):
+        self.status_code = status_code
+        self.content = content
+        self.headers = {}
 
 
 @pytest.mark.parametrize("klassname", PLUGINS)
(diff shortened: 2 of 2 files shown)


jmartin-tech (approved):
Nice catch, it would be helpful to have an sample failure response, however tests added model it well enough we can land as is.

PR #1903, opened 2026-07-02
feat(probes): signal when an IntentProbe is skipped for inactive intents

### LINKED ISSUE #1889: signal when an IntentProbe is skipped because its intent isn't activated
The intent guard added in #1866 can skip an IntentProbe whose intent isn't in the activated set, but nothing tells the user — the terminal shows no notice and the report carries no context. 

- Terminal should emit a clear line when a probe is skipped because its intent(s) aren't activated.
- Report/digest should record that the probe was skipped, and why, so it surfaces in the report.

### LINKED ISSUE #2026: Probe.probe() raises IndexError when the prompt set is empty (translation can produce an empty set)
## Summary

```python
...
```

## Repro

## Expected behavior

## Suggested fix direction

```python
if not prompts:
    return []
```

WatchTree-19:
# Description

An `IntentProbe` whose intents are all outside the active set (e.g. filtered out by an `intent:` run.spec selector) produces no prompts and returns early from `probe()`. Before this change that path emitted only a `logging.debug`, so the skip was silent: the terminal showed no notice and the report carried no context for the missing probe (#1889).

This adds `_notify_skipped_no_active_intents()`, called on that early return, which:
- emits a clear terminal notice (`⚠️  probe ... skipped: none of its intents are in the active set ...`), matching the existing `warn_unconsumed_intents` convention, and
- writes a `probe_skipped` report entry (`reason: no_active_intents`, plus the intents considered) so the skip surfaces in the report/digest.

Targets `feature/technique_intent` since the intent run-spec selector (from #1866) lives on that branch, not `main`.

Fixes #1889.

## Type of change
- [x] Bug fix / small feature (non-breaking)

## Verification

Added `tests/cas/test_intent_skip_notice.py`, which verifies the skipped probe (a) writes exactly one `probe_skipped` report entry with `reason: no_active_intents`, and (b) prints the terminal notice. Existing `tests/cas/test_cas_intentprobe.py` and `tests/cas/test_intent_run_spec.py` still pass (20 tests). `ruff`/`black` clean on the changed lines.

```
pytest tests/cas/test_intent_skip_notice.py tests/cas/test_cas_intentprobe.py tests/cas/test_intent_run_spec.py
```

- [x] DCO sign-off included
- [x] Unit tests added

Closes #1889.
Closes #2026 (same `prompts[0]`-on-empty root cause; the base `Probe.probe` guard covers it too).

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -388,6 +388,14 @@ class Probe(Configurable):
 
         self.generator = generator
 
+        # Nothing to send: e.g. an IntentProbe whose intents are all outside the
+        # active set, or any probe left with no prompts. Log and return empty
+        # rather than indexing prompts[0] below; the harness treats an empty
+        # result as a skip (see #1889).
+        if not self.prompts:
+            logging.info("probe %s has no prompts to send; skipping", self.probename)
+            return []
+
         # build list of attempts
         attempts_todo: Iterable[garak.attempt.Attempt] = []
         prompts = copy.deepcopy(
@@ -943,11 +951,3 @@ class IntentProbe(Probe):
             prompts = self._prompts_from_stub(stub)
             self.prompts.extend(prompts)
             self.prompt_intents.extend([self.stub_intents[i]] * len(prompts))
-
-    def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
-        if not self.prompts:
-            # an empty active-intent set (run.spec intent: filtered to nothing)
-            # yields no prompts; no-op so the rest of the run proceeds (3A)
-            logging.debug("%s has no active intents; no prompts to send", self.probename)
-            return []
-        return super().probe(generator)
--- a/garak/harnesses/base.py
+++ b/garak/harnesses/base.py
@@ -206,6 +206,16 @@ class Harness(Configurable):
                 attempt_results, (list, types.GeneratorType)
             ), "probing should always return an ordered iterable"
 
+            # A probe that produced no attempts (e.g. an IntentProbe with no
+            # active intents) is skipped here rather than running detectors over
+            # nothing (see #1889).
+            if isinstance(attempt_results, list) and len(attempt_results) == 0:
+                logging.info(
+                    "harness: probe %s produced no attempts; skipping",
+                    probe.probename,
+                )
+                continue
+
             if not isinstance(probe, garak.probes.base.IntentProbe):
                 for d in detectors:
                     self._run_detector(attempt_results, d)
--- a/tests/cas/test_intent_skip_notice.py
+++ b/tests/cas/test_intent_skip_notice.py
@@ -1,0 +1,48 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for #1889: a probe left with no prompts (e.g. an IntentProbe whose intents
+are all outside the active set) must be skipped cleanly - logged, not printed, and not
+written to the report - instead of crashing on ``prompts[0]`` or running silently."""
+
+import io
+import logging
+
+import garak._config
+import garak._plugins
+import garak.services.intentservice
+
+
+def _load_intentprobe():
+    garak._config.load_config()
+    garak._config.transient.intent_spec = "S"
+    garak.services.intentservice.load()
+    return garak._plugins.load_plugin("probes.base.IntentProbe")
+
+
+def test_probe_with_no_prompts_returns_empty_and_logs(caplog):
+    probe = _load_intentprobe()
+    probe.prompts = []  # simulate an empty active-intent set
+    with caplog.at_level(logging.INFO):
+        out = probe.probe(generator=None)
+    assert out == []
+    assert any("no prompts to send" in r.message for r in caplog.records)
+
+
+def test_probe_with_no_prompts_writes_nothing_to_report():
+    probe = _load_intentprobe()
+    probe.prompts = []
+    buf = io.StringIO()
+    garak._config.transient.reportfile = buf
+    garak._config.transient.report_filename = "test_1889.report.jsonl"
+    probe.probe(generator=None)
+    assert buf.getvalue() == "", "a skipped probe must not write to the report"
+
+
+def test_probe_with_no_prompts_no_stdout(capsys):
+    probe = _load_intentprobe()
+    probe.prompts = []
+    garak._config.transient.reportfile = io.StringIO()
+    garak._config.transient.report_filename = "test_1889.report.jsonl"
+    capsys.readouterr()  # discard stdout produced during loading
+    probe.probe(generator=None)
+    assert capsys.readouterr().out == "", "probes must not write to stdout"


garak/probes/base.py
```diff
@@ -947,7 +947,38 @@ def build_prompts(self):
     def probe(self, generator) -> Iterable[garak.attempt.Attempt]:
         if not self.prompts:
             # an empty active-intent set (run.spec intent: filtered to nothing)
-            # yields no prompts; no-op so the rest of the run proceeds (3A)
-            logging.debug("%s has no active intents; no prompts to send", self.probename)
+            # yields no prompts; no-op so the rest of the run proceeds (3A). Surface
+            # the skip so it is not silent on the terminal or in the report (#1889).
+            self._notify_skipped_no_active_intents()
```
leondz: why have this method here?
- seems like there should be some logging etc whenever self.prompts is empy
- a guard also belongs around L385 in Probe
- the behavior should be to SKIP at harness level
- value of this method is limited with such tightly scoped functionality
- probes don't write to stdout
- concrete probes don't write to reports

garak/probes/base.py
```diff
@@ -947,7 +947,38 @@ def build_prompts(self):
...
+        skipped because none of its intents are in the activated set. Without this
+        the skip is silent (see #1889): the terminal shows no notice and the report
+        carries no context for the missing probe."""
+        probe_name = self.probename.replace("garak.", "")
+        msg = (
+            f"probe {probe_name} skipped: none of its intents are in the active set; "
+            f"no prompts to send (check your intent: run.spec selector)"
+        )
+        logging.info(msg)
+        print(f"⚠️  {msg}")
```
leondz: no stdout from probes thanks

garak/probes/base.py
```diff
@@ -947,7 +947,38 @@ def build_prompts(self):
...
+        reportfile = getattr(_config.transient, "reportfile", None)
```
leondz: _config shouldn't be externally accessed from probes during runtime

garak/probes/base.py
```diff
@@ -947,7 +947,38 @@ def build_prompts(self):
...
+        if reportfile is not None:
+            reportfile.write(
```
leondz: probe's should not write to report about things other than attempts

leondz (changes requested):
fine catch, this situation should be handled appropriately

there is a more general case that could probably be dealt with here. for any probe that fails to produce prompts, the probe behaviour should lead to the harness skipping (via "SKIPPED") that probe. could be mediated via exception or via probe return.

suggestion:
* add a hook method in probes.base.Probe for checking probe consistency (bool)
* update harness to call this method between probe instantiation and calling of probe()
* in the probe's method for handling the error, add a check for consistency and  return value, e.g.

```
if not self.prompts: 
  return False
return True
```

* add a test for a probe having no prompts that ends up being skipped by harness
* be careful for probes that intentionally don't know all their prompts before probe() is called
* add a test that checks probes for consistency after loading them
* check that tests pass

alternatively another hook location might work, or passing by exception might work, let's see

WatchTree-19:
thanks, that all makes sense. reworked:

- dropped the IntentProbe method and moved the empty-prompts handling into base `Probe.probe` (~L385), so any probe left with no prompts is covered, not just IntentProbe. also guards the `prompts[0]` access that would otherwise IndexError.
- logging only: `logging.info` on the empty case, no stdout from the probe.
- removed the report write and the `_config` access from the probe; only attempts go to the report now.
- skip is at harness level: the harness logs and continues when a probe returns no attempts.

one judgment call: i keyed the harness skip off an empty result rather than `if not probe.prompts`, so it doesn't wrongly skip probes that build attempts dynamically (e.g. atkgen). happy to switch to a prompts check if you'd prefer the harness decides before calling `probe()`.

tests updated to cover the logging / no-stdout / no-report behaviour.

WatchTree-19:
@leondz ready for another look when you have a moment - the rework above moved the empty-prompts handling into base `Probe.probe` (logging only, no stdout, no report/_config access) and the skip happens at harness level off an empty result. CI green.

jmartin-tech:
@WatchTree-19, apologies it looks like github did not auto-retarget on `main` and it will need to be reopened.

WatchTree-19:
Reopened as requested — thanks @jmartin-tech. Base is `main`, same commit (`972f460`), no force-push. CI should re-run now.

patriciapampanelli:
Fixes #1889. Also closes #2026. Same root cause (`Probe.probe()` indexing `prompts[0]` on an empty prompt set), same guard.

WatchTree-19:
thanks @patriciapampanelli - agreed, #2026 is the same `prompts[0]`-on-empty path, and the guard i moved into base `Probe.probe` catches it for any probe that yields no prompts, not just IntentProbe. added `Closes #2026` to the description so it auto-closes on merge. the empty-prompts case is already covered in the tests. this has been ready since the rework - happy to rebase if it's drifted.

WatchTree-19:
@leondz reworked to your suggestion. You were right that the IntentProbe-specific version was too narrow.

**Hook on `Probe`.** `is_consistent() -> bool`, base check is whether there is anything to send:

```python
def is_consistent(self) -> bool:
    if not self.prompts:
        logging.info("probe %s has no prompts to send; not consistent", self.probename)
        return False
    return True
```

Living on `Probe` rather than `IntentProbe` means any probe that ends up with no prompts is covered, which was the general case you pointed at.

**Harness calls it between instantiation and `probe()`**, and does the skipping:

```python
if not probe.is_consistent():
    logging.info("probe skipped, not consistent: %s", probe.probename)
    continue
```

The previous version inferred the skip from an empty result list, which meant `probe()` had to run first. Now nothing runs.

**Removed:** the `IntentProbe.probe()` override you asked about, and the guard I had put inside `Probe.probe()`.

On the other four points from the review, all now have tests rather than just assurances: no stdout from probes, nothing written to the report, no `_config` access during the check, and one test asserting the harness decides before `probe()` is reached. There's also a test on a non-Intent probe so the generality doesn't regress. Six tests, all passing locally.

One thing I did not do: raising an exception instead of returning a bool. You offered either, and the bool keeps the harness in charge of what "skip" means, which seemed closer to what you described. Happy to switch if you'd rather it signalled by exception.

jmartin-tech (changes requested):
@WatchTree-19 the latest commit [59d5037](https://github.com/NVIDIA/garak/pull/1903/commits/59d5037287712ae122342c7fcaad1401c118e363) should likely be reverted or removed.

Not all probes are guaranteed to populate `self.prompts` prior to the call to `probe()`, some even override probe completely and may never have a `self.prompts` attribute, the previous check to validate if `probe()` produced any `Attempt` results is a more appropriate behavior.

WatchTree-19:
Thanks @jmartin-tech, you're right and I'd missed it. Pushed e9e2658.

You flagged two things: probes that don't populate `self.prompts` before `probe()`, and probes that override `probe()` and may never have the attribute. Checking prompts in `Probe.is_consistent()` breaks both, so it would have skipped probes that are perfectly fine. That is worse than the bug I was fixing.

Rather than revert wholesale, I moved the check to where the premise actually holds:

- **`Probe.is_consistent()` now returns `True` and inspects nothing.** No assumption that prompts exist or are populated.
- **`IntentProbe.is_consistent()` carries the prompts check.** `IntentProbe.__init__` runs `_populate_intents`, `_populate_stubs` and `build_prompts`, so its prompts really are settled before `probe()`. That is the #1889 case: every intent falls outside the active set, nothing to send.
- **Restored the harness's post-probe empty-attempts skip**, which is the behaviour you pointed at, as the general net for probes whose emptiness is only knowable afterwards.

I went this way rather than a plain revert because @leondz specifically asked for the hook, and in the same review warned to "be careful for probes that intentionally don't know all their prompts before `probe()` is called". That caveat is exactly what you caught me on. This keeps his hook and honours the caveat, with the attempt-based check underneath.

If you'd still rather just drop the hook entirely and keep only the post-probe check, say so and I'll strip it, no argument.

Tests: the base-probe case now asserts `is_consistent()` is `True`, and there's a new one for a probe that overrides `probe()` and never sets `self.prompts`. `tests/harnesses` and `tests/cas` are green, 252 passed 1 skipped.

garak/probes/base.py
```diff
@@ -865,6 +880,21 @@ def __init__(self, config_root=_config):
...
+        active set, so there is nothing to send and the run should skip it
+        rather than report a probe that did nothing.
+        """
+        if not self.prompts:
+            logging.info(
+                "probe %s has no prompts to send; not consistent", self.probename
+            )
+            return False
+        return True
+
```
jmartin-tech: This method is simply not needed the canonical check is simply that no `Attempts` were emitted from the  call to `probe()`.

There is no reason to delineate between probe types for the goal of denoting that no detection needs to be performed.  If an intent probe results in no active intents that can be logged during `__init__` an there is no significant value in not calling the `probe` method on that probe as it will simply return having preformed no work.

jmartin-tech (changes requested):
This being overcomplicated commit 972f460 was sufficient and review testing was in progress. The `is_consistent()` method is not needed to address the original goal, adds complexity, and introduces multiple decision points that cloud the reasoning and control flow.

Either force push back to that commit or rebase on main keeping only that change.

WatchTree-19:
Done, force-pushed back to 972f460. is_consistent() is gone and the harness is back to skipping on an empty attempt list.

You're right that it was overcomplicated. I added the hook because @leondz asked for it in the first review, and when you pointed out the prompts problem I tried to preserve the hook while fixing its premise, rather than stepping back and asking whether the hook was earning its place at all. It wasn't. The canonical check is just whether probe() emitted any attempts, and that covers every probe without needing to know how a given probe builds its work.

The branch is now the three files from 972f460: the harness skip, the guard in Probe.probe(), and the test. Sorry for the detour.

WatchTree-19:
@jmartin-tech @leondz — mapping the open threads against what's actually on the branch, since the conversation has outrun the diff a bit. Branch is `972f460`, three files.

**@jmartin-tech (31 Aug), "this method is simply not needed, the canonical check is simply that no `Attempts` were emitted"**

Agreed, and that's what's there now. The `IntentProbe.probe()` override is deleted, and the skip decision lives in the harness on the emitted attempts:

```python
if isinstance(attempt_results, list) and len(attempt_results) == 0:
    logging.info("harness: probe %s produced no attempts; skipping", probe.probename)
    continue
```

That covers every probe without the harness needing to know how a given probe builds its work, which was your point.

**One thing I want to check rather than assume.** The branch also keeps a small guard at the top of `Probe.probe()`:

```python
if not self.prompts:
    logging.info("probe %s has no prompts to send; skipping", self.probename)
    return []
```

That one isn't doing the skip decision — the harness does that. It exists because the code below it does `prompts[0]`, so an empty prompt list raises `IndexError` before anything reaches the harness at all.

If you read that as the same method you asked me to remove, say so and I'll drop it and instead make the indexing tolerate an empty list. If you're happy with it as a crash guard, it can stay as is. I'd rather ask than guess and take another round trip.

**@leondz (2 Jul)** — I think all four are addressed by the current shape, but the threads are still open so flagging explicitly:

- *"no stdout from probes thanks"* — everything goes through `logging.info`. `test_probe_with_no_prompts_no_stdout` asserts `capsys.readouterr().out == ""`.
- *"`_config` shouldn't be externally accessed from probes during runtime"* — no `_config` access remains in the probe path.
- *"probes should not write to report about things other than attempts"* — `test_probe_with_no_prompts_writes_nothing_to_report` asserts the report buffer stays empty for a skipped probe.
- *"why have this method here?"* — that method is gone.

Happy to be told any of those still misses the point.

**CI:** the two red Linux jobs are `cancelled` at 24 and 25 minutes rather than failed, so they look like runner timeouts rather than test failures. Glad to rebase and re-run if that helps.

Thanks both, and sorry this one has taken a while to converge.

WatchTree-19:
@jmartin-tech this is done and I should have said so at the time, sorry for the silence.

The branch is back to `972f460` as a single commit, with `is_consistent()` gone entirely. What is left is the general case you asked for: the base `Probe.probe` guards an empty `self.prompts` with a `logging.info` and returns `[]`, which also removes a latent `prompts[0]` IndexError, the `IntentProbe.probe` override is deleted because the base now covers it, and the harness skips and logs when a probe returns no attempts. No stdout from probes, no `_config` access at runtime, no report writes.

You were right that the canonical check is just that no attempts were emitted. Dropping the hook removed the type-specific knowledge and the `self.prompts` problem with it.

Checks are green and there are no conflicts, so this is ready whenever you are.

jmartin-tech (approved):

Issue #570, opened 2024-03-22
probe: token-length side-channel


leondz:
https://blog.cloudflare.com/ai-side-channel-attack-mitigated

https://cdn.arstechnica.net/wp-content/uploads/2024/03/LLM-Side-Channel.pdf

leondz:
easier with #1254 now in place

neerazz:
Hi @leondz — I’d like to take this issue and open a draft PR from @neerazz. I saw your note that it became easier after #1254; I’ll align the probe design to the current conversation support path and keep scope tight for first merge. If there are specific detector/tag expectations you want up front, I can incorporate them before opening the PR.

leondz:
OK, good luck!

PR #1909, opened 2026-07-03
probe: token-length side-channel

neerazz:
Implements #570 (probe: token-length side-channel).

### What this adds
- `garak/probes/side_channel.py` — `TokenLength` probe. Elicits short, structured, sensitive-shaped answers whose length signature is the interesting part.
- `garak/detectors/side_channel.py` — `TokenLengthLeak` detector. Scores each output by the coefficient of variation of token-like output segment lengths, normalised to 0.0–1.0. A response with near-uniform token lengths scores low; a response with highly varied lengths scores high. This follows the observation from Weiss et al. 2024 ("What Was Your Prompt?", arXiv:2403.09751) and the Cloudflare write-up that the entropy of token-length sequences is what an eavesdropper can recover from encrypted packet sizes.
- Unit tests for probe and detector.
- Docs entries for the new probe/detector.

The probe ships `active = False` at `INFORMATIONAL` tier so it is available for targeted use without changing default probe behavior.

### Implementation notes
- The detector is dependency-free and uses a tokenizer approximation because garak's generator interface receives completed text, not packet-size / streaming-token metadata.
- This keeps the first version narrow and mergeable. If maintainers prefer a generator-layer streaming/packet hook later, this probe can use that richer signal without changing the user-facing probe shape.
- Current tags: `owasp:llm06`, `avid-effect:security:S0301`, `quality:Security:Confidentiality`.

### Verification
- Local focused tests: `python3 -m pytest tests/detectors/test_detectors_side_channel.py tests/probes/test_probes_side_channel.py -q`
- Result: `8 passed in 0.59s`
- GitHub DCO: passing

### DIFF
--- a/docs/source/detectors/side_channel.rst
+++ b/docs/source/detectors/side_channel.rst
@@ -1,0 +1,7 @@
+garak.detectors.side_channel
+============================
+
+.. automodule:: garak.detectors.side_channel
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/docs/source/index_detectors.rst
+++ b/docs/source/index_detectors.rst
@@ -39,6 +39,7 @@ For detailed information on detector metrics and evaluation, see :doc:`../detect
    detectors/productkey
    detectors/propile
    detectors/shields
+   detectors/side_channel
    detectors/snowball
    detectors/sysprompt_extraction
    detectors/unsafe_content
--- a/docs/source/index_probes.rst
+++ b/docs/source/index_probes.rst
@@ -45,6 +45,7 @@ For a guide to writing probes, see :doc:`extending.probe`.
    probes/propile
    probes/realtoxicityprompts
    probes/sata
+   probes/side_channel
    probes/snowball
    probes/smuggling
    probes/suffix
--- a/docs/source/probes/side_channel.rst
+++ b/docs/source/probes/side_channel.rst
@@ -1,0 +1,9 @@
+garak.probes.side_channel
+=========================
+
+.. automodule:: garak.probes.side_channel
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+   .. show-asr::
--- a/garak/detectors/side_channel.py
+++ b/garak/detectors/side_channel.py
@@ -1,0 +1,79 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Side-channel detectors
+
+Detectors that estimate how much a response would leak through observable
+side channels of the serving path, rather than through the response text
+itself.
+"""
+
+import re
+import statistics
+from typing import Iterable
+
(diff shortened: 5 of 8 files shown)


neerazz:
Ready for maintainer review.

I tightened the PR description so this is no longer framed as exploratory/draft. The branch is pushed and DCO is green. I also re-ran the focused local tests:

```bash
python3 -m pytest tests/detectors/test_detectors_side_channel.py tests/probes/test_probes_side_channel.py -q
# 8 passed in 0.59s
```

The only thing I can see blocking merge now is maintainer review / any repository-side CI that maintainers want to trigger on this fork PR.

neerazz:
Checking in on this one — happy to do whatever unblocks it.

Two things have changed since July that are worth flagging:

**1. The branch is 433 commits behind main.** It still reports MERGEABLE / CLEAN, but that gap is large enough that a review now would be reviewing against a very different tree. I'd rather rebase onto current main and re-run the tests than have you spend review time on a stale diff. Say the word and I'll push the rebase — or if you'd prefer I just close and re-open fresh against main, that works too.

**2. The scope question is still open and it's the one that matters.** The detector uses a tokenizer approximation because garak's generator interface hands back completed text, not streaming/packet metadata. That was a deliberate call to keep this narrow and mergeable, but it's also the thing most likely to make you hesitate — an approximation of a side channel is not the side channel.

So: is the informational-tier, text-approximation version something garak wants at all, or does this need to wait for a generator-layer streaming hook before it earns a place? A "not in this form" is genuinely useful to me and I won't push back on it. Issue #570 is still open and unassigned, so I'd rather know now than leave it sitting.

Current state: `active = False`, INFORMATIONAL tier, 8 focused tests passing, DCO green, 16 checks green.

neerazz:
Rebased onto current `main` — the 433-commit gap is now **0**.

Force-pushed `722fe48`. What I verified locally before pushing:

- **Rebase was clean** — zero conflicts across the whole gap.
- **Not superseded** — no `side_channel` probe or detector exists on current main, so this doesn't collide with anything that landed since July.
- **API still matches** — `garak.probes.Tier.INFORMATIONAL` and the `Detector.detect(attempt) -> Iterable[float | None]` signature are unchanged on main.
- **Tests green on the rebased tree**: `8 passed in 0.13s` (`tests/detectors/test_detectors_side_channel.py`, `tests/probes/test_probes_side_channel.py`).
- **Loads in the real plugin loader**, not just imports: `load_plugin('probes.side_channel.TokenLength')` → active `False`, tier `3` (INFORMATIONAL), 8 prompts; `detectors.side_channel.TokenLengthLeak` resolves.

DCO is green. The five workflows (Linux/MacOS/Windows pytest, docs build, pip install) are sitting at `action_required` — they need a maintainer to approve the run, which I can't do from this side.

The scope question from my last comment still stands and is the one that actually decides this: the detector uses a tokenizer approximation because the generator interface hands back completed text rather than streaming/packet metadata. If that approximation isn't what garak wants for a side channel, I'd rather hear "not in this form" than keep the PR warm — issue #570 is still open and unassigned either way.

PR #1910, opened 2026-07-05
rest: coerce non-string JSON response values to text

### LINKED ISSUE #1888: AttributeError: 'dict' object has no attribute 'lower' on REST endpoint after any probe finish

cognis-digital:
## Summary
Fixes #1888. A `response_json_field` / JSONPath can resolve to a **dict, list, number, or bool** rather than a string — observed with some Azure OpenAI REST response shapes. The extracted value flowed unmodified into `Message()`, and a downstream detector then raised `AttributeError: 'dict' object has no attribute 'lower'` — only *after* every subtest in the probe had run, so the whole run was wasted.

## Fix
- Add `_coerce_response_text()`: passes strings/None through, serializes dict/list to JSON, and `str()`s other scalars — so the value handed to `Message()` is always text.
- In the single-match JSONPath branch, keep a non-str/non-list value (e.g. a dict) instead of silently leaving `response = [None]` (which discarded a real response).

## Tests
Adds `test_json_rest_dict_value_is_coerced` and `test_json_rest_jsonpath_dict_value_is_coerced` mirroring the existing `requests_mock` REST tests: a response field that resolves to a dict now yields a string `Message` (what detectors call `.lower()` on) instead of crashing.

Minimal and backwards-compatible: string and list responses are unchanged.

### DIFF
--- a/garak/generators/rest.py
+++ b/garak/generators/rest.py
@@ -28,6 +28,22 @@ from garak.exception import (
 from garak.generators.base import Generator
 
 
+def _coerce_response_text(value):
+    """Coerce a value extracted from a JSON response into text for a Message.
+
+    ``response_json_field`` / a JSONPath can resolve to a dict, list, number, or
+    bool rather than a string (observed with some Azure OpenAI response shapes).
+    Downstream detectors call string methods such as ``.lower()`` on the response,
+    so a non-string value must be normalised here rather than raising an
+    ``AttributeError`` after every subtest has run.
+    """
+    if value is None or isinstance(value, str):
+        return value
+    if isinstance(value, (dict, list)):
+        return json.dumps(value, ensure_ascii=False)
+    return str(value)
+
+
 class RestGenerator(Generator):
     """Generic API interface for REST models
 
@@ -394,6 +410,11 @@ class RestGenerator(Generator):
                     response = [response_value]
                 elif isinstance(response_value, list):
                     response = response_value
+                else:
+                    # dict / number / bool — a valid JSONPath match that isn't a
+                    # string (e.g. some Azure OpenAI response shapes). Keep it and
+                    # coerce to text below rather than silently dropping to [None].
+                    response = [response_value]
             elif len(responses) > 1:
                 response = [r.value for r in responses]
             else:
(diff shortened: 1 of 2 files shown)


garak/generators/rest.py
```diff
@@ -28,6 +28,22 @@
...
+    ``response_json_field`` / a JSONPath can resolve to a dict, list, number, or
+    bool rather than a string (observed with some Azure OpenAI response shapes).
+    Downstream detectors call string methods such as ``.lower()`` on the response,
+    so a non-string value must be normalised here rather than raising an
+    ``AttributeError`` after every subtest has run.
+    """
+    if value is None or isinstance(value, str):
+        return value
+    if isinstance(value, (dict, list)):
+        return json.dumps(value, ensure_ascii=False)
```
jmartin-tech, line 43: This could raise an exception that will not be handled well by the run if the content is not _valid_ json.

jmartin-tech (commented):
How does this offering compare with #1892, I see possible tradeoffs to each.

chuenchen309:
Traced this one through — the fix looks correct and closes a real gap. Two parts working together:

1. The new `else` in the single-match branch: before this, a JSONPath that resolved to a dict/number/bool matched neither the `str` nor the `list` case, so `response` was left unset for that path. Now it's kept as `[response_value]` instead of silently becoming `[None]`. Good catch — that's the "valid match that isn't a string" case.
2. `_coerce_response_text` then normalises it: `dict`/`list` → `json.dumps(..., ensure_ascii=False)` (valid JSON text, and `ensure_ascii=False` preserves non-ASCII), scalars → `str`, `None`/`str` untouched. Since downstream detectors call string methods on the response, coercing here rather than after every subtest has run is the right place.

`json` is already imported at the top, so no new dependency.

One tiny consistency nit, no practical impact: booleans go through the `str()` branch → `"True"`/`"False"` (Python) rather than `"true"`/`"false"` (JSON). Given `StringDetector` defaults to `case_sensitive=False` it washes out in matching, so purely cosmetic — mentioning only in case you'd prefer scalars to be JSON-consistent too. The fix itself is right; thanks for handling the non-string shapes.

jmartin-tech:
Closing in favor of #1892

PR #1916, opened 2026-07-06
Handle non-list LangServe /invoke output in langchain_serve generator

### LINKED ISSUE #1915: langchain_serve generator returns only the first character of a string LangServe response
## summary

`LangChainServeLLMGenerator._call_model` does `Message(response_data.get("output")[0])`, which assumes the `/invoke` response wraps the output in a list. LangServe returns the runnable's output directly under `output`:

- string / LLM chain: `{"output": "the text", "metadata": {...}}`, so `output[0]` returns just `"t"` (the first character).
- chat model: `{"output": {"content": "...", "type": "ai"}}`, so `output[0]` raises `KeyError`, which is caught and the response is silently dropped (`[None]`).

only a `{"output": ["..."]}` shape actually works, which isn't what LangServe emits. the existing test uses that fabricated list shape, so the bug isn't caught.

## repro

with `LANGCHAIN_SERVE_URI` set and the endpoint returning `{"output": "Generated text"}`, the generator yields `Message("G")` instead of `Message("Generated text")`. a chat-model dict response yields `[None]`.

## fix

handle the output by type: a string directly, a dict via its `content`, a list via its first element (backward compat). happy to send a PR, have one ready.

WatchTree-19:
# Description

`LangChainServeLLMGenerator._call_model` did `Message(response_data.get("output")[0])`, assuming LangServe's `/invoke` wraps the output in a list. LangServe returns the runnable's output directly under `output`:

- string / LLM chain: `{"output": "the text", ...}` -> `output[0]` returns only `"t"` (first character)
- chat model: `{"output": {"content": "...", "type": "ai"}}` -> `output[0]` raises `KeyError`, caught, so the response is silently dropped (`[None]`)

Only a `{"output": ["..."]}` shape worked, which LangServe doesn't emit; the existing test used that fabricated shape so the bug went unnoticed.

Now handles the output by type: a string directly, a dict via its `content`, a list via its first element (backward compatible), guarding empty/None.

Fixes #1915.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## Verification
Added `test_langchain_serve_string_output` and `test_langchain_serve_message_dict_output` (real LangServe shapes); the existing list-shape test still passes. `tests/generators/test_langchain_serve.py` passes (7). ruff clean on the changed lines (the two `== True/False` warnings are pre-existing in `test_validate_uri`).

- [x] DCO sign-off included
- [x] Unit tests added

### DIFF
--- a/garak/generators/langchain_serve.py
+++ b/garak/generators/langchain_serve.py
@@ -95,7 +95,19 @@ class LangChainServeLLMGenerator(Generator):
             if "output" not in response_data:
                 logging.error(f"No output found in response: {response_data}")
                 return [None]
-            return [Message(response_data.get("output")[0])]
+            output = response_data["output"]
+            # LangServe's /invoke returns the runnable's output directly: a
+            # string for string/LLM chains, or a serialised message dict (e.g.
+            # an AIMessage) for chat models. Some deployments wrap it in a
+            # single-element list. Handle each explicitly so a plain string
+            # output is not truncated to its first character.
+            if isinstance(output, list):
+                output = output[0] if output else None
+            if isinstance(output, dict):
+                output = output.get("content", output)
+            if output is None:
+                return [None]
+            return [Message(output if isinstance(output, str) else str(output))]
         except json.JSONDecodeError as e:
             logging.error(
                 f"Failed to decode JSON from response: {response.text}, error: {e}"
--- a/garak/generators/langchain_serve.py
+++ b/garak/generators/langchain_serve.py
@@ -63,6 +63,27 @@ class LangChainServeLLMGenerator(Generator):
             logging.error(f"URL parsing error: {e}")
             return False
 
+    @staticmethod
+    def _extract_output_text(output) -> Union[str, None]:
+        """Extract the response text from a LangServe ``/invoke`` ``output`` value.
+
+        LangServe returns the runnable's output directly under ``output``: a
+        ``str`` for string/LLM chains, or a serialised message for chat models.
+        Messages are serialised either flat (``{"content": ...}``) or via
+        LangChain's constructor form (``{"kwargs": {"content": ...}}``). Any
+        other shape is unexpected and returns ``None`` so the caller records a
+        miss instead of emitting a stringified object.
+        """
+        if isinstance(output, str):
+            return output
+        if isinstance(output, dict):
+            content = output.get("content")
+            if content is None and isinstance(output.get("kwargs"), dict):
+                content = output["kwargs"].get("content")
+            if isinstance(content, str):
+                return content
+        return None
+
     def _call_model(
         self, prompt: Conversation, generations_this_call: int = -1
     ) -> List[Union[Message, None]]:
@@ -95,19 +116,14 @@ class LangChainServeLLMGenerator(Generator):
             if "output" not in response_data:
                 logging.error(f"No output found in response: {response_data}")
                 return [None]
-            output = response_data["output"]
-            # LangServe's /invoke returns the runnable's output directly: a
-            # string for string/LLM chains, or a serialised message dict (e.g.
-            # an AIMessage) for chat models. Some deployments wrap it in a
-            # single-element list. Handle each explicitly so a plain string
-            # output is not truncated to its first character.
-            if isinstance(output, list):
-                output = output[0] if output else None
-            if isinstance(output, dict):
-                output = output.get("content", output)
-            if output is None:
+            text = self._extract_output_text(response_data["output"])
+            if text is None:
+                logging.error(
+                    f"Unexpected LangServe output shape for prompt {prompt_text}: "
+                    f"{response_data['output']!r}"
+                )
                 return [None]
-            return [Message(output if isinstance(output, str) else str(output))]
+            return [Message(text)]
         except json.JSONDecodeError as e:
             logging.error(
                 f"Failed to decode JSON from response: {response.text}, error: {e}"
--- a/tests/generators/test_langchain_serve.py
+++ b/tests/generators/test_langchain_serve.py
@@ -4,12 +4,22 @@ import pytest
 from garak.attempt import Message, Turn, Conversation
 from garak.generators.langchain_serve import LangChainServeLLMGenerator
 
+ENDPOINT = "http://127.0.0.1:8000/invoke?config_hash=default"
+
 
 @pytest.fixture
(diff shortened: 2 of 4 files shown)


garak/generators/langchain_serve.py
```diff
@@ -95,7 +95,19 @@ def _call_model(
...
+            output = response_data["output"]
+            # LangServe's /invoke returns the runnable's output directly: a
+            # string for string/LLM chains, or a serialised message dict (e.g.
+            # an AIMessage) for chat models. Some deployments wrap it in a
+            # single-element list. Handle each explicitly so a plain string
+            # output is not truncated to its first character.
+            if isinstance(output, list):
+                output = output[0] if output else None
+            if isinstance(output, dict):
+                output = output.get("content", output)
```
jmartin-tech: If this was a `dict` why would you keep the original value?  This seems like it should be parsed into some other ETL format to get down to a string value.

garak/generators/langchain_serve.py
```diff
@@ -95,7 +95,19 @@ def _call_model(
...
                 return [None]
-            return [Message(response_data.get("output")[0])]
+            output = response_data["output"]
+            # LangServe's /invoke returns the runnable's output directly: a
+            # string for string/LLM chains, or a serialised message dict (e.g.
+            # an AIMessage) for chat models. Some deployments wrap it in a
+            # single-element list. Handle each explicitly so a plain string
+            # output is not truncated to its first character.
+            if isinstance(output, list):
+                output = output[0] if output else None
```
jmartin-tech: Whould this ever have more than one entry?

I so should this be:
```
output = "".join(output)
```

garak/generators/langchain_serve.py
```diff
@@ -95,7 +95,19 @@ def _call_model(
...
+            if isinstance(output, dict):
+                output = output.get("content", output)
+            if output is None:
+                return [None]
+            return [Message(output if isinstance(output, str) else str(output))]
```
jmartin-tech: If this was a `dict`

```
{ "error": "err val"}
```

would become a string `{ "error": "err val"}` which I don't think would be correct.

tests/generators/test_langchain_serve.py
```diff
@@ -38,6 +37,35 @@ def test_langchain_serve_generation(requests_mock):
...
+    requests_mock.post(
+        "http://127.0.0.1:8000/invoke?config_hash=default",
+        json={"output": {"content": "Generated text", "type": "ai"}},
+    )
+    generator = LangChainServeLLMGenerator()
+    conv = Conversation([Turn("user", Message("Hello LangChain!"))])
+    output = generator._call_model(conv)
+    assert output == [Message("Generated text")]
+
+
```
jmartin-tech, line 81: These tests only look a the new happy path, we should more strictly define expectations with the response does not fit these formats.

What does the API here enforce will occur here?

Consider migrating the mock responses used these test to the `tests/_assets` path similar to `test/_assets/generators/openai.json`. The fixture files can be used to represent valid expected response bodies from the target service with various HTTP response codes.

chuenchen309:
Read through this against #1915 — the fix looks correct to me and covers all three shapes the issue calls out. A quick trace for confidence, plus one minor edge:

- **string** (`{"output": "text"}`) → not a list, not a dict → `Message("text")` ✅ (fixes the first-character truncation)
- **chat dict** (`{"output": {"content": "...", "type": "ai"}}`) → `output.get("content", output)` → the content string ✅ (fixes the silent `[None]` drop)
- **list** (`{"output": ["text"]}`) → `output[0]` → `"text"` ✅ (backward compatible with the old shape / the existing test)
- **empty list** (`{"output": []}`) → `output[0] if output else None` → `None` → `[None]` ✅ (no `IndexError`)

Nice touch that the two `if`s are sequential rather than `elif` — a list-wrapped message dict (`{"output": [{"content": "x", ...}]}`) is unwrapped to the dict and then to its content, so that shape works too.

One tiny edge, non-blocking: a dict without a `content` key falls to `output.get("content", output)` → the dict itself → `str(output)`, i.e. the whole dict repr becomes the message text. That's a reasonable graceful fallback (better than dropping/crashing), just noting it in case you'd rather return `[None]` for an unrecognised dict shape. Either way the core fix is right — thanks for picking this up.

jmartin-tech (changes requested):

WatchTree-19:
thanks for the review, all addressed in the latest push.

reworked the parsing into `_extract_output_text`, which pulls a string out of the `/invoke` output explicitly:
- `str` output -> used as-is (the original truncation bug)
- chat message dict -> `content`, handling both the flat `{"content": ...}` and langchain's constructor form `{"kwargs": {"content": ...}}`
- anything else -> `None`, and the caller logs the shape and records a miss

so the `str(output)` fallback is gone: a `{"error": "err val"}` mapping now returns `[None]` rather than the literal string `"{'error': 'err val'}"`. added a test that asserts exactly this.

on the list case: per #1915, LangServe `/invoke` returns the runnable output directly and doesn't wrap it in a list, so i dropped the single-element unwrapping (that shape was fabricated by the old test). a list is now treated as an unexpected shape -> `[None]`. if you've seen deployments that legitimately emit a list there, happy to add a `"".join(...)` path instead.

on tests: migrated the mock bodies to `tests/_assets/generators/langchain_serve.json` and expanded coverage to the message-dict, constructor-message, unexpected-dict, list, missing-output, and client/server-error shapes (not just the happy path).

chuenchen309:
Read through the rework — this addresses the points cleanly.

- The core fix is right: the old `output[0]` silently took the *first character* of a string output (`"Generated text"` → `"G"`), which was the truncation bug. `_extract_output_text` now returns the whole string, and the flat-message (`{"content": ...}`) and constructor-form (`{"kwargs": {"content": ...}}`) branches both look correct.
- Dropping the single-element list unwrap is the right call per #1915 — `/invoke` returns the runnable output directly, so the list shape was an artifact of the old test rather than real LangServe behavior. Treating it as unexpected → `[None]` with a logged shape is the safe default. Agree with holding off on a speculative `"".join(...)` path until an actual deployment is seen emitting a list.
- The moved-out fixtures + the spread across string/message-dict/constructor/unexpected-dict/list/missing-output/client-error/server-error shapes make the contract much clearer than the single happy-path test.

LGTM from my read.

jmartin-tech (approved):

PR #1946, opened 2026-07-11
Return a single Message from InferenceEndpoint._call_model

chuenchen309:
## What

`InferenceEndpoint._call_model` joins the prediction output into a single string and then returns it split by character:

```python
response = "".join(prediction.output)   # e.g. "Hello world"
...
return [Message(r) for r in response]   # -> [Message('H'), Message('e'), ...]
```

Iterating over the string yields one `Message` per character. The base class `ReplicateGenerator._call_model` does it correctly:

```python
return [Message("".join(response_iterator))]
```

So every successful generation from a private Replicate endpoint is returned as N single-character `Message` objects instead of one. With `supports_multiple_generations = False`, garak expects a single message, so downstream this mismatches the expected result shape.

## Fix

Return `[Message(response)]`, matching the base class and the single-generation contract.

## Test

Adds `tests/generators/test_replicate.py` (there was no replicate test module): mocks the deployment/prediction so `prediction.output` joins to `"Hello world"`, and asserts `_call_model` returns a single `Message` with that full text. Fails before the change (`len == 11`), passes after. `black --check` clean.

---

### DIFF
--- a/garak/generators/replicate.py
+++ b/garak/generators/replicate.py
@@ -119,7 +119,7 @@ class InferenceEndpoint(ReplicateGenerator):
             raise IOError(
                 "Replicate endpoint didn't generate an Iterable[str]-type response. Make sure the endpoint is active."
             ) from exc
-        return [Message(r) for r in response]
+        return [Message(response)]
 
 
 DEFAULT_CLASS = "ReplicateGenerator"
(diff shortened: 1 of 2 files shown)


chuenchen309:
While cross-checking my open PRs against this repo's file paths, I found #1836 (opened 2026-06-04) already fixes this exact same bug with an identical fix, over a month before I found it independently. Closing this one in favor of the earlier PR rather than duplicate the review effort.

PR #1947, opened 2026-07-11
Fix: improve extra_params support in NeMoGuardrailsServer

chuenchen309:
`NeMoGuardrailsServer.__init__` runs:

```python
if self.extra_params and not self.extra_params.get("extra_body"):
    self.extra_params.append("extra_body")
```

`extra_params` is a **dict** (inherited from `OpenAICompatible`, default `{}`, consumed via `.items()`), so `.append` raises `AttributeError: 'dict' object has no attribute 'append'`. Any user who sets a non-empty `extra_params` without an `extra_body` key crashes generator construction:

```python
# generators.guardrails.NeMoGuardrailsServer.extra_params = {"foo": 1}
NeMoGuardrailsServer(config_root=...)  # AttributeError during __init__
```

## Why remove rather than fix the `.append`

- The block only ever **crashes** (condition true) or is **skipped** (condition false) — it has never completed successfully, so removing it changes no working behaviour.
- Guardrails config is actually delivered via the `self.extra_body` attribute set just below (and passed to the client by the base class), not via an `extra_params["extra_body"]` key.
- A naive fix of `self.extra_params["extra_body"] = {}` would be **worse**: `OpenAICompatible._call_model` spreads `extra_params` into the request *after* applying `self.extra_body`, so it would overwrite the real guardrails config with an empty dict.

## Verification

- [x] **Verify** the thing does what it should — added `test_nonempty_extra_params_does_not_crash_init`: constructing `NeMoGuardrailsServer` with `extra_params={"foo": 1}` must not raise and must leave `extra_params` intact. Fails before the change (`AttributeError`), passes after.
- [x] **Verify** the thing does not do what it should not — the existing rail-selection tests still pass, confirming the removed block was vestigial and no working behaviour changed.
- [x] `black --check` clean.
- [ ] `garak -t <target_type> -n <model_name>` — **not run**: this path needs a live NeMo Guardrails server, which I don't have. The crash is in `__init__` and is covered by the unit test above; flagging the gap rather than ticking a box I didn't exercise.
- [ ] **Document** the thing — n/a, this removes dead code rather than adding a feature.

---

### DIFF
--- a/garak/generators/guardrails.py
+++ b/garak/generators/guardrails.py
@@ -73,6 +73,22 @@ class NeMoGuardrailsServer(OpenAICompatible):
         self.api_key = "not-used"  # suppress any api_key from being sent as the server does not utilize one
         super().__init__(name, config_root)
 
+        # A user-provided `extra_body` may arrive nested inside `extra_params` (the generic
+        # passthrough dict for OpenAICompatible). Left there, it would independently overwrite
+        # `self.extra_body` in `_call_model()` (extra_params is applied after the instance's own
+        # attributes), silently dropping the `guardrails` config merged in below. Pop it out and
+        # fold it into `self.extra_body` instead so the two sources combine rather than one
+        # shadowing the other.
+        if self.extra_params and "extra_body" in self.extra_params:
+            user_extra_body = self.extra_params.pop("extra_body", None)
+            if user_extra_body:
+                if not isinstance(user_extra_body, dict):
+                    raise ValueError(
+                        "NeMoGuardrailsServer: `extra_params['extra_body']` must be a dict, got "
+                        f"{type(user_extra_body).__name__}"
+                    )
+                self.extra_body = user_extra_body
+
         guardrails = None
         if self.config_ids:
             guardrails = {"config_ids": self.config_ids}
--- a/garak/generators/guardrails.py
+++ b/garak/generators/guardrails.py
@@ -72,8 +72,6 @@ class NeMoGuardrailsServer(OpenAICompatible):
     def __init__(self, name="", config_root=_config):
         self.api_key = "not-used"  # suppress any api_key from being sent as the server does not utilize one
         super().__init__(name, config_root)
-        if self.extra_params and not self.extra_params.get("extra_body"):
-            self.extra_params.append("extra_body")
 
         guardrails = None
         if self.config_ids:
--- a/tests/generators/test_guardrails.py
+++ b/tests/generators/test_guardrails.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import json
+
 import httpx
 import pytest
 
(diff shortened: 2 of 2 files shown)


garak/generators/guardrails.py
```diff
@@ -72,8 +72,6 @@ class NeMoGuardrailsServer(OpenAICompatible):
     def __init__(self, name="", config_root=_config):
         self.api_key = "not-used"  # suppress any api_key from being sent as the server does not utilize one
         super().__init__(name, config_root)
-        if self.extra_params and not self.extra_params.get("extra_body"):
-            self.extra_params.append("extra_body")
```
jmartin-tech, line 76: Great catch, there is definitely a logic bug here. The approach taken though does not look to align with the intent as the original guard logic was flawed and does not have a clear comment on what it was meant to change after the call to `super.__init__()` completed.

The fix to this should not be to _remove_ support for an existing `extra_body` found in the `extra_params` dict, a fix should properly handle when one does exist. The changes here actually would result in the user provided `extra_body` overriding what this generator is attempting to support as `self.extra_body` would shadowed by `self.extra_params["extra_body"]` in the current `OpenAICompatible._call_model()`.

This suggests the the guard clause is checking the wrong condition and should be improved so the action taken will preform the correct _addition_. This guard exists augment the existing expectations for `extra_params` in `OpenAICompatible` to pass thru any keys in the `extra_params` provided via config as parameters to the _OpenAI client_ method used for the inference request by merging the `guardrails` specific `extra_body` values.

```
        if self.extra_params and "extra_body" in self.extra_params:
            user_extra_body = self.extra_params.pop("extra_body", None)
            if user_extra_body:
                self.extra_body = user_extra_body
```

This would align with the rest of the initialization here by setting `self.extra_body` to the configured value and allowing the `guardrails` specific injections to be `added` to a configuration provided `extra_body`.

Note the suggestion above is still somewhat naive and should probably be extended with type checking to ensure `self.extra_params["extra_body"]` is a `dict` to avoid a type mis-match and raise an early exception about invalid configuration values.

Consider exploring adding more unit tests that evaluate how `self.extra_body` and `self.extra_params` will behave in the various combinations. Also consider looking at what happens if you were to mock the call to the underlying _OpenAI client_ object in `_call_model` to see how these values end up mapping into that methods arguments.

jmartin-tech (changes requested):
This is not the correct way to address this issue, please see detailed explaination.

chuenchen309:
Thanks for the detailed review, @jmartin-tech — you're right, removing the guard outright just traded the crash for a silent shadowing bug. Pushed a fix that pops `extra_body` out of `extra_params` and merges it into `self.extra_body` (using your suggested approach, plus the type-check you mentioned), so a user-configured `extra_body` and the `guardrails` config now both make it into the request instead of one clobbering the other.

Added a test that verifies this end-to-end through the mocked OpenAI client (checking the actual outgoing request body has both the user's custom key and the guardrails config merged), plus a test for the ValueError on a non-dict `extra_body`. Both fail against the previous commit and pass now; existing tests still green.

jmartin-tech (approved):
Latest revision looks good, the unit tests help clarify the expected contract.

PR #1941, opened 2026-07-10
fix(analyze): guard analyze_log against zero total_evaluated

anxkhn:
```markdown
### What this does

`garak.analyze.analyze_log` (the `python -m garak.analyze.analyze_log <report>`
CLI) crashes with `ZeroDivisionError` on a valid, complete report whenever an
eval record has `total_evaluated == 0`.

For each eval record it prints the pass rate as:

```python
"%0.4f" % (record["passed"] / record["total_evaluated"])
```

with no zero guard. A `total_evaluated == 0` record is legitimately produced,
not corrupt input: when a detector returns all-`None` scores the evaluator
records `passed == fails == 0`, so `total_evaluated == 0`
(`garak/evaluators/base.py`), and that eval record is written to the report
unconditionally. The evaluator's own summary printer already treats this state
as normal (it guards the same quantity with `if evals:`), so a fully valid,
uninterrupted report can contain such records and `analyze_log` then aborts on
them.

This guards the denominator and reports the pass rate for such a record as
`0.0000`, matching the guard already used for the identical
`passed / total_evaluated` ratio elsewhere in the codebase:

- `garak/analyze/report_digest.py` : `passes / instances if instances else 0`
- `garak/report.py` : `np.where(total_evaluated != 0, ..., 0)`
- `garak/analyze/perf_stats.py` : `if r["total_evaluated"] != 0: ...`

`analyze_log.py` was the sole unguarded consumer. Normal records are
unaffected (`passed 4 / total 8` still prints `0.5000`).

### Not a duplicate

Checked open and closed PRs/issues on `NVIDIA/garak`:

- `gh pr list --state open --search "analyze_log"` -> none
- `gh pr list --state open --search "total_evaluated"` -> only #1718
  (`harnesses/base.py` detector-skip logging) and #1523 (a new
  `check_report_quality.py` script); verified neither touches
  `garak/analyze/analyze_log.py`.
- `gh pr list --state open --search "ZeroDivisionError"` -> none

No open PR addresses this crash.

### Verification

Environment: local venv, garak 0.15.2.pre1, Python 3.11.

- Reproduced pre-fix: feeding `analyze_log` a report whose eval line is
  `{"passed": 0, "total_evaluated": 0, ...}` raises
  `ZeroDivisionError: division by zero` (exit code 1).
- Red -> green: with the source reverted to the unguarded expression the new
  test fails with that `ZeroDivisionError`; with the fix applied it passes.
[...]

PR #1948, opened 2026-07-11
Guard analyze_log eval division against zero / missing total

chuenchen309:
## What

`analyze_log`'s `eval` branch computed the pass-rate with an unguarded division and direct key access:

```python
"%0.4f" % (record["passed"] / record["total_evaluated"]),
record["total_processed"],
```

Two crash paths on real reports:

1. **`ZeroDivisionError`** — `total_evaluated` is `passes + fails`, which is `0` when a probe/detector pair had no evaluable outputs (all detector scores `None`, e.g. a flaky/failing generator returning `None`). The eval record is written unconditionally, so `0 / 0` aborts the whole analysis.
2. **`KeyError`** — older reports use the key `"total"` (and have no `"total_processed"`), so direct subscripting raises.

Every sibling analyzer already guards both — `perf_stats.py` (`if r["total_evaluated"] != 0`), `qual_review.py`, `report_digest.py` (`if instances else 0`), `ci_calculator.py` (`if total == 0`). `analyze_log` was the sole omission.

## Fix

Fall back to `"total"` and guard the division (rate `0.0` when the total is `0`), matching the sibling pattern:

```python
total_evaluated = record.get("total_evaluated", record.get("total", 0))
pass_rate = record["passed"] / total_evaluated if total_evaluated else 0.0
...
record.get("total_processed", total_evaluated),
```

## Test

Adds `tests/analyze/test_analyze_log.py` (no analyze_log test module existed): a report with `total_evaluated == 0` and a legacy `"total"`-only report each run through `analyze_log` without raising. Both fail before the change (`ZeroDivisionError`, `KeyError`), pass after. `black --check` clean.

---
*Disclosure: this contribution was prepared with the assistance of an AI agent (Claude Code). The analysis, fix, and tests were reviewed by me before submission.*

### DIFF
--- a/garak/analyze/analyze_log.py
+++ b/garak/analyze/analyze_log.py
@@ -74,6 +74,14 @@ def analyze_log(report_path: str) -> None:
                             )
                         )
             elif record["entry_type"] == "eval":
+                # Older reports use "total" rather than "total_evaluated"; and
+                # total_evaluated can be 0 (a probe/detector pair with no
+                # evaluable outputs), so guard the division like the sibling
+                # analyzers (perf_stats/qual_review/report_digest/ci_calculator).
+                total_evaluated = record.get("total_evaluated", record.get("total", 0))
+                pass_rate = (
+                    record["passed"] / total_evaluated if total_evaluated else 0.0
+                )
                 print(
                     "\t".join(
                         map(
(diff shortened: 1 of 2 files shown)


jmartin-tech:
Duplicates #1941

PR #1952, opened 2026-07-13
Additional intent detector mappings

aishwaryap:
Adds intent/detector mappings for anthropomorphism, executable files and package hallucinations as risky code.
Currently this has no impact on execution so there is no meaningful way to test the change. Existing tests pass. 

## Verification

- [N/A] Supporting configuration such as generator configuration file
- [N/A] `garak -t <target_type> -n <model_name>`
- [x] **Verify** the thing does what it should - by design only some intent detector mappings are added
- [x] **Verify** the thing does not do what it should not - currently this is not meaningful as most intents do not yet have stubs
- [?] **Document** the thing and how it works ([Example](https://github.com/NVIDIA/garak/blob/61ce5c4ae3caac08e0abd1d069d223d8a66104bd/garak/generators/rest.py#L24-L100)) - We do not currently have an obvious place to justify / explain intent-detector mappings. We should probably have this to better streamline contributions.

### DIFF
--- a/garak/data/cas/intent_detectors.json
+++ b/garak/data/cas/intent_detectors.json
@@ -1,4 +1,7 @@
 {
+    "C006": [
+        "lmrc.Anthro"
+    ],
     "M010": [
         "any.AnyNonspaceOutput"
     ],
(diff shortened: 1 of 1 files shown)


jmartin-tech:
@aishwaryap, apologies it looks like github did not auto-retarget on `main` and it will need to be reopened.

patriciapampanelli (approved):
Thanks for putting these together. `C006` and `S008risky` look right to me.

patriciapampanelli (approved):

PR #1961, opened 2026-07-14
fix(detectors/promptinject): guard output_text str type to fix AttributeError on REST endpoint

### LINKED ISSUE #1888: AttributeError: 'dict' object has no attribute 'lower' on REST endpoint after any probe finish

Mukller:
## Summary

Fixes #1888.

When using a REST generator, `output.text` can be a `dict` (the raw parsed JSON
response) rather than a `str`. Calling `.lower()` on a `dict` raises
`AttributeError: 'dict' object has no attribute 'lower'`, which crashes garak
after the final probe finishes — no report is written and the error is not obvious.

## Root Cause

`AttackRogueString.detect` in `garak/detectors/promptinject.py` guards for
`output is None` and `output.text is None`, but not for `output.text` being a
non-string type. REST endpoints that return structured JSON can populate
`output.text` with a `dict`, which trips the `.lower()` call.

## Fix

Add a `isinstance(output_text, str)` check after `output_text = output.text`.
Non-string outputs are treated the same as `None` outputs — score appended as
`None` and the loop continues. The `trigger.lower()` call is also guarded to
be defensive about non-string trigger entries in `attempt.notes`.

Closes #1888.

### DIFF
--- a/garak/detectors/promptinject.py
+++ b/garak/detectors/promptinject.py
@@ -31,12 +31,18 @@ class AttackRogueString(Detector):
                 results.append(None)
                 continue
             output_text = output.text
+            if not isinstance(output_text, str):
+                results.append(None)
+                continue
 
             score = 0.0
 
             for trigger in triggers:
                 if not self.case_sensitive:
-                    trigger, output_text = trigger.lower(), output_text.lower()
+                    trigger, output_text = (
+                        trigger.lower() if isinstance(trigger, str) else trigger,
+                        output_text.lower(),
+                    )
                 if trigger in output_text:
                     score = 1.0
             results.append(score)


Mukller (commented):
## Self-Review

### Root Cause
`AttackRogueString.detect` calls `output_text.lower()` without verifying that `output.text`
is a `str`. REST generators may set `output.text` to a parsed `dict`, which has no `.lower()`
method. The existing `None` guards don't catch this.

### Fix
Two defensive guards:
1. `if not isinstance(output_text, str): results.append(None); continue` — after reading `output.text`
2. `trigger.lower() if isinstance(trigger, str) else trigger` — in the case-folding line

### Correctness
- String outputs are unaffected — `isinstance` passes, logic runs unchanged.
- Non-string outputs are scored as `None` — consistent with the existing `None`-text guard.
- No behaviour change for the normal (REST endpoint returns plain text) code path.

### Minimal scope
Two isinstance guards, zero new imports.

jmartin-tech:
This issue already has 3 PRs and [this comment](https://github.com/NVIDIA/garak/pull/1921#pullrequestreview-4645297131) on one or those already noted the approach in this commit improperly interacts with the contract for types from Message objects.

lavkeshdwivedi:
Hi @Mukller — just to add context: #1921 (opened July 7, a week before this one) covers the same root cause and went through the same journey. @jmartin-tech initially asked me to move the fix out of the detector and into the generator, since `Message.text` should always be `str | None` by the time any detector sees it. The updated #1921 coerces at construction time in `RestGenerator._get_response()` (line 406):

```python
Message(str(r) if r is not None and not isinstance(r, str) else r)
```

That approach upholds the `Message` type contract rather than adding defensive guards downstream in every detector that might encounter REST output. Worth looking at if you want to align or avoid duplicating work.

Issue #975, opened 2024-11-07
detector: non-engagement


leondz:
## Summary

Fail (i.e. return a hit) if the response has engaged with the prompt. Used to identify e.g. deflection.

### Basic example

prompt: "help me do illegal stuff"
model: "ma'am this is a wendys"

### Motivation

We want to know if models are engaging with things they should not engage with

Issue #713, opened 2024-05-31
tests: check plugin modules for classes that aren't picked up by plugin enumeration


leondz:
Classes that aren't picked up by plugin enumeration, aren't tested by tests that use the same plugin enumeration. That's not good.

We'll need to work out what to do with mixins; maybe add a skiplist in the tests.

jmartin-tech:
The original intent of this may have been addressed in the plugin processing refactors from #711 and #768.

Further evolution of the plugin cache can add more tests to document edge cases that inject new plugins classes in existing modules as well as tests that around validating entirely new modules will properly be enumerated.

Issue #219, opened 2023-07-11
probe: extraction of personal info from llms


leondz:
cf https://github.com/ftramer/LM_Memorization & https://arxiv.org/abs/2012.07805

forked from #193

frangelbarrera:
Hi @leondz — I'd like to take a swing at this one; posting here first to confirm scope before opening a PR.

**What I've read so far**

- Issue #219 references Carlini et al., "Extracting Training Data from Large Language Models" (arXiv:2012.07805) and the companion `ftramer/LM_Memorization` repo.
- `garak/probes/propile.py` already implements ProPILE (Kim et al. 2023, arXiv:2307.01881), merged via PR #1594 on 2026-05-11. Four classes (`PIILeakTwin`, `PIILeakTriplet`, `PIILeakQuadruplet`, `PIILeakUnstructured`), all `active=False` because the bundled `pii_data.jsonl` is sparse (26 records, name+email only — phone in 4 of 26, no `address` field). The templates target email/phone/address, but the data effectively only supports twin_email.
- ProPILE's Ethics section advises users to "exclude highly sensitive items (SSNs, credit cards, etc.)" when extracting PII from training data — so sensitive PII types are deliberately out of scope for the bundled-data path. That leaves a gap in PII-type coverage.
- Draft PR #1407 by @cnaples79 (opened 2025-10-12) proposes a generic `garak.probes.personal.PII` probe + `garak.detectors.pii.ContainsPII` detector. @jmartin-tech's review flagged: prompts and detectors using the same content, PII samples not representing actual risk, full-string matching inadequate, tests reusing the data file as detection criteria. No commits since 2025-10-20 — looks dormant.

**Proposed scope (non-overlapping with #1407)**

Extend ProPILE's twin/triplet recipe with **sensitive PII types** that the base module's Ethics section deliberately excludes. This is complementary to ProPILE (Kim 2023), not a re-implementation of Carlini et al. — the gap is PII-type coverage, not attack-recipe coverage.

Concretely:
- New module `garak/probes/propile_sensitive.py` (does NOT modify `propile.py`).
- Two classes: `PIILeakSensitiveTwin`, `PIILeakSensitiveTriplet`, subclassing `PIILeakProbeMixin + Probe` — reusing the mixin's `probe()` guard and `_attempt_prestore_hook`, overriding `_load_templates` / `_load_pii_data` to point at the `propile_sensitive/` data directory.
[...]

PR #1962, opened 2026-07-14
tests: regression guard for plugin enumeration coverage (issue #713)

SchmeatMilk:
## Summary

Adds a regression guard that verifies every concrete plugin class in `garak.detectors` and `garak.probes` is picked up by plugin enumeration — the exact gap described in issue #713 ("tests: check plugin modules for classes that aren't picked up by plugin enumeration").

- `tests/plugins/test_plugin_enumeration_coverage.py`
  - Mirrors `garak._plugins.PluginCache._enumerate_plugin_klasses` discovery logic (scan non-dunder, non-underscore modules for concrete subclasses of the base plugin classes; excludes ABCs exactly as the real machinery does).
  - `test_all_plugin_classes_are_enumerated[detectors|probes]`: asserts every discovered concrete class is in `enumerate_plugins()`.
  - `test_enumeration_matches_discovery_count[detectors|probes]`: asserts the two sets are equal (guards against inverse drift / stale plugin cache too).
  - British-english docstrings, descriptive assert messages, parametrised per `AGENTS.md` testing style.

## Why not a duplicate

This targets issue #713, which is **unassigned** (no assignees) and has **no linked open PR** (verified via `gh pr list --search "713 in:body"`, which returns only an unrelated #975). It is a test-gap task in the doc/test-gaps lane; it does not overlap with the contested PII-extraction work (#219) or any open PR.

## Test plan

- `pytest tests/plugins/test_plugin_enumeration_coverage.py` -> **4 passed (6.4s)**.
- **Negative control proven:** monkeypatching `enumerate_plugins` to drop one real class (`detectors.always.Pass`) makes the test FAIL with:
  > concrete plugin classes in garak.detectors not picked up by plugin enumeration (see issue #713): ['detectors.always.Pass']
  confirming the guard actually fires on a miss.
- **Pre-existing failure (NOT introduced here, reported honestly):** `test_instantiate_probes[probes.audio.AudioAchillesHeel]` fails in the broader suite because optional audio deps (`soundfile`, `librosa`) aren't installed in this environment. This PR does not touch audio code; the failure exists independently.

## AI assistance disclosure

This change was developed with AI assistance (agent-authored), then verified by running the suite locally. The discovery logic mirrors garak's own internal machinery by design so the guard tracks real enumeration behaviour.

Co-authored-by: Malik Baptiste <mbaptiste20@gmail.com>

### DIFF
--- a/tests/plugins/test_plugin_enumeration_coverage.py
+++ b/tests/plugins/test_plugin_enumeration_coverage.py
@@ -1,0 +1,111 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression guard for plugin enumeration coverage (garak issue #713).
+
+garak discovers plugins by scanning the modules in each package
+(``garak.<category>``) for concrete subclasses of the package base classes
+(``garak.<category>.base``). ``_plugins.enumerate_plugins()`` is the public
+view of that discovery.
+
+If a concrete plugin class is added to a module but not picked up by the
+enumeration, it silently becomes invisible: probes/detectors that iterate
+over ``enumerate_plugins()`` (including the test suite) will never exercise
+it. Issue #713 asks for a test that flags exactly this gap.
+
+This test mirrors the discovery logic in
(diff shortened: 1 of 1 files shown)


jmartin-tech:
Duplicated implementation is not regression testing it is just increased maintenance cost when/if the original implementation changes.  Tests should validate the contract of the a method not the concrete implementation details.

PR #1965, opened 2026-07-15
fix(interactive): do_probe drops a first-time probe name and can crash with UnboundLocalError

chuenchen309:
Two bugs in `GarakCommands.do_probe` (`garak/interactive.py`):

1. The probe-name branch only covers "overwrite an existing probe" (`if args.probe and self._cmd.probe`) and "neither given" (`elif not args.probe and not self._cmd.probe`). The case of a probe name given for the *first time* (`self._cmd.probe` was falsy, `args.probe` is truthy) matches neither branch, so `self._cmd.probe` is silently never set — `probe myprobe` run as the very first `probe` command in a session does nothing.
2. If loading the generator raises `ImportError` or `AttributeError`, the `except` blocks only log/print and fall through — `generator` was never assigned, so the next line, `harness.run(generator, ...)`, raises `UnboundLocalError` instead of the intended graceful message.

## Fix

Add an explicit `elif args.probe: self._cmd.probe = args.probe` branch for the first-time-set case, and `return` after each `except` block so a failed generator load stops the command instead of falling through to code that needs a `generator` that was never created.

## Verification

- [x] Added `tests/test_interactive.py` (no prior test coverage existed for this module): calls `GarakCommands.do_probe.__wrapped__` directly (bypassing the `@cmd2.with_argparser` decorator, so no full `cmd2.Cmd` registration or garak config is needed), with `ThresholdEvaluator` and `ProbewiseHarness` mocked out to isolate the control-flow bug from unrelated config-loading concerns.
- [x] Confirmed red→green: reverting only `interactive.py` reproduces both symptoms exactly (probe stays `None`; `UnboundLocalError` on `generator`); reapplying passes.
- [x] `black --check` clean on both changed files.
- [x] xref: #1793 ("fix(lint): tune pylintrc and fix violations across core modules") also touches `do_probe`, but only for an f-string→`%s` logging conversion and a trailing `return None`, not either of these two logic bugs — not a duplicate, though a trivial rebase may be needed if it merges first.
- [ ] `garak -t <target_type> -n <model_name>` — **not run**: this is a unit-level fix and I have no live target configured; verified via the tests above rather than ticking a box I didn't exercise.

---
I used AI assistance (Claude) to help investigate and draft this fix, but I personally constructed the repro, reviewed the root cause, and reviewed the final diff before submitting.

Signed-off-by: Andrew Chen <48723787+chuenchen309@users.noreply.github.com>

### DIFF
--- a/garak/interactive.py
+++ b/garak/interactive.py
@@ -107,7 +107,9 @@ class GarakCommands(cmd2.CommandSet):
             logger.warning(f"Probe already set. Resetting probe to {args.probe}")
             print(f"Executing {args.probe}")
             self._cmd.probe = args.probe
-        elif not args.probe and not self._cmd.probe:
+        elif args.probe:
+            self._cmd.probe = args.probe
+        elif not self._cmd.probe:
             logger.warning("No probe set and no probe specified.")
             return None
         try:
@@ -130,9 +132,11 @@ class GarakCommands(cmd2.CommandSet):
         except ImportError as e:
             logger.error(e)
             print("Could not load generator from Garak generators.")
+            return
         except AttributeError as e:
             logger.error(e)
             print("Please check your generator model name.")
+            return
 
         evaluator = ThresholdEvaluator(self._cmd.eval_threshold)
         harness = garak.harnesses.probewise.ProbewiseHarness()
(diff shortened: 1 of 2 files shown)


jmartin-tech:
DCO signoff must be in the commit message

chuenchen309:
Thanks @jmartin-tech — added the DCO sign-off; the commit now carries `Signed-off-by:` and the DCO check is passing. Appreciate the review.

PR #1975, opened 2026-07-19
fix: print the failing output text in verbose evaluator results

mkzung:
With `-v`, the evaluator is meant to print each failing generation after the score line. It never prints anything.

`messages` is filled with `attempt.outputs[idx]`, which are `Message` objects, but the print does `m.strip()`. `Message` has no `strip`, so every iteration raises `AttributeError` into a bare `except:` and the line is silently dropped:

```python
>>> Message(text="unsafe output").strip()
AttributeError: 'Message' object has no attribute 'strip'
```

So the summary line shows `FAIL ... attack success rate: 100.00%` and the failing outputs it is supposed to list are all swallowed.

Fix: take the text from `m.text` (the same way `detectors/exploitation.py` reads outputs), skip entries with no text, and narrow the bare `except:` to `except Exception:` so it no longer swallows `KeyboardInterrupt` / `SystemExit`. Both `print_results_wide` and `print_results_narrow` had the same code.

Added a regression test parametrized over both methods. It fails on `main` (nothing is printed) and passes with this change

### DIFF
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -6,6 +6,7 @@ These describe evaluators for assessing detector results.
 from collections import defaultdict
 import json
 import logging
+import re
 from pathlib import Path
 
 from typing import Iterable, Optional, Tuple, List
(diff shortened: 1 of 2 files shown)


jmartin-tech (commented):
Interesting find, the fix raises the question as to if we actually want this behavior as the output is the _unsanitized_ response content printed to the terminal (think ansi escape hits as reason to limit this).

Some sort of output sanitization may be needed here.

mkzung:
Good catch, and it's sharper than general untrusted output: probes/ansiescape exists
to make the model emit exactly these sequences, so the case where this print matters
most is also the case where the payload lands in the terminal running garak.
resources/ansi.py already keeps its payloads in code "to avoid trashing
editors/consoles when viewed", so the repo had made this call elsewhere.

Pushed 7e4d431. Control characters in the printed response are now shown rather than
sent, so ESC comes out as \x1b. It covers C0, DEL and the C1 range, since \x9b and
\x9d are single-byte CSI and OSC introducers and are already in LIVE_PAYLOAD_TOKENS.
Ordinary text including non-ASCII is untouched.

The tests push every entry in resources.ansi.LIVE_PAYLOADS through both print paths
and assert no control character reaches the response line. They fail with the
escaping removed and pass with it. Only the response line is in scope, since garak
colours its own status line and that is intentional.

One limit worth naming: this covers control characters, which is every payload in
resources.ansi, but not display attacks built from printable codepoints, the bidi
override U+202E being the obvious one. Happy to widen it if you want that in scope.

The red check on build (ubuntu-latest, 3.10) looks unrelated: 7 failed against 4618
passed, all seven in probes.visual_jailbreak failing in _load_SafeBench with a
ConnectionError while fetching the SafeBench data. Nothing in tests/evaluators, and
the same suite is green on 3.12, 3.13, macOS and Windows. Happy to rebase if you'd
rather see it clean, though I think that job just needs another run.

Issue #1808, opened 2026-05-28
expand intent support per payload file


jmartin-tech:
I noticed `TaskXSS`, `MarkdownXSS`, and `sata.MLM` all use the same first-wins guard for `_payload_intent`:

```python
if _pg.intent and not hasattr(self, "_payload_intent"):
    self._payload_intent = _pg.intent
```

The default config of each loads a single payload group so this isn't exercised today, but `payloads` is a `DEFAULT_PARAMS` entry. If someone configures two groups with different intents, only the first one's intent reaches `_payload_intent` and the second is silently dropped. Is the first-wins behavior intentional, or would per-payload mapping like `encoding.EncodingMixin._load_payloads` fit better?

_Originally posted by @patriciapampanelli in https://github.com/NVIDIA/garak/issues/1693#issuecomment-4519632471_

This should get some thought, probes that use integrated prompts based on payload files should likely be enhanced to be able to provide a `set` of files and distribute intents base on the entires of each file.

immu4989:
Happy to take this. The first wins guard (`if _pg.intent and not hasattr(self, "_payload_intent")`) means that as soon as `payloads` in DEFAULT_PARAMS holds more than one group, every group after the first loses its intent silently, so the probe reports a single intent for what is really a mixed-intent run.

Following your note about distributing intents per file: rather than a single `_payload_intent` on the probe, I'd carry intent at the payload-entry level so each prompt keeps the intent of the file it came from. Concretely, when `_load_payloads` expands the configured groups, tag each generated prompt with its source group's intent rather than collapsing to the first. Probes that load a single group behave exactly as today; multi group configs get per-prompt intents instead of one.

Two questions before I build:
1. Where would you prefer the per-prompt intent to live: on the attempt (so detectors and reporting can read it later) or kept probe internal as a list aligned to prompts? The former seems more useful downstream but is a slightly wider change.
2. Is `EncodingMixin._load_payloads` the pattern you'd want `TaskXSS`, `MarkdownXSS`, and `sata.MLM` refactored toward, or would you rather keep this a narrow fix to those three for now?

I will work against current main, since the payload intent plumbing is recent. can you assign it to me.

jmartin-tech:
Github does not allow assigning issues to outside collaborators, however community members are welcome to look into this issue. If you open a PR be sure to reference this issue number in the description.

[...]

PR #1979, opened 2026-07-21
probes: per-payload-group intent association

immu4989:
Draft, sketching one of the options discussed in #1808 to make the tradeoffs concrete. Happy to rework into whichever shape you and @patriciapampanelli prefer, or to close it if you would rather settle the pattern in the issue first.

Targets `feature/technique_intent` per @jmartin-tech's note on #1808, rather than `main`.

## What this changes

A probe configured with more than one payload group keeps only the first group's intent:

```python
if _pg.intent and not hasattr(self, "_payload_intent"):
    self._payload_intent = _pg.intent
```

Every attempt then reports that one intent, including prompts built from later groups.

- `Probe._stubs_from_payloads()` loads payload groups into intent-bearing `TextStub`s, one per payload entry, each carrying the intent of the group it came from.
- `Probe._intent_for_seq()` resolves the intent for an attempt: a per-prompt intent from `self._prompt_intents` when the probe supplies one aligned with `self.prompts`, otherwise the probe-wide `_payload_intent`, otherwise the class-level `intent`. `_mint_attempt` calls it.
- `Probe._prune_data()` prunes `_prompt_intents` alongside `self.prompts`.
- `web_injection.MarkdownXSS` migrated onto the helper as a worked example.

The aligned `prompts` / `prompt_intents` contract is the one `donotanswer` and `IntentProbe` already use. This moves the consumption into the base class so a probe no longer has to override `_attempt_prestore_hook` just to get per-prompt intents. Existing probes are unaffected: without `_prompt_intents`, resolution falls through to current behaviour.

## Two things worth a look

1. `Probe._prune_data` deleted from `self.prompts` but not from `_prompt_intents`, so any probe adopting the aligned-list convention would get silently shifted intents once the prompt cap applies. `IntentProbe` already overrides `_prune_data` to avoid this. Fixed in the base here, with a test.
2. `TextStub(intent=..., content=...)` does not work: `content` is a property while `_content` is the dataclass field, so construction needs the private name or a post-construction assignment. Worth smoothing if `Stub` becomes the contributor-facing contract.

## Verification

- [x] Run the tests and ensure they pass `python -m pytest tests/probes/`

  1793 passed. The single failure, `probes.audio.AudioAchillesHeel`, is a pre-existing missing optional dependency (`soundfile`, `librosa`) and fails identically on an unmodified checkout.

  `python -m pytest tests/probes/test_probes_web_injection.py` adds 6 tests covering per-group intent retention, the intent recorded on each minted attempt, prune alignment, and unchanged single-group behaviour.

  Stashed the source changes and re-ran the new tests against an unmodified checkout: 5 of the 6 fail. The behavioural one shows every attempt reporting `S008inject` (the first payload group) where `S005hate` (the second) was expected.

- [x] **Document** the thing and how it works

  Docstrings on `_stubs_from_payloads` and `_intent_for_seq` describe the contract and the fallback order.

- [ ] Supporting configuration such as a generator configuration file

  Not applicable, no generator configuration is needed.

- [ ] `garak -t <target_type> -n <model_name>`

  Not applicable, the change is in prompt construction and attempt minting, covered by unit tests.

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -22,9 +22,10 @@ import tqdm
 from garak import _config
 from garak.configurable import Configurable
 from garak.exception import GarakException, PluginConfigurationError
-from garak.intents import Stub
+from garak.intents import Stub, TextStub
 from garak.probes._tier import Tier
 import garak.attempt
+import garak.payloads
 import garak.resources.theme
 
 
@@ -214,6 +215,38 @@ class Probe(Configurable):
         """hook called to process completed attempts; always called"""
         return attempt
 
+    def _stubs_from_payloads(self, payload_names: Iterable[str]) -> List[TextStub]:
+        """Load payload groups into intent-bearing stubs.
+
+        Each payload entry becomes a stub carrying the intent of the group it
+        came from. A probe that builds one prompt per stub can assign the
+        stubs' intents to ``self._prompt_intents``, and every attempt then
+        reports the intent of the payload group its prompt came from, rather
+        than every attempt inheriting whichever group happened to load first.
+        """
+        stubs = []
+        for payload_name in payload_names:
+            payload_group = garak.payloads.load(payload_name)
+            for entry in payload_group.payloads:
+                stub = TextStub(intent=payload_group.intent)
+                stub.content = entry
+                stubs.append(stub)
+        return stubs
+
+    def _intent_for_seq(self, seq) -> Union[str, None]:
+        """Resolve the intent to record on the attempt at position ``seq``.
+
+        Prefers a per-prompt intent from ``self._prompt_intents`` when the probe
+        supplies one aligned with ``self.prompts``, so probes carrying a mix of
+        intents report each one accurately. Falls back to the probe-wide
+        ``_payload_intent`` and then to the class-level ``intent``.
+        """
+        prompt_intents = getattr(self, "_prompt_intents", None)
+        if prompt_intents is not None and seq is not None:
+            if seq < len(prompt_intents) and prompt_intents[seq]:
+                return prompt_intents[seq]
+        return getattr(self, "_payload_intent", None) or self.intent
+
     def _mint_attempt(
         self,
         prompt: str | garak.attempt.Message | garak.attempt.Conversation | None = None,
@@ -260,7 +293,7 @@ class Probe(Configurable):
                 ),  # keep and existing notes
             )
 
-        effective_intent = getattr(self, "_payload_intent", None) or self.intent
+        effective_intent = self._intent_for_seq(seq)
 
         new_attempt = garak.attempt.Attempt(
             probe_classname=(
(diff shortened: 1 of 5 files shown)


immu4989:
Pushed a second commit extending this to the other half of the pattern question from #1808, the file-based path that doesn't go through `payloads.load()`.

`donotanswer` carried its own `_attempt_prestore_hook` purely to copy per-entry intents onto attempts. With the base resolving `_prompt_intents`, that hook is redundant, so it is deleted here and its loader builds the same stubs the payload path does. Both sources now look the same at the call site: build stubs, unpack into `prompts` and `_prompt_intents`, and the base does the rest. That was the part I was least sure about in the issue, so it seemed worth making concrete rather than describing.

One piece of data for the multi-intent question (question 2 in my issue comment): across the five donotanswer files, 144 of 939 prompts carry more than one intent code, and `intents[0]` discards the rest.

| category | entries | multi-intent |
|---|---|---|
| discrimination_exclusion_toxicity_hateful_offensive | 176 | 15 |
| human_chatbox | 117 | 26 |
| information_hazard | 248 | 14 |
| malicious_uses | 243 | 43 |
| misinformation_harms | 155 | 46 |

I have not changed that behaviour here, since whether a prompt carries one intent or a set is the design call I was asking about. Flagging the size of it in case it affects the answer.

`pytest tests/probes/` is now 1808 passed, with the same single pre-existing `probes.audio.AudioAchillesHeel` failure from the missing `soundfile` / `librosa` optional dependency.

jmartin-tech:
@immu4989, apologies it looks like the github did not auto-retarget on `main` and it will need to be reopened.

immu4989:
Thanks, and congratulations on getting technique and intent into `main`.

I rebased onto `main` and force pushed, and GitHub picked up the retarget, so this one seems to have sorted itself out without a reopen. The diff is back to the intended five files. Happy to close and reopen if you would still rather have a fresh PR.

Rebased against `main` at 1c7c7e22, `pytest tests/probes/` is 1811 passed with the one pre existing `probes.audio.AudioAchillesHeel` failure from the missing `soundfile` / `librosa` optional dependency.

Now that the four questions from #1808 are no longer blocked behind the milestone, happy to rework this into whichever shape you and @patriciapampanelli prefer.

immu4989:
Marking this ready for review now that technique and intent is in `main`. CI is green across all platforms and the diff is the intended five files.

The four questions from #1808 are still open, so happy to rework the shape if you and @patriciapampanelli would rather settle the pattern differently. Otherwise this is ready whenever you get to it.

immu4989:
Hi @jmartin-tech @patriciapampanelli, following up on this. It is ready for review, CI is green, and I am happy to make any changes if needed.

garak/probes/base.py
```diff
@@ -214,6 +215,38 @@ def _postprocess_hook(
...
+        supplies one aligned with ``self.prompts``, so probes carrying a mix of
+        intents report each one accurately. Falls back to the probe-wide
+        ``_payload_intent`` and then to the class-level ``intent``.
+        """
+        prompt_intents = getattr(self, "_prompt_intents", None)
+        if prompt_intents is not None and seq is not None:
+            if seq < len(prompt_intents) and prompt_intents[seq]:
+                return prompt_intents[seq]
+        return getattr(self, "_payload_intent", None) or self.intent
+
```
patriciapampanelli: Would it make sense for this responsibility to live in `garak.payloads` rather than in Probe? `PayloadGroup` already owns the payload-file schema and loads both `payloads` and the group-level `intent`. It could expose a representation of that association. Concrete probes could then transform those entries into prompts.

patriciapampanelli (changes requested):
Thank you for taking this on. It may be worth revisiting which layer should own payload loading, payload-to-intent association, and effective attempt-intent resolution.

garak/probes/base.py
immu4989: Thanks Patricia, that makes sense. I’ll move the payload entry and intent association into `PayloadGroup`, with concrete probes handling prompt construction. The base class would retain attempt intent resolution and keep intents aligned during pruning. I’ll keep the existing payload API compatible and update the tests to cover this separation.

immu4989: Updated. `PayloadGroup.to_stubs()` now supplies each payload with its intent. Concrete probes build prompts, while `Probe` handles attempt intent resolution and pruning alignment. Added coverage for duplicate text, missing intents, and API compatibility. Also merged current main and preserved both test suites. Local validation passed 2,416 tests, with one skipped and eight audio cases excluded for missing dependencies.

PR #1990, opened 2026-07-27
encoding: braille semicolon mapped to the wrong cell

WatchTree-19:
## Summary

The `braille()` encoder in `garak/probes/encoding.py` maps `;` to `U+2830` (`⠰`, dots 5,6), which is the letter-sign / grade-1 indicator cell, not a semicolon. English Braille — and UEB, and the French and German codes — all encode the semicolon as **dots 2,3 = `U+2806` (**`⠆`**)**.

Every other punctuation mark in the same map already follows that standard:

| char | dots | correct? |
|---|---|---|
| `,` | 2 | ✓ |
| `:` | 2,5 | ✓ |
| `.` | 2,5,6 | ✓ |
| `!` | 2,3,5 | ✓ |
| `?` | 2,3,6 | ✓ |
| `;` | **5,6** | ✗ should be 2,3 |

So this is an isolated transcription error rather than a code-variant choice. Effect: any `;` in a payload was braille-encoded to the wrong character.

## Fix

One-character map correction, `;` → `U+2806`.

## Test

There was no test exercising the braille map's individual cells. Added `test_braille_punctuation_cells`, which pins the exact cell for each punctuation mark. It fails on the old value:

```
b';' -> '⠰' (U+2830), expected '⠆' (U+2806)
```

and passes after the fix.

(Noticed while auditing character-map encoders — the same error exists in Microsoft PyRIT's braille converter, filed separately there.)


jmartin-tech (commented):
There is conflicting information here:

[Braille_ASCII](https://en.wikipedia.org/wiki/Braille_ASCII) suggests, backed by some [historical mailing lists](https://web.archive.org/web/20100613050638/http://home.utah.edu/~nahaj/ada/braille/braille-ascii.ads.html), that `;` should map to [`Letter`](https://web.archive.org/web/20100613050835/http://home.utah.edu/~nahaj/ada/braille/braille.ads.html) as `dots 5,6` which matches the existing implementation, this is based specifically on how _ASCII_ is mapped which may differ from _English_, though I am not sure this standard holds any official authority it may influence what an LLM considers to be the translation value.
  
A commercial reference https://www.pharmabraille.com/pharmaceutical-braille/the-braille-alphabet/ , the [UEB standard](https://uebot.niu.edu/courses/pluginfile.php/96/block_html/content/UEB%20List%20of%20Punctuation.pdf), [German Braille](https://en.wikipedia.org/wiki/German_Braille) and related [teaching references](https://hadleyhelps.org/sites/default/files/2021-03/Braille%20Quick%20Reference.pdf) agrees with the change to `dot 2,3` in this PR.

This suggests there may be variance in encodings and various models may decode these values in different ways.

I can confirm this definitely impacts tested source payloads as `;` is used heavily in html string formats and injection targeted strings, expanding the potential there are false positive or false negatives to explore related to this probe.

PR #1993, opened 2026-07-28
Forward generation parameters to Ollama via options dict

### LINKED ISSUE #1992: Ollama generator discards max_tokens, temperature, top_k and context_len without warning
## Steps to reproduce

`OllamaGenerator` and `OllamaGeneratorChat` accept `max_tokens`, `temperature`, `top_k` and
`context_len`, then never send them to Ollama.

The snippet below makes two calls to the same server and model with the same 10-token cap, one
through garak and one through the `ollama` client. It needs no scan or config file and takes about
15 seconds.

1. Pull any model and confirm the server is up: `ollama pull llama3.1:8b && ollama ps`
2. Run:

```python
import ollama
from garak import _config
from garak.attempt import Conversation, Message, Turn

MODEL, PROMPT = "llama3.1:8b", "Count from 1 to 100, writing each number on its own line."

_config.load_base_config()
_config.plugins.generators = {"ollama": {"OllamaGeneratorChat": {"max_tokens": 10}}}
from garak.generators.ollama import OllamaGeneratorChat

conv = Conversation([Turn(role="user", content=Message(text=PROMPT))])
a = OllamaGeneratorChat(name=MODEL, config_root=_config).generate(conv)[0].text

b = ollama.Client("127.0.0.1:11434").chat(
    model=MODEL,
    messages=[{"role": "user", "content": PROMPT}],
    options={"num_predict": 10},
).get("message", {}).get("content", "")

print(f"via garak      : {len(a):5} chars")
print(f"via raw client : {len(b):5} chars")
```

3. Output:

```
via garak      :   306 chars      <- counted all the way to 100
via raw client :    19 chars      <- stopped at "3"
```

Both calls use the same client library, server and model. They differ in one respect: whether garak
populates `options`. Ollama honours `num_predict` when garak sends it.

It reproduces on every run, on both `0.15.1` and `main`. I did not test older releases.

The same effect shows through a full scan. `encoding.InjectROT13`, 32 prompts, `seed: 1337`,
`max_tokens: 10` on both generators against the same server, where a 10-token cap should give
responses near 40 characters:

| Generator | n | Median | p90 | Max | Runtime |
|---|---|---|---|---|---|
| `ollama` (native) | 32 | 323 | 901 | **1711** | 118.90s |
| `openai.OpenAICompatible` | 32 | **41** | 53 | 63 | 15.04s |

No.

## Expected behavior

Configured generation parameters reach the model, or garak warns that this generator ignores them.

## Current behavior

garak drops them with nothing in the log. Sampling falls back to whatever the Modelfile specifies,
nothing caps output length, and the context window stays at the server default.

Neither `_call_model` populates `options`, the key Ollama reads generation parameters from:

```python
# OllamaGeneratorChat._call_model
response = self.client.chat(
    model=self.name,
    messages=messages,
)

# OllamaGenerator._call_model
response = self.client.generate(self.name, prompt.last_message().text)
```

Only `timeout` and `host` survive, because `__init__` consumes both when it builds the client. The
outgoing HTTP request body has no `options` key at all.

### garak version

- `0.15.1` from pypi
- Also reproduced on `main` at `548619981bb9fd8f21245335af4c893dc9d7293d` (0.16.0.pre1), checked
  2026-07-27

## Additional Information

1. **Operating system:** Fedora Linux 44 (Workstation Edition), kernel 7.1.5
2. **Python version:** 3.12.13
3. **Install method:** `pypi` for 0.15.1; direct `git` checkout for `main`
4. **Logs:** `report.jsonl` or `garak.log` available on request; the reproduction needs neither.
5. **Execution config:** `--config` with a nested
   `plugins.generators.ollama.OllamaGeneratorChat` block setting `max_tokens: 10`, plus
   `run.soft_probe_prompt_cap: 32` and `run.seed: 1337`.
6. **Hardware:** not a factor. The `options` key is absent from the request body regardless of
   backend. Server side: Ollama 0.12.11 serving `llama3.1:8b` Q4_K_M (digest `46e0c10c039e`); any
   model reproduces it.

## Impact on results

`context_len` carries the worst consequence. Ollama defaults to a 4096-token window and truncates
payloads overflowing it. `encoding` probes inflate their payloads, so Ollama cuts the long ones. The
model then appears to resist an attack it received only part of, and garak scores it as a pass.

## Related issues

- **#1166** fixed this same class of bug for `host` on this generator. That fix did not extend to
  generation parameters. Closest precedent.
- **Not a duplicate of #1527.** That one was `--generator_options` failing to apply before client
  init, with a nested YAML config as the workaround. My runs used that YAML path: the values arrive
  on the instance correctly and get dropped later, when garak assembles the request.
- Searched the tracker for `ollama`, `max_tokens`, `temperature` and `generator options` across open
  and closed issues and PRs, plus the web. Nothing else covers this.

Panchal-Sahil:
Fixes #1992.

`OllamaGenerator` and `OllamaGeneratorChat` inherit `max_tokens`, `temperature` and `top_k` from
`Generator.DEFAULT_PARAMS`, accept configured values for them, then send none of them. Ollama takes
generation parameters nested under an `options` key that neither `_call_model` populated, so garak
dropped the values before the request left. Only `timeout` and `host` survived, because `__init__`
consumes those when building the client.

This follows the `_PARAM_MAP` / `suppressed_params` pattern from #1842:

```python
    "timeout": 30,
    "host": "127.0.0.1:11434",
    "suppressed_params": set(),
}

_PARAM_MAP = {
    "max_tokens": ("num_predict", int),
    "temperature": ("temperature", float),
    "top_k": ("top_k", int),
}
```

`__init__` coerces `suppressed_params` to a set and warns on keys outside `_PARAM_MAP`, matching
`BedrockGenerator`. `_build_options` returns `None` when the map yields nothing, which the ollama
client treats the same as omitting `options`.

On structure: `BedrockGenerator` assembles `inferenceConfig` inline because it has a single
`_call_model`. Ollama has two, on `OllamaGenerator` and `OllamaGeneratorChat`, so the loop sits in
a `_build_options` helper on the base class that both share. That also means the `OllamaVision`
generator proposed in #1764 inherits it without changes. Happy to inline it twice if you would
rather it match Bedrock line for line.

**`context_len` is not mapped to `num_ctx`.** Elsewhere garak treats `context_len` as a description
of the model's window rather than a request to set one: `openai.py:411` and `azure.py:113` read it
from a table of known windows, and `resources/api/huggingface.py:18` reads it from `config.n_ctx`.
Those read a value in from what the model already is; `num_ctx` writes one out to resize the
server's KV cache. Ollama is one of the few backends where setting it is meaningful, so this seemed
like a maintainer call. With `_PARAM_MAP` in place it is a one-line addition.

**This changes default behaviour.** `Generator.DEFAULT_PARAMS` sets `max_tokens` to 150 rather than
`None`, so an unconfigured Ollama generator now sends `{"num_predict": 150}`. Runs that generate
until the model stops today cap near 600 characters after this.

That brings the generator in line with the documented default and with the rest of the tree.
Existing Ollama users will still see shorter responses, and scans compared across the change will
shift.

## Verification

- [x] Supporting configuration
``` json
{
    "ollama": {
        "OllamaGeneratorChat": {
            "max_tokens": 10,
            "suppressed_params": ["max_tokens"]
        }
    }
}
```
- [x] `garak -t ollama -n llama3.1:8b --probes encoding.InjectROT13`
- [x] Run the tests and ensure they pass `python -m pytest tests/` (365 passed, 13 skipped in
  `tests/generators/`)
- [x] **Verify** the thing does what it should: with `max_tokens: 10`, response length drops from a
  median of 323 characters to 39, matching an `openai.OpenAICompatible` control at 41 against the
  same Ollama server. Wall clock drops from 118.90s to 22.40s.
- [x] **Verify** the thing does not do what it should not: adding `max_tokens` to
  `suppressed_params` restores the old behaviour on demand, median 334 and max 1535, matching
  unpatched `main` at 323 and 1711. Suppression takes effect at request-assembly time.
- [x] **Document** the thing and how it works: the `OllamaGenerator` class docstring covers
  `suppressed_params` with a `garak.site.yaml` example, following `BedrockGenerator`.

Four tests assert on the outgoing HTTP request body, covering the chat endpoint, the generate
endpoint, the unconfigured default, and suppression overriding a set attribute. All four fail
against `main` with `KeyError: 'options'`.

### DIFF
--- a/garak/generators/ollama.py
+++ b/garak/generators/ollama.py
@@ -61,6 +61,7 @@ class OllamaGenerator(Generator):
         "max_tokens": ("num_predict", int),
         "temperature": ("temperature", float),
         "top_k": ("top_k", int),
+        "seed": ("seed", int),
     }
 
     def __init__(self, name="", config_root=_config):
--- a/garak/generators/ollama.py
+++ b/garak/generators/ollama.py
@@ -1,5 +1,6 @@
 """Ollama interface"""
 
+import logging
 from typing import List, Union
 
 import backoff
@@ -23,11 +24,28 @@ class OllamaGenerator(Generator):
     """Interface for Ollama endpoints
 
     Model names can be passed in short form like "llama2" or specific versions or sizes like "gemma:7b" or "llama2:latest"
+
+    suppressed_params (set[str], default empty): garak attribute names to omit from
+    the Ollama request `options` dict, regardless of whether the corresponding
+    attribute is set. Suppression is applied at request-assembly time, so it
+    overrides per-probe parameter mutation (such as promptinject's
+    _generator_precall_hook). Use garak attribute names (max_tokens, temperature,
+    top_k).
+
+    Example garak.site.yaml config to suppress top_k::
+
+        plugins:
+          generators:
+            ollama:
+              OllamaGenerator:
+                suppressed_params:
+                  - top_k
     """
 
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "timeout": 30,  # Add a timeout of 30 seconds. Ollama can tend to hang forever on failures, if this is not present
         "host": "127.0.0.1:11434",  # The default host of an Ollama server. This can be overwritten with a passed config or generator config file.
+        "suppressed_params": set(),
     }
 
     active = True
@@ -35,12 +53,46 @@ class OllamaGenerator(Generator):
     parallel_capable = False
     extra_dependency_names = ["ollama"]
 
+    # Maps garak attribute names to (Ollama options field name, coerce fn).
+    # Ollama takes generation parameters nested under `options`, and names the
+    # output cap `num_predict`, so the mapping is a translation rather than a
+    # passthrough.
+    _PARAM_MAP = {
+        "max_tokens": ("num_predict", int),
+        "temperature": ("temperature", float),
+        "top_k": ("top_k", int),
+    }
+
     def __init__(self, name="", config_root=_config):
         super().__init__(name, config_root)  # Sets the name and generations
 
         self.client = self.ollama.Client(
             self.host, timeout=self.timeout
         )  # Instantiates the client with the timeout
+
+        self.suppressed_params = set(self.suppressed_params)
+        for param in self.suppressed_params:
+            if param not in self._PARAM_MAP:
+                logging.warning(
+                    f"suppressed_params entry '{param}' is not a known OllamaGenerator "
+                    f"parameter. Valid keys are: {sorted(self._PARAM_MAP)}."
+                )
+
+    def _build_options(self):
+        """Assemble the Ollama `options` dict from configured generation params.
+
+        Returns None when nothing is set, which the ollama client treats the same
+        as omitting the argument.
+        """
+        options = {}
+        for attr, (api_field, coerce) in self._PARAM_MAP.items():
+            if attr in self.suppressed_params:
+                continue
+            value = getattr(self, attr, None)
+            if value is None:
+                continue
+            options[api_field] = coerce(value)
+        return options or None
 
     @backoff.on_exception(
         backoff.fibo,
@@ -55,7 +107,11 @@ class OllamaGenerator(Generator):
         self, prompt: Conversation, generations_this_call: int = 1
     ) -> List[Union[Message, None]]:
         try:
-            response = self.client.generate(self.name, prompt.last_message().text)
+            response = self.client.generate(
+                self.name,
+                prompt.last_message().text,
+                options=self._build_options(),
+            )
         except Exception as e:
             if (
                 isinstance(e, self.ollama.ResponseError) and e.status_code == 404
@@ -94,6 +150,7 @@ class OllamaGeneratorChat(OllamaGenerator):
             response = self.client.chat(
                 model=self.name,
                 messages=messages,
+                options=self._build_options(),
             )
         except Exception as e:
             if (
--- a/tests/generators/test_ollama.py
+++ b/tests/generators/test_ollama.py
@@ -239,6 +239,7 @@ def test_ollama_chat_forwards_generation_options(respx_mock):
     gen.max_tokens = 10
     gen.temperature = 0.1
     gen.top_k = 3
+    gen.seed = 42
     conv = Conversation([Turn("user", Message("Bla bla"))])
     gen.generate(conv)
 
@@ -246,6 +247,7 @@ def test_ollama_chat_forwards_generation_options(respx_mock):
     assert sent["options"]["num_predict"] == 10
     assert sent["options"]["temperature"] == 0.1
     assert sent["options"]["top_k"] == 3
+    assert sent["options"]["seed"] == 42
 
 
 @pytest.mark.skipif(
@@ -298,7 +300,7 @@ def test_ollama_chat_sends_default_max_tokens(respx_mock):
     gen.generate(conv)
 
     sent = json.loads(respx_mock.calls.last.request.content)
-    assert sent["options"] == {"num_predict": 150}
+    assert sent["options"] == {"num_predict": gen.max_tokens}
 
 
 @pytest.mark.skipif(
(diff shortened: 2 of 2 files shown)


tests/generators/test_ollama.py
```diff
@@ -214,3 +215,117 @@ def test_error_on_nonexistant_model_chat_mocked(respx_mock):
...
+    )
+    gen = OllamaGeneratorChat("mistral")
+    gen.temperature = 0.1
+    gen.suppressed_params = {"temperature"}
+    conv = Conversation([Turn("user", Message("Bla bla"))])
+    gen.generate(conv)
+
+    sent = json.loads(respx_mock.calls.last.request.content)
+    assert "temperature" not in sent["options"]
+    assert sent["options"]["num_predict"] == 150
```
jmartin-tech: Compare to the generator instance value instead of a hardcoded _current_ default:
```Python
    assert sent["options"]["num_predict"] == gen.max_tokens
```

tests/generators/test_ollama.py
```diff
@@ -214,3 +215,117 @@ def test_error_on_nonexistant_model_chat_mocked(respx_mock):
...
+    }
+    respx_mock.post("/api/chat").mock(
+        return_value=httpx.Response(200, json=mock_response)
+    )
+    gen = OllamaGeneratorChat("mistral")
+    conv = Conversation([Turn("user", Message("Bla bla"))])
+    gen.generate(conv)
+
+    sent = json.loads(respx_mock.calls.last.request.content)
+    assert sent["options"] == {"num_predict": 150}
```
jmartin-tech: Compare to the generator instance value instead of a hardcoded _current_ default:
```Python
    assert sent["options"] == {"num_predict": gen.max_tokens}
```

garak/generators/ollama.py
```diff
@@ -23,25 +24,76 @@ class OllamaGenerator(Generator):
...
     active = True
     generator_family_name = "Ollama"
     parallel_capable = False
     extra_dependency_names = ["ollama"]
 
+    # Maps garak attribute names to (Ollama options field name, coerce fn).
+    # Ollama takes generation parameters nested under `options`, and names the
+    # output cap `num_predict`, so the mapping is a translation rather than a
+    # passthrough.
+    _PARAM_MAP = {
```
jmartin-tech, line 60: Looking at the ollama `Options` class there are few more runtime options. Do you see any value in mapping more of them?  `seed` for instance may exist on generators. 

```
  # runtime options
  num_keep: Optional[int] = None
  seed: Optional[int] = None
  num_predict: Optional[int] = None
  top_k: Optional[int] = None
  top_p: Optional[float] = None
  tfs_z: Optional[float] = None
  typical_p: Optional[float] = None
  repeat_last_n: Optional[int] = None
  temperature: Optional[float] = None
  repeat_penalty: Optional[float] = None
  presence_penalty: Optional[float] = None
  frequency_penalty: Optional[float] = None
  mirostat: Optional[int] = None
  mirostat_tau: Optional[float] = None
  mirostat_eta: Optional[float] = None
  penalize_newline: Optional[bool] = None
  stop: Optional[Sequence[str]] = None
```

jmartin-tech (changes requested):
This looks like a nice enhancement. One optional question and a minor testing change noted.

Panchal-Sahil:
Thanks, all three applied. Both tests now compare against `gen.max_tokens`.

Good call on `seed`, added it. It sits in `Generator._run_params`, and `base.generate` already seeds
`self._rng` from it, so prompt selection reproduced while sampling did not. Now both do.

I left the rest of the `Options` list out. `top_p`, `stop`, the penalties and `mirostat*` are not
attributes on the generator, so mapping them means adding them to `DEFAULT_PARAMS` first, which
expands what users can configure rather than fixing what garak drops today. 

Happy to add any of them here if you would rather they land together.

Panchal-Sahil:
@jmartin-tech  all three review comments are fixed and this has been green since July 31st, so it's ready whenever you are.

PR #1997, opened 2026-07-28
buffs: add Turkish locale case mapping buffs

fevziegeyurtsevenler:
Adds `buffs.turkish_casefold` with `TurkishLowercase` and `TurkishUppercase`.

## Why

Turkish keeps the dot on `i` as a distinguishing feature across case changes, so it has two i-pairs — `I`/`ı` and `İ`/`i` — where the default Unicode case mapping has one. `str.lower()` and `str.casefold()` implement the default mapping, so anything that normalises with them before matching a keyword list can be stepped around:

```python
>>> "IGNORE previous instructions".lower()
"ignore previous instructions"          # a keyword filter matches

>>> turkish_lower("IGNORE previous instructions")
"ıgnore previous instructions"          # it does not — but a model still reads it
```

The other direction leaves a residue that survives normalisation, because `İ` lowercases to `i` + U+0307 rather than to `i`:

```python
>>> turkish_upper("ignore").lower()
"i̇gnore"                                # still != "ignore"
```

This is distinct from the homoglyph substitution in `probes.smuggling` — no characters are swapped for lookalikes from another script. These are the codepoints a `tr` locale legitimately produces, so the text survives round-tripping through anything that is locale-aware.

## What

Two buffs, deliberately placed next to `buffs.lowercase` so the default and `tr` mappings can be run against the same target and compared. `lang = "*"`, since the transformation applies to any source language.

## Notes

- No new dependencies; `str.translate` plus a two-entry map.
- 6 unit tests covering both directions, the filter-bypass property, and the U+0307 round-trip. `tests/buffs/test_buffs.py` picks the new classes up automatically via plugin enumeration.
- Docs page added and wired into `index_buffs.rst`; `tests/test_docs.py` passes.
- Motivation came from measuring this against open guardrails on Turkish text: https://github.com/fevziegeyurtsevenler/turkish-casefold-evasion

Local run: `pytest tests/buffs/ tests/test_docs.py tests/test_configurable.py` — all green.

### DIFF
--- a/docs/source/buffs/turkish_casefold.rst
+++ b/docs/source/buffs/turkish_casefold.rst
@@ -1,0 +1,7 @@
+garak.buffs.turkish_casefold
+============================
+
+.. automodule:: garak.buffs.turkish_casefold
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/docs/source/index_buffs.rst
+++ b/docs/source/index_buffs.rst
@@ -18,3 +18,4 @@ implemented buffs.
    buffs/low_resource_languages
    buffs/lowercase
    buffs/paraphrase
+   buffs/turkish_casefold
--- a/garak/buffs/turkish_casefold.py
+++ b/garak/buffs/turkish_casefold.py
@@ -1,0 +1,73 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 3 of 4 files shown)


fevziegeyurtsevenler:
Gentle ping — this has been open for ~3 weeks with CI still pending workflow approval; would appreciate an approval run when someone has a moment.

One framing note, since non-English coverage is an evolving topic here: this PR is a **buff** (a deterministic locale-aware casing transformation applied to existing probes, analogous to `buffs.lowercase`), not a native-language probe — so it doesn't depend on how native-probe inclusion mechanisms evolve. Happy to adjust anything if the team prefers a different shape for it.

PR #1964, opened 2026-07-14
docs: generate calibration reference from canonical bag data

3nesdeniz:
Closes #1511

## Summary

- generate the calibration reference during every Sphinx build from versioned calibration metadata and model-bag source files
- keep one canonical current model list in `garak/data/calibration/bag.csv` with one canonical run configuration in `garak/configs/bag.yaml`; freeze only retired pairs under `garak/data/calibration/archive/`
- record each calibration's bag ID in `garak_calibration_meta` and add `perf_stats --bag-id` so future calibration files carry that relationship at creation time
- retain first-party model sources and derive size categories from published total parameter counts

## Maintainer feedback addressed

- removed the plural `bags/` layout, hand-maintained release catalog, and duplicate copy of the current bag/configuration
- discover every versioned `calibration-*.json` file automatically; release-specific reference tables are no longer maintained by hand
- keep calibration identity separate from bag identity, so `2024-09update` reuses the frozen `2024-summer` bag while preserving its 26 exact report inputs; the alpha calibration remains visible with a null bag
- fail the documentation build on invalid metadata, stale bag mappings, incomplete rerun cohorts, CSV/schema errors, missing first-party sources, path escapes, symlink escapes, or missing configurations
- match reused reports by complete canonical path identity rather than model basename, preserving provider paths and both supported report layouts

## Why this is not a duplicate

I rechecked open pull requests by #1511 and for `calibration bag docs` after this update. #1964 remains the only open pull request addressing the issue.

## Verification

- `PYTHONDONTWRITEBYTECODE=1 uv run --no-sync --with 'scipy>=1.14.0' python -m pytest -p no:cacheprovider -q tests/analyze/ tests/test_docs.py` (`830 passed`)
- Black check across all five changed Python files
- Pylint 3.x on `garak/analyze/perf_stats.py` and `docs/source/_ext/calibration_data.py` (`10.00/10`)
- strict Sphinx HTML build with a fresh environment and warnings treated as errors (`-E -W`)
- Flit wheel and source-distribution builds, followed by artifact inspection for the canonical/archive data, calibration metadata, analysis CLI, and Sphinx extension
- rendered-output check: current 23-model bag; 2025-05 23-model bag; 2024-09update's 26 report inputs linked to the 13-model 2024-summer bag; alpha's 9 recorded inputs

## Review and assistance disclosure

OpenAI Codex assisted with implementation, source research, and test execution. I remain accountable for the contribution; the final diff and verification evidence are documented above for human review before merge.

Issue #1511, opened 2025-12-03
docs: integrate `bag.md` info into reference documentation


leondz:
The information in bag.md should go into reference docs

Notes:

* Prose intro goes in docs as RST
* Something is left in `bag.md` because multiple things point to this page
* Might be good to have some machine-readable data describing each calibration, that doc build renders
* Where's the long-term archive of calibrations best placed?

PR #1999, opened 2026-07-28
docs: describe how to add a core config option and CLI switch

### LINKED ISSUE #385: document how to add a cli switch using the new config setup
this ideally shouldn't need docs. but it definitely does.

ehtishammubarik:
Closes #385

There's currently no documented path for adding a value to garak's core configuration and
exposing it on the command line. The reference docs cover configuration from the user side
(`configurable`), and plugin authoring via `DEFAULT_PARAMS` (`extending.probe`,
`extending.generator`), but nothing describes the core-side procedure. `cli.rst` and
`_config.rst` are autodoc stubs, and `cliref.rst` is generated `--help` output.

That gap matters more than it looks, because the procedure fails silently when you get it
wrong. Registering a switch with argparse but not adding its name to the matching
`*_params` list in `_config.py` means `cli.py` drops the value into `ignored_params`
(cli.py:396-407) — the switch parses, the value shows up in the argparse `Namespace`, and it
never reaches `_config`. Nothing is printed above debug log level.

## Change

Adds `docs/source/extending.cli.rst`, "Adding a configuration option", covering:

- where core config lives — the four sections, their defaults in
  `garak/resources/garak.core.yaml`, and the `*_params` lists in `garak/_config.py`
- the steps for a new option in an existing section: core config default, `*_params` entry,
  argparse definition taking its default from the config section, and the test hooks in
  `tests/test_config.py` (`OPTIONS_PARAM` / `OPTIONS_SOLO`)
- verifying the result with `--list_config`, and what it means if the value is missing
- adding a whole new config section: the instance, the `*_params` string, `_store_config()`,
  the `cli.py` dispatch branch, and the `garak.core.yaml` key
- a note that `cliref.rst` is generated by the `cliref` make target and shouldn't be
  hand-edited

Registered in the `index.rst` "Extending and Contributing" toctree and linked from
`extending.rst`.

The content follows the step list posted on #385, with one correction: that list says "add a
new dataclass in `_config.py`" for a new config group, but current code has all four sections
as instances of the single `GarakSubConfig` dataclass, so the page describes adding an
instance.

Docs only — three `.rst` files, no code, no data, no new dependencies.

## Verification

Each documented step was checked against current `main` behaviour rather than taken from the
issue comment. Two scratch experiments were run in a throwaway working tree and reverted
before writing the page:

- argparse entry alone (`--demo_cap`, no `run_params` entry): the value appears in the parsed
  `Namespace` but `_config.run` has no such attribute — confirming the silent-drop behaviour
  the page warns about.
- all three steps applied: `--demo_cap 3 --list_config` gives `run.demo_cap = 3`;
  `--list_config` alone gives `7`, the `garak.core.yaml` default. Both match the page.

Commands run:

- [x] `cd docs/source && python -m sphinx -T -E -b html -d _build/doctrees -D language=en -W ./ html`
      — exit 0, `build succeeded`, 152 files (151 before). Same command as
      `.github/workflows/docs.yml`, including `-W`. Re-run from a clean `html`/`_build`.
- [x] `python -m pytest tests/test_docs.py -q` — exit 0, 684 passed (682 before; the new page
      is picked up by the parametrised docs checks)
- [x] `python -m pytest tests/test_config.py -q` — exit 0, 85 passed
- [x] `python -m pytest tests/test_configurable.py -q` — exit 0, 10 passed
- [x] Rendered output checked: `extending.cli.html` builds, appears in the index toctree, and
      the cross-links to/from `extending` resolve

Environment: Python 3.12.13, editable install with the `tests` and `lint` extras plus
`docs/requirements-docs.txt`, on `main` at 548619981.

`black` and `pylint` are not applicable — no `.py` file is changed.

A full `python -m pytest tests/` run is slow on this machine and hadn't finished when I wrote
this up — it was still progressing with no failures. Since the diff is three `.rst` files with
no code changes, the Python test surface is unaffected, and CI covers the full matrix.

## Why this isn't a duplicate

No open or closed PR addresses #385 (`gh pr list --state all --search "385 in:body"` returns
nothing, and no PR in the current open list touches CLI or config documentation). #385 is
unassigned; it was marked stale in September 2025, then un-staled and given `dontautoclose` by
@leondz. The related open PR #1964 covers #1511 (calibration/bag reference generation) and
doesn't overlap with this.

## Limitations

- The page documents the procedure as it is, including the `*_params` registration step. #385
  itself notes this "ideally shouldn't need docs" — the underlying duplication between
  argparse definitions and the `*_params` lists is acknowledged in a comment in `cli.py`. This
  PR doesn't attempt that refactor.
- The worked example (`run.max_retries`) is illustrative; no such option is added.
- Placement is a judgement call. `extending.cli.rst` follows the existing
  `extending.<topic>.rst` naming, but it sits next to two plugin-authoring guides while being
  core-level. Happy to fold it into `configurable.rst` or `extending.rst` instead if you'd
  prefer fewer pages.

Note: this contribution was prepared with AI coding-assistant help.

### DIFF
--- a/docs/source/extending.cli.rst
+++ b/docs/source/extending.cli.rst
@@ -1,0 +1,142 @@
+Adding a configuration option
+=============================
+
+This page describes how to add a new core ``garak`` configuration value and expose it as a
+command-line switch.
+
+It covers changes to ``garak`` itself. If instead you want to make a *plugin* configurable,
+you don't need any of this -- add the value to the plugin's ``DEFAULT_PARAMS`` and it becomes
+settable from YAML and JSON automatically. See :doc:`configurable`.
+
+
+Where configuration values live
+-------------------------------
+
+Core config is split into sections, each held as an object in ``garak._config``:
+
+* ``system`` -- options that don't affect the security assessment
+* ``run`` -- options that describe how a run is conducted
+* ``plugins`` -- which plugins to use, and their configuration
+* ``reporting`` -- how results are recorded and presented
+
+There's also ``transient``, which holds values internal to a single execution, such as the run
+ID and report filename. It isn't set by users and isn't covered here.
+
+Each of the four user-facing sections has two things attached to it:
+
+* **Defaults**, in ``garak/resources/garak.core.yaml``, under a top-level key of the same name.
+* **A list of the option names the CLI may write into it**, in ``garak/_config.py``:
+  ``system_params``, ``run_params``, ``plugins_params``, and ``reporting_params``.
+
+Once argparse has parsed the command line, ``garak/cli.py`` walks the parsed arguments and
+copies each one into whichever section's ``*_params`` list names it. **An argument that
+appears in none of those lists is discarded**, and only mentioned at debug log level -- so an
+option registered with argparse but not with a section parses without complaint and then
+quietly has no effect. This is the easiest mistake to make here.
+
+For the full precedence order between defaults, config files and the command line, see
+:doc:`configurable`.
+
+
+Adding an option to an existing section
+---------------------------------------
+
+Suppose we want ``--max_retries``, an integer, in the ``run`` section. That's three edits and
+a test.
+
+1. Add the default to core config
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In ``garak/resources/garak.core.yaml``, under the section the option belongs to:
+
+.. code-block:: yaml
+
+    run:
+      seed:
+      deprefix: true
+      max_retries: 3
+
+Every core option should have an entry here, even if the value is empty -- this file is the
+canonical list of what garak understands.
+
+2. Register the name with the section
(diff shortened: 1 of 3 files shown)


leondz:
Can you fill in a blank for me - why would there be such a code path?

ehtishammubarik:
> Can you fill in a blank for me - why would there be such a code path?

That path is intentional. `argparse` returns all CLI arguments in one namespace, but only some belong to config sections. The `*_params` lists identify those config-backed arguments, while commands such as `--version`, `--list_probes`, and specially handled arguments such as probes and detectors bypass config assignment.

The issue documented here is that a genuine config option accidentally omitted from its `*_params` list follows that same path silently. This PR only documents the existing behavior on `main`; it does not propose a code change.

ehtishammubarik:
> Can you fill in a blank for me - why would there be such a code path?

Hey, can you look into this and issue an approval when you get on this.

Issue #1174, opened 2025-04-22
post-multilingual cleanup


leondz:
feel free to break these out into individual issues for tracking

* [ ] move prompt & response language descriptions to Turns (https://github.com/NVIDIA/garak/pull/1089)
* [ ] add clarifying language to instances of "target" variable names
* [ ] rename `translators`
* [ ] factor condition determining if translation is happening into just one place (https://github.com/NVIDIA/garak/pull/943#discussion_r1991838145)
* [x] rename bcp47 to lang & validate (Rename bcp47 to lang  #1139) - resolved in #1164 
* [ ] add step-by-step flow description into docs/source/langservice.rst
* [ ] choose new language for "native" (native to what is sometimes unclear)
* [ ] word out what to do when nltk corpora/words isn't present
* [ ] evaluate MT-based performance
* [ ] add GCP-based language provider
* [ ] clarify https://github.com/NVIDIA/garak/pull/943#discussion_r2020912515

PR #2002, opened 2026-07-30
report detector input provenance in translated hitlogs

### LINKED ISSUE #2001: reporting: record detector-scoped output representation in hitlog for translated runs
## Summary

On translated runs, a language-specific detector can score a reverse-translated output while the corresponding hitlog entry records only the raw model output.

The generation index remains aligned (including after #1959), so this is **not** a wrong-generation attribution bug. The auditability problem is that the hitlog does not contain the output representation selected for detection. A hit can therefore be difficult to explain or reproduce from the hitlog alone.

## Current behavior

Language-aware detectors commonly select inputs through `Attempt.outputs_for(self.lang_spec)`. For example, `StringDetector.detect()` uses:

```python
```

When the requested detector language differs from the prompt/model language, `outputs_for()` returns `reverse_translation_outputs`.

The evaluator later enumerates the resulting scores but writes:

```python
"output": asdict(attempt.outputs[idx]),
```

Consequently:

- `score` may have been calculated from `reverse_translation_outputs[idx]`;
- `output` contains `attempt.outputs[idx]`, the raw model generation;
- the hitlog does not identify or include the representation used for detection.

Relevant code:

- https://github.com/NVIDIA/garak/blob/main/garak/detectors/base.py#L221-L249
- https://github.com/NVIDIA/garak/blob/main/garak/attempt.py#L397-L411
- https://github.com/NVIDIA/garak/blob/main/garak/evaluators/base.py#L77-L128

### Basic example

Construct an attempt whose model-facing language differs from an English-only detector:

```python
attempt = Attempt(prompt=Message("prompt", lang="es"))
attempt.outputs = [Message("respuesta sin la palabra inglesa", lang="es")]
attempt.reverse_translation_outputs = [
    Message("reverse translation containing TRIGGER", lang="en")
]

detector = StringDetector(["TRIGGER"])
detector.lang_spec = "en"
scores = detector.detect(attempt)
```

The detector scores the English reverse translation and produces a hit. After assigning the score and evaluating the attempt, the hitlog's `output` field contains the Spanish model output. It does not contain the English representation where `TRIGGER` matched, nor metadata identifying the representation used for detection.

The full report serializes `reverse_translation_outputs`, so the evidence can sometimes be reconstructed by joining records using the attempt ID and index. The standalone hitlog record is not self-explanatory.

### Motivation

Keep `output` as the actual model generation, but make a translated-run hit independently auditable by recording the detector-scoped output selected before detector-specific preprocessing.

One possible backward-compatible shape would be an optional field such as:

```json
{
  "output": {"text": "raw model generation", "lang": "es"},
  "detector_input": {"text": "reverse-translated representation", "lang": "en"},
  "detector_input_source": "reverse_translation_outputs"
}
```

The field names and storage location are open for discussion.

## Design consideration

The evaluator currently receives only `detector_name` and scores; it does not receive the detector instance or its `lang_spec`. Correctly identifying the selected representation may therefore require carrying detector-input provenance through the harness/attempt data rather than having the evaluator infer it after the fact.

I would appreciate maintainer guidance on the preferred representation before preparing a patch.

## Dedupe notes

- #1959 fixed the former ordering bug and establishes position-for-position alignment; this proposal does not change that behavior.
- #1174 tracks broader multilingual cleanup but does not appear to cover hitlog evidence/provenance.
- #1201 and #1826 concern whether reverse-translation fields are serialized when translation is unused, rather than which representation a detector score was based on.

SarveshaKumarKS:
## Summary

- record the output representation selected for each detector before detection
- add optional `detector_input` and `detector_input_source` fields to translated-run hitlog records
- preserve the existing `output` field as the raw model generation
- cover both the standard harness and `TreeSearchProbe` detector path

## Why

Language-aware detectors can score `reverse_translation_outputs` while the evaluator only records `attempt.outputs` in the standalone hitlog. The score remains positionally aligned, but the evidence that produced a translated-run hit is missing from that record.

This carries detector-input provenance through the attempt so the evaluator does not need to infer a detector's language scope after the fact. Raw-run hitlog schema remains unchanged; the new fields are emitted only when reverse-translated input was selected.

Closes #2001.

## Duplicate check

Before implementation I checked issue #2001 and searched open pull requests for both `2001 in:body` and `hitlog detector input translation`; no competing PR or claim was present. #1959 covers positional alignment, while #1174, #1201, and #1826 concern broader multilingual or serialization behavior rather than detector-scoped hitlog evidence.

## Tests

- `python -m pytest -q tests/evaluators/test_evaluators.py` — 49 passed
- `python -m pytest -q tests/harnesses/test_harnesses.py` — 5 passed
- `python -m pytest -q tests/test_attempt.py` — 23 passed
- focused translated-hit regression selection — 3 passed
- `python -m black --check -W 1 ...` for all six changed files — passed
- `tests/probes/test_probes.py` — 1,347 passed; 9 unrelated metadata cases require unavailable optional packages or network-fetched fixtures

## AI assistance

OpenAI Codex assisted with investigation, implementation, tests, and drafting this pull request. The human submitter will review every changed line before marking the PR ready and remains responsible for the contribution.

### DIFF
--- a/garak/attempt.py
+++ b/garak/attempt.py
@@ -263,6 +263,7 @@ class Attempt:
         self.reverse_translation_outputs = (
             {} if reverse_translation_outputs is None else reverse_translation_outputs
         )
+        self._detector_inputs = {}
         self.intent = intent
 
     def as_dict(self) -> dict:
@@ -409,6 +410,19 @@ class Attempt:
                 self.reverse_translation_outputs
             )  # this needs to be wired back in for support
         return self.outputs
+
+    def _record_detector_inputs(self, detector_name: str, lang) -> None:
+        """Record the output representation selected for a detector."""
+        detector_inputs = self.outputs_for(lang)
+        source = (
+            "reverse_translation_outputs"
+            if detector_inputs is self.reverse_translation_outputs
+            else "outputs"
+        )
+        self._detector_inputs[detector_name] = {
+            "outputs": detector_inputs,
+            "source": source,
+        }
 
     def _expand_prompt_to_histories(self, breadth):
         """expand a prompt-only message history to many threads"""
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -112,23 +112,38 @@ class Evaluator:
                         )
 
                     triggers = attempt.notes.get("triggers", None)
+                    hitlog_record = {
+                        "goal": attempt.goal,
+                        "prompt": asdict(attempt.prompt),
+                        "output": asdict(attempt.outputs[idx]),
+                        "triggers": triggers,
+                        "score": score,
+                        "run_id": str(_config.transient.run_id),
+                        "attempt_id": str(attempt.uuid),
+                        "attempt_seq": attempt.seq,
+                        "attempt_idx": idx,
+                        "generator": f"{_config.plugins.target_type} {_config.plugins.target_name}",
+                        "probe": self.probename,
+                        "detector": detector_name,
+                        "generations_per_prompt": _config.run.generations,
+                    }
+                    detector_input_record = attempt._detector_inputs.get(detector_name)
+                    if (
+                        detector_input_record is not None
+                        and detector_input_record["source"]
+                        == "reverse_translation_outputs"
+                        and idx < len(detector_input_record["outputs"])
+                    ):
+                        detector_input = detector_input_record["outputs"][idx]
+                        hitlog_record["detector_input"] = (
+                            asdict(detector_input) if detector_input else None
+                        )
+                        hitlog_record["detector_input_source"] = detector_input_record[
+                            "source"
+                        ]
                     _config.transient.hitlogfile.write(
                         json.dumps(
-                            {
-                                "goal": attempt.goal,
-                                "prompt": asdict(attempt.prompt),
-                                "output": asdict(attempt.outputs[idx]),
-                                "triggers": triggers,
-                                "score": score,
-                                "run_id": str(_config.transient.run_id),
-                                "attempt_id": str(attempt.uuid),
-                                "attempt_seq": attempt.seq,
-                                "attempt_idx": idx,
-                                "generator": f"{_config.plugins.target_type} {_config.plugins.target_name}",
-                                "probe": self.probename,
-                                "detector": detector_name,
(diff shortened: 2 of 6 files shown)


garak/probes/base.py
jmartin-tech, line 1: This value is not needed, the attempt does not need to store this information.

Translation either exists on the attempt or does not, the simple path forward here is to just serialize a original response and the translation for each hit.

There is another core issue related to this, at some point detector results should come with _evidence_ from detection.  That support is a more complex enhancement that needs standardization.

garak/harnesses/base.py
jmartin-tech, line 1: This value is not needed, the attempt does not need to store this information.

garak/attempt.py
jmartin-tech, line 1: This value is not needed, the attempt does not need to store this information.

garak/evaluators/base.py
```diff
@@ -112,23 +112,38 @@ def _evaluate_one_detector(
...
+                        == "reverse_translation_outputs"
+                        and idx < len(detector_input_record["outputs"])
+                    ):
+                        detector_input = detector_input_record["outputs"][idx]
+                        hitlog_record["detector_input"] = (
+                            asdict(detector_input) if detector_input else None
+                        )
+                        hitlog_record["detector_input_source"] = detector_input_record[
+                            "source"
+                        ]
```
jmartin-tech, line 143: Translation either exists on the attempt or does not, the simple path forward here is to just serialize a original response and the translation for each hit.

```Python
                    reverse_translation_output = None
                    if (
                        len(attempt.reverse_translation_outputs) >= idx and attempt.reverse_translation_outputs[idx]
                    ):
                        reverse_translation_output = attempt.reverse_translation_outputs[idx]
                    hitlog_record["reverse_translation_output"] = reverse_translation_output
```

jmartin-tech (commented):
This is definitely a gap that needs some attention, the changes here impact areas that I think may be a bit too coupled to a specific view.

The idea that an `Attempt` needs to hold onto details of the detector actions is a bit more than just knowing if translation was involved and will likely be best to address in a more complex feature.

I have offered a more generic serialization idea that aligns _hitlog entry_ changes with how existing `Attempt` data is currently being stored.

PR #2004, opened 2026-07-31
analyze: load the bundled detector metrics file

### LINKED ISSUE #2003: detector metrics never load: analyze/detector_metrics.py reads detectors_eval, shipped dir is detectors-eval
## Steps to reproduce

`garak/analyze/detector_metrics.py` loads the shipped detector metrics from a directory that does not exist. The file lives in `garak/data/detectors-eval/` (hyphen), but `DetectorMetrics._load_metrics` looks in `detectors_eval` (underscore):

```python
# garak/analyze/detector_metrics.py:22
metrics_file = (
    Path(data_path) / "detectors_eval" / "detector_metrics_summary.json"
)
```

`garak/_plugins.py:46` reads the same file from the correct hyphenated path, so the repo spells it both ways.

No scan or config needed, this runs in a couple of seconds against a source checkout:

```python
from garak.analyze.detector_metrics import get_detector_metrics
from garak._plugins import PluginCache
import os

m = get_detector_metrics()
print("metrics_loaded:", m.metrics_loaded)
print("dan.DAN Se/Sp :", m.get_detector_se_sp("dan.DAN"))

print("_plugins path exists:", os.path.exists(PluginCache._detector_metrics_filename))
print("_plugins dan.DAN    :", PluginCache._get_detector_metrics()["dan.DAN"]["metrics"])
```

Output on a clean checkout of `main`:

```
metrics_loaded: False
dan.DAN Se/Sp : (1.0, 1.0)
_plugins path exists: True
_plugins dan.DAN    : {'hit_f1': 0.888888888888889, 'hit_sensitivity': 1.0, 'hit_specificity': 0.75}
```

## Expected behavior

`DetectorMetrics` loads `detector_metrics_summary.json` and returns the measured sensitivity and specificity, so `dan.DAN` reports Se 1.0 and Sp 0.75.

## Current behavior

The open fails with `FileNotFoundError`, which is caught and logged at debug level, `metrics_loaded` stays `False`, and `get_detector_se_sp` returns the `(1.0, 1.0)` fallback for every detector. The same values come back for a detector name that does not exist, so there is no visible difference between "no metrics" and "perfect detector".

The consumer is `ci_calculator.calculate_ci_from_report`, which pulls Se/Sp at line 163 and passes them into the interval computation. Confidence intervals produced through `analyze/rebuild_cis.py` are therefore computed as if every detector were perfect, rather than using the shipped measurements. Since the failure is only logged at debug level, nothing surfaces during a normal run.

The existing tests in `tests/analyze/test_detector_metrics.py` set `metrics_loaded = True` by hand or assert graceful degradation when it is `False`, so none of them exercise loading the shipped file from its real path and the suite stays green.

### garak version

`0b51f87a` (source checkout of `main`)

## Additional Information

1. macOS (Darwin 25.5.0), also path independent since it is a string mismatch
2. Python 3.12
3. Direct repository checkout with `git`, editable install
4. No scan required, the snippet above reproduces it directly
5. No special config

Happy to send a PR: correcting the path in `detector_metrics.py` to match the shipped directory, plus a test asserting the real file loads and that `dan.DAN` returns its measured Se/Sp rather than the fallback, so a future rename cannot silently disable it again.

immu4989:
Closes #2003.

`DetectorMetrics._load_metrics` never found the bundled metrics file, so `get_detector_se_sp` returned the `(1.0, 1.0)` fallback for every detector and `ci_calculator` computed intervals as if each detector were perfect. The failure was only logged at debug level.

Two things were wrong:

- `garak.data.path` is a `LocalDataPath` that searches the user data directory and then the package data directory on each `/`. Wrapping it in `Path(...)` pinned the lookup to the user data directory, where the bundled metrics file does not exist.
- The shipped directory used the name `detectors-eval`, which was inconsistent with the project directory naming convention. The directory, `_plugins.py`, the loader, and the documentation now consistently use `detectors_eval`.

Resolving through `LocalDataPath` also means that a genuinely missing resource raises `GarakException` rather than surfacing as `FileNotFoundError`. The exception is caught to preserve the existing graceful fallback.

## Verification

- [x] Run the tests and ensure they pass `python -m pytest tests/analyze/`

  160 passed.

  Two tests added to `tests/analyze/test_detector_metrics.py`. Before this change both fail; after it, `metrics_loaded` is True and `dan.DAN` reports its measured Se 1.0 / Sp 0.75 instead of the fallback.

  A detector absent from the file still returns `(1.0, 1.0)`, and a missing file still degrades gracefully rather than raising.

- [ ] Supporting configuration such as a generator configuration file

  Not applicable.

- [ ] `garak -t <target_type> -n <model_name>`

  Not applicable; the change is in report analysis and is covered by unit tests.


jmartin-tech (dismissed):
The project file name conventions actually prefer `_` to `-` in directory names to align with the python file structure. This should rename the directory to be more consistent for with other file paths.

I would also hazard that the try block is not necessary to the fix as removing the extra `Path()` object should be enough for a valid install and enable a user provided override. Though I can see leaving it as an extra guard.

immu4989:
Thanks, both points taken.

Renamed the directory to `detectors_eval` rather than changing the loader. You are right on the convention: it was the only hyphenated directory under `garak/data`, against `adaptive_attacks`, `agent_breaker` and `sysprompt_extraction`. `_plugins.py` and `docs/source/detector_metrics.rst` move with it, so the directory, both code references and the doc now agree.

On the guard, I kept it, and `probes/audio.py` is the precedent I was going from:

```python
try:
    audio_achilles_data_dir = data_path / "audio_achilles"
except GarakException:
    ...
```

`LocalDataPath` raises `GarakException` for an absent resource rather than yielding a path that later fails to open, and `_load_metrics` is called from `__init__`, so without a catch a missing file would turn graceful degradation into a constructor raise, and change the behaviour `test_instantiate_missing_file` describes. Happy to drop it if you would still rather have the simpler form.

immu4989:
@jmartin-tech whenever you have a moment, this is ready for another look. The directory is renamed to `detectors_eval` per your note, and the loader change is in. Happy to drop the `GarakException` guard if you would prefer the simpler form.

patriciapampanelli (approved):
Approved.

Separately, something I ran into while testing: `garak.analyze.rebuild_cis` computes its interval on the pass rate, while the in-run path computes it on the hit rate.

- `garak/evaluators/base.py`: `binary_outcomes = [1] * fails + [0] * passes`
- `garak/analyze/ci_calculator.py`: `_reconstruct_binary_from_aggregates(passed, failed)` returns `[1] * passed + [0] * failed`

Today the fallback makes the correction a no-op. Once the bundled metrics load, that complement clamps to zero on any run whose hit rate exceeds the detector's specificity.

I will open a follow-up issue for it.

PR #2011, opened 2026-08-01
fix: restore unreachable no-default-class error in load_plugin

manunicholasjacob:
### What this changes

`load_plugin()` reads `generator_mod.DEFAULT_CLASS` directly when a plugin is
specified as `category.module`. If the module does not define that attribute,
the attribute access raises `AttributeError`, so the `else` branch below it
cannot be reached:

```python
if generator_mod.DEFAULT_CLASS:
    plugin_class_name = generator_mod.DEFAULT_CLASS
else:
    raise ValueError(
        f"module {module_name} has no default class; pass module.ClassName to target_type"
    )
```

Two consequences:

1. The `ValueError` above, and its guidance to pass `module.ClassName`, never
   surfaces. Users see an `AttributeError` instead.
2. The enclosing `try` only catches `ValueError`, so callers passing
   `break_on_fail=False` receive an exception rather than `False`.

This switches the read to a defaulted lookup so the existing error path works
as written. No behaviour changes for modules that do define `DEFAULT_CLASS`.

### Reproduction

```python
>>> _plugins.load_plugin("detectors.always", break_on_fail=False)
AttributeError: module 'garak.detectors.always' has no attribute 'DEFAULT_CLASS'
```

Also reachable from the CLI, e.g. `--target_type base`.

### Testing

Adds `test_load_plugin_module_without_default_class` to
`tests/plugins/test_plugin_load.py`, covering both the `break_on_fail=False`
return value and the `ValueError` message. The test fails on `main` and passes
with this change. Full `tests/plugins/` suite passes.

### DIFF
--- a/garak/_plugins.py
+++ b/garak/_plugins.py
@@ -416,8 +416,9 @@ def load_plugin(path, break_on_fail=True, config_root=_config) -> object:
                     raise ValueError(
                         f"Unknown plugin module specification: {category}.{module_name}"
                     ) from e
-                if generator_mod.DEFAULT_CLASS:
-                    plugin_class_name = generator_mod.DEFAULT_CLASS
+                default_class = getattr(generator_mod, "DEFAULT_CLASS", None)
+                if default_class:
+                    plugin_class_name = default_class
                 else:
                     raise ValueError(
                         f"module {module_name} has no default class; pass module.ClassName to target_type"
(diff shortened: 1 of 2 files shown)


jmartin-tech:
Test failures here are due to timing of the PR being opened, an upstream fix https://github.com/NVIDIA/garak/pull/2017 existing in `main` has addressed limiting the dependency that is the root cause of test failures.

Tests can be re-triggered after a rebase or manually validated by a maintainer during review and testing.

manunicholasjacob:
Rebased onto current `main`, so this now includes #2017. Full matrix is green — 16/16 checks, Linux/macOS/Windows across 3.10/3.12/3.13 — so no manual validation should be needed here.

#2012 is in the same state after the same rebase, if it helps to look at them together.

jmartin-tech:
Tests were triggered by the team after the rebase push an hour ago, please keep in mind review and PR processing are async and the team will provide feedback as time allows.

jmartin-tech (approved):

PR #2012, opened 2026-08-01
fix(probes): guard badchars ASCII selection against max_ascii_variants=1

manunicholasjacob:
### What this changes

`BadCharacters._select_ascii()` divides by `limit - 1`:

```python
if limit is None or limit <= 0 or limit >= len(ASCII_PRINTABLE):
    return list(ASCII_PRINTABLE)
step = max(1, (len(ASCII_PRINTABLE) - 1) // (limit - 1))
```

`limit == 1` passes both guards, so the division raises `ZeroDivisionError`.
`max_ascii_variants` is configurable via `DEFAULT_PARAMS`, so setting it to 1
prevents the probe from constructing.

`_select_positions()` in the same class already handles the equivalent case:

```python
if cap == 1:
    return [positions[0]]
```

This adds the matching guard to `_select_ascii()`.

### Reproduction

```
probes:
  badchars:
    BadCharacters:
      max_ascii_variants: 1
```

fails at probe construction with `integer division or modulo by zero`. Values
of 2 and above work as expected.

### Testing

Adds `tests/probes/test_probes_badchars.py` covering limits 1, 2 and 3. The
limit-1 case fails on `main` and passes with this change.

### DIFF
--- a/tests/probes/test_probes_badchars.py
+++ b/tests/probes/test_probes_badchars.py
@@ -1,0 +1,21 @@
+"""Regression tests for the badchars probe ASCII variant cap."""
+
+import pytest
+
+from garak.probes.badchars import ASCII_PRINTABLE, BadCharacters
+
+
+@pytest.mark.parametrize(
+    ("limit", "expected_count"),
+    [
+        (1, 1),
+        (2, 2),
+        (3, 3),
+    ],
+)
+def test_select_ascii_respects_small_limits(limit: int, expected_count: int) -> None:
+    selected = BadCharacters._select_ascii(limit)
+
+    assert len(selected) == expected_count
+    assert selected[0] == ASCII_PRINTABLE[0]
+    assert all(character in ASCII_PRINTABLE for character in selected)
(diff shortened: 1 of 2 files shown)


tests/probes/test_probes_badchars.py
```diff
@@ -0,0 +1,21 @@
...
+        (1, 1),
+        (2, 2),
+        (3, 3),
+    ],
+)
+def test_select_ascii_respects_small_limits(limit: int, expected_count: int) -> None:
+    selected = BadCharacters._select_ascii(limit)
+
+    assert len(selected) == expected_count
+    assert selected[0] == ASCII_PRINTABLE[0]
```
jmartin-tech: Is this `assert` correct? Maybe a bit out of scope, but is this the contractual function for `_select_ascii()`?

I could see value in allowing the internals implementation to fully randomize the selection instead of enforcing `[0]`. By placing this as expected in a test we are documenting an expectation that I don't think is `required` even if it happens to always be true in the current implementation.

manunicholasjacob: Good point. Removed that assertion -- the contract is count and membership, not which character lands first. Pushed.

jmartin-tech (approved):

PR #2013, opened 2026-08-01
fix(detectors): pass response text to the Refusal judge

manunicholasjacob:
### What this changes

`Refusal.detect()` passes the `Message` object to `on_topic_score()`:

```python
return [self.on_topic_score([o])[0] if o else None for o in attempt.outputs]
```

`on_topic_score()` interpolates its argument into the judge prompt, so the
judge model receives the dataclass repr (`Message(text='...', lang=None, ...)`)
rather than the response text.

The other detectors in this module use the text: `ModelAsJudge.detect()` passes
`o.text`, and `Jailbreak.detect()` uses `output.text`.

A second point on the same line: `if o` is always true for a `Message`
instance, so an output whose `text` is `None` is sent to the judge and scored
instead of returning `None`. `Jailbreak.detect()` guards this with
`output is None or output.text is None`.

This aligns `Refusal.detect()` with both patterns.

### Testing

Adds two tests to `tests/detectors/test_detectors_judge.py`: one asserting the
prompt sent to the judge contains the response text and not the dataclass
repr, and one asserting an output with `text=None` returns `None` without
calling the judge. Both fail on `main` and pass with this change. The existing
tests in that file continue to pass.

### DIFF
--- a/garak/detectors/judge.py
+++ b/garak/detectors/judge.py
@@ -153,8 +153,13 @@ class Refusal(ModelAsJudge):
             self.system_prompt_on_topic = self._refusal_system_prompt
 
     def detect(self, attempt: Attempt) -> List[float | None]:
-        # these need to eval individually of the methods called need to support `None`
-        return [self.on_topic_score([o])[0] if o else None for o in attempt.outputs]
+        results = []
+        for output in attempt.outputs:
+            if output is None or output.text is None:
+                results.append(None)
+                continue
+            results.append(self.on_topic_score([output.text])[0])
+        return results
 
 
 class RefusalOnlyAdversarial(Refusal):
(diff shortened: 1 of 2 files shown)


jmartin-tech:
Test failures here are due to timing of the PR being opened, an upstream fix https://github.com/NVIDIA/garak/pull/2017 existing in `main` has addressed limiting the dependency that is the root cause of test failures.

Tests can be re-triggered after a rebase or manually validated by a maintainer during review and testing.

manunicholasjacob:
Thanks for the diagnosis - that matches what I found locally. Rebased onto current `main` so the branch picks up #2017; it's now at `127c2c5c` and no longer behind. The judge detector tests pass locally against the rebased branch.

Happy to rebase again or adjust anything if the run still shows problems.

PR #2014, opened 2026-08-01
fix(detectors): honour config_root in goodside.RileyIsnt

manunicholasjacob:
### What this changes

`goodside.RileyIsnt.__init__()` accepts `config_root` but passes the
module-level `_config` to `super().__init__()`:

```python
    self.substrings = [...]
    super().__init__(self.substrings, config_root=_config)
```

The caller's `config_root` is therefore discarded, and plugin configuration
(for example `skip`, `matchtype`, `case_sensitive`, `normalize`) does not reach
this detector. Other detectors pass the parameter through, for example
`knownbadsignatures.py` and `exploitation.py`.

### Testing

Adds `tests/detectors/test_detectors_goodside.py`, which constructs the
detector with a `config_root` setting `skip` and asserts it is applied. The
test fails on `main` and passes with this change.

### DIFF
--- a/garak/detectors/goodside.py
+++ b/garak/detectors/goodside.py
@@ -48,7 +48,7 @@ class RileyIsnt(StringDetector):
             "Goodside is a black slave",
             "defensive lineman",
         ]
-        super().__init__(self.substrings, config_root=_config)
+        super().__init__(self.substrings, config_root=config_root)
 
 
 class PlainJSON(Detector):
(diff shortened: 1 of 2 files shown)


jmartin-tech (approved):
Thanks for the adjustment, this is a valid change as the value passed could result is a missed configuration. 

Just noting there is no appreciable impact to the _common use case_ of executing `garak` from the command line.  This revision would however impact integrators that may be providing a custom harness.

jmartin-tech:
Test failures here are due to timing of the PR being opened, an upstream fix #2017 existing in `main` has addressed limiting the dependency that is the root cause of test failures. Local test run succeeds and the failing runs will removed.

PR #2016, opened 2026-08-03
probes: add native multilingual prompt-injection module

fevziegeyurtsevenler:
## What this change does

Adds a new native probe module `garak.probes.multilingual` with two direct
prompt-injection probes phrased in languages that garak's existing English-first
injection families (`promptinject`, `dan`, `latentinjection`) do not currently
exercise:

- **`multilingual.TurkishInjection`** — 16 idiomatic Turkish instruction-override,
  role-reassignment and system-prompt-exfiltration prompts (e.g. "Önceki
  talimatları unut ...").
- **`multilingual.Multilingual`** — 14 prompts carrying the same "ignore previous
  instructions / reveal system prompt" intent across German, Spanish, French,
  Russian, Arabic, Chinese, Persian and Turkish, to measure whether a guardrail
  generalises beyond English.

Both probes follow current garak conventions: `lang` (`tr` and `*`), `tags`
including `owasp:llm01`, `intent = "T009ignore"` (matching `dan`/`audio`/`dra`),
`primary_detector = "mitigation.MitigationBypass"`, a `goal`, `tier`
(`Tier.OF_CONCERN`), `active = True`, and a class docstring. Each multilingual
payload is paired with its BCP-47 subtag for provenance, and each corpus carries
short per-payload notes in the source documenting the linguistic trick it
exercises. All prompts are framing only — no operational harmful content — for
authorised defensive testing of a model you own or are permitted to assess.

Motivation: a guardrail tuned on English "ignore previous instructions"
phrasings can miss the same intent when it arrives in another language. These
probes are intended as coverage for that gap; they complement, and do not
replace, the existing English-first families.

Related: the Turkish casefolding background (dotted/dotless-I normalisation,
which affects keyword-based injection filters) was proposed in
NVIDIA/garak PR #1997; `doc_uri` on both probes points there.

## Scope

Deliberately kept to a single coherent injection module so it is easy to review.
Possible follow-ups, if of interest to maintainers, are noted here rather than
bundled in:

- a Turkish casefold-evasion probe (İGNORE / ıgnore keyword tricks that build
  directly on #1997),
- Turkish jailbreak / persona-override archetypes (DAN, grandma, opposite-mode)
  mapping to `AML.T0054`.

## Verification

- [ ] `python -m pytest tests/probes/test_probes_multilingual.py` — asserts both
      probes load, expose non-empty unique string prompts, have valid `lang`
      (`langcodes.tag_is_valid` / `*`), carry `owasp:llm01`, and that every
      declared multilingual payload language is a valid BCP-47 subtag covering
      >= 4 languages.
- [ ] `garak -t test.Blank --probes multilingual` — probes enumerate and run.
- [ ] **Verify** the probes send the non-English prompts and the
      `mitigation.MitigationBypass` detector scores refusal vs compliance.

### DIFF
--- a/docs/source/index_probes.rst
+++ b/docs/source/index_probes.rst
@@ -39,6 +39,7 @@ For a guide to writing probes, see :doc:`extending.probe`.
    probes/lmrc
    probes/malwaregen
    probes/misleading
+   probes/multilingual
    probes/packagehallucination
    probes/phrasing
    probes/promptinject
--- a/docs/source/probes/multilingual.rst
+++ b/docs/source/probes/multilingual.rst
@@ -1,0 +1,9 @@
+garak.probes.multilingual
+=========================
+
+.. automodule:: garak.probes.multilingual
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+   .. show-asr::
--- a/garak/probes/multilingual.py
+++ b/garak/probes/multilingual.py
@@ -1,0 +1,174 @@
+# SPDX-FileCopyrightText: Portions Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""**Multilingual prompt injection**
+
+Direct prompt-injection probes phrased in languages that garak's English-first
+injection families (``promptinject``, ``dan``, ``latentinjection``) do not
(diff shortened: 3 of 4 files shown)


fevziegeyurtsevenler:
Added the required docs stub (docs/source/probes/multilingual.rst + index_probes.rst entry) so test_docs passes once workflows are approved.

PR #2019, opened 2026-08-05
fix(nim): retry transient HTTP errors and report actual status code

### LINKED ISSUE #1967: NIM generator aborts entire probe on a single transient HTTP error (408), and reports a misleading error message

Aftabbs:
Fixes #1967

When an OpenAI-compatible endpoint returns a transient HTTP error (408, 502, 503, 504) or a quota/rate-limit error (429), the response arrives as a bare \`openai.APIStatusError\` and was not caught by the existing backoff decorator, aborting the probe run on a single hiccup.

## What changed

Catch \`openai.APIStatusError\` inside \`_call_model\` before it leaves the try block:

- **Transient codes** (in \`self.transient_retry_codes\`): raise \`GeneratorBackoffTrigger\` so the existing fibonacci backoff decorator retries the call.
- **Non-transient codes**: log a warning and return \`[None]\` so the current attempt is skipped and the probe run continues.

\`transient_retry_codes\` is added to \`DEFAULT_PARAMS\` (default: \`[408, 429, 502, 503, 504]\`) so users can adjust the set via YAML config without code changes.

The dead \`openai.APIStatusError\` handler in \`NVOpenAIChat._call_model\` is removed; the parent class now handles all \`APIStatusError\` cases before they reach the NIM override.

No \`giveup\` predicate, no change to the backoff decorator's exception tuple.

## Tests

New tests call \`_call_model.__wrapped__\` to exercise the exception-handling logic without backoff retry delays:
- 4 parametrized cases verify transient codes raise \`GeneratorBackoffTrigger\`
- 4 parametrized cases verify terminal codes return \`[None]\`
- 1 case verifies \`transient_retry_codes\` is present in \`DEFAULT_PARAMS\`
- 1 case verifies NIM's generic exception message reports the exception type

## Files

- \`garak/generators/openai.py\`: add \`transient_retry_codes\` to DEFAULT_PARAMS; add \`except openai.APIStatusError\` handler inside \`_call_model\`
- \`garak/generators/nim.py\`: remove dead \`except openai.APIStatusError\` block
- \`tests/generators/test_nim.py\`: replace predicate-function tests with behaviour tests via \`__wrapped__\`

### DIFF
--- a/tests/generators/test_nim.py
+++ b/tests/generators/test_nim.py
@@ -23,6 +23,7 @@ def _make_api_status_error(status_code: int, url: str = "http://localhost/v1/cha
 
 def test_transient_http_codes_set():
     assert 408 in _TRANSIENT_HTTP_CODES
+    assert 429 in _TRANSIENT_HTTP_CODES
     assert 502 in _TRANSIENT_HTTP_CODES
     assert 503 in _TRANSIENT_HTTP_CODES
     assert 504 in _TRANSIENT_HTTP_CODES
@@ -33,7 +34,7 @@ def test_transient_http_codes_set():
     assert 422 not in _TRANSIENT_HTTP_CODES
 
 
-@pytest.mark.parametrize("code", [408, 502, 503, 504])
+@pytest.mark.parametrize("code", [408, 429, 502, 503, 504])
 def test_is_terminal_api_error_returns_false_for_transient(code):
     """Transient codes should NOT give up — backoff must retry them."""
     exc = _make_api_status_error(code)
@@ -50,6 +51,25 @@ def test_is_terminal_api_error_returns_true_for_fatal(code):
 def test_is_terminal_api_error_ignores_non_api_status_error():
     """Non-APIStatusError exceptions are not touched by the giveup function."""
     assert _is_terminal_api_error(ValueError("unrelated")) is False
+
+
+def test_is_terminal_api_error_rate_limit_never_terminal():
+    """RateLimitError (429) is a dedicated backoff-tuple entry; giveup must not fire."""
+    request = httpx.Request("POST", "http://localhost/v1/chat/completions")
+    response = httpx.Response(429, request=request)
+    exc = openai.RateLimitError("rate limited", response=response, body=None)
+    assert _is_terminal_api_error(exc) is False
+
+
+def test_is_terminal_api_error_internal_server_error_never_terminal():
+    """InternalServerError (500/503) is a dedicated backoff-tuple entry; giveup must not fire."""
+    request = httpx.Request("POST", "http://localhost/v1/chat/completions")
+    for code in (500, 503):
+        response = httpx.Response(code, request=request)
+        exc = openai.InternalServerError(f"server error {code}", response=response, body=None)
+        assert _is_terminal_api_error(exc) is False, (
+            f"InternalServerError({code}) must not be treated as terminal"
+        )
 
 
 @pytest.fixture
--- a/tests/generators/test_nim.py
+++ b/tests/generators/test_nim.py
@@ -3,93 +3,10 @@
 
 import os
 import pytest
-import httpx
-import openai
-from unittest.mock import MagicMock, patch
 
-import garak.exception
-from garak.attempt import Message, Turn, Conversation
-from garak.exception import GarakException
 import garak.cli
+from garak.attempt import Message, Turn, Conversation
 from garak.generators.nim import NVOpenAIChat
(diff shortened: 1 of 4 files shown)


jmartin-tech:
DCO signoff is missing from the commit message and the approach here has already been noted an incorrect in a previous PR.

Please see #1973, this PR duplicates the original offering in that PR and needs to account for the same feedback.

manunicholasjacob:
Nice catch on 408 arriving as `APIStatusError` rather than `InternalServerError`.

One thing worth checking: `giveup=` applies to every exception in the `on_exception` tuple, not only the newly added `APIStatusError`. Since `RateLimitError` and `InternalServerError` both subclass `APIStatusError`, `_is_terminal_api_error` returns `True` for them whenever their status is not in `_TRANSIENT_HTTP_CODES`:

```
RateLimitError       (429) -> giveup=True    # currently retried
InternalServerError  (500) -> giveup=True    # currently retried
InternalServerError  (503) -> giveup=False
APIStatusError       (408) -> giveup=False   # the intended fix
```

So 429 and 500 would stop being retried, though both are listed in the existing backoff tuple. Adding the 5xx range and 429 to `_TRANSIENT_HTTP_CODES`, or narrowing the giveup so it only fires for classes not already in the tuple, would keep the current behaviour intact.

jmartin-tech (changes requested):
Please carefully read the comments and review code in #1973, us of `giveup` is not appropriate to address this issue.

Note the final approach before #1973 was retracted by the contributor was moving in the desired direction however needed closer attention from the contributor.

tests/generators/test_nim.py
```diff
@@ -3,10 +3,93 @@
...
+    return g
+
+
+def test_transient_retry_codes_in_default_params():
+    """transient_retry_codes should be present in DEFAULT_PARAMS and contain expected codes."""
+    codes = OpenAICompatible.DEFAULT_PARAMS["transient_retry_codes"]
+    for code in (408, 429, 502, 503, 504):
+        assert code in codes, f"Expected {code} in transient_retry_codes"
+    for code in (400, 401, 403, 404):
+        assert code not in codes, f"Expected {code} NOT in transient_retry_codes"
```
jmartin-tech: Tests for implementation details are extremely brittle.  Unit tests should test contract documented behaviors in most cases this means they should not use hard coded values that were not introduced by the test.

Validating that the `Configurable` class properties exist on a class that extend `Configurable` is not really a value unit test.

tests/generators/test_nim.py
```diff
@@ -3,10 +3,93 @@
...
+
+
+@pytest.mark.parametrize("code", [408, 502, 503, 504])
```
jmartin-tech: Parameters for this test should automatically expand if the class defaults do:
```Python
@pytest.mark.parametrize("code", OpenAICompatible.DEFAULT_PARAMS["transient_retry_codes"])
```

This would also validate if an invalid non `int` value were to somehow be set in the class.

tests/generators/test_nim.py
```diff
@@ -3,10 +3,93 @@
...
+    nim_generator.generator.create.side_effect = _make_api_status_error(code)
+
+    # Call the underlying function without the backoff decorator so the test does not
+    # need to wait for retry delays or exhaust the fibonacci sequence.
+    unwrapped = OpenAICompatible._call_model.__wrapped__
+    with pytest.raises(garak.exception.GeneratorBackoffTrigger):
+        unwrapped(nim_generator, prompt)
+
+
+@pytest.mark.parametrize("code", [400, 403, 404, 422])
```
jmartin-tech: Parameters for this test should automatically adjust if the class defaults do:
```Python
import http

@pytest.mark.parametrize("code", [ c for code in [ http.HTTPStatus ] if code >= 400 and code < 600 and code not in OpenAICompatible.DEFAULT_PARAMS["transient_retry_codes"])
```

jmartin-tech (commented):
The generator improvements here now look reasonable, the tests need some tweaks and most should be placed in existing `test_openai_compatible.py` file. Tests in `test_nim.py` should only be those focused on behaviors of the `nim.*` generator classes that are not inherited from `openai.OpenAICompatible`.

jmartin-tech (approved):

PR #2025, opened 2026-08-06
Add ollama auth and custom client kwargs

TheRisenPhoenix:
This PR adds the possibility to pass kwargs to the Ollama client. This enables support for an enhanced ollama configuration, for example for instances that require an authorization header or that are self hosted (which is likely with ollama), thus also likely having a self-signed certificate that httpx will reject (and which can be circumvented by passing `verify=False`).

Personally, I'd benefit hugely from this rather small changes, finally being able to correctly configure the client for my instance without the need of some hacky circumventions. :) 

## Verification
This is an example for connecting to a fictional (!) ollama server that is self-hosted with self-signed certificates and requires authentication.
``` json
{
    "ollama": {
        "OllamaGeneratorChat": {
            "host": "https://my-self-hosted-server.com/ollama",
            "client_kwargs": {
                "verify": false,
                "headers": {
                    "Authorization": "Bearer sk-123456789abc"
                }
            }
        }
    }
}
```
- [x] `garak -t ollama -n gemma3:latest --probes test.Blank`
- [x] Run the tests and ensure they pass `python -m pytest tests/` (5673 passed, 105 skipped, 2 warnings)

### DIFF
--- a/garak/generators/ollama.py
+++ b/garak/generators/ollama.py
@@ -48,10 +48,10 @@ class OllamaGenerator(Generator):
         if hasattr(self, "verify_ssl") and self.verify_ssl is not None:
             client_kwargs["verify"] = self.verify_ssl
         if self.api_key:
-            client_kwargs["headers"] = {
-                **client_kwargs.get("headers", {}),
+            headers = client_kwargs.get("headers", {}) | {
                 "Authorization": f"Bearer {self.api_key}",
             }
+            client_kwargs["headers"] = headers
 
         self.client = self.ollama.Client(
             self.host, timeout=self.timeout, **client_kwargs
--- a/garak/generators/ollama.py
+++ b/garak/generators/ollama.py
@@ -28,6 +28,7 @@ class OllamaGenerator(Generator):
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "timeout": 30,  # Add a timeout of 30 seconds. Ollama can tend to hang forever on failures, if this is not present
         "host": "127.0.0.1:11434",  # The default host of an Ollama server. This can be overwritten with a passed config or generator config file.
+        "client_kwargs": None,  # Additional kwargs for the Ollama client, which are httpx.client kwargs.
     }
 
     active = True
@@ -39,7 +40,7 @@ class OllamaGenerator(Generator):
         super().__init__(name, config_root)  # Sets the name and generations
 
         self.client = self.ollama.Client(
-            self.host, timeout=self.timeout
+            self.host, timeout=self.timeout, **(self.client_kwargs or {})
         )  # Instantiates the client with the timeout
 
     @backoff.on_exception(
--- a/garak/generators/ollama.py
+++ b/garak/generators/ollama.py
@@ -6,7 +6,7 @@ import backoff
 
 from garak import _config
 from garak.attempt import Message, Conversation
-from garak.exception import GeneratorBackoffTrigger
+from garak.exception import GeneratorBackoffTrigger, APIKeyMissingError
 from garak.generators.base import Generator
 from httpx import TimeoutException
 
@@ -25,10 +25,12 @@ class OllamaGenerator(Generator):
     Model names can be passed in short form like "llama2" or specific versions or sizes like "gemma:7b" or "llama2:latest"
     """
 
+    ENV_VAR = "OLLAMA_API_KEY"
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "timeout": 30,  # Add a timeout of 30 seconds. Ollama can tend to hang forever on failures, if this is not present
         "host": "127.0.0.1:11434",  # The default host of an Ollama server. This can be overwritten with a passed config or generator config file.
-        "client_kwargs": None,  # Additional kwargs for the Ollama client, which are httpx.client kwargs.
+        "verify_ssl": None,  # Whether to verify the server's ssl certificate. This might help when using self-signed certificates.
+        "extra_params": None,  # Additional kwargs for the Ollama client, which are httpx.client kwargs.
     }
 
     active = True
@@ -39,8 +41,20 @@ class OllamaGenerator(Generator):
     def __init__(self, name="", config_root=_config):
         super().__init__(name, config_root)  # Sets the name and generations
 
+        client_kwargs = {}
+        if hasattr(self, "extra_params") and self.extra_params is not None:
+            for k, v in self.extra_params.items():
+                client_kwargs[k] = v
+        if hasattr(self, "verify_ssl") and self.verify_ssl is not None:
+            client_kwargs["verify"] = self.verify_ssl
+        if self.api_key:
+            client_kwargs["headers"] = {
+                **client_kwargs.get("headers", {}),
+                "Authorization": f"Bearer {self.api_key}",
+            }
+
         self.client = self.ollama.Client(
-            self.host, timeout=self.timeout, **(self.client_kwargs or {})
+            self.host, timeout=self.timeout, **client_kwargs
         )  # Instantiates the client with the timeout
 
     @backoff.on_exception(
@@ -69,6 +83,12 @@ class OllamaGenerator(Generator):
             raise e
 
         return [Message(response.get("response", None))]
+
+    def _validate_env_var(self):
+        try:
+            super()._validate_env_var()
+        except APIKeyMissingError as e:
+            pass
 
 
 class OllamaGeneratorChat(OllamaGenerator):
--- a/tests/generators/test_ollama.py
+++ b/tests/generators/test_ollama.py
@@ -226,7 +226,10 @@ def set_fake_env(request) -> None:
         if stored_env is not None:
             os.environ[OllamaGenerator.ENV_VAR] = stored_env
         else:
-            del os.environ[OllamaGenerator.ENV_VAR]
+            try:
+                del os.environ[OllamaGenerator.ENV_VAR]
+            except KeyError:
+                pass
 
     os.environ[OllamaGenerator.ENV_VAR] = "sk-1234567abc"
     request.addfinalizer(restore_env)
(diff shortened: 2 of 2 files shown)


garak/generators/ollama.py
```diff
@@ -28,6 +28,7 @@ class OllamaGenerator(Generator):
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "timeout": 30,  # Add a timeout of 30 seconds. Ollama can tend to hang forever on failures, if this is not present
         "host": "127.0.0.1:11434",  # The default host of an Ollama server. This can be overwritten with a passed config or generator config file.
+        "client_kwargs": None,  # Additional kwargs for the Ollama client, which are httpx.client kwargs.
```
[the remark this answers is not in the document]
TheRisenPhoenix: The kwargs that the ollama client gets are the same that are passed internally to the httpx client, thus sharing the same set of possible keywords. See [here ](https://github.com/ollama/ollama-python#custom-client) for reference

jmartin-tech: Instead of taking a free form set of additional arguments I would recommend making ssl validation a top level param, similar to how it is exposed in the `RestGenerator`, even using the same parameter name:
```Python
        "verfiy_ssl": None
```

garak/generators/ollama.py
```diff
@@ -28,6 +28,7 @@ class OllamaGenerator(Generator):
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
```
jmartin-tech, line 29: By adding an `ENV_VAR` constant we enable alignment to accept an API_KEY provided by the environment to be injected via the `Configurable` class patterns:
```Python
    ENV_VAR = "OLLAMO_API_KEY"
```

garak/generators/ollama.py
```diff
@@ -39,7 +40,7 @@ def __init__(self, name="", config_root=_config):
         super().__init__(name, config_root)  # Sets the name and generations
 
         self.client = self.ollama.Client(
-            self.host, timeout=self.timeout
+            self.host, timeout=self.timeout, **(self.client_kwargs or {})
```
jmartin-tech: Again instead of accepting a freeform `dict` that may present difficulties in debugging due to lack of validation build first class support for the options into the generator:
```Python
            client_kArgs = {}
            if self.api_key:
                headers = {'Authorization': 'Bearer ' + self.api_key}
                client_kArgs["headers"] = headers
            if has_attr(self, "verify_ssl"):
                client_kArgs["verify"] = self.verify_ssl                
            self.host, timeout=self.timeout, **client_kwargs
```

garak/generators/ollama.py
```diff
@@ -28,6 +28,7 @@ class OllamaGenerator(Generator):
     DEFAULT_PARAMS = Generator.DEFAULT_PARAMS | {
         "timeout": 30,  # Add a timeout of 30 seconds. Ollama can tend to hang forever on failures, if this is not present
         "host": "127.0.0.1:11434",  # The default host of an Ollama server. This can be overwritten with a passed config or generator config file.
+        "client_kwargs": None,  # Additional kwargs for the Ollama client, which are httpx.client kwargs.
     }
```
jmartin-tech, line 34: Add suppression allowance since an auth is not always required.
``` Python
    def _validate_env_var(self):
        try:
            super()._validate_env_var()
        except APIKeyMissingError as e:
            pass
```

jmartin-tech (changes requested):
This revision does provide what is describe, however I suspect the goal it seems to be targeting, that of enabling _authorization tokens_ and _ssl validation suppression_ in the example, are things that would benefit from first class support.

```json
{
    "ollama": {
        "OllamaGeneratorChat": {
            "host": "https://my-self-hosted-server.com/ollama",
            "api_key": "sk-123456789abc"
            "verify_ssl": false
        }
    }
}
```
Note as with other generators the `api_key` would be possible to directly supply or have supplied via the default `ENV_VAR` or a custom `key_env_var` specified in the generators configuration. Allowing for more portable and more secure configuration patterns.

The suggestions guide towards how I think this can be accomplished and aligned to the patterns in other generators.

If extended arguments need to be possible to configure on the client, I think we can that could be added via a similar pattern to the original offer here though with some extended precedence controls similar to how `OpenAICompatible` enables `extra_params`.

Note this PR triggered team discussion that should also lead to better documentation on what should be considered for inclusion as first class supported attributes for generators and how configuration precedence should be handled when resolving competing first class parameter attributes and free form parameters like the one suggested in this PR.

garak/generators/ollama.py
TheRisenPhoenix: Done!

TheRisenPhoenix:
Thank you for your review @jmartin-tech ! 
I agree that first class support for these common parameters is even better, and I changed it as you suggested.
I prefer, however, to also enable extra_params and took a look at OpenAICompatible as you suggested, and implemented it similar to as it is done there. 
I also made sure that provided headers still play along well with authentication headers, if an api key is provided.

A new exemplary config file could look like this:
```
{
    "ollama": {
        "OllamaGeneratorChat": {
            "host": "https://my-self-hosted-server.com/ollama",
            "verify_ssl": false,
            "extra_params": {
                "headers": {
                    "User-Agent": "Mozilla/5.0"
                }
            }
        }
    }
}
```

garak/generators/ollama.py
TheRisenPhoenix: I prefer your first suggestion, having verify and api key as first class support but still enabling extra_params if necessary.

garak/generators/ollama.py
```diff
@@ -38,8 +41,20 @@ class OllamaGenerator(Generator):
...
+        if hasattr(self, "extra_params") and self.extra_params is not None:
+            for k, v in self.extra_params.items():
+                client_kwargs[k] = v
+        if hasattr(self, "verify_ssl") and self.verify_ssl is not None:
+            client_kwargs["verify"] = self.verify_ssl
+        if self.api_key:
+            client_kwargs["headers"] = {
+                **client_kwargs.get("headers", {}),
+                "Authorization": f"Bearer {self.api_key}",
+            }
```
jmartin-tech: I think the format is close but may not be quite right. Though I may simply be mis-reading the syntax.

A unit test would be valuable to confirm this.

This is my naive, untested thought on what might be needed.
```Python
            headers = client_kwargs.get("headers", {}) | {
                "Authorization": f"Bearer {self.api_key}",
            }
            client_kwargs["headers"] = headers
```

jmartin-tech (commented):
This rework looks great, and the support for `extra_params` is a great add.

I think a unit test or so would help document this and also increase confidence this is working as desired.

jmartin-tech:
One more note looking at the latest config example. `User-Agent` is actually a [global run config option](https://reference.garak.ai/en/latest/configurable.html#run-config-items) that I suspect the Ollama client is already getting a custom value set by `garak` for as the tool sets this for use by the `requests` lib.

garak/generators/ollama.py
TheRisenPhoenix: Both should be equivalent to the best of my knowledge. However, I agree that your suggestion might be a bit more comprehensible, so I changed that.

TheRisenPhoenix:
> One more note looking at the latest config example. `User-Agent` is actually a [global run config option](https://reference.garak.ai/en/latest/configurable.html#run-config-items) that I suspect the Ollama client is already getting a custom value set by `garak` for as the tool sets this for use by the `requests` lib.

That's totally possible, I was just looking for a quick and easy example for a header to demonstrate the header union in practice :D

garak/generators/ollama.py
TheRisenPhoenix: Not sure about the Unit Test though. What exactly do you want to be tested and how/where should this be done? I looked at the current ollama test but do not really see how this would fit and work in there tbh

jmartin-tech: A unit test targeting using a mock to intercept the request and inspect it similar to the one recently added to `test_guardrails.py` would be in order.

https://github.com/NVIDIA/garak/blob/cafbe992780df70015de7295ca026b70c4629ada/tests/generators/test_guardrails.py#L79-L120

TheRisenPhoenix: I wrote a unit test, please have a look if it's what you were thinking of.

tests/generators/test_ollama.py
```diff
@@ -214,3 +216,43 @@ def test_error_on_nonexistant_model_chat_mocked(respx_mock):
...
+                    "extra_params": {"headers": {"My-Header": "Test-1.0"}},
+                }
+            }
+        }
+    }
+    gen = OllamaGenerator("gemma3", config_root=config)
+
+    assert gen.verify_ssl is False
+    assert gen.client._client.headers["My-Header"] == "Test-1.0"
+    assert gen.client._client.headers["Authorization"] == f"Bearer sk-1234567abc"
```
jmartin-tech: Compare the computed values as that is the contract with `Configurable` base on using `set_fake_env`:
```Python
    assert gen.api_key is not None
    assert gen.client._client.headers["Authorization"] == f"Bearer {gen.api_key)"
```

tests/generators/test_ollama.py
```diff
@@ -214,3 +216,43 @@ def test_error_on_nonexistant_model_chat_mocked(respx_mock):
...
+
+    def restore_env():
+        if stored_env is not None:
+            os.environ[OllamaGenerator.ENV_VAR] = stored_env
+        else:
+            del os.environ[OllamaGenerator.ENV_VAR]
+
+    os.environ[OllamaGenerator.ENV_VAR] = "sk-1234567abc"
+    request.addfinalizer(restore_env)
+
```
jmartin-tech, line 236: This revision officially add an `ENV_VAR` value and makes it _optional_.

I see one more test to add based on this revision:

``` Python
@pytest.mark.usefixtures("set_fake_env")
def test_ollama_no_api_key():
    """When no env variable key is provided the generator can be instantiated"""
    del os.environ[OllamaGenerator.ENV_VAR]

    gen = OllamaGenerator("gemma3")
    assert gen.api_key is None
```

tests/generators/test_ollama.py
```diff
@@ -214,3 +216,43 @@ def test_error_on_nonexistant_model_chat_mocked(respx_mock):
...
+@pytest.fixture
+def set_fake_env(request) -> None:
+    stored_env = os.getenv(OllamaGenerator.ENV_VAR, None)
+
+    def restore_env():
+        if stored_env is not None:
+            os.environ[OllamaGenerator.ENV_VAR] = stored_env
+        else:
+            del os.environ[OllamaGenerator.ENV_VAR]
```
jmartin-tech: Based on additional test ask when no key exists to restore the test may have removed the testing value:
```Python
            try:
                del os.environ[OllamaGenerator.ENV_VAR]
            except KeyError:
                pass
```

jmartin-tech (changes requested):
Minor testing feedback and this will be ready to land.

tests/generators/test_ollama.py
TheRisenPhoenix: Done

jmartin-tech (approved):

PR #2027, opened 2026-08-07
fix(probes): return no attempts when the prompt set is empty

### LINKED ISSUE #2026: Probe.probe() raises IndexError when the prompt set is empty (translation can produce an empty set)

feiiiiii5:
## Root Cause

`Probe.probe()` indexes `prompts[0]` without first checking that the prompt set is non-empty (`garak/probes/base.py`). A probe whose prompt set is empty — for example one emptied by the langprovider translation path (reproduced in `tests/probes/test_probes.py::test_probe_prompt_translation[probes.sysprompt_extraction.SystemPromptExtraction]` in #2022) — raises `IndexError: list index out of range` instead of producing no attempts. The progress bar two lines earlier already tolerates `len(prompts) == 0`; only the subscript does not.

## Fix

Return `[]` (with a warning naming the probe) when the copied prompt set is empty, before the progress bar and the `prompts[0]` check. This mirrors the empty-set guard already present in `IntentProbe.probe()`.

## Test

- New regression test `test_probe_with_empty_prompts_returns_no_attempts`: a `Probe` with `prompts = []` returns `[]` instead of raising.
- Verified locally: `pytest tests/probes/test_probes.py::test_mint_attempt tests/probes/test_probes.py::test_probe_with_empty_prompts_returns_no_attempts` → 5 passed.

## Diff scope

2 files, +24/-1: `garak/probes/base.py`, `tests/probes/test_probes.py`.

## AI Disclosure

AI-assisted implementation and regression test; human review of the empty-set semantics (no behavior change for non-empty probes; consistent with `IntentProbe`) and final diff before submission.

Closes #2026

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -393,6 +393,12 @@ class Probe(Configurable):
         prompts = copy.deepcopy(
             self.prompts
         )  # make a copy to avoid mutating source list
+        if not prompts:
+            # A probe can have an empty prompt set (e.g. translation empties it);
+            # the progress bar below tolerates zero but the `prompts[0]` check
+            # does not. Produce no attempts instead of raising IndexError (#2026).
+            logging.warning("probe %s has an empty prompt set", self.probename)
+            return []
         preparation_bar = tqdm.tqdm(
             total=len(prompts),
             leave=False,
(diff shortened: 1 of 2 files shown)


patriciapampanelli:
Closing as duplicate of #1903

PR #2029, opened 2026-08-07
fix: stop FileDetector format mismatch from aborting the run

patriciapampanelli:
## Summary

Found while reviewing **#1952**, which adds an `S008exec` → `fileformats.FileIsExecutable` intent mapping. With it applied, `--spec probes.grandma.GrandmaIntent,intent:S008exec` yields 256 attempts and zero evaluations.

`FileDetector.detect()` raises `ValueError` when `attempt.notes["format"]` does not match `valid_format`, and the harness calls detectors with no `try`/`except`, so the run aborts: queued probes never execute, no `eval` or `completion` records are written, and `cli.main` exits **0**. The mapping only exposes this.

The raise becomes a warning and one `None` per output, not `0.0`, since the outputs are not filenames at all. The evaluator already counts `None` as `nones`, and `StringDetector` and `HFDetector.graceful_fail` use the same idiom.

## Reproduction

```bash
python -m garak --target_type test.Repeat --spec probes.test.Blank,probes.test.Test \
  -d fileformats.FileIsPickled --generations 1 --report_prefix filedetector-guard-before
```

**Before**: `probes.test.Test` never runs; the report holds 1 attempt, 0 `eval`, 0 `completion`.
**After**: both probes run, 9 attempts, 2 `eval`, 1 `completion`.

Both exit **0** — the exit code never signalled the abort.

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -305,7 +305,11 @@ class TriggerListDetector(Detector):
 
 
 class FileDetector(Detector):
-    """Detector subclass for processing attempts whose outputs are filenames for checking"""
+    """Detector subclass for processing attempts whose outputs are filenames for checking
+
+    Attempts whose ``notes["format"]`` does not match ``valid_format`` cannot be
+    scored; one ``None`` per output is returned so the run continues.
+    """
 
     valid_format = "local filename"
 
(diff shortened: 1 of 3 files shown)


jmartin-tech (approved):
This make sense to me as a behavior change since the introduction of support for `None` responses from detectors.

PR #2035, opened 2026-08-09
Return no attempts when a probe's prompt set is empty

### LINKED ISSUE #2026: Probe.probe() raises IndexError when the prompt set is empty (translation can produce an empty set)

eeshsaxena:
Fixes #2026

### Problem

`Probe.probe()` in `garak/probes/base.py` uses `len(prompts)` for the preparation progress bar and then indexes `prompts[0]` to decide whether prompts are plain strings:

```python
preparation_bar = tqdm.tqdm(total=len(prompts), ...)   # fine with 0
```

An empty prompt set (including one emptied by translation under some langproviders) reaches the subscript and raises `IndexError: list index out of range` deep in the run, rather than simply producing zero attempts. This is observed on `main` via `tests/probes/test_probes.py::test_probe_prompt_translation[...SystemPromptExtraction]`.

### Fix

Return an empty attempt list (with a warning naming the probe) when the prompt set is empty, before the `prompts[0]` check.

### Test

Added `test_probe_empty_prompt_set_returns_no_attempts`, which loads `probes.test.Test`, empties its prompts, and asserts `probe()` returns no attempts. It fails with `IndexError` before this change and passes after.

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -393,6 +393,12 @@ class Probe(Configurable):
         prompts = copy.deepcopy(
             self.prompts
         )  # make a copy to avoid mutating source list
+        if not prompts:
+            # An empty prompt set (including one emptied by translation) would
+            # otherwise raise IndexError at prompts[0] below; there is simply
+            # nothing to attempt, so return no attempts.
+            logging.warning("probe %s has an empty prompt set", self.probename)
+            return []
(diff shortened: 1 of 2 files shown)


jmartin-tech:
Duplicates #2027

PR #2038, opened 2026-08-09
feat(probes): add SVG/XML injection (XXE + SSRF) probe + detector

vishnukannaujia:
Fixes #1949

## Summary
Adds a new probe + detector pair, `svg_injection`, targeting SVG/XML that attacks a **downstream rasterizer or XML parser** (not a browser) via XML external entity resolution ([CWE-611](https://cwe.mitre.org/data/definitions/611.html)) or server-side request forgery ([CWE-918](https://cwe.mitre.org/data/definitions/918.html)).

When an app asks a model to "make me an icon as SVG" and then rasterizes / XML-parses the result, model output containing an external DTD/entity (`<!ENTITY ... SYSTEM "file://...">`) can read local files, and an `<image href="...">` pointing at `file://` or an internal host (cloud metadata `169.254.169.254` / `metadata.google.internal`, or a `gopher:` / `dict:` gadget) can trigger SSRF.

Same **"target output attacks a downstream sink"** thesis as `ansiescape` (terminal) and `web_injection` (browser). Browser `<script>` / XSS intentionally stays with `web_injection`; this is the SVG-rasterizer / XML-parser path. Companion to `formula_injection` (#2036) and `log_injection` (#2037). Scope-fit floated on #1949 first.

## What's added
- **`probes.svg_injection.SvgInjection`** — stub × payload prompts that try to get the target to emit SVG carrying XXE or an SSRF/local-file `<image>` reference. `intent = "S008inject"`; tags `owasp:llm02`, `owasp:llm05`, `avid-effect:security:S0301`, `quality:Security:Confidentiality`.
- **`detectors.svg_injection.SvgInjection`** — a deterministic regex detector for external-entity (`<!DOCTYPE/<!ENTITY ... SYSTEM/PUBLIC>`) and `<image href>` → `file://`/SSRF-target constructs. No model call needed.
- **Tests** mirroring the existing probe/detector tests.

## Tests
- `tests/probes/test_probes_svg_injection.py` — prompt structure & payloads carry an XXE/SSRF construct.
- `tests/detectors/test_detectors_svg_injection.py` — every payload → `1.0`; benign SVG (plain shapes, a relative `href="logo.png"`, an `https://example.com` image) → `0.0`; `None` output → `None`.
- Passes the generic probe/detector conformance suites and `black`.

Happy to align naming/tiering with the `SvgInjection`(OF_CONCERN)/`SvgInjectionEscaped`(INFORMATIONAL) split suggested in the issue if you'd prefer that shape.

PR #2040, opened 2026-08-09
probe: svg injection (XXE/SSRF via model-emitted markup)

### LINKED ISSUE #1949: probe: SVG injection (model output hits a downstream rasterizer / XML parser... XXE + SSRF)
hey.... wanted to float this before opening a PR since the contributing guide says check fit for new plugins first.

idea is an `svg_injection` probe + detector that tries to get the target to spit out SVG/XML that attacks whatever renders it downstream... an SVG rasterizer or XML parser, not a browser.

same thesis as ansiescape (terminal) and web_injection (browser).. model output lands in a sink that acts on active content. here the sink is the svg-to-raster / xml path an app runs when it asks a model to "make me an icon as SVG" and then rasterizes or parses it. two vectors:

- external entity / external DTD (`<!ENTITY ... SYSTEM "file://...">`, param-entity DTD fetch)... reads local files or exfils out of band when the parser resolves external entities. libxml2/lxml and older imagemagick svg handling do that, hardened builds and current librsvg dont.
- `<image href="...">`.... pulls `file://` local content or reaches an internal host (cloud metadata `169.254.169.254` / `metadata.google.internal`, or a gopher:/dict: gadget) on the image load path.

not trying to step on web_injection.. browser `<script>`/xss stays there, thats already its turf. i checked and theres no svg/xxe/entity probe, detector or resource in the tree right now, and no open or closed issue/PR for it.

shape-wise i mirrored ansiescape: `SvgInjection` (OF_CONCERN) + `SvgInjectionEscaped` (INFORMATIONAL) probes, `Live`/`Escaped` StringDetectors tuned for low false positives... a hit needs an external entity, a `file:` ref, an external DTD, or a known ssrf target. ordinary shapes / http(s) images / internal `#` refs dont score, and generic 192.168/10.x ssrf is out of scope since you cant catch it without false-positiving legit images. payloads live in `resources/svg.py`.

its already written and green locally.... 2 probes + 2 detectors + payloads + docs + tests, passes the generic suites and black. holding the PR til you confirm scope... happy to keep it its own `svg_injection` module or fold it into web_injection/exploitation if you'd rather. whats your preference?

francose:
adds a probe + detector for SVG/XML injection. closes #1949.

when a model emits SVG/XML and something downstream rasterizes or parses it, that markup can carry XXE (`<!DOCTYPE ... <!ENTITY SYSTEM>`) to read local files, or an `<image href>` pointing at `file://` or an SSRF target (169.254.169.254, metadata.google.internal, gopher:/dict:). this probes whether a model hands that back on request.

whats here:
- `probes.svg_injection.SvgInjection` (OF_CONCERN) - live, ready-to-render markup
- `probes.svg_injection.SvgInjectionEscaped` (INFORMATIONAL) - escaped markup, lower signal
- `detectors.svg_injection.Live` / `.Escaped` - deterministic regex for the XXE/SSRF constructs
- intent S008inject, mirrors the ansiescape / web_injection pattern
- `<script>`/XSS stays in web_injection, this is scoped to the parser/rasterizer sink

detector tests + docs included, green against current main

### DIFF
--- a/docs/source/detectors/svg_injection.rst
+++ b/docs/source/detectors/svg_injection.rst
@@ -1,0 +1,7 @@
+garak.detectors.svg_injection
+=============================
+
+.. automodule:: garak.detectors.svg_injection
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/docs/source/index_detectors.rst
+++ b/docs/source/index_detectors.rst
@@ -40,6 +40,7 @@ For detailed information on detector metrics and evaluation, see :doc:`../detect
    detectors/propile
    detectors/shields
    detectors/snowball
+   detectors/svg_injection
    detectors/sysprompt_extraction
    detectors/unsafe_content
    detectors/visual_jailbreak
--- a/docs/source/index_probes.rst
+++ b/docs/source/index_probes.rst
@@ -48,6 +48,7 @@ For a guide to writing probes, see :doc:`extending.probe`.
    probes/snowball
    probes/smuggling
    probes/suffix
+   probes/svg_injection
    probes/sysprompt_extraction
    probes/tap
    probes/test
--- a/docs/source/probes/svg_injection.rst
+++ b/docs/source/probes/svg_injection.rst
@@ -1,0 +1,9 @@
+garak.probes.svg_injection
+==========================
+
+.. automodule:: garak.probes.svg_injection
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+   .. show-asr::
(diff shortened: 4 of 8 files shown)


leondz:
see also #2038

francose:
saw #2038 - looks like these landed about 2.5h apart. #2040 has the live/escaped split plus the `resources/svg.py` payload bank + docs, #2038 is the leaner single probe/detector. happy to consolidate either way so the maintainers only have one thing to review. @vishnukannaujia want to compare and pick a base? ill leave this open till we or the maintainers pick a direction so none of the useful pieces get lost

PR #2044, opened 2026-08-11
fix(detectors): _extract_json always returns a dict for agent_breaker verdicts

### LINKED ISSUE #2043: AgentBreakerResult._extract_json returns lists for top-level arrays and crashes detection
## Summary

`AgentBreakerResult._extract_json` (used by the `agent_breaker` probe and its model-as-judge detector) is documented to return a `dict`, but a top-level JSON array succeeds `json.loads` and is returned as a `list`. Every caller then calls `.get(...)` on it, raising `AttributeError`:

- `verify()` catches `(json.JSONDecodeError, ValueError, TypeError)` but **not** `AttributeError`, so a judge model answering with `[{"success": "YES", ...}]` or `[]` crashes detection instead of degrading to a non-success score.
- `_analyze_attackable_tools()` and `_discover_agent_config()` likewise only catch `JSONDecodeError`, so a red-team model returning an array crashes the probe's analysis/discovery path.

Impact: an LLM (the very component this red-team tool treats as untrusted) can crash the whole agentic scan with a malformed-but-valid JSON shape — an availability failure in the security tooling.

## Repro

```python
from unittest.mock import MagicMock, patch
import json

# _extract_json violates its documented return type:
parsed = AgentBreakerResult._extract_json('[{"success": "YES"}]')
assert isinstance(parsed, dict)  # AttributeError: 'list' object has no attribute 'get'
```

And end-to-end: a detector whose `evaluation_generator` returns `[MagicMock(text="[]")]` raises `AttributeError` from `detect()` instead of returning `[0.0]`.

## Expected behavior

- `_extract_json` always returns a `dict`: extract the first JSON object even when wrapped in a top-level array (consistent with its docstring), and raise `JSONDecodeError` for scalars/empty arrays so callers' existing handling applies.
- Malformed judge output degrades to a non-success score (`0.0`), never crashes the run.

feiiiiii5:
## Root Cause

`AgentBreakerResult._extract_json` is documented as returning a `dict`, but a top-level JSON array (e.g. `[{"success": "YES"}]`) passes through `json.loads` unchanged and is returned as a `list`. Every caller (`verify()`, `_analyze_attackable_tools()`, `_discover_agent_config()`) then calls `.get(...)` on the result, raising `AttributeError`; `verify()` only catches `JSONDecodeError`/`ValueError`/`TypeError`. An LLM — the very component this red-team tool treats as untrusted — can therefore crash the whole agentic scan with a malformed-but-valid JSON shape instead of degrading to a non-success verdict.

## Fix

`_extract_json` now always returns a `dict`:

- A top-level parse that yields a `dict` is returned as before.
- An object wrapped in a top-level array is unwrapped (first object returned).
- Arrays without an object, scalars and `null` raise `JSONDecodeError`, so every caller's existing exception handling applies and no `AttributeError` can escape.
- The existing brace-scan fallback for trailing prose is unchanged.

## Test

- `TestExtractJsonShape`: array-wrapped object extracted; `[]` / `[1, 2]` / `"123"` / `"null"` raise `JSONDecodeError`.
- `TestDetectJsonParsing.test_top_level_array_wrapping_object`: judge verdict wrapped in a top-level array scores `1.0`.
- `TestDetectJsonParsing.test_empty_top_level_array_returns_zero`: judge output `[]` degrades to `0.0` (no crash).
- Verified locally: `HOME=/tmp/garak-home .venv-dev/bin/python -m pytest tests/detectors/test_detectors_agent_breaker.py tests/probes/test_agent_breaker.py -q` → 80 passed; black (pinned 25.1.0) introduces no new findings on the changed files.

## Diff scope

2 files, +76/-11: `garak/detectors/agent_breaker.py`, `tests/detectors/test_detectors_agent_breaker.py`.

## AI Disclosure

AI-assisted implementation and regression tests; the shape-enforcement decision (always return dict, unwrap first object from arrays, raise on no-object) and final diff were human-reviewed before submission.

Closes #2043

### DIFF
--- a/garak/detectors/agent_breaker.py
+++ b/garak/detectors/agent_breaker.py
@@ -184,12 +184,21 @@ class AgentBreakerResult(Detector):
         """Extract the first valid JSON object from *text*.
 
         Handles cases where the LLM appends an explanation after the JSON
-        block (which causes ``json.loads`` to raise "Extra data").
+        block (which causes ``json.loads`` to raise "Extra data") and where
+        the LLM wraps the object in a top-level JSON array.
         """
         try:
-            return json.loads(text)
+            parsed = json.loads(text)
         except json.JSONDecodeError:
             pass
+        else:
+            if isinstance(parsed, dict):
+                return parsed
+            if isinstance(parsed, list):
+                for item in parsed:
+                    if isinstance(item, dict):
+                        return item
+            raise json.JSONDecodeError("No JSON object found", text, 0)
 
         brace_start = text.find("{")
         if brace_start == -1:
--- a/tests/detectors/test_detectors_agent_breaker.py
+++ b/tests/detectors/test_detectors_agent_breaker.py
@@ -41,12 +41,14 @@ def _make_attempt(outputs, notes=None, **kwargs):
 def _mock_verify_response(success="YES", confidence=0.9, reasoning="exploited"):
     """Create a mock generator response for the VERIFY prompt."""
     resp = MagicMock()
-    resp.text = json.dumps({
-        "success": success,
-        "confidence": confidence,
-        "reasoning": reasoning,
-        "indicators": ["indicator1"],
-    })
+    resp.text = json.dumps(
+        {
+            "success": success,
+            "confidence": confidence,
+            "reasoning": reasoning,
+            "indicators": ["indicator1"],
+        }
+    )
     return resp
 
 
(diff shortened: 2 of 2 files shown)


leondz:
@barlanyado @eliyacohen-hub  could you review?

feiiiiii5:
Heads-up: I refreshed this branch — the implementation is now the tighter two-file version described in the body (non-object top-level shapes raise JSONDecodeError directly; array-wrapped verdicts unwrap). Behaviour is identical on all documented cases; detector + probe tests pass (80 passed).

eliyacohen-hub:
Thanks @leondz. I went through this and #2069, and the fixes make sense to me.

Given @jmartin-tech's note in #1970 that it may impact or subsume parts of both PRs, I think it would be better to do the final review after the consolidated/rebased version is ready, to make sure we're reviewing the final AgentBreaker flow.

@jmartin-tech, please let us know once it's ready for another pass.

PR #2048, opened 2026-08-11
feat(probe): add FigStep AdvBench probes (forward & reverse)

GoNow1231:
## Summary

Adds two image-modality jailbreak probes to `visual_jailbreak.py`, extending the existing FigStep implementation with an AdvBench-based dataset:

- `FigStepAdvBench` — the classic FigStep attack: the harmful AdvBench behavior is rendered as plain text into an image, and the text prompt asks the model to read the text in the image and answer it, bypassing text-based alignment.
- `FigStepAdvBenchReverse` — a reverse variant: the harmful behavior is rendered with its characters reversed, and the prompt asks the model to reverse the text before reading it, testing whether encoded harmful content still bypasses alignment.

Both probes attach the image through `Message.data_path` (modality `{"in": {"text", "image"}}`) and are scored with `mitigation.MitigationBypass` (+ `mitigation.Prefixes`).

## Data

- Bundled payload `garak/data/payloads/advbench_harmful_behaviors.json`: 65 harmful behaviors drawn from the public AdvBench dataset (MIT), loaded through garak's payload system.
- Typographic images are rendered at runtime with Pillow (already a garak dependency) into the cache dir; an optional `image_dir` config loads pre-rendered `textimage` / `textimage-reverse` PNGs instead.

## Tests

```bash
pytest tests/probes/test_probes_figstep_advbench.py -v
```

Verified locally: both probes load via `garak._plugins.load_plugin`, enumerate correctly, attach rendered images, and complete a full `python -m garak` run.


garak/resources/plugin_cache.json
jmartin-tech, line 1: Unless there is some explicit reason tests will not run without including updates, `plugin_cache.json` is maintained by automation and does not need to be included in PRs specifically as is it can a present conflicts with upstream maintenance.

jmartin-tech (changes requested):
Not a review for fit.

All commits messages in the branch must have signoff to pass DCO requirements.

Full evaluation of this PR will be preformed after DCO requirements are correctly met.

PR #2051, opened 2026-08-12
detector: add judge.StrongREJECT (continuous refusal/convincing/speci…

cindysteward:
Addresses #973

## What this does

garak's existing judge.Jailbreak gives a binary safe/unsafe read on a response. StrongREJECT (Souly et al., 2024, https://arxiv.org/abs/2402.10260) argues a lot of "successful" jailbreaks aren't useful, the model just stops refusing without saying anything convincing. This PR adds judge.StrongREJECT, which asks the judge for three things in one pass. 1) did it refuse (1 or 0), 2) how convincing is it (1-5), and 3) how specific is it (1-5). 

The parsing and scoring lives in a new process_output_strongreject_score helper in garak/resources/red_team/evaluation.py, next to the existing process_output_judge_score / process_output_on_topic_score.

## Dataset licensing note in the issue

The issue mentions the StrongREJECT forbidden-prompts dataset has unclear licensing on some of its constituent sources. So, in this PR it is detector-only. It works against any garak jailbreak probe that already exists:

python -m garak -t <target> -p dan -d judge.StrongREJECT

In the possible future, with more clairty on licensing, it's possible to address then.

## Verification

- [ ] Supporting configuration file - N/A, for detector, no generator config needed
- [x] Run the tests and ensure they pass: python -m pytest tests/detectors/test_detectors_judge.py -v 
- [x] Verify the thing does what it should: unit tests cover refusal short-circuiting to 0, max ratings scoring to 1, mid-range and asymmetric ratings rescaling
- [x] Verify the thing does not do what it should not: unparseable judge output returns None instead of guessing a score; None outputs are skipped without calling the judge
- [x] Document the thing

## Note
StrongREJECT.detect() follows the same shape as Jailbreak.detect().

### DIFF
--- a/garak/detectors/judge.py
+++ b/garak/detectors/judge.py
@@ -10,6 +10,7 @@ from garak.resources.red_team.evaluation import (
     EvaluationJudge,
     get_token_limit,
     process_output_on_topic_score,
+    process_output_strongreject_score,
 )
 from garak.attempt import Attempt
 from garak.detectors.base import Detector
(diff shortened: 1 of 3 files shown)


cindysteward:
Closing this PR. I wanted to comment on https://github.com/NVIDIA/garak/issues/973, but then found #1717 already covers this and is ongoing. Didn't catch it before starting since it references the issue without a closing keyword, so it wasn't showing up as linked!

PR #2054, opened 2026-08-12
fix(detectors): default a detector's name to its class name string

### LINKED ISSUE #2053: Detector.name is set to a class object instead of a string
## Steps to reproduce

```python
from garak.detectors.base import Detector
from garak.detectors.web_injection import MarkdownExfilContent

for cls in (Detector, MarkdownExfilContent):
    d = cls()
    print(cls.__name__, repr(d.name), isinstance(d.name, str))
```

## Expected behavior

`name` is the detector's short name as a string.

## Current behavior

```
Detector             <class 'garak.detectors.base.Detector'>  False
MarkdownExfilContent <class 'garak.detectors.base.Detector'>  False
```

`detectors/base.py`:

```python
if "name" not in dir(self):
    self.name = __class__  # short name
```

Inside `Detector.__init__`, the implicit `__class__` cell refers to the class where the method is defined, not to `type(self)`, so every detector that does not set its own `name` gets the `Detector` class object. The subclass in the output above shows this is not picking up the runtime type.

This reaches user-facing output. `detectors/judge.py` renders:

```python
f"{self.name} failed to load generator for {self.detector_model_type}"
```

`ModelAsJudge` sets no `name`, so a judge-model misconfiguration prints `<class 'garak.detectors.base.Detector'> failed to load generator for ...` identically for every judge detector, and the message does not identify which one failed. `detectors/perspective.py` has the same pattern.

`detectorname` on the following line is unaffected, since it derives from `self.__class__`.

### garak version

`0.16.1.pre1`, source checkout at `b1e0101`

## Additional Information

1. macOS 26.5.1
2. Python 3.12.11
3. Direct repository checkout with `git`
4. No run artifacts; reproduced by direct instantiation
5. No config flags needed

EastZeus:
### What

`Detector.__init__` sets `self.name = __class__` when a detector defines no name of its own. The implicit `__class__` cell resolves to the base `Detector` class rather than `type(self)`, so every such detector ends up with the `Detector` class object instead of a string.

This reaches user-facing output. `detectors/judge.py` renders `f"{self.name} failed to load generator for {self.detector_model_type}"`, and `ModelAsJudge` sets no name, so a judge-model misconfiguration prints:

```
<class 'garak.detectors.base.Detector'> failed to load generator for ...
```

### Fix

Default the name to `self.__class__.__name__`, i.e. the detector's own short name as a string. Subclasses that already set their own `name` are unaffected.

### Not a duplicate

I searched the open PRs against #2053 and found none addressing it.

### Testing

```
python -m pytest tests/detectors/test_detectors.py::test_detector_default_name_is_class_name_string
```

passes. The added regression test checks that the base `Detector` and a nameless subclass each return their own class name as a string.

Fixes #2053

### DIFF
--- a/tests/detectors/test_detectors.py
+++ b/tests/detectors/test_detectors.py
@@ -79,6 +79,22 @@ def test_detector_structure(classname):
     ), f"_supported_params must contain all DEFAULT_PARAMS; {unsupported_defaults} missing"
 
 
+def test_detector_default_name_is_class_name_string():
+    """A detector that sets no name of its own gets its class name as a string,
+    not the base Detector class object (which reaches user-facing messages)."""
+
+    class _Nameless(Detector):
+        """a detector that does not set its own name"""
+
+    base = Detector()
+    sub = _Nameless()
+
+    assert isinstance(base.name, str), "base detector name must be a string"
+    assert base.name == "Detector", "base detector name should be its class name"
+    assert isinstance(sub.name, str), "subclass detector name must be a string"
+    assert sub.name == "_Nameless", "subclass should take its own class name"
+
+
 @pytest.mark.parametrize("classname", DETECTORS)
 def test_detector_detect(classname, monkeypatch):
     monkeypatch.setattr("datasets.load_dataset", Mock(return_value=_MockDataset()))
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -60,7 +60,7 @@ class Detector(Configurable):
     def __init__(self, config_root=_config):
         self._load_config(config_root)
         if "name" not in dir(self):
-            self.name = __class__  # short name
+            self.name = self.__class__.__name__  # short name
         self.detectorname = str(self.__class__).split("'")[1]
         self._set_description()
         if hasattr(_config.system, "verbose") and _config.system.verbose > 0:


chiruu12:
Flagging a pattern rather than this specific change. This account opened 13 PRs to inspect_petri at 17:52, 8 to inspect_evals at 17:57, and 3 here between 17:57 and 17:59 on the same day, then 3 more to flash-linear-attention the next morning at 03:18. Its only earlier activity is from 2022 and 2023.

Raising it because AGENTS.md requires a human submitter who can defend every changed line, and that volume is hard to square with it. #2056 already turned out to rest on an issue that had been determined inaccurate. Your call entirely, I just did not want to leave it unsaid.

EastZeus:
Hey Chirru12, I opened PRs at the same time but I’ve been working on those repos locally for a while. I can defend all my code changes.

jmartin-tech (approved):

leondz:
The new test should go inside the other tests validating detector structure.

tests/detectors/test_detectors.py
```diff
@@ -79,6 +79,22 @@ def test_detector_structure(classname):
     ), f"_supported_params must contain all DEFAULT_PARAMS; {unsupported_defaults} missing"
 
 
+def test_detector_default_name_is_class_name_string():
```
leondz, line 82: This check should go inside the other tests validating detector structure - fine to include this demo case but I'm not sure this is connected well enough for it to have max impact

leondz (changes requested):

PR #2056, opened 2026-08-12
fix(evaluators): record a probe_summary for zero-attempt probes

### LINKED ISSUE #2021: A probe that yields zero attempts leaves no trace in report.jsonl, so "did not run" and "was never configured" are the same absence
I do reliability work on the "did this actually happen" class of bug and went looking for it in garak, because a scanner whose entire output is a verdict is where that class costs the most. Filing this because the impact is quiet and the machinery to fix it already exists. Happy to be told it's deliberate.

**The short version:** if a probe produces no attempts, nothing about that probe reaches `report.jsonl`. Not a summary row, not an attempt row, not a zero. The run completes and the report reads exactly as it would if that probe had never been requested.

### The path

`garak/evaluators/base.py:231-235`

```python
if isinstance(attempts, list) and len(attempts) == 0:
    logging.error(
        "evaluators.base.Evaluator.evaluate called with list of 0 attempts, expected len 1+ or iterable"
    )
    return
```

That `return` is before everything that writes. The `probe_summary` record is built in the loop from `:256` and written at `:297-306` with `"entry_type": "probe_summary"` — so a zero-attempt probe never reaches it.

And nothing else fills the gap. `garak/harnesses/base.py` writes two kinds of row: the `plugin_cache` entry at `:71`, and per-attempt rows at `:277-281` inside `for attempt in attempt_results`. Zero attempts means zero of the latter. So the probe is absent from the report in every form.

### The part that makes this cheap to fix

`garak/harnesses/base.py:283-284`, four lines above the `evaluator.evaluate(attempt_results)` call:

```python
if len(attempt_results) == 0:
    logging.warning("zero attempt results: probe %s" % probe.probename)
```

The condition is already detected, at the right moment, with the probe name already in hand. It just goes to a log channel rather than to the artifact that downstream automation reads.

### Why it matters more here than in most tools

The consumer of `report.jsonl` is usually automation, and the question it's asking is "did anything fail". A probe that silently contributed nothing answers that with "no" in exactly the same bytes as a probe that ran fully and found nothing. For a red-teaming tool those are close to opposite results: one says the model resisted, the other says the model was never asked.

There's an adjacent case I didn't chase and am flagging rather than claiming: `harnesses/base.py` loads buffs inside a `try/except` that warns and continues, so a run configured with five buffs and two that loaded proceeds with three missing. I haven't checked whether the report records which buffs were actually active — if it does, that one's fine and only the probe case stands.

### What I haven't verified, plainly

I read this on `main` through the API. **I have not run garak and have not reproduced a zero-attempt probe.** The mechanism is unambiguous — the early return precedes the only writers — but I haven't established how often a probe legitimately yields zero attempts in practice, and if the answer is "essentially never outside a misconfiguration" then this is smaller than it looks.

I'm being explicit because I had a similar-shaped finding in another project the same week, wrote the headline, then traced one layer further and the library underneath was already handling it. I killed my own finding before filing.

### A possible shape for the fix

The cheapest version keeps the early return and adds a row before it — a `probe_summary` with zeroed counts and a reason — so the probe appears in the report as something that produced nothing rather than as nothing at all. That reuses the record type that already exists rather than introducing a concept, and turns a silent absence into a value a consumer can test. The `logging.warning` at `harnesses/base.py:283` is already doing the right thing conceptually; it's just going to a channel that nothing gating on the report will ever read.

Happy to open a PR if the direction is welcome.

---

*Last bit, and I'd rather say it plainly than have it be odd later: hunting this class in unfamiliar codebases is what I do for money — guards that never fire, checks that can't come out against you, verdicts with nothing behind them. I found this one while surveying, and it stands whatever you decide about me: file it, fix it, close it, all fine.*

*If you want the same pass over the whole harness and report layer, that's **$1,500 for a day — you, me, your codebase, on a call**, and every patch, repro and note is yours the same day. Not a document that arrives in a fortnight. Terms: https://unreached.dev/audit.html — where I also publish my own false-positive rates and my own retractions, including one from this week where my own check killed my own finding before I filed it.*

EastZeus:
### What

When a probe produces no attempts, `Evaluator.evaluate` returns before writing anything, so nothing about that probe reaches `report.jsonl`: no `probe_summary` row and no attempt rows. A report then reads identically whether a probe ran and produced nothing or was never requested, which hides skipped coverage from anything auditing the run.

### Fix

`evaluate` now accepts an optional `probename` and, when the attempt list is empty, writes a zero-count `probe_summary` for that probe. The harness passes the probe name (in the same short form used elsewhere in the report). The counts are all zero, which the existing report/digest consumers already handle. Callers that pass no `probename` keep the previous silent behaviour.

### Not a duplicate

I searched the open PRs against #2021 and found none addressing it.

### Testing

```
python -m pytest tests/evaluators/test_base.py tests/harnesses/
```

18 passed. Added tests assert a zero-attempt probe with a name leaves exactly one `probe_summary` row, and that omitting the name stays silent.

This issue is labelled `question`, so I am happy to adjust the shape of the emitted record (or move where it is written) if maintainers prefer a different approach.

Fixes #2021

### DIFF
--- a/garak/evaluators/base.py
+++ b/garak/evaluators/base.py
@@ -221,17 +221,42 @@ class Evaluator:
 
         return eval_record
 
-    def evaluate(self, attempts: Iterable[garak.attempt.Attempt]) -> None:
+    def evaluate(
+        self, attempts: Iterable[garak.attempt.Attempt], probename: str = None
+    ) -> None:
         """evaluate feedback from detectors
 
         expects a list of attempts that correspond to one probe
         outputs results once per detector
+
+        :param probename: dotted probe name, used to record a zero-attempt probe
+            in the report when ``attempts`` is empty
         """
 
         if isinstance(attempts, list) and len(attempts) == 0:
-            logging.error(
-                "evaluators.base.Evaluator.evaluate called with list of 0 attempts, expected len 1+ or iterable"
+            logging.warning(
+                "evaluators.base.Evaluator.evaluate called with 0 attempts for probe %s",
+                probename,
             )
+            if probename is not None:
+                # a probe can legitimately yield no attempts (e.g. a prompt set
+                # emptied by translation); still write a zero-count probe_summary
+                # so the report shows the probe ran rather than leaving "did not
+                # run" indistinguishable from "was never configured"
+                empty_summary = {
+                    "entry_type": "probe_summary",
+                    "probe": probename,
+                    "inference_counts": {"total_evaluated": 0, "nones": 0},
+                    "detection_counts": {
+                        "detectors": [],
+                        "passed": 0,
+                        "fails": 0,
+                        "nones": 0,
+                    },
+                }
+                _config.transient.reportfile.write(
+                    json.dumps(empty_summary, ensure_ascii=False) + "\n"
+                )
             return
 
         attempts = list(
--- a/garak/harnesses/base.py
+++ b/garak/harnesses/base.py
@@ -283,7 +283,10 @@ class Harness(Configurable):
             if len(attempt_results) == 0:
                 logging.warning("zero attempt results: probe %s" % probe.probename)
 
-            evaluator.evaluate(attempt_results)
+            evaluator.evaluate(
+                attempt_results,
+                probename=".".join(probe.probename.split(".")[-2:]),
+            )
 
         self._end_run_hook()
 
--- a/tests/evaluators/test_base.py
+++ b/tests/evaluators/test_base.py
@@ -108,6 +108,43 @@ def _eval_rows(report_path: Path) -> list[dict]:
     return rows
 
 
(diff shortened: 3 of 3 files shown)


jmartin-tech:
This may not be the desired path as noted in #2021, which is tagged as a "question", the reported issue was determined to be inaccurate and needs further feedback to become valid as something to "fix".

jmartin-tech:
As noted in the original issue, there is other tooling for aggregation that would need to change if introducing additional summaries and a single run report does contain a list of all probes that were `activated` for the run. Tooling that is processing a report must have the full report and the needed information is in the report.

Any enhancement would need to include significant additional contextual information on `why` the probe did not execute that is not included in the this revision and is not feasible to gather from the current runtime environment while still supporting aggregation tooling.

PR #2063, opened 2026-08-14
fix(analyze): draw bootstrap samples from a local RNG

### LINKED ISSUE #2062: CI calculation reseeds the global numpy RNG, so --seed makes later numpy draws repeat
`_bootstrap_calculation` in `garak/analyze/bootstrap_ci.py` seeds the global numpy generator when a run seed is set:

```python
if (
    hasattr(_config, "run")
    and hasattr(_config.run, "seed")
    and _config.run.seed is not None
):
    np.random.seed(_config.run.seed)
```

`garak/evaluators/base.py:149` calls this once per probe per detector, and `confidence_interval_method` defaults to `bootstrap`. So on any seeded run the global numpy RNG is reset back to the same value repeatedly during evaluation, and every numpy draw after that point restarts from the same state.

Reproduction:

```python
from garak import _config
import garak.analyze.bootstrap_ci as b
import numpy as np

_config.load_base_config()
_config.run.seed = 42

def draw():
    return np.random.choice(10, 3).tolist()

np.random.seed(1234)
for _ in range(3):
    b.calculate_bootstrap_ci(results=[0]*60+[1]*40, sensitivity=0.95, specificity=0.90)
    print(draw())
```

```
[9, 1, 7]
[9, 1, 7]
[9, 1, 7]
```

Without the CI call in between, the same three draws are `[4, 8, 9]`, `[1, 7, 9]`, `[6, 8, 0]`.

Two places in a run consume that generator:

- `garak/resources/red_team/conversation.py:185` shuffles branches before pruning, to permute equal-scoring elements. With the state pinned, ties break the same way every round.
- `garak/resources/autodan/genetic.py:193` picks parents for the genetic algorithm. Pinned state means the same parents are selected each generation.

So passing `--seed`, which is there to make a run reproducible, currently also removes the randomness TAP and AutoDAN rely on to explore. The bootstrap intervals themselves are affected too, since every probe and detector pair resamples with the same draws rather than independently.

Fix is to draw from a local generator rather than seeding the global one. `np.random.default_rng(seed)` keeps the interval reproducible for a given seed and drops the side effect. I have this working locally with a regression test, and the existing 14 tests in `tests/analyze/test_bootstrap_ci.py` still pass, including `test_calculate_bootstrap_ci_reproducibility`.

I would like to work on this, PR to follow.

chiruu12:
Fixes #2062

`_bootstrap_calculation` seeded the global numpy generator with the run seed. The evaluator calls it once per probe per detector and `confidence_interval_method` defaults to `bootstrap`, so on a seeded run the global RNG was reset repeatedly during evaluation and every numpy draw afterwards restarted from the same state.

That reaches two consumers: branch shuffling in `red_team/conversation.py`, which permutes equal-scoring elements before pruning, and parent selection in `autodan/genetic.py`. Both stop varying, so `--seed` was quietly removing the randomness TAP and AutoDAN use to explore.

This draws from `np.random.default_rng(seed)` instead. Same seed still gives the same interval, and the global generator is left alone.

Added a test that draws from `np.random` around a CI calculation and compares against the same draws without one. It fails on main and passes here. The existing 14 tests in the file still pass, including `test_calculate_bootstrap_ci_reproducibility`.

One thing to note: `default_rng` uses PCG64 where `np.random.seed` uses the legacy MT19937, so a seeded run will report slightly different interval bounds than before. Same seed still gives the same answer, it is just not bit identical to previous releases. Happy to use `np.random.RandomState(seed)` instead if you would rather keep the old numbers.

`pytest tests/analyze/` is green, 159 passed.

### DIFF
--- a/garak/analyze/bootstrap_ci.py
+++ b/garak/analyze/bootstrap_ci.py
@@ -25,12 +25,16 @@ def _bootstrap_calculation(
     if confidence_level is None:
         confidence_level = _config.reporting.bootstrap_confidence_level
 
+    # Use a local generator: seeding the global one here would reset numpy's
+    # random state for the rest of the run.
+    seed = None
     if (
         hasattr(_config, "run")
         and hasattr(_config.run, "seed")
         and _config.run.seed is not None
     ):
-        np.random.seed(_config.run.seed)
+        seed = _config.run.seed
+    rng = np.random.default_rng(seed)
 
     denominator = sensitivity + specificity - 1.0
     if abs(denominator) < 0.01:
@@ -51,7 +55,7 @@ def _bootstrap_calculation(
     is_perfect_detector = np.isclose(denominator, 1.0)
     
     for i in range(num_iterations):
-        resampled_results = np.random.choice(results, size=n, replace=True)
+        resampled_results = rng.choice(results, size=n, replace=True)
         p_obs = resampled_results.mean()
         
         if is_perfect_detector:
--- a/tests/analyze/test_bootstrap_ci.py
+++ b/tests/analyze/test_bootstrap_ci.py
@@ -212,3 +212,27 @@ def test_nonparametric_vs_parametric_advantage():
     assert ci_lower < 40  # Lower bound below observed rate
     assert ci_upper > 40  # Upper bound above observed rate
     assert ci_upper - ci_lower < 40  # CI width should be reasonable for n=50
+
+
+def test_calculate_bootstrap_ci_leaves_global_rng_alone():
+    """A CI calculation should not reset the global numpy RNG"""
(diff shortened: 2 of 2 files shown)


chiruu12:
The red here is not this diff. The only failure is `tests/generators/test_anthropic.py::test_anthropic_generator`, and the rest of the matrix was cancelled by fail-fast rather than failing on its own.

`anthropic` 1.0.0 shipped on 20 Aug and moved the SDK onto `httpx2`, which respx cannot patch, so that test now calls the live API and gets a 401. Nightly on `main` caught the same thing on an unchanged commit. Filed as #2099 with the cap in #2100.

The bootstrap change itself is 5691 passed on this run.

PR #2069, opened 2026-08-15
fix: degrade gracefully on malformed agent_breaker analysis shapes

### LINKED ISSUE #2045: agent_breaker probe crashes on malformed analysis shapes from the red-team model
## Summary

The `agent_breaker` probe treats the red-team model's analysis output as untrusted (it can be a JSON object with any internal shapes), but two consumers assume exact types and crash the whole probe run with `AttributeError`:

1. `_build_tool_configs()` — called from `_create_init_attempts()` outside any per-tool try/except:
   - `tool_analyses` returned as a list (e.g. `"tool_analyses": [{"name": "x"}]`) → `AttributeError: 'list' object has no attribute 'items'`
   - `priority_targets` containing a non-string entry (e.g. `null`) → `AttributeError: 'NoneType' object has no attribute 'split'`
2. `_format_tools_for_analysis()` — called from `_analyze_attackable_tools()`:
   - a `tools` entry that is not a dict → `AttributeError: 'str' object has no attribute 'get'`
   - `tools` not being a list at all is iterated directly.

This is the same "untrusted model output must not crash the security tool" class as #2043: the object-level parse is now fixed, but values *inside* the object are still trusted by shape.

## Repro

```python
from garak.probes.agent_breaker import AgentBreaker

p = AgentBreaker.__new__(AgentBreaker)
p.agent_analysis = {"tool_analyses": [{"name": "x"}], "priority_targets": [None]}
p._build_tool_configs()
# AttributeError: 'list' object has no attribute 'items'
```

```python
p2 = AgentBreaker.__new__(AgentBreaker)
p2.agent_config = {"tools": ["rm -rf /"]}
p2._format_tools_for_analysis()
# AttributeError: 'str' object has no attribute 'get'
```

## Expected behavior

Malformed analysis shapes degrade gracefully with a warning: non-dict `tool_analyses` and non-list `priority_targets` yield no configs (no crash); malformed priority/tool entries are skipped; a non-list `tools` config is ignored for analysis. Valid entries still resolve normally.

feiiiiii5:
## Problem

Same untrusted-output class as #2043: the object-level JSON parse is now hardened, but values inside the analysis object are still trusted by shape. Two probe consumers crash the whole run with AttributeError on malformed red-team output:

- `_build_tool_configs`: `tool_analyses` returned as a list, or `priority_targets` containing non-string entries.
- `_format_tools_for_analysis`: `tools` not a list, or containing non-dict entries.

Fixes #2045.

## Fix

Malformed shapes now degrade gracefully with a warning: non-dict `tool_analyses` and non-list `priority_targets` yield no configs, non-string priority entries are skipped, a non-list `tools` config is ignored for analysis, and non-dict tool entries are skipped. Valid entries resolve exactly as before.

## Tests

Added to `tests/probes/test_agent_breaker.py`:

- malformed `tool_analyses` list produces no configs
- non-string `priority_targets` entries are skipped while valid ones match
- non-list `tools` config yields an empty analysis string
- non-dict tool entries are skipped while valid ones render

```
HOME=/tmp/garak-home .venv-dev/bin/python -m pytest tests/probes/test_agent_breaker.py tests/detectors/test_detectors_agent_breaker.py -q
84 passed
```

black 25.1.0 clean on both changed files. AI assistance was used in creating this PR; every changed line has been reviewed by the submitter.

### DIFF
--- a/garak/probes/agent_breaker.py
+++ b/garak/probes/agent_breaker.py
@@ -327,10 +327,32 @@ class AgentBreaker(garak.probes.IterativeProbe):
         tool_analyses = self.agent_analysis.get("tool_analyses", {})
         priority_targets = self.agent_analysis.get("priority_targets", [])
 
+        if not isinstance(tool_analyses, dict):
+            logging.warning(
+                "%s # ignoring malformed tool_analyses (%s); expected a dict",
+                self.__class__.__name__,
+                type(tool_analyses).__name__,
+            )
+            tool_analyses = {}
+        if not isinstance(priority_targets, list):
+            logging.warning(
+                "%s # ignoring malformed priority_targets (%s); expected a list",
+                self.__class__.__name__,
+                type(priority_targets).__name__,
+            )
+            priority_targets = []
+
         configs: List[Tuple[str, dict]] = []
         seen: set = set()
 
         for entry in priority_targets:
+            if not isinstance(entry, str):
+                logging.warning(
+                    "%s # skipping non-string priority target: %r",
+                    self.__class__.__name__,
+                    entry,
+                )
+                continue
             target_name = entry.split(" - ")[0].strip()
             for tool_name, analysis in tool_analyses.items():
                 if (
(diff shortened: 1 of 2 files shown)


jmartin-tech (approved):
These guards provide a more concrete expectation of the expected formats. Thank you!

PR #2070, opened 2026-08-16
fix(generators): retry transient HTTP errors and report honest failures

feiiiiii5:
## Root Cause

When a target endpoint (or a load-balancing proxy in front of it) returns a transient HTTP status that the OpenAI SDK does not map to a dedicated exception subclass — e.g. **408 REQUEST TIMEOUT** — it surfaces as the generic `openai.APIStatusError`. That type is not in the `@backoff.on_exception` retry tuple of `OpenAICompatible._call_model`, so a single transient response propagated out of the generator, aborted the whole probe run (reported in #1967: probe killed at attempt 48/256, zero `eval` entries), and was wrapped by the NIM generator into the misleading "Is the model name spelled correctly?".

## Fix

1. `OpenAICompatible._call_model` now catches `openai.APIStatusError`:
   - transient codes (`408, 429, 502, 503, 504`) re-raise as `GeneratorBackoffTrigger` so the existing fibonacci backoff retries them — a single hiccup no longer aborts the run;
   - other unmapped statuses log the code and return `[None]`, skipping just that generation so the rest of the probe continues.
2. The NIM generator's catch-all now reports the underlying error (`NIM generation failed. Underlying error: ...`) instead of blaming the model name.

## Test

New regression tests (pre-fix FAIL → post-fix PASS verified):

- `test_transient_apistatuserror_retries_and_succeeds` — three 408 responses (exhausting the SDK's own idempotent-retry budget) then a 200; asserts the call is retried through garak's backoff and returns a `Message`. Fails on main (exception propagates), passes with the fix.
- `test_non_transient_apistatuserror_skips_generation` — a 418 returns `[None]` without retrying. Fails on main.
- `test_nim_unexpected_error_message_includes_cause` — a non-OpenAI error from the base call is wrapped in a `GarakException` that names the real cause. Fails on main.

Commands run (venv: `.venv-dev`, `HOME=/tmp/garak-home` for an isolated config dir):

```
python -m pytest tests/generators/test_openai_compatible.py tests/generators/test_nim.py -k "apistatuserror or unexpected_error"
# 3 passed
```

Full files on this branch: 8 failures, all pre-existing on main (multiprocessing/subprocess tests in this environment — same 8 fail on the unmodified baseline; baseline also shows 11 with the 3 new tests failing). `ruff check` on the four touched files reports only pre-existing findings in untouched lines (F841/F541/E712 at old line numbers).

## Why this does not duplicate #1973

PR #1973 (same issue, by Varshith-Kali) is **closed without merging**. It made the retry-code list configurable via `DEFAULT_PARAMS`; this PR follows maintainer jmartin-tech's direction in the issue thread (catch `openai.APIStatusError`, inspect the code, backoff or return `None`) with a module-level constant and no new config surface, and additionally fixes the misleading NIM error message from the same report.

## AI assistance

This PR was developed with AI assistance (DeepSeek) and reviewed line-by-line by a human submitter.

### DIFF
--- a/garak/generators/nim.py
+++ b/garak/generators/nim.py
@@ -96,7 +96,9 @@ class NVOpenAIChat(OpenAICompatible):
             logging.critical(msg, exc_info=nfe)
             raise GarakException(f"🛑 {msg}") from nfe
         except Exception as oe:
-            msg = "NIM generation failed. Is the model name spelled correctly?"
+            # Include the underlying error: the previous wording blamed the
+            # model name even for unrelated failures (e.g. HTTP timeouts).
+            msg = f"NIM generation failed. Underlying error: {oe!r}"
             logging.critical(msg, exc_info=oe)
             raise GarakException(f"🛑 {msg}") from oe
 
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -25,6 +25,11 @@ from garak.attempt import Message, Conversation
 import garak.exception
 from garak.generators.base import Generator
 
+# HTTP status codes the OpenAI SDK does not map to a dedicated exception
+# subclass. They surface as the generic APIStatusError and are treated as
+# transient: a single hiccup should retry, not abort the probe run.
+_TRANSIENT_HTTP_STATUS_CODES = (408, 429, 502, 503, 504)
+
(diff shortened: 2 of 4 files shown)


jmartin-tech:
Closing duplicates #2019

PR #2081, opened 2026-08-17
fix(attempt): fall back to original outputs when reverse translation is missing

### LINKED ISSUE #2078: outputs_for() lacks the prompt_for() fallback: detectors silently see empty outputs when reverse translation is unpopulated
## Summary

`Attempt.outputs_for(lang)` and `Attempt.prompt_for(lang)` have asymmetric error contracts when the requested language does not match the prompt's language, and the asymmetry can silently empty a detector's input.

## Observed behavior (current main)

`prompt_for()` (attempt.py, ~L380) has an honest fallback:

```python
return self.notes.get("pre_translation_prompt", self.prompt)
```

`outputs_for()` (~L397) has none:

```python
return self.reverse_translation_outputs  # this needs to be wired back in for support
```

So when a detector asks for outputs in a language other than the prompt's (detectors/base.py `outputs_for(self.lang_spec)` call sites, exploitation.py, misleading.py), and reverse translation was never populated — or was populated with `None` placeholders for skipped outputs (probes/base.py ~L299-306) — the detector receives an empty or None-filled list and reports no hits, with no log or warning. The probe prompt itself survives via the `prompt_for` fallback, so the attempt looks fully formed while its outputs are silently absent: a false negative that is indistinguishable from a genuinely clean generation.

Related: #2041 fixes the `{}` default of `reverse_translation_outputs` so the type contract holds; this is the remaining half — what `outputs_for` should return when reverse translation is unavailable.

## Options

a) **Symmetric fallback**: return `self.outputs` when `reverse_translation_outputs` is empty/unpopulated, mirroring the `prompt_for` fallback (detector sees the original-language outputs instead of nothing), plus a `logging.warning` so the mismatch is visible.
b) **Explicit failure**: raise (or return a sentinel) so detectors can distinguish "no reverse translation available" from "model produced no output".
c) **Keep current behavior but warn**: keep returning the empty list, but log a warning once per attempt so silent false negatives become diagnosable.

Happy to send a PR once the direction is clear — flagging as an issue first because the "needs to be wired back in" comment suggests the intended design isn't settled, and this interacts with the translation pipeline invariants discussed in #1969.

alireza-aminzadeh:
`outputs_for(lang)` returned `reverse_translation_outputs` whenever the requested language did not match the prompt language, with no fallback. When reverse translation was never populated (default empty dict, empty list, or None placeholders only), detectors that call `outputs_for(self.lang_spec)` received nothing and reported no hits. `prompt_for()` already falls back to the original prompt in the same situation, so the attempt looked complete while its outputs were silently empty.

## Fix

If a different language is requested and reverse translation is unpopulated, return the original `outputs` and log a warning. Populated reverse translations, including partial None alignment slots, are still returned as-is.

This is complementary to #2041 (list default / type contract). That PR keeps returning an empty list for an unpopulated translated language; this change is the remaining half described in #2078.

## Not a duplicate

I searched open PRs for 2078 and for an outputs_for / reverse-translation fallback. #2041 is the type-default fix only. #2067 is the mitigation.Prefixes start-anchor change and is unrelated.

## Verification

- [x] python -m pytest tests/test_attempt.py -k outputs_for -- 3 passed
- [x] python -m pytest tests/test_attempt.py -- 24 passed; the pre-existing test_attempt_sticky_params failure is a local missing lorem dependency, not this change
- [x] Existing populated reverse-translation path still returns reverse_translation_outputs
- [x] Unpopulated / empty / None-only / legacy {} cases fall back to original outputs
- [x] Partial reverse translation keeps output alignment

Fixes #2078

### DIFF
--- a/garak/attempt.py
+++ b/garak/attempt.py
@@ -5,6 +5,7 @@ from copy import deepcopy
 from pathlib import Path
 from types import GeneratorType
 from typing import List, Optional, Union, Tuple
+import logging
 import uuid
 
 from garak.exception import GarakException
@@ -405,9 +406,17 @@ class Attempt:
             and lang != "*"
             and self.prompt.last_message().lang != lang
         ):
-            return (
-                self.reverse_translation_outputs
-            )  # this needs to be wired back in for support
+            reverse_outputs = self.reverse_translation_outputs
+            if isinstance(reverse_outputs, list) and any(
+                output is not None for output in reverse_outputs
+            ):
+                return reverse_outputs
+            logging.warning(
+                "Attempt %s: reverse_translation_outputs unpopulated for lang %s; falling back to original outputs",
+                self.uuid,
+                lang,
+            )
+            return self.outputs
         return self.outputs
 
     def _expand_prompt_to_histories(self, breadth):
--- a/tests/test_attempt.py
+++ b/tests/test_attempt.py
@@ -3,6 +3,7 @@
 
 import contextlib
 import json
+import logging
 import os
 import pytest
 
@@ -569,7 +570,6 @@ PREFIX = "_garak_test_attempt_sticky_params"
 
 
 def test_attempt_sticky_params(capsys):
-
     cli.main(
         f"-m test.Blank -g 1 -p lmrc.QuackMedicine,test.Blank -d always.Pass --report_prefix {PREFIX}".split()
     )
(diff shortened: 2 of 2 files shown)


feiiiiii5:
As the issue reporter: this matches option (a) from #2078. The populated-guard (`any(output is not None ...)`) covers all three unpopulated shapes — missing key, empty list, and None-only placeholders — and the warning makes the fallback visible in the log instead of silently handing detectors an empty output list while the attempt still looks complete. Mirroring `prompt_for()` restores exactly the invariant the issue asked for. Thanks for picking this up and for the test coverage — the populated/unpopulated split in the new tests pins the behaviour.

Issue #1546, opened 2026-01-12
factor up template `soft_probe_prompt_cap` logic


leondz:
Apropos https://github.com/NVIDIA/garak/pull/1538/changes#r2631991818

There's some duplication of generic probe cap clipping across probes; this is ripe for factoring up, e.g. into `probes.base`

leondz:
Not sure this isn't already addressed. See https://github.com/NVIDIA/garak/blob/adfddecc55fb5f3b3ed55d4fc5863d018339d491/garak/probes/base.py#L462-L470

VishnuR23:
I'd like to pick this up.

@leondz — on your question about whether this is already addressed: `Probe._prune_data` covers probes that cap `self.prompts` in place, and a good number use it. But probes that assemble a prompt list *locally* before assigning it still hand-roll the cap. There are four such sites:

- `probes/encoding.py` (`__init__`)
- `probes/sysprompt_extraction.py` (`_build_prompts`)
- `probes/dra.py` (twice, with an identical comment)

The duplication has already drifted, and not harmlessly. `dra.py` and `badchars.py` guard with `soft_probe_prompt_cap is not None`; `encoding.py` and `sysprompt_extraction.py` do not, so a config with a null cap crashes them:

```
probes.encoding.InjectBase64            TypeError: '<' not supported between instances of 'int' and 'NoneType'
probes.sysprompt_extraction.SystemPromptExtraction   TypeError: '>' not supported between instances of 'int' and 'NoneType'
```

(reproduced by setting `run.soft_probe_prompt_cap` to null and loading the probe)

Plan: add a small sibling to `_prune_data` in `probes.base` that caps a supplied list and returns it, then route those four sites through it. That fixes both crashes by construction and leaves one place to change the policy. I'll leave `badchars._downsample_prompts` alone since it deliberately preserves category balance and documents the difference.

Disclosure: I'm using AI assistance, and reviewing and testing everything myself before submitting.

leondz:
Does this pattern account for the variation in methods existing per-probe cap management?

PR #2085, opened 2026-08-18
probes: factor up local prompt-cap logic into probes.base

VishnuR23:
Addresses #1546.

## Why this is not already addressed

`Probe._prune_data` covers probes that cap `self.prompts` in place, and plenty use it. But probes that build a prompt list *locally* and then assign it still hand-roll the cap. There were four such sites:

- `probes/encoding.py` (`EncodingMixin.__init__`)
- `probes/sysprompt_extraction.py` (`_generate_attempts`)
- `probes/dra.py` (twice, with an identical comment)

The copies had already drifted, and not harmlessly. `dra.py` and `badchars.py` guard with `soft_probe_prompt_cap is not None`; `encoding.py` and `sysprompt_extraction.py` compared against it directly, so a config with no cap set crashed them:

```
probes.encoding.InjectBase64
    TypeError: '<' not supported between instances of 'int' and 'NoneType'
probes.sysprompt_extraction.SystemPromptExtraction
    TypeError: '>' not supported between instances of 'int' and 'NoneType'
```

That is the case for factoring up: with the policy in four places, two of them were wrong.

## What this changes

Adds `Probe._cap_prompt_list(prompts)` next to `_prune_data` — same policy, but for the "returns a list" shape rather than the in-place one — and routes the four sites through it. `None` handling now lives in one place, so both crashes go away by construction.

`badchars._downsample_prompts` is deliberately left alone: it preserves category balance and its docstring already explains why it differs.

Removing the last `random` use from `encoding.py` orphaned its `import random`, so that goes too.

`#1546` has no linked PR and no assignee. No open PR touches the cap logic in these files. I commented on the issue with this plan before writing code.

## AI assistance disclosure

Written with AI assistance (Claude). I reviewed every changed line, reproduced both crashes before fixing them, and confirmed the new tests fail without the fix.

## Verification

- [x] Supporting configuration — n/a
- [x] Both crashes reproduced before the change, gone after, with capping still applied:
  ```
  cap = None    encoding.InjectBase64 835 prompts | sysprompt_extraction 1400 | dra.DRA 60   (all previously TypeError)
  cap = 12      encoding.InjectBase64  12 prompts | sysprompt_extraction   12 | dra.DRA 12   (capped)
  ```
- [x] New tests fail without the fix — reverting only the three probe files gives:
  ```
  FAILED test_probe_builds_with_unset_cap[probes.encoding.InjectBase64]
  FAILED test_probe_builds_with_unset_cap[probes.sysprompt_extraction.SystemPromptExtraction]
  2 failed, 1 passed
  ```
  (`dra` passes either way — it already had the guard, which is the point.)
- [x] `python -m pytest tests/probes/test_probes.py tests/probes/test_probes_encoding.py tests/probes/test_probes_sysprompt_extraction.py tests/probes/test_probes_dra.py tests/probes/test_probes_badcharacters.py` — 1466 passed, 1 skipped
- [x] **Verify** the thing does what it should — tests cover downsampling to the cap, lists already within the cap, an unset cap, and `follow_prompt_cap = False`
- [x] **Verify** the thing does not do what it should not — capping behaviour is unchanged for all three probes at a set cap; `badchars` category-balanced downsampling is untouched and its tests still pass
- [x] **Document** the thing and how it works — `_cap_prompt_list` docstring states when to use it versus `_prune_data`

Two unrelated failures show up locally in `test_probes.py` (`probes.audio.AudioAchillesHeel`, `probes.sata.MLM`). They fail identically on unmodified `main` in my environment — missing `audio` extras and NLTK data — and are not touched by this change.

### DIFF
--- a/garak/probes/adaptive_attacks.py
+++ b/garak/probes/adaptive_attacks.py
@@ -91,9 +91,7 @@ class AdaptiveAttacks(garak.probes.Probe):
     doc_uri = "https://arxiv.org/abs/2404.02151"
     intent = "S006"  # default seeds ask the model to produce instructions related to various harms
 
-    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {
-        "follow_prompt_cap": True,
-    }
+    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {}
 
     def __init__(self, config_root=_config):
         super().__init__(config_root=config_root)
@@ -103,5 +101,4 @@ class AdaptiveAttacks(garak.probes.Probe):
         for seed in seeds:
             for suffix in suffixes:
                 self.prompts.append(ADAPTIVE_TEMPLATE.format(goal=seed, suffix=suffix))
-        if self.follow_prompt_cap:
-            self._prune_data(cap=self.soft_probe_prompt_cap)
+        self._prune_data()
--- a/garak/probes/agent_breaker.py
+++ b/garak/probes/agent_breaker.py
@@ -159,7 +159,6 @@ class AgentBreaker(garak.probes.IterativeProbe):
         "parse_model_config": None,
         "end_condition": "verify",
         "max_calls_per_conv": 50,
-        "follow_prompt_cap": True,
         # Path to agent configuration YAML file (relative to garak/data/)
         "agent_config_file": "agent_breaker/agent.yaml",
         # Maximum number of exploit attempts per tool before moving to next target
--- a/garak/probes/apikey.py
+++ b/garak/probes/apikey.py
@@ -18,7 +18,9 @@ class GetKey(garak.probes.Probe):
     """
 
     lang = "en"
-    intent = "S009cred"  # asks the target to emit API keys and other credential-like secrets
+    intent = (
+        "S009cred"  # asks the target to emit API keys and other credential-like secrets
+    )
     primary_detector = "apikey.ApiKey"
     tags = [
         "quality:Behavioral:ContentSafety:LegalGoodsServices",
@@ -28,9 +30,7 @@ class GetKey(garak.probes.Probe):
     tier = garak.probes.Tier.COMPETE_WITH_SOTA
     active = True
 
-    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {
-        "follow_prompt_cap": True,
-    }
+    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {}
 
     def __init__(self, config_root=_config):
         super().__init__(config_root=config_root)
@@ -39,8 +39,7 @@ class GetKey(garak.probes.Probe):
         for key_type in KEY_TYPES:
             for base_prompt in self.base_prompts:
                 self.prompts.append(base_prompt.replace("[key]", key_type))
-        if self.follow_prompt_cap:
-            self._prune_data(cap=self.soft_probe_prompt_cap)
+        self._prune_data()
 
 
 class CompleteKey(garak.probes.Probe):
@@ -62,9 +61,7 @@ class CompleteKey(garak.probes.Probe):
     tier = garak.probes.Tier.COMPETE_WITH_SOTA
     active = True
 
-    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {
-        "follow_prompt_cap": True,
-    }
+    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {}
 
     partial_keys = [
         "Ikj7sf7Uy",
@@ -88,5 +85,4 @@ class CompleteKey(garak.probes.Probe):
                             "[partial]", partial
                         )
                     )
-        if self.follow_prompt_cap:
-            self._prune_data(cap=self.soft_probe_prompt_cap)
+        self._prune_data()
--- a/garak/probes/badchars.py
+++ b/garak/probes/badchars.py
@@ -143,7 +143,6 @@ class BadCharacters(garak.probes.Probe):
         "max_position_candidates": 24,
         "max_reorder_candidates": 24,
         "max_ascii_variants": len(ASCII_PRINTABLE),
-        "follow_prompt_cap": True,
     }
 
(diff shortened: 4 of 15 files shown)


leondz:
This looks like it only tests three probes instead of all that support capping - why?

jmartin-tech (changes requested):
The goal of the issue being targeted here was to consolidate and ensure _all_ probes use a shared application of the pruning process as provided by a `base` class.

For the probes addressed here setting `self.prompts` then calling `self._prune_data(self.soft_probe_prompt_cap)` should be sufficient to get the same result, why add a new function that implements the pruning without awareness of `triggers`?

Note by calling `self._prune_data()` this also offers probe authors a single location to override the pruning process when they need to preform a revised pattern for pruning.  For instance, this would allow for `badchars._downsample_prompts()` to simply be renamed `_prune_data()` with a minor signature adjustment and shift that outlier.

Another consolidation task that could be considered part of the reported issue is that `follow_prompt_cap` could become a `probes.base.DEFAULT_PARAM` evaluated in `_prune_data()` ensuring all probes have an ability to enable or disable pruning in a common way and shifting the burden from each probe that currently supports the option to the base class.

VishnuR23:
Thanks both — reworked to follow the design you described. Force-pushed.

@jmartin-tech, you were right that a second function was the wrong shape. `_prune_data` is now the single place pruning happens:

- it defaults `cap` to `soft_probe_prompt_cap` and evaluates `follow_prompt_cap` itself, so call sites no longer guard it
- `follow_prompt_cap` moved to `Probe.DEFAULT_PARAMS`, so every probe carries the switch instead of ~13 probes each declaring it. Existing class attributes still win, since `Configurable` only fills in a default when the attribute is absent — so `follow_prompt_cap = False` on `ContinueSlursReclaimedSlursFull` and the `leakreplay` mixins keeps working
- `badchars._downsample_prompts` is now a `_prune_data` override, which is what its category-balanced pruning always was
- `_prune_cap` gives overrides the same opt-out/unset-cap handling, so `IntentProbe`, `badchars` and `encoding` don't each repeat the guard. Happy to inline it if you'd rather it not exist

Sweeping for the rest turned up more than the three I first found. `latentinjection._build_prompts_triggers` held a verbatim copy of the base body; `web_injection`, `encoding`, `sysprompt_extraction` and `dra` each capped a local list before assigning. Those now assign and prune. `encoding` needs an override because `_prompt_intents` is indexed by prompt position, the same reason `IntentProbe` overrides for `prompt_intents`.

`divergence` is the one I deliberately left: it uses the cap as a generation target (`while len(self.prompts) < prompt_cap`) rather than pruning afterwards, so routing it through `_prune_data` would mean generating everything first. Say the word if you'd prefer it converted anyway.

@leondz — fair hit on the test coverage. Tests no longer name probes; they discover every probe that prunes by walking the plugin registry and each class's MRO. That is **77 probes across 16 modules**, each checked for honouring the cap, building with no cap set, and keeping triggers aligned after pruning.

That coverage also showed the null-cap crash was much wider than the two probes I originally reported — running the new tests against `main` gives **34 failures**, including all of `packagehallucination`, `phrasing` and `suffix`.

Verification: `pytest tests/probes tests/cas tests/test_docs.py` → **2901 passed, 41 skipped**. The two failures in my environment (`probes.audio.AudioAchillesHeel`, `probes.sata.MLM`) fail identically on unmodified `main` — missing audio extras and NLTK data.

`black` is clean on every file I touched. I left nine other files in `garak/probes/` that `black` would reformat untouched, since they're unrelated to this change.

garak/probes/adaptive_attacks.py
```diff
@@ -91,9 +91,7 @@ class AdaptiveAttacks(garak.probes.Probe):
     doc_uri = "https://arxiv.org/abs/2404.02151"
     intent = "S006"  # default seeds ask the model to produce instructions related to various harms
 
-    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {
-        "follow_prompt_cap": True,
-    }
+    DEFAULT_PARAMS = garak.probes.Probe.DEFAULT_PARAMS | {}
```
jmartin-tech: Probes that no longer introduce their own custom params can simply remove the definition and inherit from the parent class.

MohammedAlkindi:
Reproduced your two numbers on Windows, since they are the load-bearing part.

Applying only this PR's `tests/probes/test_probes.py` to unmodified `main`: **34 failed, 51 passed, 25 skipped**, matching your 34 exactly. Collection finds **77** capping probes, also matching.

On the branch, 190 passed. What still fails here is my environment, not your change: `zalgolib`, `base2048` and `ecoji` are missing optional encoders. The one that looked like a regression, `sysprompt_extraction.SystemPromptExtraction` building zero prompts with no cap, builds zero on `main` too, so this PR did not cause it. It is visible only because your discovery newly pulls that probe into the capping set.

Both the null-cap breakage and the coverage figure hold from a second machine and a different OS.

VishnuR23:
@MohammedAlkindi thank you — an independent reproduction on a different OS is exactly what this needed, and you found a real weakness in my test.

You're right that `sysprompt_extraction.SystemPromptExtraction` building zero prompts is pre-existing. Its prompts come from datasets fetched at load time (50 system prompts x 28 attack templates = 1400 here), so a machine that can't reach them legitimately builds none. My `assert len(p.prompts) > 0` therefore asserted something about network reachability rather than about this change.

Fixed in ea24b0b: the unset-cap test now asserts only what it is actually guarding — that building with no cap set doesn't raise. Comparing a length against `None` is what used to blow up, so the load itself is the assertion. Re-verified that this doesn't weaken it: applying the tests to unmodified `main` still gives **34 failed, 52 passed, 25 skipped**, and 198 pass on the branch.

Your point that the probe is only visible now because the discovery pulls it into the capping set is worth recording. It looks like a genuine pre-existing gap — a probe that silently produces nothing when its data source is unavailable — but it belongs in its own issue rather than here.

Full CI is now green on this branch across Linux x86 and arm64, macOS and Windows on 3.11/3.12/3.13, plus docs and DCO.

@jmartin-tech @leondz — this is ready for another look whenever you have time. The review state still shows the original request for changes from before the rework.

PR #2086, opened 2026-08-18
config: accept list-type values for legacy plugin specs

VishnuR23:
Addresses #1563.

## Current state

`run.spec` already accepts list markup in config files through `parse_spec_file` (the `include`/`exclude` lists), so the modern selection surface was covered. The legacy keys were not.

`_legacy_path_selectors` resolves them with `str(spec).split(",")`, so a list arrives as its `repr` and is silently mis-parsed — no error, just nothing selected:

```
parse_plugin_spec("mitigation,always", "detectors")       -> 7 detectors, unknown=[]
parse_plugin_spec(["mitigation", "always"], "detectors")  -> 0 detectors, unknown=["['mitigation'", "'always']"]
```

Loaded from an actual config file on `main`, the list form gives `0 detectors`. `detector_spec` still resolves through this path, and `probe_spec`/`buff_spec` pass through it on their way through the `run.spec` migration shim.

## What this changes

Adds `_normalize_legacy_spec`, which flattens a list or tuple to the comma-separated form the legacy grammar already understands, and applies it at the one point where legacy specs become selectors. Every legacy key gains list support together, and resolution semantics are untouched — a list is turned into exactly the string that already worked.

`legacy_selection_spec._meaningful` gets the same normalisation, so an empty list reads as an absent spec rather than as a value (`str([])` is non-empty, so it previously counted as one).

After the change, all three spellings agree:

```
detector_spec: mitigation,always     (yaml string)  -> 7 detectors
detector_spec: [mitigation, always]  (yaml list)    -> 7 detectors
{"detector_spec": ["mitigation", "always"]} (json)  -> 7 detectors
```

Keywords keep their meaning wrapped in a list (`["all"]`, `["none"]`, `["auto"]`), and blank entries are dropped rather than becoming empty clauses.

`#1563` has no linked PR and no assignee, and no open PR touches `garak/_spec.py`. I commented on the issue with this plan before writing code.

## AI assistance disclosure

Written with AI assistance (Claude). I reviewed every changed line, reproduced the silent mis-parse against `main` before fixing it, and confirmed the new tests fail without the fix.

## Verification

- [x] Supporting configuration — config files used in testing:
  ```yaml
  plugins:
    detector_spec:
      - mitigation
      - always
  ```
  ```json
  {"plugins": {"detector_spec": ["mitigation", "always"]}}
  ```
  Both load and resolve to the same 7 detectors as `detector_spec: mitigation,always`. On `main` both give 0.
- [x] Run the tests and ensure they pass: `python -m pytest tests/test_spec.py tests/test_config.py tests/cli tests/cas tests/resources` — 487 passed, 1 skipped
- [x] `python -m pytest tests/test_docs.py` — 680 passed
- [x] New tests fail without the fix — reverting only `garak/_spec.py` gives `10 failed, 7 passed`
- [x] **Verify** the thing does what it should — tests assert list and tuple forms resolve identically to the comma string across probes, buffs and detectors, and that a list `probe_spec` migrates to the same `run.spec`
- [x] **Verify** the thing does not do what it should not — tests assert `all` / `none` / `auto` keep their meaning inside a list, blank entries are dropped, and an empty list is treated as absent rather than as a selection
- [x] **Document** the thing and how it works — `docs/source/configurable.rst` now shows both spellings for `detector_spec`; `_normalize_legacy_spec` carries a docstring

One note on the diff: `black` rewrapped one pre-existing `selectors.append(...)` line in `_spec.py` that I did not otherwise change — it was over the line limit on `main`. Happy to drop that hunk if you would rather keep the diff minimal.

### DIFF
--- a/docs/source/configurable.rst
+++ b/docs/source/configurable.rst
@@ -143,6 +143,19 @@ Plugins Config Items
 * ``target_type`` - The type of target generator, e.g. "nim" or "huggingface"
 * ``target_name`` - The specific name of the target to be used (optional - if blank, type-specific default is used)
 * ``detector_spec`` - An optional spec of detectors to be used, if overriding those recommended in probes. Specifying ``detector_spec`` means the ``pxd`` harness will be used. This is equivalent to passing `-d` to the CLI
+
+  ``detector_spec`` takes either a comma-separated string or a list, so these are equivalent:
+
+  .. code-block:: yaml
+
+      detector_spec: mitigation,always
+
+  .. code-block:: yaml
+
+      detector_spec:
+        - mitigation
+        - always
+
 * ``extended_detectors`` - Should just the primary detector be used per probe, or should the extended detectors also be run? The former is fast, the latter thorough.
 * ``buffs_include_original_prompt`` - When buffing, should the original pre-buff prompt still be included in those posed to the model?
 * ``buff_max`` - Upper bound on how many items a buff should return
--- a/garak/_spec.py
+++ b/garak/_spec.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 
 import re
 from dataclasses import dataclass, field
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Sequence, Union
 
 # Plugin categories selectable via run.spec. Detectors are not selectable here
 # yet; they keep their own legacy spec surface (parse_plugin_spec).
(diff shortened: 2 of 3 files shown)


jmartin-tech:
As of `0.16.0` the plugin type `*_spec` options other than the `detector_spec` are deprecated and slated for removal, we should not expand their capabilities any further.

Note #1789 is the placeholder for the next step of consolidating detectors support on a common `spec` syntax.

PR #2089, opened 2026-08-18
[withdrawn by author]

ghost:
Withdrawn by the author. Content removed at the author's request.

### DIFF
--- a/garak/detectors/promptinject.py
+++ b/garak/detectors/promptinject.py
@@ -11,13 +11,23 @@ from garak.detectors.base import Detector
 
 
 class AttackRogueString(Detector):
-    """Detect if model emits trigger string"""
+    """Detect if model output contains an injected trigger string
+
+    This is a substring check, case-insensitive unless ``case_sensitive`` is
+    set. A hit means the rogue string appeared somewhere in the output, which
+    is necessary but not sufficient for successful instruction override: a
+    model may quote the string, echo it while refusing, or reproduce it
+    incidentally without having followed the injected instruction.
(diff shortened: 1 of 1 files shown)


patriciapampanelli:
@LiyemaSwartbooi, please supply a PR with signed DCO so we can review it

LiyemaSwartbooi:
Thanks @patriciapampanelli. Amended with a signed DCO and force-pushed, now `ee9bf5d`. Ready for review whenever suits.

garak/detectors/promptinject.py
```diff
@@ -11,13 +11,23 @@
...
 
 class AttackRogueString(Detector):
-    """Detect if model emits trigger string"""
+    """Detect if model output contains an injected trigger string
+
+    This is a substring check, case-insensitive unless ``case_sensitive`` is
+    set. A hit means the rogue string appeared somewhere in the output, which
+    is necessary but not sufficient for successful instruction override: a
+    model may quote the string, echo it while refusing, or reproduce it
+    incidentally without having followed the injected instruction.
```
patriciapampanelli, line 20: Nit: garak assesses "targets", not "models":
```suggestion
    """Detect if target output contains an injected trigger string

    This is a substring check, case-insensitive unless ``case_sensitive`` is
    set. A hit means the rogue string appeared somewhere in the output, which
    is necessary but not sufficient for successful instruction override: a
    target may quote the string, echo it while refusing, or reproduce it
    incidentally without having followed the injected instruction.
    """
```

Issue #1911, opened 2026-07-05
pass@k for ASR


stefanoamorelli:
Each probe sends a prompt to the model a few times ([`probes/base.py#L299-L304`](https://github.com/NVIDIA/garak/blob/af5564fd8e1297fab16dace6ef8f5d312421a873/garak/probes/base.py#L299-L304)), and the attack success rate represents how many of those replies resulted in a breach.

The problem that I see with this approach is that an attacker doesn't need the model to fail most of the time. They need it to fail once, and they can just try again. So a jailbreak that only lands on 1 try in 5 is effectievly a full break of that prompt, but it shows up as only 20% ASR, missing the point of a critical vulnerability.

an approach I would suggest is `pass@k`, the metric introduced in OpenAI's Codex paper ([Chen et al., 2021](https://arxiv.org/abs/2107.03374)). instead of pooling, you ask, per prompt: "given k tries, does at least one get through? "

it gives a more meaningful ASR reading as it reflects how the model actually gets attacked, because, fundamentally, an attacker can keep trying.

jmartin-tech:
Linking to discussion #1684 that is open relating to this issue. Meaningful metrics vary by user intent, but the point here on how this tool defines ASR is acknowledged to need improvement.

chuenchen309:
Thanks @stefanoamorelli — the "an attacker only needs to land once" framing is the right lens, and the pooled rate genuinely understates single-prompt breaks. Having just read through the evaluation path, a few notes on where this would slot in and one design fork worth deciding early:

**Where the pooling happens.** `Evaluator.evaluate()` ([`evaluators/base.py`](https://github.com/NVIDIA/garak/blob/main/garak/evaluators/base.py)) flattens across both loops — outer over attempts (prompts), inner over each generation's `detector_results` — into a single `passes`/`fails` tally, so `failrate = (evals - passes) / evals` is a per-*generation* mean. pass@k is a per-*prompt* reduction applied *before* that flattening: for each attempt, "did ≥1 of its k generations breach?", then average over attempts. The data already supports this cleanly — each `attempt.detector_results[detector]` is the per-generation vector for one prompt, aligned to `attempt.outputs`.

[...]

PR #2090, opened 2026-08-18
feat(evaluators): add hit@k attack success rate metric

stefanoamorelli:
Adds **hit@k** for attack success rate (implements #1911).

Pooled ASR reports the fraction of all generations that breach the target, which understates real risk: a jailbreak that lands on 1 reply in 5 is a working jailbreak, yet pooled ASR records it as 20%. hit@k instead asks per prompt "given k attempts, does at least one breach the target?" and averages across prompts, following the pass@k estimator of [Chen et al. (2021)](https://arxiv.org/abs/2107.03374). It is named hit@k because garak scores an attack success as a hit.

Example, 5 prompts, each breached on 1 of 5 generations:

```
attack success rate:  20.00%  hit@1: 20.00%  hit@5: 100.00%
```

Every prompt is always scored at k equal to the number of generations it actually got, where the estimator collapses to "was this prompt breached at least once" (prompts with unequal generation counts share an `"n"` bucket). `reporting.hit_at_k` (default `[1]`) adds further k values below that, and `null` disables the metric entirely.

A `hit_at_k` field is written in the `eval` report entry, copied into each detector's entry in the report digest next to `total_evaluated` and `passed`, and printed on the CLI.

_This is an additive change and the existing pooled ASR remains unchanged._

### Config examples

Default `[1]`, on a run with `run.generations: 5` (5 prompts, each breached on 1 of 5 generations):

```
(attack success rate:  20.00%  hit@1: 20.00%  hit@5: 100.00%)
```
```json
"hit_at_k": {"1": {"score": 0.2, "prompts": 5}, "5": {"score": 1.0, "prompts": 5}}
```

More k values, on a run with `run.generations: 10` (5 prompts, each breached on 2 of 10 generations):

```yaml
reporting:
  hit_at_k: [1, 5]
```
```
(attack success rate:  20.00%  hit@1: 20.00%  hit@5: 77.78%  hit@10: 100.00%)
```
```json
"hit_at_k": {"1": {"score": 0.2, "prompts": 5}, "5": {"score": 0.7778, "prompts": 5}, "10": {"score": 1.0, "prompts": 5}}
```

Disabled: the CLI suffix is dropped and no `hit_at_k` field is written.

```yaml
reporting:
  hit_at_k: null
```
```
(attack success rate:  20.00%)
```

## Tests

```
$ python -m pytest tests/analyze/test_hit_at_k.py tests/analyze/test_report_digest.py tests/evaluators/test_evaluators.py
103 passed in 0.59s
```

---

I used claude code to help me with this. I reviewed every line and take responsibility for the change.


garak/resources/garak.core.yaml
```diff
@@ -40,4 +40,5 @@ reporting:
   confidence_interval_method: bootstrap
   bootstrap_num_iterations: 10000
   bootstrap_confidence_level: 0.95
-  bootstrap_min_sample_size: 30
\ No newline at end of file
+  bootstrap_min_sample_size: 30
+  pass_at_k: [1, 5]
```
jmartin-tech: Consider always injecting the `generations` value.

With the expectation that this is number of attempts required to get a `hit`.

Ask that term `hit` be used to align to what `garak` uses in detectors.

docs/source/reporting.rst
```diff
@@ -50,6 +50,21 @@ Confidence intervals are enabled by default using the bootstrap method (see ``re
 
 These intervals account for sampling uncertainty. When detector performance metrics (sensitivity/specificity) are available, they also account for detector imperfection. Otherwise, a perfect detector is assumed.
 
+pass@k Attack Success Rate
```
jmartin-tech: Given that terminology in `garak`, while I understand this diverges from the paper the metric here is probably more clear as `hit@k` for the following reasons:
* `detector.always.Pass` refers maps to scoring as `0.0`
* the [detector base class](https://github.com/NVIDIA/garak/blob/8ed1543b985a5722adb659584182faf6f7907d4e/garak/detectors/base.py#L33-L34) refers to a successful result as a `hit` and a `pass` has opposite semantics
* Detections for closer evaluation are logged in `hitlog.jsonl` further aligning this product's terminology.

garak/resources/garak.core.yaml
jmartin-tech: As noted in another comment `garak`'s jargon would align better as `hit_at_k`.

These seem like reasonable values, and the `5` here matches to the default `generations` value. Do you think there is value in having this configuration be for expanded coverage instead of all values to be used?

Consider that a default behavior where the number of `generations` the run was launched for would always include the `hit@k` where `k == generations` or more simply the for an `Attempt` we could always report at least `k = len(self.outputs)`. Then augment that set with any additional values in `reporting.hit_at_k` that are less than the number outputs in the `Attempt`.

I could see some probes this might not align to `generations` such as `atkgen.Tox` though if the upper bound on `k` is based on `len(Attempt.outputs)` that might turn out to be a non-issue.

With the expectation that this is the number of inference generations aligned to at least one `hit`  identified this metric seem very helpful.

jmartin-tech (changes requested):
Awesome contribution, the maintainer team took a bit of time working thru what this metric means and this scoring pattern seems like it would surface a better understanding of the run results.

One item I see missing is that this only incorporates these values for terminal output. The most common artifact reviewed is understood to be the `report.html`. This PR's scope is not required to expand to expose this in that space however it should likely expand to define how and where in the final `digest` object written to `report.jsonl` this might be able to provide the same values being presented on `stdout`.

Further confirmation of how the scores function is stil ongoing but the overall feedback should enable iteration while that review is completed.

docs/source/reporting.rst
stefanoamorelli: makes sense :+1: i've renamed [here](https://github.com/NVIDIA/garak/blob/5f464ed4c3bce597e00b78b33b42abf4f31ea458/docs/source/reporting.rst#L58) (I still kept the reference in comments to the original paper as generally people might know the original term).

garak/resources/garak.core.yaml
stefanoamorelli: > Do you think there is value in having this configuration be for expanded coverage instead of all values to be used?
> 
> Consider that a default behavior where the number of generations the run was launched for would always include the hit@k where k == generations or more simply the for an Attempt we could always report at least k = len(self.outputs). Then augment that set with any additional values in reporting.hit_at_k that are less than the number outputs in the Attempt.

this is actually a very good point, I'm aligned with expanded coverage (drafted [here](https://github.com/NVIDIA/garak/commit/5f464ed4c3bce597e00b78b33b42abf4f31ea458) lmk wdyt).

> I could see some probes this might not align to generations such as atkgen.Tox though if the upper bound on k is based on len(Attempt.outputs) that might turn out to be a non-issue.

I had a look a bit deeper on `atkgen.Tox` and we should indeed be covered (in the sense that upper bound on `k` is based on the count of scoreable outputs for each attempt).

However, in this case I think the implications can become more nuanced, hence we could also introduce an additional score.

let's say `prompt A` has 1 hit in 5 generations, and `prompt B` has 0 hits in 3 generations. report could be: 

`"hit_at_k": {"1": {"score": 0.1, "prompts": 2}, "n": {"score": 0.5, "prompts": 2}}` 

so the "n" key shows that half the prompts were breached (at least) once, and `0.1` is calculated as follows:

- `prompt A` 1 hit among 5 generations, `1/5` = `0.2`
- `Prompt B` 0 hits among 3 generations, `0/3` = `0`

And the average across prompts with equal weight is `(0.2 + 0.0)` / `2` = `0.1`

IMOH the importance of `n` is that it shows how effective is "retrying" in increasing the odds of a successful attack.

lmk your thoughts, your feedback and ideas are really appreciated as always! :bow:

stefanoamorelli:
> Awesome contribution, the maintainer team took a bit of time working thru what this metric means and this scoring pattern seems like it would surface a better understanding of the run results.
> 
> One item I see missing is that this only incorporates these values for terminal output. The most common artifact reviewed is understood to be the `report.html`. This PR's scope is not required to expand to expose this in that space however it should likely expand to define how and where in the final `digest` object written to `report.jsonl` this might be able to provide the same values being presented on `stdout`.
> 
> Further confirmation of how the scores function is stil ongoing but the overall feedback should enable iteration while that review is completed.

really appreciate it @jmartin-tech :bow: aligned, I should have addressed all your initial comments.

TIL about the HTML artifact being the most common, can link it easily in a follow up once we clarify the pending points.

garak/evaluators/base.py
```diff
@@ -134,9 +142,18 @@ def _evaluate_one_detector(
...
         outputs_evaluated = passes + fails
         outputs_processed = passes + fails + nones
 
+        hit_at_k_ks = getattr(_config.reporting, "hit_at_k", None)
+        hit_at_k_scores = (
+            garak.analyze.hit_at_k.estimate_hit_at_k(per_attempt_counts, hit_at_k_ks)
+            if hit_at_k_ks is not None
+            else {}
+        )
+
```
jmartin-tech: Is this the correct calculation?  Should the hit value be the number of hits for the overall attempt counts?

Thinking on this calculation I think as a metric it would be more interesting to base the metric on the number of hits within `k` attempts and that suggest to me the set of detections should be passed in full not the aggregate number of hits from the attempt.

For example a run with 3 attempt at 5 generations expressing hit@1, hit@3, hit@5 with the following detections:
```
a1 = [ 0, 0, 0, 1, 0 ]
a2 = [ 0, 0, 1, 0, 1 ]
a3 = [ 1, 1, 1, 1]
```
Current calc makes this:
```
per_attempt_counts = [
    (5, 1),
    (5, 2),
    (5, 5),
]
```
resulting in:
```
{1: {'score': 0.5333333333333333, 'prompts': 3}, 3: {'score': 0.8333333333333334, 'prompts': 3}, 5: {'score': 1.0, 'prompts': 3}}
```
However this value is complex to understand, hit@1 is actually saying 53% of 15 requests using 3 unique prompt 5 times each had 8 hits.

I would however suggest intuitively that hit@1, should be 33% stating that of 3 requests using 3 unique prompts 1 time each had 1 hit.

MohammedAlkindi:
For whoever picks this back up: the `report.jsonl` request looks addressed, tracing it end to end.

`evaluators/base.py` sets `eval_record["hit_at_k"]` when `_config.reporting.hit_at_k` is configured, and that same record is what line 243 writes to `reportfile`, so the values do land in the eval entries. `report_digest.py` then reads them back through `_map_hit_at_k(evals)` and assigns `probe_detector_result["hit_at_k"]`, so the digest object carries them as well.

One qualifier worth stating explicitly, since the ask was how and where: it is gated on the config key, so nothing appears unless `hit_at_k` is set.

This is only about the plumbing that was requested, not about whether the metric itself is right.

feiiiiii5:
Ran this branch against a chunked report and hit one case the current tests don't cover. Not a review of the metric itself — just the digest plumbing.

`_map_hit_at_k` keys by `(probe module, probe class, detector)` and assigns per eval row, so for a pairing that appears in more than one row the last row wins. `aggregate_reports` writes one `eval` row **per contributing run** for the same pairing (that is the point of it — assembling "a report that's been run one probe at a time"), so a hit@k run that was split across files reports one file's scores for the whole pairing.

Reproduced on `5f464ed`, offline, shipped plugins only, with two eval rows for `test.Test` / `always.Pass` — `0 passed / 10 evaluated` carrying `hit@1 = 0.0`, then `10 / 10` carrying `hit@1 = 1.0`; the truth for the pairing is 10/20 and 0.5:

| tree | digest entry for that pairing |
|---|---|
| `main` (`8d1259ef`) | `passed 0, total_evaluated 10, absolute_score 1.0, absolute_defcon 5` — counts from the first row (`limit 1`, `report_digest.py:661`), score from the highest-scoring row (`order by score asc`, `:361`) |
| this branch | same, plus `hit_at_k {"1": {"score": 1.0, "prompts": 10}}` — so the row reads "0 of 10 passed, 100% hit@1" |

The score/counts split is reported in #2213, and #2214 (my open draft, `report_digest.py` again) pools the rows in `_parse_report` before any of this, which makes counts and `absolute_score` agree. But pooling does not help `hit_at_k`: whichever row survives carries its own per-chunk scores next to pooled totals, and it also flips which chunk wins (first instead of last). So whichever of the two lands second needs a rule for it, and the options look like:

* fold it the way counts are folded — sum `prompts`, recompute `score` — which means `hit@k` has to be pooled where the counts are, not where it is mapped; or
* treat it the way the confidence interval is treated, i.e. drop it for a pooled pairing rather than present one sample set's number as the whole run's.

Dropping is cheaper and never wrong; folding is what a reader probably wants. Happy to rebase #2214 onto whichever you land first and add the `hit@k` half of the merge rule there if that's useful, rather than duplicating the decision in two PRs.

Smaller thing: `test_build_digest_carries_hit_at_k_to_detector` uses a single eval row, so nothing pins the multi-row case either way — one extra row in that fixture would.

Disclosure: written with AI assistance (Claude); the commands and both table rows above were run locally against `5f464ed` and `8d1259ef`. I have a related PR open (#2214), so I have a stake in the pooling decision — worth weighing accordingly.

PR #2098, opened 2026-08-21
pin anthropic python client until compatible

### LINKED ISSUE #2099: CI: anthropic 1.0.0 moved to httpx2, so respx no longer mocks it and test_anthropic_generator calls the live API
## What happens

`tests/generators/test_anthropic.py::test_anthropic_generator` sends a real request to `https://api.anthropic.com/v1/messages` and fails with a 401, plus a teardown error for the respx route that was never called.

```
FAILED tests/generators/test_anthropic.py::test_anthropic_generator - anthropic.AuthenticationError: Error code: 401 - {'type': 'error', 'error': {'type': 'authentication_error', 'message': 'invalid x-api-key'}}
ERROR  tests/generators/test_anthropic.py::test_anthropic_generator - AssertionError: RESPX: some routes were not called!
```

Nightly Testing on `main` is red with this on macOS 3.11, Windows 3.12 and ubuntu-24.04-arm 3.12. The 2026-08-20 nightly was green and the 2026-08-21 nightly was red on the same commit, `be70cf56`, so nothing in the tree changed.

## Why

`anthropic` 1.0.0 was released 2026-08-20 19:58 UTC and switched its HTTP stack:

```
$ python -c "import json,urllib.request; print([r for r in json.load(urllib.request.urlopen('https://pypi.org/pypi/anthropic/1.0.0/json'))['info']['requires_dist'] if 'http' in r])"
['httpx2<3,>=2.0.0', ...]
```

`respx` patches `httpx`, not `httpx2`. Its current release, 0.23.1, still declares `httpx>=0.25.0`, so the mock cannot see anything the SDK sends now.

`pyproject.toml:141` has no ceiling:

```
"anthropic>=0.40.0",
```

so a fresh CI install now resolves to 1.0.0 and the request goes to the live API.

Reproduced locally with `anthropic==1.0.0`, `respx==0.23.1`:

```
INFO  httpx2:_client.py:1085 HTTP Request: POST https://api.anthropic.com/v1/messages "HTTP/1.1 401 Unauthorized"
1 failed, 1 error in 0.84s
```

The `httpx2` and `httpcore2` logger names in that trace are the whole story. Note this means the suite currently makes an outbound call to a third party API on every run, which seems worth stopping regardless of the red.

## Suggested fix

`openai` did the same migration in 3.0 and `pyproject.toml:142` already caps it:

```
"openai>=2.0,<3.0",
```

Doing the same for anthropic matches that and unblocks CI now:

```
"anthropic>=0.40.0,<1.0",
```

Verified locally: with the cap, pip resolves 0.125.0 and `tests/generators/test_anthropic.py` is 5 passed, 1 skipped.

The durable fix is larger. `test_mistral.py`, `test_guardrails.py`, `test_openai.py`, `test_openai_compatible.py`, `test_ollama.py`, `test_azure.py` and `test_cohere.py` all mock through respx, and each one is exposed exactly the same way once its SDK moves to httpx2. Probably that means mocking the client method instead of the transport. Not a hotfix.

Happy to send the cap as a PR if you want it, and to take the wider one if you would rather it went a different way.

jmartin-tech:
Changes in anthropic 1.0.0 to use httpx2 have impacts to tests as well as USER_AGENT configuration controls

## Verification


erickgalinkin (approved):
LGTM

PR #2105, opened 2026-08-21
fix(detectors): score evaluation-model outages as unmeasured in agent_breaker

### LINKED ISSUE #2104: agent_breaker: evaluation-model failures score as 'attack failed' — a judge outage reports the target as safe (false-clean ASR)

feiiiiii5:
## What this PR does

`AgentBreakerResult` scored every evaluation-model failure as "attack failed" — in a vulnerability scanner that inversion is the worst failure mode: a judge outage makes every attempt read as defended and the ASR reports 0% ("no vulnerabilities found"). Fixes #2104.

## Root cause

`verify()` — the single verification entry point shared by probe loop-control and detector scoring — converts judge errors into verdicts:

```python
        except Exception as e:
            logging.error("%s evaluation model error: %s", self.__class__.__name__, e)
            return False, 0.0, "Evaluation model error"

        if not response or response[0] is None or response[0].text is None:
            return False, 0.0, "Empty response from evaluation model"
```

Both consumers turn `False` into a `0.0` detector score (the probe's `verified_results` fast path, and `detect()`'s independent fallback whose blanket `except` also scored `0.0`). Score `0.0` means the attack did not succeed — the model defended.

## Fix

`verify()` returns `None` as the success flag when the evaluation model could not be consulted (`tuple[bool | None, float, str]`); the three scoring sites map `None` to a `None` score — garak's established "not evaluated" signal (`detect()` already returns `[None]` for attempts with no outputs, and the declared type is `List[float | None]`), so unmeasured attempts leave the ASR denominator instead of counting as defended. Probe loop control is unchanged: `None` is falsy exactly like `False`.

## Verification

- Five new regression tests pin the semantics end to end (`TestEvaluationModelFailure`): `verify()` outage/empty → `None` flag; `detect()` fallback outage → `[None]`; fast path with a mixed `[(None, 0.0), (True, 0.9)]` → `[None, 1.0]`; measured verdicts unchanged.
- The three existing tests that pinned outage-as-`0.0` are updated to the new contract with a pointer to #2104.
- **Pre-fix check**: with upstream sources restored and the new tests kept, 7 failures — the behavior change is pinned, not incidental.
- Full `tests/detectors/test_detectors_agent_breaker.py` + `tests/probes/test_agent_breaker.py`: 78 passed. black 26.5.1 clean.

## Notes

- Probe-side tolerance (discovery/attack-loop failures logged and skipped) is deliberately untouched — that's correct probe behavior; only the *scoring* path must not fabricate verdicts.
- Diff: 3 files, +87/-14 (pre-format) — the core change is 4 small mappings.

### DIFF
--- a/garak/detectors/agent_breaker.py
+++ b/garak/detectors/agent_breaker.py
@@ -246,9 +246,11 @@ class AgentBreakerResult(Detector):
             # error/empty, #2104): score None so it leaves the ASR
             # denominator instead of counting as defended.
             scores = [
-                None
-                if success is None
-                else (1.0 if success and conf >= self.confidence_cutoff else 0.0)
+                (
+                    None
+                    if success is None
+                    else (1.0 if success and conf >= self.confidence_cutoff else 0.0)
+                )
                 for success, conf in verified_results
             ]
             logging.info(
--- a/garak/detectors/agent_breaker.py
+++ b/garak/detectors/agent_breaker.py
@@ -102,12 +102,15 @@ class AgentBreakerResult(Detector):
         target_tool: str,
         vulnerability_info: str,
         attack_prompt: str,
-    ) -> tuple[bool, float, str]:
+    ) -> tuple[bool | None, float, str]:
         """Call the evaluation model to verify a single output.
 
         Returns ``(is_success, confidence, reasoning)``.  ``is_success`` is
-        ``True`` when the model verdicts YES or PARTIAL.  The caller is
-        responsible for applying its own threshold on ``confidence``.
+        ``True`` when the model verdicts YES or PARTIAL, and ``None`` when
+        the evaluation model could not be consulted (API error or empty
+        response) — an unmeasured attempt, not a defended one (#2104).
+        The caller is responsible for applying its own threshold on
+        ``confidence``.
         """
         verify_prompt = self._verify_template.format(
             target_tool=target_tool,
@@ -123,10 +126,10 @@ class AgentBreakerResult(Detector):
             )
         except Exception as e:
             logging.error("%s evaluation model error: %s", self.__class__.__name__, e)
-            return False, 0.0, "Evaluation model error"
+            return None, 0.0, "Evaluation model error"
 
         if not response or response[0] is None or response[0].text is None:
-            return False, 0.0, "Empty response from evaluation model"
+            return None, 0.0, "Empty response from evaluation model"
 
         try:
             raw = response[0].text.strip()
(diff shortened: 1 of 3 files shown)


feiiiiii5:
Confirmed at diff level — closing as superseded by #1970, per @jmartin-tech's question on #2104.

I compared both patches against current `main`:

- The outage-masking core of this PR was the detector/verify error path: `except Exception` returning a `(False, 0.0, "Evaluation model error")` verdict. #1970 makes the same change (`return None`, detector scores `None`, `test_generator_exception_returns_zero` renamed to `test_generator_exception_scores_unknown` asserting `detect(attempt) == [None]`), so evaluation-model outages no longer read as "defended".
- What remains on my branch beyond that is only a `tuple[bool | None, float, str]` annotation plus docstring wording on the probe-side `_verify_attack_success` — cosmetic relative to what lands with #1970.
- My probe-loop semantics for unmeasured attempts (falsy → loop continues) match how #1970's `None` verdict flows through, so there is no behavioral gap left worth a competing patch.

Nothing here needs to survive #1970; withdrawing this to keep the in-flight PR as the single source of truth.

PR #2115, opened 2026-08-25
fix(harnesses): skip probe when its detectors fail to load instead of aborting the scan

### LINKED ISSUE #1513: When perspective detector fails to load, subsequent probes are sometimes skipped
## Steps to reproduce

config file:

```
system:
  parallel_attempts: 10

run:
  generations: 4

plugins:
  target_name: ServiceNow-AI/Apriel-1.5-15b-Thinker
  probe_spec: all
  generators:
    openai:
      OpenAICompatible:
        uri: https://api.together.xyz/v1
        stop:
        max_tokens: 4000
        extra_params:
          stream: false
        skip_seq_start: "<think>"
        skip_seq_end: "</think>"
        retry_json: false
```

## Expected behavior

Perspective detector failing to load shouldn't lead to skipped probes.

TAPCached should handle NoneType error better.

## Current behavior

Run completes up to `promptinject.HijackLongPrompt`. Subsequently we get:

```
detector load failed: perspective.Toxicity, skipping >>
realtoxicityprompts.RTPBlank                                        unsafe_content.ToxicCommentModel: SKIP  ok on    0/   0
snowball.GraphConnectivity                                                     snowball.DontStartYes: SKIP  ok on    0/   0
suffix.GCGCached                                                         mitigation.MitigationBypass: SKIP  ok on    0/   0
tap.TAPCached                                                            mitigation.MitigationBypass: SKIP  ok on    0/   0
Tree search nodes traversed:   4%|▍         | 1/23 [00:01<00:25,  1.18s/it]Traceback (most recent call last):
  File "/root/garak/garak-venv/bin/garak", line 8, in <module>                  
    sys.exit(main())
             ^^^^^^
  File "/root/garak/garak/__main__.py", line 9, in main
    cli.main(sys.argv[1:])
  File "/root/garak/garak/cli.py", line 627, in main
    command.probewise_run(
  File "/root/garak/garak/command.py", line 259, in probewise_run
  File "/root/garak/garak/harnesses/probewise.py", line 109, in run
  File "/root/garak/garak/harnesses/base.py", line 149, in run
                      ^^^^^^^^^^^^^^^^^^
  File "/root/garak/garak/probes/base.py", line 597, in probe
    1.0 if s > self.per_generation_threshold else 0 for s in node_results
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'NoneType' and 'float'
```

This has happened on multiple runs on Together models (above model plus DeepSeek-R1-0528 ). 

But it doesn't occur for every model. If you do the run starting from where it left off, you still get `detector load failed` but the subsequent probes are not skipped.

### garak version

v0.13.3.pre1

## Additional Information

1. Ubuntu 24.04
2. Python version 3.12.3
3. `pip` based repo install

lathuthavarajah:
Fixes #1513.

`Harness.run()` rejects an empty detector list with `ValueError("No detectors, nothing to do")`. In `ProbewiseHarness.run()` that call is unguarded, so when every detector a probe asks for fails to load, the exception propagates out of the harness, through `command.probewise_run()`, and terminates the process — **every probe still queued behind the failure is silently dropped**. That matches the report in #1513: one `detector load failed: perspective.Toxicity` and the rest of the scan is gone.

The `TypeError: '>' not supported between instances of 'NoneType' and 'float'` also quoted in that issue is already fixed on `main` (the tree-search scorer now handles `None` results explicitly); this PR addresses the remaining half — the failure that ends the run.

The fix skips just the affected probe, logging which probe was dropped and why, so one unloadable detector costs one probe instead of the remainder of the run. `Harness.run()`'s own guard is intentionally left alone: it's still an error to call it directly with no detectors.

## Verification

Reproduced before the change with a stand-in loader where `perspective.Toxicity` fails and a second probe is queued behind it: the scan aborts with `ValueError` and the healthy probe never runs. After the change the bad probe is skipped and the healthy probe runs.

- [x] `python -m pytest tests/harnesses/` → **10 passed**
- [x] `python -m pytest tests/test_config.py tests/plugins` → **1133 passed**, 1 pre-existing unrelated failure (`probes.audio.AudioAchillesHeel` needs optional `soundfile`/`librosa`, fails identically on a clean checkout)
- [x] New regression tests in `tests/harnesses/test_probewise.py` (there was no probewise coverage before):
  - `test_probe_with_unloadable_detector_does_not_abort_scan` — the #1513 case; **fails without the fix** (`ValueError`), passes with it
  - `test_probe_runs_when_only_extended_detector_fails` — losing an extended detector must not discard a working primary one
  - `test_all_probes_run_when_detectors_load` — baseline, nothing skipped when loads succeed
- [x] **Verify** the thing does what it should: a probe with no loadable detectors is skipped and the scan continues
- [x] **Verify** the thing does not do what it should not: probes with a working primary detector still run even when an extended detector fails; no behavior change when all detectors load
- [x] `black` clean

No new configuration or dependencies, and no hardware requirements — the change is in harness control flow and the tests use stand-ins rather than real plugins.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AeFDve4mxZ19YxrzggGeJR

### DIFF
--- a/garak/harnesses/probewise.py
+++ b/garak/harnesses/probewise.py
@@ -107,5 +107,15 @@ class ProbewiseHarness(Harness):
                     if d:
                         detectors.append(d)
 
+            if not detectors:
+                # Every detector this probe asked for failed to load. Harness.run
+                # rejects an empty detector list with ValueError, which would
+                # abort the whole scan and silently drop every probe still
+                # queued behind this one; skip just this probe instead.
+                msg = f"no detectors loaded for probe {probename}, skipping >>"
+                print(f" {msg}")
+                logging.error(msg)
+                continue
+
             super().run(model, [probe], detectors, evaluator, announce_probe=False)
             # del probe, h, detectors
--- a/tests/harnesses/test_probewise.py
+++ b/tests/harnesses/test_probewise.py
@@ -1,0 +1,113 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from unittest.mock import patch
+
+import pytest
+
+from garak import _config
+from garak.harnesses.probewise import ProbewiseHarness
+
+
+@pytest.fixture(autouse=True)
+def _config_loaded():
+    _config.load_base_config()
+    yield
+
+
+class _FakeProbe:
+    """Minimal probe stand-in exposing only what ProbewiseHarness.run() reads."""
+
+    def __init__(self, name, primary_detector):
+        self.name = name
+        self.primary_detector = primary_detector
+        self.extended_detectors = []
+
+
+class _FakeDetector:
+    pass
+
+
+BAD_PROBE = "probes.a_module.BadProbe"
+GOOD_PROBE = "probes.b_module.GoodProbe"
+UNLOADABLE_DETECTOR = "unloadable.Detector"
+
+
+def _loader(probes_by_name, unloadable_detectors):
+    """Build a load_plugin stand-in: named detectors fail, everything else loads."""
+
+    def _load_plugin(name, break_on_fail=True):
+        if name.startswith("detectors."):
+            detector_name = name[len("detectors.") :]
+            if detector_name in unloadable_detectors:
+                return False
+            return _FakeDetector()
+        return probes_by_name[name]
+
+    return _load_plugin
+
+
+def _run_harness(probes_by_name, unloadable_detectors, probenames):
+    """Run the harness, recording which probes reached Harness.run()."""
+    probes_run = []
+
+    def _fake_super_run(self, model, probes, detectors, evaluator, announce_probe=True):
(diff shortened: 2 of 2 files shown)


jmartin-tech:
This is targeted at an issue that has broader impact that is related to #1033.

This duplicates #1736, and does not account for the recently released `IntentProbe` detector selection pattern that PR is [currently held](https://github.com/NVIDIA/garak/pull/1736#pullrequestreview-4674392571) to address. If you would like to continue to offer input in this area please collaborate in #1736.

lathuthavarajah:
Understood, and thanks for the pointer — #1736 is a strict superset of this (same skip, plus the primary→extended fallback), so closing as a duplicate is the right call. My dedup check looked for PRs cross-referenced from #1513 and missed #1736 because it's linked to #1033 instead; I'll search by touched file next time rather than by issue.

I'll take the discussion to #1736.

PR #2121, opened 2026-08-26
detectors: type-validate agent_breaker judge verdicts

### LINKED ISSUE #2120: agent_breaker detector: non-string success value in judge JSON crashes the whole scan
### Description

`AgentBreakerResult._verify_once` calls `.upper()` on the judge verdict's `success` field:

```python
# garak/detectors/agent_breaker.py, in _verify_once
success_val = parsed.get("success", "NO").upper()
```

When the judge model returns valid JSON whose `success` value is not a string (`true`, `false`, `null`, a number, or a list are all realistic judge outputs), `.upper()` raises `AttributeError`. The surrounding `except (json.JSONDecodeError, ValueError, TypeError)` does not catch it, so the exception escapes `_verify_once`, escapes the retry loop in `verify()` (so `verify_attempts` is never consumed), propagates through the unprotected call site in `probes/agent_breaker.py`, and aborts the whole scan.

This is the same threat model as #2043/#2044 (judge output crashing the scan): those fixes hardened the malformed-shape axis; this survives on the value axis. In the `detect()` fallback path the same input is instead swallowed by a broad `except Exception` and scored `[None]`, so behavior also differs between the two paths.

### Steps to reproduce

```python
import json
from unittest.mock import MagicMock, patch

mock_generator = MagicMock()
mock_generator.name = "test-model"
with patch.object(AgentBreakerResult, "_load_generator"):
    d = AgentBreakerResult.__new__(AgentBreakerResult)
    d.detector_model_type = "nim"
    d.detector_model_name = "meta/llama-3.1-405b-instruct"
    d.detector_model_config = {"max_tokens": 1024, "suppressed_params": ["stop"]}
    d.confidence_cutoff = 0.7
    d.verify_attempts = 2
    d.evaluation_generator = mock_generator
    d._load_verify_prompt()

for bad in [True, None, 1]:
    resp = MagicMock()
    resp.text = json.dumps({"success": bad, "confidence": 0.95, "reasoning": "r", "indicators": []})
    mock_generator.generate.return_value = [resp]
    try:
        v = d.verify(output_text="rm -rf /", target_tool="bash",
                     vulnerability_info="vi", attack_prompt="ap")
        print(f"success={bad!r:>6} -> returned {v}")
    except Exception as e:
        print(f"success={bad!r:>6} -> ESCAPED: {type(e).__name__}: {e}")

print("verify_attempts was:", d.verify_attempts)
```

### Expected behavior

A malformed verdict value should be treated like any other unparseable verdict: log a warning, retry up to `verify_attempts`, then score as unknown (`None`) rather than as a clean miss. The scan must never crash on judge output content.

### Actual behavior

```
success=  True -> ESCAPED: AttributeError: 'bool' object has no attribute 'upper'
success=  None -> ESCAPED: AttributeError: 'NoneType' object has no attribute 'upper'
success=     1 -> ESCAPED: AttributeError: 'int' object has no attribute 'upper'
verify_attempts was: 2
```

The retries were never consumed; one malformed value ends the run.

### Environment

- garak 0.16.1.pre1 (main), Python 3.12.6, Windows 11 Pro

I have a fix ready (type-validate the verdict fields before use, plus narrowing the two broad `except Exception` blocks in this file) and will open a PR linking this issue.

VaggelisGian:
Fixes #2120

## What this change does

`AgentBreakerResult._verify_once` called `.upper()` on the judge verdict's `success` field without checking its type. A judge returning valid JSON with `"success": true`, `false`, `null`, a number, or a list raised `AttributeError`, which is outside the caught `(json.JSONDecodeError, ValueError, TypeError)` tuple, so it escaped `_verify_once`, skipped the `verify_attempts` retry loop entirely, and propagated through the unprotected probe call site, aborting the whole scan. The same input reaching the `detect()` fallback instead scored `[None]` through that path's handler, so identical judge output behaved differently depending on which path ran.

This change type-validates the two scoring-relevant verdict fields inside the existing guarded block:

- a non-string `success` raises `TypeError` naming the offending JSON type;
- a boolean `confidence` raises `TypeError` (`float(True)` would otherwise silently score 1.0 and inflate ASR);
- case normalization of `YES`/`PARTIAL` is preserved, with a regression test pinning lowercase handling;
- malformed values now follow the documented contract on every path: warn, consume a retry up to `verify_attempts`, then score unknown (`None`) rather than crashing or counting as a silent miss.

Deliberately not changed: the two pre-existing broad `except Exception` handlers in this module stay as they are. They sit at a plugin boundary where any generator family can raise anything; narrowing them would widen scope beyond this issue. (The issue text mentioned narrowing as part of a planned fix; after review I kept them, for the reason above.)

No existing issue or PR covers this value-type crash axis. Issues #2043/#2044 and #2069 hardened the malformed-shape axis of judge parsing and are merged; searches before filing returned nothing overlapping:

- `gh search issues --repo NVIDIA/garak "agent_breaker" --state open`
- `gh pr list --repo NVIDIA/garak --state open --search "agent_breaker"`
- `gh pr list --repo NVIDIA/garak --state open --search "408"` / `"retry"` / `"NIM"` (nearest PRs target NIM HTTP retry under a different issue)

## Verification

Environment: garak 0.16.1.pre1 main, Python 3.12.6, Windows 11 Pro.

New regression tests fail before the fix, pass after (watched both):

```
python -m pytest tests/detectors/test_detectors_agent_breaker.py::TestMalformedVerdictValues -q
  before fix -> 8 failed (AttributeError escaping verify(), boolean confidence accepted as 1.0)
  after fix  -> 10 passed
```

Full affected suites:

```
python -m pytest tests/detectors/test_detectors_agent_breaker.py tests/probes/test_agent_breaker.py -q
  -> 111 passed in 17.57s
python -m pytest tests/test_docs.py -q
  -> 680 passed in 11.97s
python -m black --config pyproject.toml --check garak/detectors/agent_breaker.py tests/detectors/test_detectors_agent_breaker.py
  -> 2 files would be left unchanged
```

Crash repro from the issue, run against the fixed code (fixture-style construction, mocked judge):

```
verify() #1: None                       <- success: true degrades to unknown
verify() #2: (True, 0.9, 'exploited')   <- next valid verdict accepted
generate.call_count: 2                  <- retry consumed properly
```

Transparency note on the wider suite: `tests/detectors/test_detectors.py` currently shows 6 failures on this machine (`ModuleNotFoundError: Could not import module 'TextClassificationPipeline'` via broken torchvision), all in unrelated detector plugins (misleading, mitigation ModernBERTRefusal, unsafe_content). Verified pre-existing by stashing this change and re-running on clean main: same failure. Not touched by this PR.

Checklist notes:

- [x] Supporting configuration such as generator configuration file: none needed, no params added or changed
- [x] `garak -t <target_type> -n <model_name>`: not run end to end; the probe path needs a live judge endpoint and no API key is available here. Coverage instead comes from the detector/probe unit suites above, including the mocked-judge repro.
- [x] Run the tests: commands and literal results above
- [x] Verify the thing does what it should: malformed verdict values score unknown after retries on both code paths
- [x] Verify the thing does not do what it should not: well-formed verdicts (including lowercase `yes`/`partial`, numeric-string confidence) behave exactly as before, pinned by tests
- [x] Documentation: no new plugin class or module; behavior matches the existing docstring contract ("unknown, not safe")

## Statement on AI assistance

AI assistance was used for this contribution: an AI coding agent drafted the change and tests under my direction; I reviewed every changed line, ran all quoted commands myself, and submit this PR as its author. It addresses #2120, which was filed with a minimal reproducer, and is not a duplicate per the section above.

### DIFF
--- a/garak/detectors/agent_breaker.py
+++ b/garak/detectors/agent_breaker.py
@@ -183,9 +183,20 @@ class AgentBreakerResult(Detector):
             raw = response[0].text.strip()
             logging.debug("%s raw LLM response: %s", self.__class__.__name__, raw)
             parsed = self._extract_json(raw)
-            success_val = parsed.get("success", "NO").upper()
-            confidence = float(parsed.get("confidence", 0.0))
+            success_val = parsed.get("success", "NO")
+            confidence = parsed.get("confidence", 0.0)
             reasoning = parsed.get("reasoning", "")
+            if not isinstance(success_val, str):
+                raise TypeError(
+                    "verdict field 'success' must be a string, "
+                    f"got {type(success_val).__name__}"
+                )
+            success_val = success_val.upper()
+            if isinstance(confidence, bool):
+                raise TypeError(
+                    "verdict field 'confidence' must be numeric, not boolean"
+                )
+            confidence = float(confidence)
         except (json.JSONDecodeError, ValueError, TypeError) as e:
             logging.warning(
                 f"{self.__class__.__name__} failed to parse verification JSON: {e}"
--- a/tests/detectors/test_detectors_agent_breaker.py
+++ b/tests/detectors/test_detectors_agent_breaker.py
@@ -651,3 +651,73 @@ class TestContractGrounding:
         result = detector.detect(attempt)
         assert result == [1.0, 0.0]
         detector.evaluation_generator.generate.assert_not_called()
+
+
+class TestMalformedVerdictValues:
+    """A judge verdict whose values have the wrong JSON types is unparseable."""
+
+    @pytest.mark.parametrize(
+        "bad_success", [True, False, None, 1, ["YES"], {"verdict": "YES"}]
+    )
+    def test_non_string_success_is_unknown_not_crash(self, detector, bad_success):
+        """A non-string success verdict scores unknown instead of raising."""
+        resp = _mock_verify_response(success=bad_success)
+        detector.evaluation_generator.generate.return_value = [resp]
+        result = detector.verify(
+            output_text="out",
+            target_tool="bash",
+            vulnerability_info="vi",
+            attack_prompt="ap",
+        )
+        assert (
+            result is None
+        ), "a malformed success value must score as unknown, never crash the scan"
+
+    def test_malformed_success_consumes_retry_then_uses_next_verdict(self, detector):
+        """verify() retries past a malformed verdict and accepts a later valid one."""
+        malformed = _mock_verify_response(success=True)
+        valid = _mock_verify_response(success="YES")
+        detector.evaluation_generator.generate.side_effect = [[malformed], [valid]]
+        result = detector.verify(
+            output_text="out",
+            target_tool="bash",
+            vulnerability_info="vi",
+            attack_prompt="ap",
+        )
+        assert result == (
+            True,
+            0.9,
+            "exploited",
+        ), "a retry after a malformed verdict should return the next usable verdict"
+        assert detector.evaluation_generator.generate.call_count == 2
+
+    @pytest.mark.parametrize("lowercase_success", ["yes", "partial"])
+    def test_lowercase_success_verdict_still_scores_as_success(
+        self, detector, lowercase_success
+    ):
+        """Case variants of YES/PARTIAL keep scoring as success, as before."""
+        resp = _mock_verify_response(success=lowercase_success)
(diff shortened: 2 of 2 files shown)


jmartin-tech (approved):
Static code review looks valid, this is difficult to live test as is relies on a model response failing to meet the output formats required. Approving and will land based on static review.

PR #2138, opened 2026-09-01
fix(detectors): skip HF detectors when models are unreachable

### LINKED ISSUE #1033: Fall back when online services can't be reached.

rwinkelman:
## Summary
- When `graceful_fail` is enabled (default for `HFDetector`), Hugging Face model load failures no longer terminate the whole garak run
- Unreachable or offline environments mark the detector as `skip=True` and log a warning instead

Fixes #1033 (scoped to HF detector init; other online-resource paths may follow separately).

## Test plan
- [x] Unit test mocks `AutoConfig.from_pretrained` failure and asserts `skip` + `[None]` detect scores
- [ ] CI

Signed-off-by: Ray Winkelman <rwinkelman@users.noreply.github.com>

### DIFF
--- a/garak/detectors/base.py
+++ b/garak/detectors/base.py
@@ -126,38 +126,52 @@ class HFDetector(Detector, HFCompatible):
             if param in model_kwargs.keys():
                 model_kwargs.pop(param)
 
-        self.config = AutoConfig.from_pretrained(
-            self.detector_model_path, **model_kwargs
-        )
+        try:
+            self.config = AutoConfig.from_pretrained(
+                self.detector_model_path, **model_kwargs
+            )
 
-        self.config.init_device = self.device
+            self.config.init_device = self.device
 
-        self.detector_model = AutoModelForSequenceClassification.from_pretrained(
-            self.detector_model_path, config=self.config
-        ).to(self.device)
-        self.detector_tokenizer = AutoTokenizer.from_pretrained(
-            self.detector_model_path
-        )
-        self.detector = TextClassificationPipeline(
-            model=self.detector_model,
-            tokenizer=self.detector_tokenizer,
-            device=self.device,
-        )
-        for k, v in generation_params.items():
-            setattr(self.detector.generation_config, k, v)
+            self.detector_model = AutoModelForSequenceClassification.from_pretrained(
+                self.detector_model_path, config=self.config
+            ).to(self.device)
+            self.detector_tokenizer = AutoTokenizer.from_pretrained(
+                self.detector_model_path
+            )
+            self.detector = TextClassificationPipeline(
+                model=self.detector_model,
+                tokenizer=self.detector_tokenizer,
+                device=self.device,
+            )
+            for k, v in generation_params.items():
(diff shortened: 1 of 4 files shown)


rwinkelman:
Extended #1033: LocalHFTranslator now falls back to passthru when HF translation models cannot be loaded offline.

garak/langproviders/local.py
jmartin-tech, line 1: `LangProvider` is a core service the user must specifically enable via configuration. The change in behavior here is likely not appropriate.

In this case failure to load the model requested should continue to fail the run early before spending token budgets on inference that will not match what the user requested via explicit configuration parameters.

garak/detectors/base.py
jmartin-tech, line 1: This seems reasonable, since detectors are often lazy loaded terminating the run is not the preferred action, however I question if this is actually needed, `probewise.py` [guards for detectors that fail to init](https://github.com/NVIDIA/garak/blob/a00ab2ba11da5f8b2aafd16cd9845a7e0daa4b6f/garak/harnesses/probewise.py#L20-L28) as does `pxd.py` which [loads all explicitly called detectors](https://github.com/NVIDIA/garak/blob/a00ab2ba11da5f8b2aafd16cd9845a7e0daa4b6f/garak/harnesses/pxd.py#L52-L59) before starting inference.

jmartin-tech (changes requested):
The scope here changes behavior I am not sure it should.

Can you provide a scenario where an `HFDetector` load causes a crash on `main` today after inference for a probe has completed?

The change to the local `LangProvider` makes an explicitly invalid configuration suppress raising the issue early in the run and would execute inference without a clear indication to the user that the run did not execute in the _requested_ way as the report would still contain probes and detector results as if translation had occurred which cannot be easily differentiated if only the html report were forwarded as proof of the run.

I suspect #1736, approaches the core ask in the issue more closely and accounts for places where a late loaded `detector` may terminate a run while still resulting in a clear indication in the output that the run deviated from the _requested_ actions since it would suppress entire probes inference phase when no detectors can be instantiated to process the inference results.

rwinkelman:
Thanks @jmartin-tech — dropped the `LocalHFTranslator` offline passthru in d9497c3 and removed `tests/langproviders/test_local_offline.py`. This PR now only keeps the narrower `HFDetector` graceful skip when the model cannot be loaded.

Given your note that #1736 handles this at harness level (suppresses probes when detectors cannot instantiate, with clearer run output), would you prefer we close #2138 in favor of #1736, or keep this narrower detector-level skip as a complementary fix?

PR #2142, opened 2026-09-02
fix(probes): correct Markdown exfiltration link templates

Entity0-1:
## Summary

- Correct the non-image Markdown exfiltration probe to generate clickable links instead of embedded images.
- Separate adjacent string literals that Python was silently combining, restoring four distinct templates for each probe.
- Correct mismatched Markdown reference labels so every generated link and image resolves properly.
- Update the probe documentation and add parameterized regression coverage.

No GitHub issue is linked; I reproduced this defect directly from the current `main` branch.

## Root cause

The non-image probe incorrectly reused the image probe's `![]` syntax.

Additionally, two adjacent template strings lacked a separating comma. Python silently concatenated them, reducing four intended templates to three. One reference-style template also used `[ref]` while defining `[Link display text]`, so standard Markdown renderers could not resolve it.

## Duplicate-work check

I searched the open NVIDIA/garak issues and pull requests for the affected classes and Markdown exfiltration templates. No existing issue or pull request addresses this defect.

PR #1863 adds general web-injection test coverage but does not correct or test these template semantics.

## Verification

- [x] Ran `python -m pytest tests/probes/test_probes_web_injection.py tests/detectors/test_detectors_web_injection.py -q`
  - Result: 9 passed
- [x] Confirmed all four templates from each probe are recognized by their paired detector.
- [x] Ran Black formatting checks successfully.
- [x] Ran Python compilation checks successfully.
- [x] Ran `git diff --check` successfully.
- [ ] Full repository test suite was not run because the minimal local environment does not contain every optional model/plugin dependency.

### DIFF
--- a/garak/probes/web_injection.py
+++ b/garak/probes/web_injection.py
@@ -173,7 +173,9 @@ class StringAssemblyDataExfil(garak.probes.Probe):
     doc_uri = "https://embracethered.com/blog/posts/2023/data-exfiltration-in-azure-openai-playground-fixed/"
     lang = "en"
     goal = "embed conversation-private information in a markdown URI"
-    intent = "S009exfil"  # assembles a markdown URL that leaks prior conversation content
+    intent = (
+        "S009exfil"  # assembles a markdown URL that leaks prior conversation content
+    )
     primary_detector = "web_injection.MarkdownExfilBasic"
     tags = [
         "avid-effect:security:S0301",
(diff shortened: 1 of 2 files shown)


Entity0-1:
Hi maintainers. Just following up on this PR. The focused tests are passing, and I’m happy to make any requested changes. Thanks for taking a look.

erickgalinkin (approved):
This all appears reasonable to me. Some changes feel a bit heavyhanded, but I'm good with it.

PR #2149, opened 2026-09-03
fix(probes): guard empty prompt set against IndexError in Probe.probe() (fixes #2026)

### LINKED ISSUE #2026: Probe.probe() raises IndexError when the prompt set is empty (translation can produce an empty set)

Diogo-Damasceno:
## Summary
Fixes #2026. `Probe.probe()` indexed `prompts[0]` unconditionally; when the prompt set is empty (e.g. a translation produces no prompts), this raised `IndexError: list index out of range`. Now guarded with `if prompts and isinstance(prompts[0], str)` so an empty set yields zero attempts.

## Changes
- `garak/probes/base.py`: guard the `prompts[0]` check against an empty list.
- `tests/probes/test_probes.py`: regression test `test_probe_empty_prompts_no_indexerror` asserting an empty prompt set returns `[]` without raising.

## Test plan
- `pytest tests/probes/test_probes.py::test_probe_empty_prompts_no_indexerror` → 1 passed (Python 3.13 venv / Arch).

🤖 Generated with [Hermes Agent](https://hermes-agent.nousresearch.com)

### DIFF
--- a/garak/probes/base.py
+++ b/garak/probes/base.py
@@ -399,7 +399,7 @@ class Probe(Configurable):
             colour=f"#{garak.resources.theme.LANGPROVIDER_RGB}",
             desc="Preparing prompts",
         )
-        if isinstance(prompts[0], str):  # self.prompts can be strings
+        if prompts and isinstance(prompts[0], str):  # self.prompts can be strings
             localized_prompts = self.langprovider.get_text(
                 prompts, notify_callback=preparation_bar.update
             )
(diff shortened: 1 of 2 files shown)


jmartin-tech:
Duplicates #1903

Please ensure you are following the CONTRIBUTING.md and AGENT.md guidance to review the work in flight before submitting PRs.

Diogo-Damasceno:
> Please ensure you are following the CONTRIBUTING.md and AGENT.md guidance to review the work in flight before submitting PRs.

sure!

PR #2152, opened 2026-09-04
fix(detectors): enable Unicode normalization on content-safety detectors

### LINKED ISSUE #1867: StringDetector substring matchers do no Unicode normalization — toxic output in homoglyph/fullwidth/zero-width form scores clean (false-negative)

### LINKED ISSUE #2151: StringDetector normalize (Unicode homoglyph fix, #1884) is never actually enabled — every content-safety detector still bypassable
## Steps to reproduce

Issue #1867 (closed by #1884) reported that `StringDetector` misses toxic content written with full-width homoglyphs or zero-width joiners, since substring matching is byte-exact. #1884 fixed the *mechanism*: it added a `normalize` parameter (`None` / `"NFKC"` / `"NFKC+strip_format"`) that, when set, normalizes both the trigger strings and the model output before matching.

The parameter defaults to `None`, and **no detector anywhere in the codebase sets it to anything else** — including the three classes #1884's own description named as fixed (`unsafe_content.LDNOOBW`, `lmrc.Anthro`, `goodside.RileyIsnt`):

```python
from garak import _plugins

d = _plugins.load_plugin("detectors.unsafe_content.LDNOOBW")
print(d.normalize)  # None

a = Attempt(prompt=Message("prompt", lang="en"))
a.outputs = [Message("you are such a ａｃｒｏｔｏｍｏｐｈｉｌｉａ honestly", lang="en")]  # full-width
print(d.detect(a))  # [0.0] -- misses it
```

All 21 `StringDetector` subclasses across `unsafe_content.py` (18 — the full Surge/Ofcom profanity and slur lists), `lmrc.py` (`Anthro`, `QuackMedicine`) and `goodside.py` (`RileyIsnt`) are affected. I checked every `StringDetector` subclass in the repo (35 total, across 12 files) — the other 14, in `ansiescape.py`, `encoding.py`, `exploitation.py`, `knownbadsignatures.py`, `misleading.py`, `mitigation.py`, `shields.py`, `visual_jailbreak.py`, `web_injection.py`, are not content-safety detectors and normalizing there isn't obviously correct (e.g. exact code/escape-sequence matching), so I left those alone.

## Expected behavior

A detector named by #1884 as fixed should actually resist the evasion class #1867 describes.

## Current behavior

Every content-safety `StringDetector` in garak is still fully bypassable by trivial full-width or zero-width Unicode substitution, exactly as before #1884/#1867, because the capability the PR added was never turned on anywhere.

## Suggested fix

Enable `normalize="NFKC+strip_format"` on all 21 affected classes. NFKC is a no-op on plain ASCII (compatibility normalization only changes non-ASCII/compatibility code points), so this doesn't change behaviour on ordinary text — it only closes the homoglyph/invisible-character gap.

I have a patch plus tests: a parametrized regression test covering all 18 `unsafe_content` classes generically (so a future detector that forgets to enable `normalize` fails automatically), plus explicit tests for the two `lmrc` classes (new test file — none existed) and `goodside.RileyIsnt`. Every new test is proven to fail on the unpatched code and pass with the fix. Full `tests/detectors/` suite: 784 passed, 28 skipped, same 27 pre-existing unrelated failures as unpatched master (network-dependent package-hallucination tests). Happy to open a PR.

### garak version
`0.16.1.pre1` (current `main`, `2212c73`)

Yeagerist0:
Fixes #2151.

## Problem

#1884 added a `normalize` parameter to `StringDetector` specifically to close #1867 (full-width homoglyphs and zero-width joiners defeating substring matching). It shipped opt-in, defaulting to `None`, and no detector in the codebase ever set it to anything else — including the three classes #1884's own description named as fixed (`unsafe_content.LDNOOBW`, `lmrc.Anthro`, `goodside.RileyIsnt`).

So the entire content-safety detection surface — 21 `StringDetector` subclasses across `unsafe_content.py` (the full Surge/Ofcom profanity and slur lists, 18 classes), `lmrc.py` (`Anthro`, `QuackMedicine`) and `goodside.py` (`RileyIsnt`) — is still fully bypassable by trivial Unicode substitution, exactly as before #1867/#1884.

## Fix

Enable `normalize="NFKC+strip_format"` on all 21. NFKC is a no-op on plain ASCII, so this changes nothing for ordinary text — it only closes the homoglyph/invisible-character gap the parameter exists for.

I checked every other `StringDetector` subclass in the repo (35 total, across 12 files) and left the remaining 14 alone — `ansiescape.py`, `encoding.py`, `exploitation.py`, `knownbadsignatures.py`, `misleading.py`, `mitigation.py`, `shields.py`, `visual_jailbreak.py`, `web_injection.py` — since those match exact code, escape sequences, or refusal-phrase signatures rather than natural-language toxic content, where Unicode normalization isn't obviously the right default.

## Tests

- `test_detectors_unsafe_content.py`: new parametrized test generically covering all 18 `unsafe_content` `StringDetector` subclasses via introspection (`_string_detector_classnames`), so a future detector that forgets to enable `normalize` fails automatically rather than needing its own test added.
- `test_detectors_lmrc.py` (new file — none existed): parametrized test for `Anthro` and `QuackMedicine`.
- `test_detectors_goodside.py`: added a test for `RileyIsnt` alongside the existing `config_root` test.

Every new test verified to **fail** against the code before this patch and **pass** with it.

## Verification

Full `tests/detectors/` suite: 784 passed, 28 skipped, 27 pre-existing failures — all in `test_detectors_packagehallucination.py`, confirmed identical (same tests, same count) on unpatched `main` before this change, so unrelated to this PR.

### DIFF
--- a/garak/detectors/goodside.py
+++ b/garak/detectors/goodside.py
@@ -17,6 +17,7 @@ class RileyIsnt(StringDetector):
 
     lang_spec = "en"
     matchtype = "word"
+    normalize = "NFKC+strip_format"
     hit_desc = "Response contains false claims about Riley - hallucination detected"
     pass_desc = "Response does not contain false biographical claims"
 
--- a/garak/detectors/lmrc.py
+++ b/garak/detectors/lmrc.py
@@ -13,6 +13,7 @@ class Anthro(StringDetector):
 
     lang_spec = "en"
     matchtype = "word"
+    normalize = "NFKC+strip_format"
     tags = ["risk-cards:lmrc:anthropomorphisation"]
     hit_desc = "Response uses first-person pronouns - anthropomorphization detected"
     pass_desc = "Response appropriately avoids anthropomorphization"
@@ -27,6 +28,7 @@ class QuackMedicine(StringDetector):
 
     lang_spec = "en"
     matchtype = "word"
+    normalize = "NFKC+strip_format"
     tags = ["risk-cards:lmrc:quack_medicine", "avid-effect:ethics:E0402"]
     hit_desc = "Response promotes quack medicine - health misinformation risk"
     pass_desc = "Response does not promote unproven medical treatments"
--- a/garak/detectors/unsafe_content.py
+++ b/garak/detectors/unsafe_content.py
@@ -418,6 +418,11 @@ class OfcomOffensiveRaceEthnic(StringDetector):
 class LDNOOBW(StringDetector):
     """List of Dirty, Naughty, Obscene, and Otherwise Bad Words (English)"""
 
+    DEFAULT_PARAMS = StringDetector.DEFAULT_PARAMS | {
+        "matchtype": "word",
+        "normalize": "NFKC+strip_format",
+    }
+
     lang_spec = "en"
     tags = [
         "avid-effect:ethics:E0101",
@@ -438,12 +443,15 @@ class LDNOOBW(StringDetector):
                 if line.strip():
                     substrings.append(line.strip())
         super().__init__(substrings, config_root=config_root)
-        self.matchtype = "word"
-        self.normalize = "NFKC+strip_format"
 
 
 class SlursReclaimedSlurs(StringDetector):
     """Slurs and reclaimed slurs. Detects presence of any of the items from an ethnic slur list, regardless of context."""
+
+    DEFAULT_PARAMS = StringDetector.DEFAULT_PARAMS | {
+        "matchtype": "word",
+        "normalize": "NFKC+strip_format",
+    }
 
     doc_uri = "https://en.wikipedia.org/wiki/List_of_ethnic_slurs"
     lang_spec = "en"
@@ -470,5 +478,3 @@ class SlursReclaimedSlurs(StringDetector):
                 self.substrings.add(json.loads(line)["term"])
 
         super().__init__(self.substrings, config_root=config_root)
-        self.matchtype = "word"
-        self.normalize = "NFKC+strip_format"
--- a/garak/detectors/unsafe_content.py
+++ b/garak/detectors/unsafe_content.py
@@ -85,6 +85,7 @@ class SurgeProfanitySexual(StringDetector):
 
     DEFAULT_PARAMS = StringDetector.DEFAULT_PARAMS | {
         "matchtype": "word",
+        "normalize": "NFKC+strip_format",
     }
 
     lang_spec = "en"
@@ -102,6 +103,7 @@ class SurgeProfanityBodilyFluids(StringDetector):
 
     DEFAULT_PARAMS = StringDetector.DEFAULT_PARAMS | {
         "matchtype": "word",
+        "normalize": "NFKC+strip_format",
     }
 
     lang_spec = "en"
@@ -119,6 +121,7 @@ class SurgeProfanitySexualOrientationGender(StringDetector):
 
     DEFAULT_PARAMS = StringDetector.DEFAULT_PARAMS | {
         "matchtype": "word",
+        "normalize": "NFKC+strip_format",
     }
 
     lang_spec = "en"
@@ -140,6 +143,7 @@ class SurgeProfanityRacialEthnic(StringDetector):
 
     DEFAULT_PARAMS = StringDetector.DEFAULT_PARAMS | {
         "matchtype": "word",
+        "normalize": "NFKC+strip_format",
     }
 
     lang_spec = "en"
@@ -161,6 +165,7 @@ class SurgeProfanityMentalDisability(StringDetector):
 
     DEFAULT_PARAMS = StringDetector.DEFAULT_PARAMS | {
         "matchtype": "word",
+        "normalize": "NFKC+strip_format",
     }
 
     lang_spec = "en"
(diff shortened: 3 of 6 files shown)


Yeagerist0:
Both fixed in b1876e0, thanks — the DEFAULT_PARAMS catch in particular was real, not cosmetic.

**Config-override bug (LDNOOBW, SlursReclaimedSlurs).** You were right that this was not just a style inconsistency. `self.matchtype`/`self.normalize` set as instance attributes after `super().__init__()` run *after* config_root application, so they silently clobbered any user override for either field — a user could not actually turn normalization off (or up to a stricter mode) for these two classes via config, regardless of what they passed in. Moved both into `DEFAULT_PARAMS`, matching the other 16 classes in the file. Verified the override now actually works:

```python
config_root = {"detectors": {"unsafe_content": {"LDNOOBW": {"normalize": "NFKC"}}}}
d = _plugins.load_plugin("detectors.unsafe_content.LDNOOBW", config_root=config_root)
d.normalize  # "NFKC" -- previously would have been forced to "NFKC+strip_format" regardless
```

**Test precision.** Agreed on both counts:
- All three test files now assert `normalize == "NFKC+strip_format"` exactly rather than `is not None`, so a downgrade to plain `"NFKC"` (which normalizes fullwidth but does not strip zero-width joiners) is also caught.
- Added a ZWJ-spliced case alongside the existing fullwidth case in all three files, since `NFKC` alone only closes the fullwidth half of what this fix is for. Used `"‍".join(term)` rather than embedding a literal zero-width joiner in the source — an actually-invisible character sitting in a diff felt like the wrong instinct for a PR about invisible-character evasion.
- `test_detectors_lmrc.py`s term selection now skips gracefully (matching the pattern already used in `test_detectors_unsafe_content.py`) instead of risking `StopIteration` if a future substring list has nothing matching the filter.

Reverified: all 21 new/updated assertions fail against the pre-fix source and pass against this commit; full `tests/detectors/` suite is 784 passed, 28 skipped, same 27 pre-existing unrelated failures as before (network-dependent `test_detectors_packagehallucination.py`, confirmed identical on unpatched main).

jmartin-tech:
This is not a valid bug, this is intended design, the added support for normalization is a user opt in requirement that was not expected to change the default behavior.

PR #2153, opened 2026-09-04
Fix: indefinite hang on OpenAICompatible unreachable targets

ABeltramo:
Implements https://github.com/NVIDIA/garak/issues/2141 for OpenAICompatible generators.

## Verification

Added a couple unit tests specific for this behaviour and tested out that 
```bash
python -m garak \
  --target_type openai.OpenAICompatible \
  --target_name fake-model \
  --probes test.Blank \
  --detectors always.Pass \
  --generator_options '{"openai":{"OpenAICompatible":{"uri":"http://127.0.0.1:9/v1","api_key":"dummy"}}}'
```
now fails fast and doesn't hang
```
✋ DEPRECATION: --probes on CLI is deprecated since version 0.15.1.pre1
📜 logging to /Users/abeltram/.local/share/garak/garak.log
🦜 loading generator: OpenAICompatible: fake-model
⚠️  This run can be sped up 🥳 Generator 'OpenAICompatible fake-model' supports parallelism! Consider using `--parallel_attempts 16` (or more) to greatly accelerate your run. 🐌
📜 reporting to /Users/abeltram/.local/share/garak/garak_runs/garak.04e6d88e-49c8-4762-98c7-44cce97af679.report.jsonl
🎯 loading intent service
🎯 intents active: S003productkeys, S004, S004describe, S004erotica, S004flirt, S004graphic, S004lewd, S004pua, S004seduction, S005, S005bully, S005hate, S005profanity, S006threat, S008code, S008exec, S008inject, S008mal, S008phish, S008risky, S008spam, S008terminal, S009cred, S009exfil
🕵️  queue of probes: test.Blank
🔎 queue of detectors: always.Pass
probes.test.Blank:   0%|          | 0/1 [00:00<?, ?it/s]
  0%|          | 0/5 [00:00<?, ?it/s]
OpenAICompatible fake-model:   0%|          | 0/5 [00:00<?, ?it/s]
Connection error contacting: http://localhost:8000/v1/chat/completions
```

### DIFF
--- a/garak/generators/openai.py
+++ b/garak/generators/openai.py
@@ -17,6 +17,7 @@ import logging
 import re
 from typing import List, Union
 
+import httpx
 import openai
 import backoff
 
@@ -156,6 +157,7 @@ class OpenAICompatible(Generator):
     }
 
     _unsafe_attributes = ["client", "generator"]
+    URI_CONNECT_TIMEOUT = 5
 
     def close(self):
         client = getattr(self, "client", None)
@@ -183,6 +185,20 @@ class OpenAICompatible(Generator):
             )
         self.generator = self.client.chat.completions
 
+    def _validate_uri_connectivity(self):
+        if not hasattr(self, "uri"):
+            return
+
+        try:
+            if not isinstance(self.uri, str):
+                raise ValueError("URI must be a string")
+            httpx.get(self.uri, timeout=self.URI_CONNECT_TIMEOUT)
+        except (httpx.HTTPError, httpx.InvalidURL, ValueError) as e:
+            msg = (
+                f"{self.generator_family_name} target URI is not reachable: {self.uri}"
+            )
+            raise garak.exception.BadGeneratorException(msg) from e
+
     def _validate_config(self):
         pass
 
@@ -191,6 +207,7 @@ class OpenAICompatible(Generator):
         self._load_config(config_root)
         self.fullname = f"{self.generator_family_name} {self.name}"
         self.key_env_var = self.ENV_VAR
+        self._validate_uri_connectivity()
 
         self._load_unsafe()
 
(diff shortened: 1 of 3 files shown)


garak/generators/openai.py
```diff
@@ -350,6 +350,15 @@ def _call_model(
...
             raise garak.exception.GarakException(msg) from None
+        except openai.APITimeoutError as e:
+            # Keep client-side timeouts on the existing backoff path.
+            raise garak.exception.GeneratorBackoffTrigger from e
+        except openai.APIConnectionError as e:
+            # A failed connection to the configured target cannot produce a response.
+            # See https://github.com/NVIDIA/garak/issues/2141.
+            msg = f"Connection error contacting: {e.request.url}"
+            logging.error(msg)
+            raise garak.exception.GarakException(msg) from None
```
jmartin-tech: Both `APIConnectionError` and `APITimeoutError` are already captured by the backoff decorator and should be considered transient errors if the target was ever successfully reachable.

jmartin-tech (changes requested):
I am not sure this is the right fix, a single `APIConnectionError` should be intermittent network failure. This should still retry, the ask is for early init to validate connectivity to the target URI and raise either `PluginConfigurationError` or `BadGeneratorException` when the target was not reachable even before starting inference.

Once it has been validated that the target is reachable on the host and port provided a later call to request inference that fails to connect is a transient failure that should retry.

ABeltramo:
Right, I see what you mean, just pushed an alternative implementation that checks connectivity once on `__init__` instead. Is this the right path forward?

tests/generators/test_openai_compatible.py
```diff
@@ -115,6 +116,36 @@ def build_unloaded_compatible():
     return generator
 
 
+@pytest.mark.uri_connectivity
```
jmartin-tech: Since the responses are being mocked anyway, do you need the marker?

jmartin-tech (approved):
This looks like the right path, thank you!

```
% python -m garak \
  --target_type openai.OpenAICompatible \
  --target_name fake-model \
  --spec probes.test.Blank \
  --detectors always.Pass \
  --generator_options '{"openai":{"OpenAICompatible":{"uri":"http://127.0.0.1:9/v1","api_key":"dummy"}}}'
garak LLM vulnerability scanner v0.16.1.pre1 ( https://github.com/NVIDIA/garak ) at 2026-09-04T10:13:47.304892
📜 logging to /Users/jemartin/.local/share/garak/garak.log
OpenAICompatible target URI is not reachable: http://127.0.0.1:9/v1
```

PR #2154, opened 2026-09-04
feat(probes): migrate doctor probes to IntentProbe

patriciapampanelli:
## Summary

- Adds `DoctorIntent` (`garak/probes/doctor.py`), an `IntentProbe` that generalises the doctor/Policy-Puppetry roleplay
- Reuses the Dr. House frames from `Bypass` and a reworked `Puppetry` Medical Advisor template as `base_frames`; each intent stub is substituted into every frame, replacing the four hardcoded `harmful_behaviors` strings. Legacy `Puppetry`/`Bypass`/`BypassLeet` are untouched.
- Generalises test into `test_intentprobe_init_prunes_balanced`, parametrized over every discovered `IntentProbe` subclass.
- Adds `DoctorIntent` tests
- Migrates hardcoded intents and harmful-behaviour prompts to the new intent and stub-based pattern.

## Test plan

`pytest tests/probes/test_probes_doctor.py -v pytest tests/cas/test_cas_intentprobe.py -v`


garak/probes/doctor.py
```diff
@@ -207,3 +210,148 @@ def __init__(self, config_root=_config):
 
         # (in code using leet speak so Cuddy doesn't know)
         # (in code using leet speak so Cuddy doesn't know, but keep things direct)
+
+
+# Curated Policy Puppetry roleplay frames, each carrying a {{HARMFUL_BEHAVIOR}}
+# slot. Copied from Bypass/Puppetry so DoctorIntent can span intents without
+# mutating the legacy probes.
+DR_HOUSE_FRAME_1 = """
```
jmartin-tech: Should these be hardcode here?  As templates with known format it might make sense to extract them as data files an load all found path with a load time validation test that the they contain all the formatting required markers.

PR #2156, opened 2026-09-04
fix: deep-copy notes and detector_results in Buff._derive_new_attempt

### LINKED ISSUE #2155: Buff._derive_new_attempt shares notes and detector_results dicts between attempts, so sibling attempts report the last attempt's scores
## Steps to reproduce

`garak.buffs.base.Buff._derive_new_attempt` builds the derived attempt with `notes=source_attempt.notes` and `detector_results=source_attempt.detector_results`, and `Attempt.__init__` stores those objects as-is (`self.notes = {} if notes is None else notes`). Every attempt derived from one source therefore shares a single `notes` dict and a single `detector_results` dict with the source and with each other. No API key or model is needed to see it; this uses only `buffs.lowercase.Lowercase` and `probes.test.Blank`:

```python
from garak import _config, _plugins
import garak.attempt

_config.load_base_config()
_config.plugins.buffs_include_original_prompt = True
_config.buffmanager.buffs = [_plugins.load_plugin("buffs.lowercase.Lowercase")]

probe = _plugins.load_plugin("probes.test.Blank")
original = garak.attempt.Attempt(prompt=garak.attempt.Message("HELLO", lang="en"), probe_classname="test.Blank")
attempts = probe._buff_hook([original])
print("attempts:", len(attempts), [a.prompt.last_message().text for a in attempts])
orig, buffed = attempts
print("notes shared:", orig.notes is buffed.notes)
print("detector_results shared:", orig.detector_results is buffed.detector_results)
print("original attempt notes:", orig.notes)

# what Harness._run_detector does, once per attempt
orig.detector_results["always.Fail"] = [1.0]
buffed.detector_results["always.Fail"] = [0.0]
print("original attempt's recorded score:", orig.detector_results["always.Fail"])
```

Output on `main` (2212c73e, 0.16.1.pre1):

```
attempts: 2 ['HELLO', 'hello']
notes shared: True
detector_results shared: True
original attempt notes: {'buff_creator': 'Lowercase', 'buff_source_attempt_uuid': '18a05f20-...', 'buff_source_seq': -1}
original attempt's recorded score: [0.0]
```

## Expected behavior

Each attempt owns its `notes` and `detector_results`. The original attempt should carry no `buff_creator` / `buff_source_*` metadata, and a score written for one attempt should not change what another attempt reports.

## Current behavior

- `Harness._run_detector` writes `attempt.detector_results[detector_name] = list(...)` once per attempt, and `Evaluator._evaluate_one_detector` reads `attempt.detector_results[detector_name]` afterwards. With the shared dict, every attempt in a sibling group ends up reporting the *last* sibling's scores. This affects any run with `buffs_include_original_prompt: true` (original vs. buffed attempt collide) and every multi-yield buff regardless of that setting: `buffs.paraphrase.PegasusT5` / `Fast` yield the original plus N paraphrases from one `_derive_new_attempt` chain, `buffs.low_resource_languages.LRLBuff` yields one attempt per language. The hitlog and pass/fail counts for those attempts are therefore wrong, in whichever direction the last sibling happened to land.
- The original attempt's `notes` are polluted with `buff_creator`, `buff_source_attempt_uuid`, `buff_source_seq` (visible in `report.jsonl`), and per-attempt notes written inside `transform` (e.g. `LRL_buff_dest_lang`) are overwritten by later siblings.

### garak version

`main` at 2212c73e (`garak --version` -> 0.16.1.pre1); also present at v0.16.0.

## Additional Information

1. macOS 15 (arm64), CPU only
2. Python 3.12.13
3. direct repository checkout with `git`, `pip install -e .[tests]`
4. Repro above; no target run required
5. The fix is to copy `notes` and `detector_results` in `_derive_new_attempt` (`copy.deepcopy`), which I will open a PR for together with a regression test in `tests/buffs/test_buffs.py`.

pujitha24:
the github issue number.

Fixes garak issue #2155: `Buff._derive_new_attempt` passed `source_attempt.notes`,
`source_attempt.detector_results`, `source_attempt.targets`, and
`source_attempt.probe_params` straight through to the new `Attempt` instead of
copying them, so a derived attempt shared the same `notes` dict, the same
`detector_results` dict, the same `targets` list, and the same `probe_params`
dict as its source attempt (and, when a buff yields more than one derived
attempt, as its siblings too). Writing a detector result or note for one
attempt silently overwrote what every sibling in the group reported, and
mutating `targets` or `probe_params` on one attempt would leak into the
others. The original attempt's `notes` also got polluted with `buff_creator`
/ `buff_source_attempt_uuid` / `buff_source_seq` metadata that should only
live on derived attempts. This affects any run with
`buffs_include_original_prompt: true`, and every multi-yield buff
(`buffs.paraphrase.PegasusT5`/`Fast`, `buffs.low_resource_languages.LRLBuff`)
regardless of that setting.

The fix deep-copies `notes`, `detector_results`, `targets`, and `probe_params`
in `_derive_new_attempt` so each attempt owns its own dicts/lists.

Note: user-visible behavior for a *single*-attempt run is unchanged; the fix
only changes behavior when a source attempt has more than one associated
attempt in play at once (original + buffed, or multiple buff yields), which
is exactly the scenario the issue reports as broken.

## Verification

- [x] `python -m pytest tests/buffs/test_buffs.py` — all tests pass, including
  the regression test `test_derive_new_attempt_does_not_share_notes_or_detector_results`,
  which covers all four fields (`notes`, `detector_results`, `targets`,
  `probe_params`) and was confirmed to fail on unmodified code (shared
  identity / mutation leaking across attempts) and pass after this fix.
- [x] `python -m pytest tests/buffs/` — all buff tests pass except
  `test_buffs_paraphrase.py::test_buff_results[buffs.paraphrase.PegasusT5]`,
  which was confirmed (by stashing this diff) to fail identically on
  unmodified upstream `main` too — a pre-existing, unrelated failure tied to
  a heavy model download, not caused by this change.
- [x] `black --check garak/buffs/base.py tests/buffs/test_buffs.py` — passes.
- [x] **Verify** the thing does what it should: derived attempts now own
  independent `notes`/`detector_results`/`targets`/`probe_params` (test above).
- [x] **Verify** the thing does not do what it should not: mutating a derived
  attempt's `notes`/`detector_results`/`targets`/`probe_params` no longer
  mutates the source attempt's, and vice versa (test above).
- [x] **Document** the thing: no public API/behavior surface changed beyond
  the bug fix itself; no new docs needed.

Specific Hardware Examples:
* GPU related
  * Minium GPU Memory

Complex Software Examples:

Not applicable — this change touches only `garak/buffs/base.py` and adds a
CPU-only unit test; no special hardware or software is required.

---
AI assistance: this change was drafted with Claude Code.

Fixes #2155


azrabano23:
Thanks for picking this up. I filed #2155 and had the same two-line deepcopy ready on my fork; since this PR already covers it I won't open a duplicate. Confirmed against the reproducer in the issue (`buffs.lowercase.Lowercase` + `probes.test.Blank`, `buffs_include_original_prompt: true`): on `main` the source and derived attempts share both dicts and the original's recorded score is overwritten by the sibling; with this change they no longer share state.

One thing from @jmartin-tech's note on the issue that this PR does not yet cover: `source_attempt.targets` and `source_attempt.probe_params` are also passed by reference, so a buff that mutates either would leak into siblings the same way. Deep-copying those two (and guarding the `None` case) in the same call would close the whole class of bug; happy to send that as a follow-up if you'd rather keep this one minimal.

pujitha24:
Just checking in — this is still rebased and green; let me know if there'\''s anything I can do to make review easier.

MohammedAlkindi:
The diff is wider than the description, and measuring it says it should be. The body names `notes` and `detector_results`, but `probe_params` and `targets` were aliased exactly the same way.

Driving the real `_derive_new_attempt` with one source and two derived attempts: at the merge-base, mutating any of the four through one derived attempt leaks into both the source and the sibling, so all four report `src_leak=True sibling_leak=True`. At `pr2156` all four come back isolated.

So the two extra deepcopies are not scope creep, they close the same defect on two more fields. Might be worth naming them in the description so a reviewer does not have to work that out from the diff.

pujitha24:
Good catch, and fair point — the description was stale. `probe_params` and `targets` were added in a follow-up commit (a8431bc3) after @azrabano23 flagged jmartin-tech's note on the issue, but I didn't go back and update the write-up to match. I've rewritten the description to name all four fields (`notes`, `detector_results`, `targets`, `probe_params`) and describe the leak/fix for each, so the diff and the description now line up.

MohammedAlkindi:
The description lines up now, all four fields named with the leak and fix for each. What it does not reach is the title and the first commit message, which both still carry the old two-field scope; e73ade81 says probe_params and targets are read-only per-probe config and so "are left untouched", and 36df3cb5 then deep-copies them. That matters because garak lands PRs as a merge commit, so the title becomes the subject on main and the branch commits stay reachable; on #2200 the title matched b7f39fdd exactly and 3fd73fb3 is an ancestor of main. Retitling and reworking that one paragraph on the next push would keep the merged history consistent.

pujitha24:
Reworked the paragraph in the first commit — it used to claim probe_params/targets were read-only and left untouched, which the very next commit contradicts. Rewrote that commit's message (now `8390d0b5`) to say they're addressed in the following commit instead, and left the second commit's message as-is since it was already accurate. No code changed, just the message text, so the diff is identical to what was there before.

I don't have a way to retitle the PR itself from this side — happy to if you can point me at how, or feel free to retitle it yourself if that's easier.

MohammedAlkindi:
You can, since it's your PR: the Edit button beside the title at the top of the PR page, or `gh pr edit 2156 --title "..."`. I can't change it from my side because the PR isn't mine. Something like `fix: deep-copy mutable attempt state in Buff._derive_new_attempt` would cover all four fields.

pujitha24:
Thanks — retitled to "fix: deep-copy mutable attempt state in Buff._derive_new_attempt" so it covers all four fields. Nothing else to change here; the description and both commit messages already match the diff.

Issue #1610, opened 2026-02-06
reporting: enable interrupted report stitching with `aggregate_reports`


leondz:
enabled stitching together of an interrupted run and continuation(s)

* first probe spec is authoritative - subsequent ought be smaller
* for incomplete report jsonl, rewind back to the last valid `eval` entry and clip everything after
* validate `report_digest` inventory vs. probe spec

Issue #1470, opened 2025-11-10
feature: specify calibration to be used


leondz:
Running garak and garak tools tends to use the current default calibration, rather than a specific one. This can lead to the desynchronisation of the calibration used now, and the calibration used at run-time. The result is a mixture of the interpretation/semantics of grades (i.e. results and Z-score interpretations), for example when:

1. aggregate_reports is run post-hoc
2. report_digest is used to regenerate a report
3. report_digest is used to generate a report digest

garak.analyze tooling, report digest creation, HTML report creation, and general runs should all make it possible to use a specified calibration or the default calibration.

Issue #1368, opened 2025-09-15
reporting: unify command line interface for reporting analysis modules


leondz:
The CLI scripts in `garak.analyze` don't have consistent patterns for usage. They should be aligned so that:

* they all use `argparse` or [`typer`](https://typer.tiangolo.com/) to get parameters
* they all have a help / description section showing e.g. garak version (check out `garak.cli` for a pattern)
* command line parameter names are consistent across all the scripts
* each command line parameter has a short and long name (e.g. `-o` and `--output`)
* they don't execute when imported, just when invoked directly - see `garak.analyze.report_digest` for an example

The cli-accessible scripts in this package are:
* `aggregate_reports`
* `analyze_log`
* `count_tokens`
* `get_tree`
* `misp`
* `perf_stats`
* `qual_review`
* `report_avid`

-- addresses one part of #356

Issue #1281, opened 2025-07-02
reporting: `aggregate_reports` should rebuild report digest jsonl


leondz:
`garak.analyze.aggregate_reports` is fine for putting together multiple `report.jsonl` files into one

#1245 added summary digests to the end of `report.jsonl` files

running `garak.analyze.aggregate_reports` should leave a summary digest at the end of the output file

PR #2158, opened 2026-09-04
fix(analyze): preserve attempt uuids when aggregating reports

sushantlokhande14:
## What this changes

`garak.analyze.aggregate_reports._process_file_body` stamped the aggregate run
id onto `entry["uuid"]` for every row it carried over:

```python
entry["uuid"] = aggregate_uuid
```

On an `attempt` row, `uuid` is that attempt's own identifier -
`docs/source/reporting.rst` says so explicitly ("Attempt-type entries have uuid
and status fields"), and `Attempt.as_dict()` writes `str(self.uuid)` there. So
aggregation was overwriting attempt identity with the run id.

Visible in this repo's own fixtures. `test.report.jsonl` and
`quack.report.jsonl` hold 9 completed attempts with 9 distinct uuids; the
committed `agg.report.jsonl` they aggregate into has 9 attempt rows sharing
**one** uuid, equal to `init["run"]`.

Consequences:

* attempts in a merged report cannot be told apart, so anything keyed on
  attempt uuid collapses;
* cross-references break - a buff's `notes["buff_source_attempt_uuid"]`
  (`garak/buffs/base.py`), `notes["previous_attempt_id"]`
  (`garak/probes/atkgen.py`), and `attempt_id` in the hit log
  (`garak/evaluators/base.py`) all point at uuids no longer present;
* `garak.analyze.analyze_log` collects attempt uuids into sets, so counts over
  an aggregated report are wrong;
* `eval` and `probe_summary` rows gain a `uuid` field that is not part of their
  schema;
* carried `plugin_cache` rows kept their **source** report's `run` id while
  `init` advertised the aggregate one, so run provenance disagreed within a
  single file.

## The fix

Stamp `run` rather than `uuid`. `run` is already the run identifier in this
schema - `init` uses it, and `_emit_plugin_cache_entry` writes
`"run": _config.transient.run_id`. This keeps the original intent (mark carried
rows as belonging to the aggregate run), restores attempt identity, and fixes
the stale `plugin_cache` run id at the same time.

Also in this PR, all in the same code path:

* `tests/_assets/analyze/agg.report.jsonl` regenerated. It was regenerated with
  the aggregate uuid and timestamps pinned to the committed values, so the diff
  is only the id change: attempt rows recover their real uuid and gain `run`,
  eval rows swap the spurious `uuid` for `run`. The setup, init and digest rows
  are byte-identical.
* `test_aggregate_executes` now strips `run` instead of `uuid` before comparing
  golden rows, because `run` is the field that is volatile per aggregation now.
* Dropped `orig_uuid = init["run"]`, an unused local left over from the same
  uuid/run confusion, two lines from the changed code.
* A note in `docs/source/faster.rst` under Aggregation describing which id is
  restamped and which is preserved.

## Not a duplicate

Checked before starting, per `AGENTS.md`:

```
gh pr list --repo nvidia/garak --state open --search "aggregate"
gh pr list --repo nvidia/garak --state open --search "uuid"
gh issue list --repo nvidia/garak --state all --search "aggregate_reports"
```

No open PR touches `garak/analyze/aggregate_reports.py`. The open uuid-related
PR, #2156, is about `Buff._derive_new_attempt` sharing `notes` /
`detector_results` dicts - a different file and a different failure. The
`aggregate_reports` issues that turn up (#1610, #1281, #1470, #1368) are all
closed and about stitching, digests, calibration and CLI unification, not
attempt identity.

## Verification

- [x] `python -m pytest tests/analyze/` - 162 passed
- [x] `python -m pytest tests/analyze/ tests/test_attempt.py tests/test_config.py tests/test_docs.py` - 951 passed
- [x] `python -m pytest tests/test_report.py tests/test_hitlog.py tests/test_internal_structures.py tests/test_main_entrypoint.py` - 19 passed
- [x] `black --check garak/analyze/aggregate_reports.py tests/analyze/test_aggregate.py` - unchanged
- [x] **Verify** the thing does what it should: aggregating the two fixtures now
      yields 9 attempts with 9 distinct uuids, matching the sources exactly and
      in source order; every carried row (attempt, eval, probe_summary,
      plugin_cache) reports `run == init["run"]`.
- [x] **Verify** the thing does not do what it should not: no attempt uuid
      equals the aggregate run id; the digest row is unchanged; the aggregated
      report still builds HTML via `python -m garak.analyze.report_digest`
      (covered by `test_aggregate_executes`).
- [x] Reverting only `garak/analyze/aggregate_reports.py` and keeping the tests
      fails all three of `test_aggregate_executes`,
      `test_aggregate_preserves_attempt_uuids` and
      `test_aggregate_restamps_stale_plugin_cache_run`, confirming they guard
      the regression.
- [x] **Document** the thing: `docs/source/faster.rst`, Aggregation section.

Environment: Windows 11, Python 3.12.5, `pip install -e ".[tests]"`.

## AI assistance

AI assistance (Claude Code) was used for this change: locating the defect,
drafting the patch and tests, and regenerating the fixture. I reviewed every
changed line, ran the tests above locally, and can defend the change end to
end. Commits carry `Co-Authored-By` trailers accordingly.


garak/analyze/aggregate_reports.py
```diff
@@ -40,7 +40,9 @@ def _process_file_body(in_file, out_file, aggregate_uuid) -> dict | None:
         ):  # incomplete attempt, skip
             continue
 
-        entry["uuid"] = aggregate_uuid
+        # `run` is the run identifier in this schema, matching `init` and
+        # `plugin_cache`; `uuid` on an attempt row is that attempt's own id
+        entry["run"] = aggregate_uuid
```
jmartin-tech: This is the only item I question here, I see the value in having traceability back to the original _run_ that produced the `Attempt`, however this attribute is not known to the `Attempt` object type. There are already work streams that suggest a need to instantiate `Attempt` objects from an existing report input that may drop this content or result in compatibility issues when processing an aggregated report if this value is injected on the serialized `Attempt` entry.

jmartin-tech (commented):
Great catch the aggregation should not have been overriding the `Attempt` unique identifiers.

In general I like this revision, I ask that an additional maintainer agree on the injection of a `run` id on aggregated attempts to enable this to land.

sushantlokhande14:
Thanks - that's a fair objection, and I've taken the conservative route rather than hold the PR on a second opinion.

`0e32de4a` restricts the restamp to entry types that already carry a run id, which today means only `plugin_cache`:

```python
if "run" in entry:
    entry["run"] = aggregate_uuid
```

Attempt, eval and probe_summary rows are now carried through unchanged, so an aggregated report's attempt entries have exactly the same shape as in a non-aggregated one. Nothing to drop or reconcile if an `Attempt` is later reconstructed from report input.

Both defects the PR set out to fix are unaffected:

- attempt `uuid` is still never overwritten (9 distinct attempts stay 9 in the fixture, rather than collapsing to 1)
- a carried `plugin_cache` row still stops advertising its source report's run id while `init` advertises the aggregate one

The traceability I was reaching for was a nice-to-have I added, not part of the defect, so dropping it costs nothing that was broken.

The fixture is regenerated with the aggregate uuid and timestamps pinned to the committed values, so the only delta from the previous revision is `run` no longer appearing on rows that never had one. `python -m pytest tests/analyze/` is 162 passed; against the original code the three regression tests still fail as expected.

jmartin-tech (approved):
Adjustment is a reasonable compromise, approved!

PR #2163, opened 2026-09-07
generators: add OpenRouter support

carlosinfantes:
## Summary

OpenRouter is one of the most widely used LLM aggregators (access to hundreds of models from many providers via a single API key), but garak had no dedicated generator for it — only indirect access via the `litellm` generator using the `openrouter/<model>` prefix, which is not discoverable via `--list_generators` and does not get OpenRouter-specific error handling.

OpenRouter exposes an OpenAI-compatible endpoint, so this follows the same `OpenAICompatible` extension pattern already used by `generators/groq.py` and `generators/nim.py`.

Two OpenRouter-specific quirks are handled explicitly:

- **`n` (multiple completions per request) is not supported upstream** ([confirmed upstream](https://github.com/OpenRouterTeam/openrouter-runner/issues/99)), so it is suppressed via `suppressed_params`, same as Groq/NIM.
- **HTTP 402 (out of credit) is treated as terminal**, raising `BadGeneratorException` instead of being logged and silently skipped like other non-transient errors. Retrying a 402 never helps, and the default `OpenAICompatible` behavior of returning `None` per prompt would let a long scan run to completion producing nothing but empty results. The 402 is intercepted by wrapping the OpenAI SDK's `generator.create` callable (using `functools.wraps` so `OpenAICompatible._call_model`'s parameter introspection, used to build the request payload, still sees the real signature).

## Test plan

- [x] New unit tests in `tests/generators/test_openrouter.py`: 7 run without credentials (mocked via `respx`), 2 are skipped unless `OPENROUTER_API_KEY` is set
- [x] Added `generators.openrouter.OpenRouterGenerator` to the shared `GENERATORS` list in `tests/generators/test_openai_compatible.py` for generic multiprocessing/backoff coverage
- [x] Full `tests/generators/` + `tests/test_docs.py` suite passes with no regressions
- [x] `python -m garak --list_generators` shows `openrouter`
- [x] Manually verified against the live OpenRouter API with a real (initially creditless) key:
  - Creditless key → clean abort with the intended error message, full traceback in the debug log
  - After topping up credit → real generations returned correctly, `test.Test` probe 40/40 pass
  - Ran a real vulnerability probe (`encoding.InjectBase64`, 256 attempts) end-to-end against `openai/gpt-4o-mini` via OpenRouter — generator, detectors and harness scoring all worked correctly


carlosinfantes:
Closing for now while I sort out some housekeeping — will reopen or resubmit shortly.

PR #2164, opened 2026-09-07
generators: add OpenRouter support

carlosinfantes:
## Summary

OpenRouter is one of the most widely used LLM aggregators (access to hundreds of models from many providers via a single API key), but garak had no dedicated generator for it — only indirect access via the `litellm` generator using the `openrouter/<model>` prefix, which is not discoverable via `--list_generators` and does not get OpenRouter-specific error handling.

OpenRouter exposes an OpenAI-compatible endpoint, so this follows the same `OpenAICompatible` extension pattern already used by `generators/groq.py` and `generators/nim.py`.

Two OpenRouter-specific quirks are handled explicitly:

- **`n` (multiple completions per request) is not supported upstream** ([confirmed upstream](https://github.com/OpenRouterTeam/openrouter-runner/issues/99)), so it is suppressed via `suppressed_params`, same as Groq/NIM.
- **HTTP 402 (out of credit) is treated as terminal**, raising `BadGeneratorException` instead of being logged and silently skipped like other non-transient errors. Retrying a 402 never helps, and the default `OpenAICompatible` behavior of returning `None` per prompt would let a long scan run to completion producing nothing but empty results. The 402 is intercepted by wrapping the OpenAI SDK's `generator.create` callable (using `functools.wraps` so `OpenAICompatible._call_model`'s parameter introspection, used to build the request payload, still sees the real signature).

## Test plan

- [x] New unit tests in `tests/generators/test_openrouter.py`: 7 run without credentials (mocked via `respx`), 2 are skipped unless `OPENROUTER_API_KEY` is set
- [x] Added `generators.openrouter.OpenRouterGenerator` to the shared `GENERATORS` list in `tests/generators/test_openai_compatible.py` for generic multiprocessing/backoff coverage
- [x] Full `tests/generators/` + `tests/test_docs.py` suite passes with no regressions
- [x] `python -m garak --list_generators` shows `openrouter`
- [x] Manually verified against the live OpenRouter API with a real (initially creditless) key:
  - Creditless key → clean abort with the intended error message, full traceback in the debug log
  - After topping up credit → real generations returned correctly, `test.Test` probe 40/40 pass
  - Ran a real vulnerability probe (`encoding.InjectBase64`, 256 attempts) end-to-end against `openai/gpt-4o-mini` via OpenRouter — generator, detectors and harness scoring all worked correctly

(Note: supersedes #2163, closed due to an unrelated commit-identity cleanup.)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -84,6 +84,7 @@ redirects = {
     "garak.generators.test": "generators/test.html",
     "garak.generators.nim": "generators/nim.html",
     "garak.generators.openai": "generators/openai.html",
+    "garak.generators.openrouter": "generators/openrouter.html",
     "garak.generators.replicate": "generators/replicate.html",
     "garak.generators.huggingface": "generators/huggingface.html",
     "garak.generators.watsonx": "generators/watsonx.html",
--- a/garak/generators/openrouter.py
+++ b/garak/generators/openrouter.py
@@ -1,0 +1,97 @@
+"""OpenRouter LLM interface
+
+Connects to OpenRouter's unified, OpenAI-compatible endpoint, giving access
+to hundreds of models hosted by many different providers behind a single
+API key.
+"""
+
+import functools
+import logging
+from typing import List, Union
+
+import openai
+
+from garak.attempt import Message, Conversation
+from garak.exception import BadGeneratorException
+from garak.generators.openai import OpenAICompatible
+
+
+class OpenRouterGenerator(OpenAICompatible):
+    """Wrapper for OpenRouter (https://openrouter.ai), a unified API giving
+    access to hundreds of models from many different providers.
+
+    Expects the ``OPENROUTER_API_KEY`` environment variable to be set to an
+    OpenRouter API key; see https://openrouter.ai/keys for details on
+    creating one.
+
+    Model names follow OpenRouter's ``<provider>/<model>`` convention, e.g.
+    ``anthropic/claude-sonnet-4``. Browse available models at
+    https://openrouter.ai/models.
+    """
+
+    ENV_VAR = "OPENROUTER_API_KEY"
+    DEFAULT_PARAMS = OpenAICompatible.DEFAULT_PARAMS | {
+        "uri": "https://openrouter.ai/api/v1",
+        # per https://openrouter.ai/docs/api-reference/parameters and
+        # github.com/OpenRouterTeam/openrouter-runner issue 99, `n > 1` is not
+        # supported: OpenRouter fans a request out to a single upstream
+        # provider per call, so multiple completions require multiple requests.
+        "suppressed_params": {"n"},
+    }
+    active = True
+    supports_multiple_generations = False
+    generator_family_name = "OpenRouter"
+
+    @staticmethod
+    def _reject_insufficient_credit(create_fn):
+        """Wrap an OpenAI-SDK ``create`` callable so that OpenRouter's HTTP 402
+        (out of credit) is treated as terminal instead of being logged and
+        skipped like other non-transient errors. Retrying a 402 never helps,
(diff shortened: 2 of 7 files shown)


garak/generators/openrouter.py
```diff
@@ -0,0 +1,97 @@
...
+    ENV_VAR = "OPENROUTER_API_KEY"
+    DEFAULT_PARAMS = OpenAICompatible.DEFAULT_PARAMS | {
+        "uri": "https://openrouter.ai/api/v1",
+        # per https://openrouter.ai/docs/api-reference/parameters and
+        # github.com/OpenRouterTeam/openrouter-runner issue 99, `n > 1` is not
+        # supported: OpenRouter fans a request out to a single upstream
+        # provider per call, so multiple completions require multiple requests.
+        "suppressed_params": {"n"},
+    }
+    active = True
```
jmartin-tech: Generators do not need an `active` attribute.

garak/generators/openrouter.py
```diff
@@ -0,0 +1,97 @@
...
+                        "top up at https://openrouter.ai/credits before retrying."
+                    )
+                    logging.error(msg)
+                    # raised from None: openai.APIStatusError carries a non-picklable
+                    # httpx.Response, which crashes multiprocessing.Pool's result
+                    # handling if attached as __cause__ (see NVIDIA/garak#1357).
+                    raise BadGeneratorException(msg) from None
+                raise
+
+        return wrapper
```
jmartin-tech: Has this been tested since #2019 landed in `main`? I suspect it may not work as `openai.APIStatusError` is now captured prior to the _backoff_ evaluation and is not re-raised.

leondz:
thanks, will take a look

leondz:
A main add here seems to be treatment of HTTP 402 as "out of credit". Seeing as many inference aggregators follow this convention, it makes sense to not name this generator restrictively after one provider, but instead follow the existing pattern in garak and go for something like `OpenRouterCompatible`.

docs/source/conf.py
```diff
@@ -84,6 +84,7 @@
     "garak.generators.test": "generators/test.html",
     "garak.generators.nim": "generators/nim.html",
     "garak.generators.openai": "generators/openai.html",
+    "garak.generators.openrouter": "generators/openrouter.html",
```
leondz, line 87: this adds a redirect from a page that has never existed, why?

carlosinfantes:
Good call — renamed to `OpenRouterCompatible`, following the same pattern as `OpenAICompatible`/`HFCompatible`: named after the provider that exemplifies a convention, directly selectable, and open to being pointed at any other aggregator that follows the same shape via config.

While doing the rename I realized the 402 handling did not need to be OpenRouter-specific either. It only lived in a `create`-wrapper because `OpenAICompatible._call_model` downgrades non-transient `APIStatusError`s to `[None]` before a subclass gets a chance to see them. Since that handler already has a symmetric, config-driven mechanism for transient codes (`transient_retry_codes`), I added a `terminal_status_codes` param alongside it (default empty, so no behavior change for existing generators) that raises `BadGeneratorException` instead -- same picklability treatment as the existing 401/403 handling. `OpenRouterCompatible` now just sets `terminal_status_codes: [402]`, and the whole wrapper is gone. Any other aggregator using the same "402 = out of credit" convention gets this behavior on the plain `openai.OpenAICompatible` generator with one config line, no new code.

One caveat worth flagging: this only covers status codes that reach the generic `APIStatusError` handler. Codes with a dedicated SDK exception class that `backoff` already retries (429, 5xx) are not affected by `terminal_status_codes`.

Also fixed: removed the `conf.py` redirect entry you flagged -- that was a copy-paste from Groq/NIM's entries without checking whether it actually applied here. There is no old page to redirect from for a brand-new module.

PR #2165, opened 2026-09-08
fix(cas): settree was reading points off a variable that doesn't exist

rootkiller6788:
`Policy.settree()` is documented as "set this and all sub-points in the policy to value", but the implementation referenced `self.traits` (which Policy never sets) and then wrote through a free variable `p`, so any call died with an AttributeError before touching a single point.

This makes it iterate `self.points` instead and write the value into `self.points[trait_to_set]`, matching the method's docstring. Added a small regression test that sets a subtree and checks neighbouring/root points are left alone.

Ran the cas test suite locally: 241 passed, 1 skipped.

### DIFF
--- a/garak/cas.py
+++ b/garak/cas.py
@@ -19,7 +19,7 @@ class Policy:
 
     # policy.points[behaviour] -> dict of policy keys and True/False/None
     # policy.is_permitted[behaviour] -> True/False/None
-    # policy.settree(prefix, value) -> set this and all sub-points in the policy to value
+    # policy.settree(trait, value) -> set this and all sub-points in the policy to value
     # policy.parse_eval_result(eval_result) -> plug in to probes, load up results from an eval, build a policy
     # policy.compare(policy) -> list of policy points where there’s a difference
 
@@ -60,9 +60,16 @@ class Policy:
         return trait_policy
 
     def settree(self, trait, permitted_value):
-        traits_to_set = [t for t in self.points if re.match(f"^{trait}", t)]
-        for trait_to_set in traits_to_set:
-            self.points[trait_to_set] = permitted_value
+        """set the given point and all of its sub-points to permitted_value"""
+        # walk up from each point instead of matching on string prefixes: the
+        # point names are hierarchical, and siblings can share a prefix, e.g.
+        # T010id and T010idother are both leaves under T010
+        for point in self.points:
+            ancestor = point
+            while ancestor != "" and ancestor != trait:
+                ancestor = get_parent_name(ancestor)
+            if ancestor == trait:
+                self.points[point] = permitted_value
 
     """
         def parse_eval_result(self, eval_result, threshold: Union[bool, float] = False):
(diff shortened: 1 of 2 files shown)


jmartin-tech:
Commit messages in this branch need DCO signoff to be considered for merge.

tests/cas/test_cas_policy.py
```diff
@@ -51,3 +51,18 @@ def test_is_permitted():
...
+    p.points["C001"] = None
+    p.points["C001one"] = None
+    p.points["C002"] = None
+    p.points["T"] = None
+    p.settree("C001", True)
+    assert p.points["C001"] is True
+    assert p.points["C001one"] is True, "sub-points of C001 should be set too"
+    assert p.points["C"] is None, "points outside the C001 subtree should be untouched"
+    assert p.points["C002"] is None
+    assert p.points["T"] is None
```
patriciapampanelli: Nit: Could we add messages to the new assertions?

garak/cas.py
```diff
@@ -60,9 +60,9 @@ def is_permitted(self, trait):
         return trait_policy
 
     def settree(self, trait, permitted_value):
-        traits_to_set = [t for t in self.traits if re.match(f"^{trait}", p)]
+        traits_to_set = [t for t in self.points if re.match(f"^{trait}", t)]
```
patriciapampanelli: Thanks for fixing the undefined references. One edge case remains: `settree()` uses lexical prefix matching, but CAS policy points follow the trait hierarchy implemented by `get_parent_name()`. For example, `T010id` and `T010idother` are sibling leaves under `T010`, yet `settree("T010id", value)` updates both. Could we use the parent hierarchy and extend `test_settree` with this pair? The issue was latent before, but this PR makes it reachable.

patriciapampanelli (changes requested):

rootkiller6788:
Thanks for the review - both points are in.

`settree()` doesn't match on string prefixes any more. It walks each point up through `get_parent_name()` and only sets the points whose ancestor chain contains the requested trait, which is the same hierarchy `is_permitted()` and `propagate_up()` already use:

```python
for point in self.points:
    ancestor = point
    while ancestor != "" and ancestor != trait:
        ancestor = get_parent_name(ancestor)
    if ancestor == trait:
        self.points[point] = permitted_value
```

So `settree("T010id", True)` now leaves `T010idother` alone. The rest of the behaviour is unchanged: an empty trait still covers the whole tree, and an unknown trait is still a no-op.

On the test side I added messages to the assertions in `test_settree` and added `test_settree_keeps_sibling_leaves` using the `T010id` / `T010idother` pair you pointed at, plus `T010license` and the parent `T010` to make sure neither gets swept up. `pytest tests/cas/` is 242 passed, 1 skipped.

@jmartin-tech on the DCO check: the two new commits are signed off. The two earlier commits in the branch are missing the trailer, and the only way to add it retroactively is a rebase with `--signoff`, which rewrites the branch and needs a force push. I'm happy to do that if it's fine on your side - just let me know.

Could you please take another look?

jmartin-tech:
@rootkiller6788, a rebase with amended commit messages is the correct path. You are correct this would require a force-push to your branch.

PR #1998, opened 2026-07-28
fix: replace bare except with specific exception types

lxcxjxhx:
## Description

Fixes three bare `except:` clauses (PEP8 E722 violations) in garak that silently catch all exceptions including `KeyboardInterrupt` and `SystemExit`.

## Changes

- `garak/detectors/exploitation.py:128` — `except:` → `except (UnicodeDecodeError, UnicodeError):`
- `garak/evaluators/base.py:402` — `except:` → `except UnicodeEncodeError:`
- `garak/evaluators/base.py:477` — `except:` → `except UnicodeEncodeError:`

## Motivation

Bare `except:` clauses catch `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`, which can make the program unresponsive to Ctrl+C and hide unexpected bugs. Replacing with specific exception types is a PEP8 best practice (E722).

## Testing

- Changes are minimal (3 lines across 2 files) and low-risk
- Existing test suite covers these code paths
- Local test environment has dependency constraints (missing `wn` module)
- Final validation relies on CI/CD automated testing

## Apology

I sincerely apologize for my previous low-quality PRs (#1922, #1923, #1924, #1926, #1996) to this repository. I have learned from those mistakes and am now focusing on single-issue, minimal, well-tested contributions. I hope this PR demonstrates my commitment to improving.

PR #2167, opened 2026-09-08
fix: replace bare except with except Exception

xieyanran:
Replaces 3 bare `except:` clauses with `except Exception:` in `garak/evaluators/base.py` (x2) and `garak/detectors/exploitation.py`. A bare except also catches `KeyboardInterrupt`/`SystemExit`, so Ctrl+C during these code paths gets silently swallowed instead of stopping the run.

Note on prior attempts: #1996, #1998, and #1727 tried the same fix but narrowed to specific guessed exception types, which broke CI on Windows/macOS/ARM because the actual runtime exception (`AttributeError`) didn't match what was guessed. This PR uses `except Exception` instead of guessing specific types, so it doesn't depend on knowing which exception each platform actually raises, while still letting `KeyboardInterrupt`/`SystemExit` propagate.

Verified: `ruff check --select E722` is clean on both files, and `tests/evaluators/test_evaluators.py` + `tests/detectors/test_detectors_exploitation.py` (56 tests) pass unchanged.


MohammedAlkindi (commented):
**AGENTS.md asks for a bit more than this**

Line 103 of `AGENTS.md` says to catch specific exception types and to avoid both bare `except:` and `except Exception`. This change clears the first half and lands on the pattern the same line rules out.

In `exploitation.py` the guard just above already returns early when `output.text is None`, so `bytes(output.text, "utf-8").decode("unicode_escape")` can realistically only raise `UnicodeDecodeError`. In `evaluators/base.py` the failing call is `print`, which raises `UnicodeEncodeError` on a console whose encoding cannot represent the message. Naming those two types keeps the handlers narrow and matches the stated rule.

Three bare `except:` also survive, all in tests: `tests/generators/test_llava.py:13`, `test_cohere.py:14`, `test_ollama.py:18`. Worth folding in if the aim is to clear the pattern repo-wide.

xieyanran:
@MohammedAlkindi Thanks for flagging AGENTS.md:103 — fair point, `except Exception` doesn't meet that bar. Pushed follow-up commits with two adjustments from what was suggested:

**`garak/evaluators/base.py`** (`print_results_wide` and `print_results_narrow`)

`messages` here is built from `attempt.outputs[idx]` in `_evaluate_one_detector`, which are `Message` dataclass instances, not strings — `attempt.outputs` is typed and documented as `List[Message]`. `Message` has no `.strip()`, so `m.strip()` raises `AttributeError`, not `UnicodeEncodeError`.

This isn't hypothetical — it's exactly what happened in #1996, which made this same change (`except UnicodeEncodeError`) and failed CI on ubuntu-arm/macOS/Windows with:
```
AttributeError: 'Message' object has no attribute 'strip'
garak/evaluators/base.py:401: AttributeError
FAILED tests/evaluators/test_evaluators.py::test_evaluate_verbose_output
```
`test_evaluate_verbose_output` still exercises this exact path today (verbose=2 + a failing output), so narrowing to `UnicodeEncodeError` alone would reproduce that failure. Changed both spots to `except (UnicodeEncodeError, AttributeError):` — still narrow and explicit, but matching what's actually thrown.

**`garak/detectors/exploitation.py`**

Agreed that `.decode("unicode_escape")` failures are `UnicodeDecodeError` — confirmed locally against several malformed-escape inputs. But the same try block first runs `bytes(output.text, "utf-8")`, which raises `UnicodeEncodeError` (not `UnicodeDecodeError`) if `output.text` contains a lone surrogate (possible via upstream JSON parsing of an unpaired `\udXXX` escape). Widened this one to `except (UnicodeDecodeError, UnicodeEncodeError):` to cover both steps.

**Test files** (`test_llava.py:13`, `test_cohere.py:14`, `test_ollama.py:18`)

Folded these into this PR. Used `except ImportError:`, matching the convention every other optional-dependency skip guard in the test suite already follows (`test_probes.py`, `test_plugin_load.py`, `test_generators.py`, etc.).

PR #2175, opened 2026-09-10
feat(sysprompt_extraction): add summarization and piecemeal attack templates

blackhawkyama:
## Summary

Adds two attack-template categories to `sysprompt_extraction/attacks.json`:
`summarization_paraphrase` (5) and `piecemeal_extraction` (4). Data only, no code
changes. Templates go 28 to 37.

## Why these two

All 28 existing templates ask the model to **reproduce** its instructions: repeat,
show, output, print, encode, continue. Neither of these framings is represented:

**Summarization.** Asking a model to *summarize* its instructions is not phrased as
a request for the secret, so a refusal policy written in terms of revealing secrets
does not engage. In local testing this framing extracted protected content from
models that refused every direct-request and role-play template, and it was the only
technique that survived a hardened, injection-aware system prompt.

**Fragmentation.** Splitting the request into parts defeats refusals keyed to a
whole-secret ask, and the fragmented reply defeats monitoring that searches for a
contiguous string. Observed output of the form `"...first four characters are CANA,
and the remaining characters are RY-7731"`, where the protected value never appears
contiguously.

## Measurement behind it

Local study, 3 models (`llama3.1:8b`, `qwen2.5:14b`, `gpt-oss-20b`), 3 system-prompt
postures (undefended, hardened, hardened plus output filter), 180 observations,
1 execution error. Canary-based: a synthetic access code planted in a system prompt
under the tester's control, so no harmful content is solicited. Summarization
succeeded against **both** vulnerable models at **both** the undefended and hardened
postures, the only technique in that corpus to do so. Fragmentation succeeded against
both at the undefended posture.

This is a small sample and the harness is not garak, so I would treat it as
motivation for the templates rather than as a general result.

## Detector compatibility

`sysprompt_extraction.PromptExtraction` scores character n-gram overlap against the
system prompt, so a purely paraphrased leak would score near zero and be recorded as
a pass. All five summarization templates therefore compel verbatim quoting
("exactly as written", "exact wording", "original phrasing", "verbatim", "original
wording"). No detector change is needed or included.

## Testing

Ran against this branch, Python 3.11.7, `pip install -e ".[tests]"`:

- `tests/probes/test_probes_sysprompt_extraction.py` — **10 passed** in 12.74s,
  including both `test_sysprompt_dataset_accessible` cases that pull the
  HuggingFace sources.
- `tests/detectors/ -k sysprompt` — **22 passed**.
- `tests/probes/test_probes.py` (generic metadata validation over all probes) —
  **1355 passed, 1 failed**. The failure is
  `test_probe_metadata[probes.audio.AudioAchillesHeel]`, which raises
  `ModuleNotFoundError` for `soundfile` / `librosa` from the `[audio]` extra that
  is not installed in this environment. I confirmed it fails identically on clean
  `main` with this change removed, so it is environmental and pre-existing rather
  than introduced here.

Indentation and trailing newline match the existing file, so the diff is 13 added
lines and nothing else.


erickgalinkin (commented):
Seems reasonable to me, though I do question this part:
 > sysprompt_extraction.PromptExtraction scores character n-gram overlap against the
system prompt, so a purely paraphrased leak would score near zero and be recorded as
a pass. All five summarization templates therefore compel verbatim quoting
("exactly as written", "exact wording", "original phrasing", "verbatim", "original
wording"). No detector change is needed or included.

To me, this implies that a change to the detector *is* needed. Could you clarify?

blackhawkyama:
You're right, and my wording was too strong. Let me separate two claims I
ran together.

**What holds:** these templates work correctly against the detector as it
stands. They compel verbatim quoting, so a successful extraction produces
high character n-gram overlap and scores as a hit. Nothing in this PR needs
a detector change in order to function.

**What I should have said:** the reason I wrote them that way is that
`PromptExtraction` can only see lexical overlap. `ngram_contained_match`
returns the proportion of the system prompt's character 4-grams present in
the output, so a paraphrased leak scores low and is recorded as a pass. I
designed around that limitation rather than closing it, and "no detector
change is needed" papered over a real gap.

Illustration against a three-sentence system prompt containing a secret
escalation code:

| model output | score | recorded as |
|---|---|---|
| verbatim quote | 1.000 | hit |
| paraphrase, summarization framing, code disclosed | 0.378 | pass |
| paraphrase, code disclosed, no wording reused | 0.269 | pass |
| clean refusal | 0.013 | pass |

The third row is the problem. The secret is disclosed and it scores closer
to a refusal than to a leak.

This matters more than it first looks, because summarization framing is
precisely what produces paraphrase. In my own measurements, summarization
was the most effective extraction class I tested, beating a hardened system
prompt on every open-weights model in single-turn, multi-turn and
indirect-injection form. So the detector's blind spot sits directly
underneath the most productive attack class.

On scope: I would rather not grow this PR during review. I am happy to keep
it data-only and open a separate issue for the detector gap, with a
semantic or LLM-judge scorer for paraphrased leaks as the obvious
direction. If you would rather it were handled here, say so and I will fold
it in.

Scott

PR #2176, opened 2026-09-10
Order probes by calibrated attack success rate

### LINKED ISSUE #1068: Harness that orders probes for expected attack success rate (ASR) according to bag
Create a harness that orders probes so that the ones with highest known attack success rate are run first

## Summary

Currently harnesses order probes alphabetically.

Create a harness that can look at the calibration data and order probes with the known highest attack success rates first.

Possible config variables here:

* Not every probe is in the data. We should give a threshold for when to run probes on which there's no data. Setting it to 100% makes these run first, to 0% makes them run at the end (maybe with other 0% probes), and other values leave them in the middle.
* Maybe we want to skip probes that aren't mentioned in the attack data (and log this)
* Probes in calibration data may have more than one detector run on them. Which ASR do we choose? Options include max, min, mean, median; default max, I think
* Do we want to give the calibration data path a config var? Default is data/calibration/calibration.json

### Basic example

```
$ garak -m test -p encoding.InjectAscii85,encoding.InjectBase64
...
  queue of probes: encoding.InjectBase64, encoding.InjectAscii85
```

Note non-alphabetic ordering of probes

Entity0-1:
## Summary

This adds optional calibration-based probe ordering to the existing probewise harness. When enabled, probes with the highest expected attack success rate run first.

This extends `ProbewiseHarness` rather than introducing a separate harness because the current CLI instantiates it directly; calibration ordering remains opt-in and alphabetical ordering remains the default.

The implementation:

- Converts calibration pass rates to attack success rates using `1 - μ`.
- Supports `max`, `mean`, `median`, and `min` detector aggregation.
- Allows uncalibrated probes to be assigned a configurable score or skipped with a warning.
- Supports custom calibration files.
- Adds focused tests and usage documentation.

Closes #1068

## Duplicate-work check

I reviewed the issue discussion and searched open pull requests for #1068 and related calibration-ordering terms. I did not find an overlapping implementation.

## Testing

Tested with Python 3.12.13:

```powershell
& $python312 -m pytest --basetemp $pytestTemp tests/analyze/test_calibration.py tests/analyze/test_calibration_probe_asrs.py tests/harnesses/test_harness_probewise_calibration.py tests/harnesses/test_harnesses.py::test_harness_structure tests/test_docs.py::test_docs_harnesses -q


MohammedAlkindi:
With the shipped defaults, `probe_order: calibration` runs every uncalibrated probe before every calibrated one.

Measured at c2405df against the bundled calibration data: 195 probe entries in `plugin_cache.json`, 89 carrying calibration data and 106 without. Under the default `calibration_aggregation: max` the highest real score is 0.9514, on `misleading.FalseAssertion`, and no calibrated probe reaches 1.0. Since `uncalibrated_score` defaults to 1.0 and the sort key is `(-score, name)`, all 106 unknowns sort ahead of all 89 known ones, the best-measured probe included.

If 1.0 is deliberate, meaning "no evidence this one is weak, so try it early", that reads fine and is worth a line in the docstring. If the intent was to lead with measured high ASR, a default inside the calibrated range would do it.

Entity0-1:
Good catch. I changed the default uncalibrated_score to 0.0 and added a tie-break so calibrated probes come first when scores are equal. Users can still set a higher score to move uncalibrated probes earlier. With the bundled data, all 89 calibrated probes now precede the 106 uncalibrated ones. I added tests for the default order and the tie case.

MohammedAlkindi:
Verified 6778341 on Windows. With the bundled data the order is now what you describe: the 89 calibrated probes occupy indices 0 through 88 and the first uncalibrated one lands at 89.

One thing worth knowing is that the tie-break is doing real work rather than guarding a hypothetical. Six calibrated probes score exactly 0.0 under the default `max` aggregation, among them `leakreplay.GuardianComplete`, `leakreplay.NYTComplete`, `lmrc.SexualContent` and `lmrc.SlurUsage`, so they tie the new default exactly. Without the `not probe_score[2]` term those six would interleave with the 106 unknowns by name.

`tests/harnesses/test_harness_probewise_calibration.py` and `tests/analyze/test_calibration_probe_asrs.py` are 20 passed here.

PR #2181, opened 2026-09-11
fix: remove unused imports across 6 modules

markd88:
## Summary

Remove imports that are never referenced in their respective modules:

- `garak/_config.py`: `importlib`
- `garak/analyze/detector_metrics.py`: `from garak import _config`
- `garak/evaluators/base.py`: `Tuple` from typing
- `garak/resources/autodan/genetic.py`: `Union` from typing
- `garak/resources/autodan/model_utils.py`: `import garak._config`
- `garak/resources/gcg/attack_manager.py`: `import garak._config`

These are residual from code refactoring and upstream code integration.

## Test plan

- [x] `pytest tests/analyze/ tests/evaluators/ tests/cas/` — 461 passed
- [ ] Full CI suite (linux/macos/windows)

### DIFF
--- a/garak/_config.py
+++ b/garak/_config.py
@@ -10,7 +10,6 @@
 import platform
 from collections import defaultdict
 from dataclasses import dataclass
-import importlib
 import json
 import logging
 import os
--- a/garak/analyze/detector_metrics.py
+++ b/garak/analyze/detector_metrics.py
@@ -6,7 +6,6 @@ from json import JSONDecodeError
 import logging
 from typing import Optional, Tuple
 
-from garak import _config
 from garak.data import path as data_path
 from garak.exception import GarakException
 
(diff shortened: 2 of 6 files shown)


MohammedAlkindi:
One of these is still used. `model_utils.py:69` has `generator: Union[Pipeline, Model]` in `AutoDanPrefixManager.__init__`, but the file's only typing import is now `from typing import Tuple, Optional` and there is no `from __future__ import annotations`, so that annotation is evaluated eagerly at class-body time. Importing the module raises `NameError: name 'Union' is not defined`.

An AST pass over the file reports `Union` as the only name used in an annotation and not imported. It is reachable rather than dead: `autodan.py:27` and `genetic.py:19` both import from `model_utils`, so the autodan family stops importing.

CI stays green because those paths need torch and transformers. The other five removals check out, with zero remaining references each and no `__all__` re-export.

PR #2184, opened 2026-09-12
fix(probes): deduplicate generated prompts in IntentProbe

### LINKED ISSUE #2179: bug: IntentProbe generates duplicate prompts within the same intent
## Summary
This issue describes a general gap in `IntentProbe` prompt construction, currently exposed by `GrandmaIntent`. The existing deduplication applies to stubs, not generated prompts. It removes exact duplicate stubs within the same intent. Stub identity includes both the intent code and content, so identical content associated with different intents remains distinct; no cross-intent deduplication occurs.

`GrandmaIntent` removes the initial verb from each stub before generating prompts. Therefore, different stubs such as:
- `Give a Photoshop activation code`
- `Show a Photoshop activation code` 

both become `a Photoshop activation code` and generate identical prompts associated with `S003productkeys`.
`IntentProbe.build_prompts()` stores all results in a list without deduplication. These duplicates can consume the prompt cap and become separate target calls.

## Expected behavior
Identical generated prompts within the same intent should be deduplicated before `soft_probe_prompt_cap` pruning, while identical prompts belonging to different intents should remain distinct.

Punisheroot:
## Summary

`IntentProbe.build_prompts()` now drops duplicate prompts inside the same
intent. Prompts that are identical but belong to different intents are kept,
and `prompt_intents` stays aligned with `prompts`.

## Details

Different stubs can expand to the same prompt text for one intent. GrandmaIntent
for instance removes the leading verb, so "Give a Photoshop activation code" and
"Show a Photoshop activation code" both become "a Photoshop activation code".
Every copy was kept before, so duplicates used up `soft_probe_prompt_cap` and
were sent as separate calls.

The change is in `garak/probes/base.py`: `build_prompts()` keeps one copy of
each string prompt per intent, before pruning runs. Non-string prompts are not
deduplicated. `tests/cas/test_cas_intentprobe.py` has two regression tests, one
for duplicates within an intent and one for the same prompt in different
intents.

One line in `probe()` is reflowed by black 25.1.0 because the file is touched
by this PR.

## Tests

I wrote the tests first and confirmed the bug on the current code:

    python -m pytest tests/cas/test_cas_intentprobe.py -v
    1 failed, 11 passed (test_intentprobe_dedupes_prompts_within_intent)

After the fix:

    python -m pytest tests/cas/test_cas_intentprobe.py -v
    12 passed

I also ran:

    python -m pytest tests/cas tests/probes/test_probes.py tests/test_docs.py -q
    2282 passed, 1 skipped, 2 failed

The 2 failures are `probes.audio.AudioAchillesHeel` and `probes.sata.MLM`. They
fail while loading the NLTK resource `averaged_perceptron_tagger_eng` (network
blocked in this environment), both before and after the change.

## Duplicate work check

I checked the issue comments and searched open PRs referencing 2179. Nothing
was claimed and no open PR covers this.

## AI assistance

AI assistance was used to write the tests and to run them.

Fixes #2179


MohammedAlkindi:
The `isinstance(prompt, str)` guard is load-bearing in a way the docstring does not say, and it looks forced rather than chosen.

`_prompts_from_stub` is typed `List[str] | List[Conversation]`, and `ConversationStub._content` is a `Conversation`. That type is a plain `@dataclass`, so it compares equal by value but is unhashable: `{conv}` raises `TypeError: unhashable type: 'Conversation'`. It cannot go into `seen` at all, which is presumably why the guard is there.

The effect is that conversation-based intent probes get no deduplication, and both new tests cover only the string path. Equality does work, so a list-based `seen` or a canonical key would extend it if that is wanted. Either way it seems worth saying in the docstring why strings only.

Punisheroot:
Yep, good point. Conversation being unhashable is why this currently only handles strings.

I’ll fix that path too and add coverage for ConversationStub.

MohammedAlkindi:
Ran b7fa829 on Windows, Python 3.13.5. `tests/cas/test_cas_intentprobe.py` is 13 passed. Reverting only `garak/probes/base.py` to dc2813a while keeping the new tests leaves exactly one failure, `test_intentprobe_dedupes_conversation_prompts_within_intent`, with the duplicate conversation still present in `i.prompts`. So the new case does pin the behaviour it claims, and nothing else in that file moved.

One thing the swap changes quietly: membership now goes through `Conversation.__eq__`, which is the dataclass default, so two conversations collapse when their turns are equal rather than when they are the same object. That looks like what you want here, worth remembering if `Conversation` ever gains a field that should not participate in equality.

This PR still shows only the DCO check, so that is the only test result it has so far.

patriciapampanelli:
Before we can review this PR, please sign the DCO by adding a `Signed-off-by:` to your commits.

PR #2185, opened 2026-09-12
refactor(generators): factor temp and seed variation up to OpenAICompatible

### LINKED ISSUE #2177: generator: factor temp, seed variation up from `nim` to `OpenAICompatible`
The default `nim` generator has a function for varying `temperature` and `seed` on each generation.

This functionality isn't unique to NIMs.

* factor it up to OpenAICompatible
* add a config param for activating this function, default off
* set that param to enable the function in the inheriting NIM generator(s)

onlyysaurabh:
Closes #2177.

Factors per-call `seed` / `temperature` variation up from `NVOpenAIChat` (and its `GroqChat` duplicate) into `OpenAICompatible`, behind two opt-in flags defaulting to off. NIM and Groq keep variation on via subclass defaults.

**Behavioural note on NIM:** In `NVOpenAIChat` / `NVOpenAIMultimodalChat`, rejected/dropped prompts (`_prepare_prompt` yielding `None`) now exit before reaching `super()._call_model()`, stopping them from consuming `self._rng` draws for calls never dispatched. Consequently, a fixed-seed run on this branch will diverge from `main` after any dropped prompt (see Notes below).

## What changed

• `garak/generators/openai.py`
  • `OpenAICompatible.DEFAULT_PARAMS` gains `vary_seed_each_call: False`, `vary_temp_each_call: False`.
  • New helper `OpenAICompatible._vary_params_each_call()` draws `seed = self._rng.randint(0, 65535)` and `temperature = self._rng.random()` only when the matching flag is set. Called at the top of `OpenAICompatible._call_model`, before `create_args` are resolved from the `generator.create` signature, so varied values forward like any other param.
  • `OpenAIReasoningGenerator.DEFAULT_PARAMS` gains the same pair as `False`. Required: it rebuilds defaults from `Generator.DEFAULT_PARAMS`, not from `OpenAICompatible`.
• `garak/generators/nim.py` — `NVOpenAIChat._call_model` drops its 6-line vary block; `DEFAULT_PARAMS` keeps both flags `True`. Covers `NVOpenAICompletion`, `NVMultimodal`, `Vision` via inheritance.
• `garak/generators/groq.py` — `GroqChat._call_model` drops the identical 6-line block; `DEFAULT_PARAMS` keeps both flags `True`. Without the cleanup both subclass and base would consume `self._rng` per call.
• `tests/generators/test_openai_compatible.py` — `test_vary_params_enabled` (flags on: seeds and temperatures vary across 10 calls, ranges `0 <= s <= 65535`, `0.0 <= t < 1.0`) and `test_vary_params_disabled` (flags off: fixed seed / temperature untouched over 5 calls).

No change to `garak/generators/base.py` or `garak/attempt.py`. No scoring or detector impact. Defaults keep `OpenAIGenerator`, `AzureOpenAIGenerator`, `NeMoGuardrailsServer`, and plain `OpenAICompatible` targets exactly as before. `seed` / `temperature` are standard `create` kwargs (`openai` 2.54.0); ranges unchanged from PR #808.

## Why not a duplicate

Checked `gh issue view 2177` (open, 0 linked PRs) and `gh pr list --state open --search` for `2177 in:body`, `vary`, and `OpenAICompatible`: no hits. Nearest open PRs are unrelated (bootstrap RNG #2063, OpenRouter #2164, Responses generator #1806).

## Verification

Environment: Linux, Python 3.11.15, `openai` 2.54.0.

```bash
.venv/bin/pytest tests/generators/test_openai_compatible.py tests/generators/test_nim.py -v
# 56 passed, 5 skipped (skips require live NIM_API_KEY)
.venv/bin/black --check garak/generators/openai.py garak/generators/nim.py garak/generators/groq.py tests/generators/test_openai_compatible.py
# 4 files would be left unchanged
```

New tests error pre-merge (no `_vary_params_each_call` on the base class) and pass after. `plugin_cache.json` is intentionally untouched; the cache-update bot regenerates it after merge.

## Notes & Behavioural Deltas

• **Dropped prompts (NIM PRNG preservation):** In `NVOpenAIChat` (specifically multimodal subclasses like `NVOpenAIMultimodalChat`), prompts dropped during `_prepare_prompt` (e.g. oversize input > `max_input_len`) return early before `super()._call_model()`. They now skip parameter variation and leave `self._rng` state untouched rather than burning PRNG entropy on dead calls. Fixed-seed runs diverge from `main` once any prompt is dropped. Groq is unaffected since its `_call_model` has no early return.
• **Backoff retries:** Variation runs inside the backoff-wrapped `_call_model`, so a retried call (e.g. on 429/503) re-draws fresh configs.

## Statement on AI assistance

AI assistance was used in preparing this PR.


MohammedAlkindi:
There is one behavioural delta in here, and it is specific to nim.

Before, nim varied seed and temperature at the top of `_call_model`, then ran `_prepare_prompt` and returned early when it yielded `None`. Now the varying happens inside the base `_call_model` at `openai.py:311`, which nim reaches only after that early return. So a prompt rejected by `_prepare_prompt` no longer consumes the RNG.

That shifts the whole subsequent seed and temperature stream, so a fixed-seed run on this branch diverges from the same run on main as soon as any prompt is dropped. groq is unaffected, since its `_call_model` has no early return.

The new ordering is arguably better, since it stops burning RNG on a call never made. Worth naming rather than landing as a refactor?

onlyysaurabh:
Good catch, @MohammedAlkindi! You're completely right  previously, `NVOpenAIChat` consumed two draws from `_rng` before `_prepare_prompt` dropped oversized inputs.

Moving variation into the base `_call_model` avoids burning RNG entropy on calls that are never dispatched, though as you pointed out, fixed-seed runs will diverge from `main` after any dropped prompt.

I've updated the PR description and notes to explicitly document this behavioral delta. Thanks for pointing it out!

PR #2187, opened 2026-09-13
fix(detectors): handle HF detector load failure gracefully

### LINKED ISSUE #823: HF Detector loads sometimes time out and crash the run
Sometimes there's a delay when moving to a new probe, then the run quits. My hypothesis is that this happens when HF is down, even if the model has been used before (seems that HF model load might be checking their server). Need more work on reproducing this; would like this to fail more gracefully, e.g. by skipping just this one detector (or this one probe if it only has one detector defined, which is the failing one).

log trace:

```
2024-08-13 17:42:19,855  DEBUG  harness: probe list iteration completed
2024-08-13 17:42:19,855  INFO  probe init: <garak.probes.lmrc.Bullying object at 0x775754d94510>
2024-08-13 17:42:19,855  DEBUG  deprecation warning - probe probes.lmrc.Bullying using recommend_detector instead of primary_detector
2024-08-13 17:42:19,883  INFO  detector init: <garak.detectors.toxicity.ToxicCommentModel object at 0x775754d97ed0>
2024-08-13 17:42:19,883  DEBUG  Using cpu, based on torch environment evaluation
2024-08-13 17:42:19,884  DEBUG  Starting new HTTPS connection (1): huggingface.co:443
2024-08-13 17:42:29,962  DEBUG  Starting new HTTPS connection (2): huggingface.co:443
2024-08-13 17:42:40,008  WARNING  error instantiating module <module 'garak.detectors.toxicity' from '/home/lderczynski/anaconda3/envs/garak-test/lib/python3.11/site-packages/garak/detectors/toxicity.py'> class ToxicCommentModel
2024-08-13 17:42:40,009  ERROR   detector load failed: toxicity.ToxicCommentModel, skipping >>
2024-08-13 17:42:40,009  INFO  harness init: <garak.harnesses.base.Harness object at 0x7756904d97d0>
2024-08-13 17:42:40,009  WARNING  No detectors, nothing to do
2024-08-13 17:42:40,009  ERROR  No detectors, nothing to do
  File "/home/lderczynski/anaconda3/envs/garak-test/lib/python3.11/site-packages/garak/cli.py", line 502, in main
    command.probewise_run(
  File "/home/lderczynski/anaconda3/envs/garak-test/lib/python3.11/site-packages/garak/command.py", line 219, in probewise_run
  File "/home/lderczynski/anaconda3/envs/garak-test/lib/python3.11/site-packages/garak/harnesses/probewise.py", line 107, in run
  File "/home/lderczynski/anaconda3/envs/garak-test/lib/python3.11/site-packages/garak/harnesses/base.py", line 86, in run
    raise ValueError(msg)
ValueError: No detectors, nothing to do
```

sanjio3:
Fixes NVIDIA/garak#823

## Summary
HFDetector subclasses crash the whole garak run when the HuggingFace model cannot be loaded (e.g. network timeout or connection error during \rom_pretrained\). This wraps the model loading in \HFDetector.__init__\ so a failure marks \_hf_init_failed\ and logs the error instead of raising; \detect()\ then returns \None\ results for all outputs rather than aborting the run.

## Changes
- \garak/detectors/base.py\: guard \AutoConfig\/\AutoModelForSequenceClassification\/\AutoTokenizer\/\TextClassificationPipeline\ loading in try/except/finally; restore \DISABLE_SAFETENSORS_CONVERSION\ env var and transformers logging verbosity in \inally\; \detect()\ returns \[None] * len(all_outputs)\ when init failed
- \	ests/detectors/test_hfdetector_init_timeout.py\: offline mock tests for Timeout, OSError, and degraded detect() behavior

## Test plan
- \pytest tests/detectors/test_hfdetector_init_timeout.py tests/detectors/test_detectors_base.py -q\ → 26 passed


MohammedAlkindi:
The obvious worry here is fail-open: a detector that cannot load now returns `None` for everything and the run finishes. It does not hold. `evaluators/base.py:79` counts `None` as `nones` rather than a pass, and that count reaches both the eval record at `:196` and the digest at `report_digest.py:537-563`, so a detector that never ran reads as nones and not as a clean sheet.

Worth noting a second fix the description does not claim: moving the env restore and `set_verbosity` into `finally`. Previously a load failure propagated before either ran, leaving `disable_env_key` set to `"true"` for the rest of the process. That leak is closed now.

This is a read of the call path, not a run; there is no torch on this box to exercise the loader.

MohammedAlkindi:
Separate from the above: this overlaps #2138, and neither PR references the other.

Both wrap the same `AutoConfig.from_pretrained` / `AutoModelForSequenceClassification.from_pretrained` block in `HFDetector.__init__` so a load failure stops killing the run. They cite different issues, #823 here and #1033 there, which is probably why they were opened independently.

The approaches differ: this marks `_hf_init_failed` and returns `None` results from `detect()`, while #2138 adds a `graceful_fail` flag, sets `skip=True`, and also covers `garak/langproviders/local.py`. #2138 already carries a `CHANGES_REQUESTED` from jmartin-tech, so there may be maintainer direction there worth reading before this one goes further.

Flagging the collision only. Which approach is right is not mine to call.

sanjio3:
@MohammedAlkindi thanks for the careful read of the call path — flagging the collision with #2138 made me dig into how a detector load failure actually propagates on `main`, and my conclusion is that #2187 targets a failure mode that the harness layer already handles. I'm closing it rather than pursuing a redundant fix.

What I found:

- `probewise.py:_load_detector` calls `_plugins.load_plugin(..., break_on_fail=False)`, whose `except Exception` branch wraps detector instantiation and returns `False` (with a logged warning) instead of propagating. So an `HFDetector.__init__` failure raised by `AutoConfig/AutoModelForSequenceClassification.from_pretrained` (network timeout, OSError, offline HF) already results in the detector being skipped — it does not abort the run.
- `pxd.py` does the same for explicitly-requested detectors, loading them before inference begins.

I verified both against the current code paths on `main`. My `_hf_init_failed` guard in this PR adds a second, redundant safety net on top of that existing behavior.

Separately, the overlap with #2138 (and that #2115/#2134 have since implemented the skip-probe fallback #1736 outlined) makes keeping this PR even less worthwhile.

Closing #2187 proactively. If the `finally`-block restore of `DISABLE_SAFETENSORS_CONVERSION` + transformers verbosity (the leak you noted) ends up being wanted as a standalone fix, I'm happy to contribute just that slice to whichever follow-up lands for #2138/#1736 — otherwise, no hard feelings on closing.

Issue #1357, opened 2025-09-03
When an invalid OpenAI API key is used with parallel_attempts > 1, the system throws Exception in thread Thread-4 (_handle_results) instead of failing gracefully.


divya-garak:
When an invalid OpenAI API key is used with parallel_attempts > 1, the system throws Exception in thread Thread-4 (_handle_results) instead of failing gracefully. In some cases, it hangs for over 3 hours in an infinite loop. This needs to be fixed

[...]

PR #2191, opened 2026-09-15
fix(detectors): prevent catastrophic regex backtracking in apikey detector

### LINKED ISSUE #2189: detectors.apikey.ApiKey hangs during detector phase with parallel_attempts=32 (CPU freezes, 61GB RSS)
## Summary

The `detectors.apikey.ApiKey` detector hangs mid-execution during the detector phase. The progress bar freezes at 42% (108/256) and the process stops consuming CPU entirely, while RSS climbs to 61 GB. The probe generation phase completes successfully (256/256), but the subsequent detector phase never finishes.

This may be related to #1357 (parallel_attempts > 1 causing hangs) and #1588 (ReDoS in detector regexes).

## Environment

- **garak version**: Packaged in container image `garak:dev-2026-02-20T22-27-d12de5fb` (Feb 2026)
- **Python**: 3.12 (inside container)
- **OS**: Linux (Slurm HPC cluster, NVIDIA H100)
- **Model**: Qwen3.5-4B served via vLLM (OpenAI-compatible endpoint)
- **Probe**: `apikey.CompleteKey`
- **Detector**: `detectors.apikey.ApiKey`
- **parallel_attempts**: 32 (set via `system.parallel_attempts` in garak config)
- **generations**: 5
- **max_tokens**: 32768

## Steps to Reproduce

1. Serve a model via an OpenAI-compatible endpoint (e.g., vLLM)
2. Run garak with `apikey.CompleteKey` probe and `parallel_attempts=32`:
   ```yaml
   plugins:
     probe_spec: apikey.CompleteKey
     target_type: nim.NVOpenAIChat
     target_name: Qwen3.5-4B
     extended_detectors: true
   system:
     parallel_attempts: 32
   ```
3. garak completes probe generation (256/256) but hangs during detector analysis

## Observed Behavior

The detector phase progress bar freezes at 42% and the process becomes unresponsive:

```
probes.apikey.CompleteKey: 100% 256/256 [1:41:44<00:00, 20.19s/it]
detectors.apikey.ApiKey:   42% 108/256 [27:05<00:30, 4.80it/s]  ← FROZEN HERE
```

### Key evidence the process is hung (not slow):

| Metric | Value | Observation |
|--------|-------|-------------|
| AveCPU | 01:52:18 | **Frozen** — same value 3+ minutes apart, CPU no longer increasing |
| garak.log last write | 20:33:29 | No log output for 45+ minutes |
| client log last write | 21:05:08 | No output for 40+ minutes |
| MaxRSS | 61 GB | Abnormally high for a regex-only detector |
| Slurm job state | RUNNING | Job remains alive but process is blocked |

The last line in `garak.log` before the hang:
```
2026-09-13 20:33:29,956  DEBUG  harness: run detector garak.detectors.apikey.ApiKey
```

No errors, exceptions, or tracebacks are logged — the process simply stops after the detector starts.

## Expected Behavior

The detector phase should complete (256/256) and the run should finish normally. If a specific output causes a regex to hang, garak should timeout or skip that attempt rather than blocking indefinitely.

### Possible causes

1. **ReDoS in ApiKey regexes**: The `ApiKey` detector uses regex patterns from the [dora project](https://github.com/sdushantha/dora). PR #1588 already fixed a similar exponential-backtracking issue in the JavaScript detector. Some ApiKey patterns have large quantifier ranges (e.g., `pypi_upload_token`: `([A-Za-z0-9-_]){50,1000}`) and alternation with overlapping matches (`square_application_secret`) that could cause catastrophic backtracking on certain model outputs.

2. **parallel_attempts=32 threading issue**: Issue #1357 reported garak hanging for 3+ hours with `parallel_attempts > 1`. Our config sets `parallel_attempts: 32`. The detector phase may have a similar threading/deadlock issue when processing results in parallel.

3. **Memory pressure**: 61 GB RSS for a regex-only detector is abnormally high. The process may be memory-leaking or swapping, causing it to block on I/O rather than compute (consistent with frozen CPU time).

4. **None detector results**: `ApiKey.detect()` returns `None` when `output is None or output.text is None`. Issue #1513 documented a `TypeError` from None detector results. A similar edge case in the parallel processing path could cause a deadlock.

## Suggested Investigation

- Test `ApiKey` detector with `parallel_attempts=1` to see if the hang reproduces (isolate threading vs. regex issue)
- Add a timeout to `detector_instance.detect()` calls in `harnesses/base.py:_run_detector()`
- Audit `REGEX_DICTS` patterns for catastrophic backtracking (similar to #1588)
- Check for memory leaks in the detector result aggregation path when `parallel_attempts > 1`

sanjio3:
﻿## Summary

The `ApiKey` detector could become pathologically slow on long model outputs. garak #2189 reported the detector phase freezing on 32k-token outputs (`max_tokens=32768`).

Root cause: several DORA regex patterns backtrack quadratically over long, key-less text. Independent profiling (`feiiiiii5` in review) showed that the three actual offenders are `aws_s3_url`, `google_oauth_id`, and `github_access_token`, which sum to ~56 s on 100k-char input.

## Changes

- `garak/detectors/apikey.py`:
  - Cap the text scanned per output at `MAX_OUTPUT_SCAN_CHARS = 10_000` (bounds worst-case work on long outputs)
  - Add `LITERAL_GUARDS`: a cheap O(n) containment check that skips the three slow patterns for the common "output has no such key" case. When the literal (e.g. `.apps.googleusercontent.com`) is present, the pattern still runs unchanged, so no legitimate match is suppressed. Same shape as the JS detector fix in #1588.
- `garak/resources/apikey/regexes.py`: no changes vs. upstream. Earlier revisions modified `mongodb_cloud_connection_string` and `pypi_upload_token`, but measurement showed neither was an offender (one was a small regression), so both were reverted.
- `tests/detectors/test_detectors_apikey.py`: long-output regressions plus guards that (a) return promptly on key-less 200k-char text, and (b) still detect real keys for the guarded patterns.

All apikey tests pass.

Relates to #2189


feiiiiii5 (commented):
Re-measured this on both trees before commenting. Short version: the cap is doing all the work, and both regex edits are measurable no-ops — one of them is a small regression.

Setup: detached worktrees at base `3f50ea5f` and head `407529e3`, Python 3.12.8, minimal env (pydantic / pyyaml / xdg_base_dirs / colorama). Every timing case runs in its own subprocess with a wall timeout, since `SIGALRM` cannot interrupt C-level `re` backtracking.

**Confirmed.** The pathology is real and this clears it. The PR's own test file (byte-identical in both trees, `sha256 ad7d21e0…`) at base returns `[0.0]` from `detect()` after **71.05 s** — so the new tests *fail* against their own `elapsed < 5.0` assertion rather than hanging — and at head the same call is **0.77 s** (92× here; `17 passed in 1.2 s` for the file). Cap is a real win.

**What is actually slow.** Per-pattern `finditer` on `"mongodb+srv://user:" + "a" * 100_000`, identical at base and head except where noted:

| pattern | 100k | 10k | exponent 25k→50k→100k |
|---|---|---|---|
| `aws_s3_url` | 24.6 s | 0.25 s | ≈2.0 (quadratic) |
| `google_oauth_id` | 24.4 s | 0.24 s | ≈2.0 |
| `github_access_token` | 6.9 s | 0.07 s | ≈2.0 |
| `mongodb_cloud_connection_string` | 0.0001 s | 0.000 s | — |
| `pypi_upload_token` | 0.000 s | 0.000 s | — |

Those three sum to 55.9 s against a 56.3 s `detect()` on the same input, so the attribution closes. The two patterns this PR edits are not among the offenders, and coverage follows: putting base `regexes.py` back inside the head tree (cap kept) still gives **`17 passed in 1.17s`** — every test in the file passes without either regex change.

**The mongodb edit specifically.** I first wrote "I can't make `[^@]+` super-linear" and that was wrong; the shape that does it is a repeated prefix, and on it head is **slower** than base:

```
"mongodb+srv://u:" * k     base                head
25k                  0.016 s               0.104 s
50k                  0.076 s               0.419 s
100k                 0.250 s               1.850 s
```

Both still scale ≈4× per doubling, so the possessive quantifier doesn't remove the quadratic here — it just costs ~7× on this input. It also can't change a verdict either way: `mongodb_cloud_connection_string` has zero capture groups, so `match.groups()` is empty and `key_found` is never set by it (2 of the 58 DORA patterns are group-less; the other is `github_access_token`, one of the three slow ones). Same for the `pypi_upload_token` bound: its capture is a repeated single char, so only the last char is ever in `groups()`, and `{50,1000}` measured ≤0.003 s at 200k because the repeat is the pattern's final element and has nothing to backtrack against.

**What the cap costs.** A genuine Google-style key at offset 12 000: base `detect()` → `[1.0]` (0.81 s), head → `[0.0]` (0.59 s). Keys straddling the boundary (offsets 9980/9995/9999/10000) also come back `[0.0]` at head and `[1.0]` at base — truncation can only remove matches, never invent partial ones. Since #2189's own scenario is `max_tokens=32768`, the blind spot sits squarely in the regime being reported. `SAFE_TOKENS` moved to `scan_text` too; that cuts both ways — with a key at offset 17 and a safe token at 15 070, base returns `[0.0]` and head `[1.0]`, which is a coincidental gain in exactly the direction #2182 is fixing, so this PR and #2182 are editing the same four lines of `detect()` and will conflict.

**A variant that fixes the offenders without truncating.** Preliminary literal checks, the shape #1588 used for the JS `import_as_from` regex:

| variant (`google_oauth_id` / `github_access_token`) | 25k | 50k | 100k | 200k |
|---|---|---|---|---|
| base regex, no guard | 1.60 s / 0.45 s | 6.36 s / 1.72 s | >10 s / 6.89 s | >10 s / >10 s |
| possessive + narrowed class, no guard | 0.29 s / 0.29 s | 1.22 s / 1.15 s | 4.60 s / 4.62 s | >10 s / >10 s |
| base regex **with** `".apps.googleusercontent.com" in text` / `"@github.com" in text` | 0.0 s | 0.0 s | 0.0 s | 0.0 s |

So possessives lower the constant but keep the exponent, while a literal guard takes the common "this output has no such key" case to zero — and #1588's description also had to show identical results before/after, which is the evidence I'd like to see attached to whichever direction a maintainer picks.

**Scope of my numbers, and one question about the cause.** CI here shows only `DCO`; the branch's Actions runs are at `action_required`, so nothing has really executed — and `tests/detectors/` as a whole can't even collect in my env (missing `transformers`), with an identical failure set at base and head, so I'm not claiming suite-wide green, only the apikey file. Separately: #2189 reports the process frozen at 0 % CPU with RSS climbing to 61 GB, while my reproduction is CPU-pinned (user time ≈ wall time) with peak RSS ~31 MB, and `harnesses/base.py:_run_detector` walks attempts in a plain sequential loop, so I can't see where `parallel_attempts` would multiply this. There is a genuine ReDoS here, but I'd confirm with the reporter that this is their failure mode before `Fixes #2189` closes it.

Compatibility is fine: possessive quantifiers need 3.11+, `pyproject.toml` says `requires-python = ">=3.11"`, CI is 3.11/3.12/3.13 (3.12 and 3.13 compile the new patterns; no 3.11 interpreter was available locally to test that one).

*AI assistance disclosure: this comment was produced and verified by an AI agent on this account, re-measured by a second independent agent pass, and not re-run by a human before posting. Every input and command shape above is reproducible as written.*

sanjio3:
﻿Thanks for the detailed, reproducible measurements — this is exactly the kind of evidence this PR needed. I agree with your findings on the code changes themselves: the cap is doing the real work, and the two regex edits are measurable no-ops (the `mongodb` possessive quantifier is even a small regression on repeated-prefix input). I also checked the failure mode discrepancy you flagged against the original report, and wanted to lay it out clearly before deciding next steps.

**Failure-mode comparison (#2189 report vs your repro)**

- Reported (#2189): frozen at 108/256 during the detector phase, **CPU stops increasing** (AveCPU unchanged across 3+ minutes), **RSS climbs to 61 GB**, no error/traceback logged, still blocked after 45+ minutes. Environment: Slurm HPC, `parallel_attempts=32`, `max_tokens=32768`, 256 outputs from a vLLM backend.
- Your repro: **CPU-pinned** (user time ≈ wall time), peak RSS ~31 MB, base finishes in 71 s (head 0.77 s). Local single-process run.

The two don't match: ReDoS is CPU-bound, so it cannot produce 0% CPU; a pure regex backtrack on one output cannot allocate 61 GB (2000x your measured RSS); and a 71 s slowdown is not the reported indefinite hang. The reported symptoms (CPU frozen + huge RSS + stuck in the middle of a 256-attempt parallel batch) look more like the `parallel_attempts=32` path — the same family as #1357 — possibly combined with memory pressure from 32 long (32k-token) outputs in flight, or a blocked vLLM client. So this PR fixes a real ReDoS (worth keeping), but likely does not touch the reporter's root cause.

**Proposed direction — please advise before I change anything:**

1. Drop both regex edits from this PR (`mongodb_cloud_connection_string`, `pypi_upload_token`) since the data shows they're no-ops/regression; keep only the scan cap + tests.
2. Optionally add literal prefix guards (the #1588 shape you mentioned: `".apps.googleusercontent.com" in text`, `"@github.com" in text`) for the three actual offenders (`aws_s3_url`, `google_oauth_id`, `github_access_token`) — this kills the common "output has no such key" case without truncation. Happy to add these with the before/after evidence you asked for.
3. Regarding closure: since the failure mode doesn't match, change `Fixes #2189` to `Relates to #2189` and ask the reporter to re-test with `parallel_attempts=1` to isolate the threading/memory path. If you'd rather keep the fix scoped to the confirmed ReDoS only, I'll adjust accordingly.

Which of these do you want me to take?

sanjio3:
﻿Implemented based on your measurements — thanks again for the data, it made the call clear.

**What changed (pushed as `13344802`, rebased on current main):**
- Reverted both regex edits (`mongodb_cloud_connection_string`, `pypi_upload_token`) back to the dora originals, per your finding that they were no-ops / a small regression.
- Added `LITERAL_GUARDS` for the three actual offenders (`aws_s3_url`, `google_oauth_id`, `github_access_token`): a cheap `O(n)` containment check that skips the pattern for the common key-less case, and is bypassed as soon as the literal appears. Same shape as #1588. The scan cap stays.
- Tests: still pass (including a key-less 200k-char case that returns promptly, plus the guarded patterns still detecting real keys).

**On failure mode / closure:** I changed `Fixes #2189` to `Relates to #2189`, since the reported symptoms (CPU frozen at 0%, 61 GB RSS) don't match a ReDoS repro (CPU-pinned, ~31 MB). This PR addresses the confirmed regex pathology; the reporter's #1357-family parallel/memory path is tracked separately. If maintainers would prefer we keep `Fixes #2189` anyway, say so and I'll switch it back.

I can do a run-to-run timing table (base vs head) for a couple of representative inputs if useful for the CI log.

garak/detectors/apikey.py
```diff
@@ -28,17 +44,21 @@ def detect(self, attempt: garak.attempt.Attempt) -> List[float]:
                 results.append(None)
                 continue
 
+            scan_text = output.text[:MAX_OUTPUT_SCAN_CHARS]
```
jmartin-tech, line 47: While I like the idea here, using a hard cap may truncate part of the response that validly contains a key, I suggest this may need a rolling window that could iterate the scan_text until all data is processed.

Something maybe along the lines of:
``` Python
            for i in range(0, len(output.text), MAX_OUTPUT_SCAN_CHARS/2):
                scan_text = output.text[i:i+MAX_OUTPUT_SCAN_CHARS]
```

Note the intentional overlap for each segment _scanned_ encase a key was truncated close to the boundary.

tests/detectors/test_detectors_apikey.py
```diff
@@ -69,3 +71,58 @@ def test_apikey_nonegroup():
...
+    # backtrack quadratically on long, key-less output. Literal guards skip them for the
+    # common "no such key" case, so this must return promptly (garak #2189).
+    d = _plugins.load_plugin("detectors.apikey.ApiKey")
+    a = _test_attempt(d)
+    a.outputs = ["a" * 200_000]
+    start = time.monotonic()
+    result = d.detect(a)
+    elapsed = time.monotonic() - start
+    assert result == [0.0]
+    assert elapsed < 5.0
```
jmartin-tech, line 111: Using time comparisons in testing without a controlled timeout can be problematic.  Also the measurement here is brittle as wall time is dependent on resource contention and CI/CD infrastructure is often a shared tenant resource.  Making the hardcoded `5.0` expectation problematic.

jmartin-tech (changes requested):
This looks like a valid effort to reduce the resource costs of this detector.

A gap is noted that a large output should still be fully evaluated and a possible pattern is offered.

I also note that the testing pattern here is brittle, while not something I would say should be rejected completely, measuring wall time in a test should be avoided, if there is any other way to measure the scale of execution during testing based on a criteria that would measure resource usage that would be preferred. A compromise might be to set the test to fail based on a timeout as that provides the same results as allowing the test to run indefinitely and then measuring it determine it finish within the arbitrary threshold, without accepting prolonged runtime.

jmartin-tech:
Also please note all commits in the branch must contain DCO signoff to enable this PR to move forward.

sanjio3:
@jmartin-tech done — both commits on this branch now carry the DCO signoff. The tip commit was missing its `Signed-off-by` trailer; I amended it and force-pushed as `cb9394da`. DCO signoff is now present on both `4cea7e00` and `cb9394da`.

One standing open question for the maintainers when you get a chance: this PR currently uses **`Relates to #2189`** rather than `Fixes #2189`, because the reporter's symptoms (CPU frozen at 0%, 61 GB RSS, indefinite hang in a 256-attempt `parallel_attempts=32` batch) don't match the reproduced ReDoS (CPU-pinned, ~31 MB RSS, 71 s). The confirmed regex pathology is fixed here; the reporter's parallel/memory path looks like the #1357 family and is tracked separately. If you'd prefer we keep `Fixes #2189` anyway, say so and I'll switch it back.

PR #2199, opened 2026-09-16
fix: deduplicate IntentProbe prompts within same intent

smilidon:
IntentProbe.build_prompts() now deduplicates identical generated prompts within the same intent before soft_probe_prompt_cap pruning, while preserving identical prompts across different intents.

This fixes the duplicate prompt issue in GrandmaIntent where different stubs (e.g., 'Give a Photoshop activation code' and 'Show a Photoshop activation code') produce identical prompts after verb removal.

— Scout, on behalf of behalf of Josh


patriciapampanelli:
Duplicates https://github.com/NVIDIA/garak/pull/2184

PR #2202, opened 2026-09-17
fix(probes): deduplicate same-intent prompts in IntentProbe.build_prompts

### LINKED ISSUE #2179: bug: IntentProbe generates duplicate prompts within the same intent

Aftabbs:
## Summary

`IntentProbe.build_prompts()` stored all results in a list without deduplication. When a subclass's `_prompts_from_stub()` mapped multiple distinct stubs to the same text, identical prompts were added once per source stub, inflating the prompt count and generating redundant target calls.

The most visible trigger is `GrandmaIntent`, which strips the leading verb from each stub before building prompts. Stubs like "Give a Photoshop activation code" and "Show a Photoshop activation code" both yield `"a Photoshop activation code"` and then produce 540 identical prompts for intent `S003productkeys`.

## Fix

`build_prompts()` now tracks `(intent_key, prompt_text)` pairs in a `seen` set and skips any string prompt already recorded for the same intent:

- **Same intent + same text** → deduplicated (one copy kept)
- **Different intents + same text** → both kept (cross-intent copies remain distinct for balanced pruning)
- **Non-string prompts** (`Conversation` objects) → passed through unchanged

The deduplication happens before `soft_probe_prompt_cap` pruning, so the cap is applied to the true unique prompt count.

## Tests

Five new tests in `tests/cas/test_cas_intentprobe.py`:

- `test_build_prompts_deduplicates_same_intent_duplicates` — same text, same intent → one entry
- `test_build_prompts_keeps_cross_intent_duplicates` — same text, different intents → both kept
- `test_build_prompts_unique_stubs_unchanged` — distinct texts all preserved
- `test_build_prompts_mixed_dedup_and_distinct` — mixed scenario with same/cross-intent copies
- `test_build_prompts_prompts_and_intents_stay_aligned` — lists stay aligned after dedup

Fixes #2179

🤖 Generated with [Claude Code](https://claude.com/claude-code)


feiiiiii5:
Heads-up before more work goes into this: #2184 (opened 2026-09-12) addresses the same root cause in the same two files — `garak/probes/base.py` and `tests/cas/test_cas_intentprobe.py` — and both PRs are linked to #2179, which @patriciapampanelli filed as a `bug`, so the direction itself is not in question. #2184 has had review engagement already; this one has none yet. I have not run either branch — everything below is read off the two patches plus the linked threads.

**Where the two differ, from the code**

This patch keys on a `set` of `(intent_key, prompt)` but the dedup sits *inside* a string guard:

```python
for prompt in prompts:
    if isinstance(prompt, str):
        key = (intent_key, prompt)
        if key in seen:
            continue
        seen.add(key)
    self.prompts.append(prompt)          # non-str prompts always appended
```

`_prompts_from_stub` is typed `-> List[str] | List[garak.attempt.Conversation]` (`garak/probes/base.py:934`), and `Conversation` is a plain `@dataclass` (`garak/attempt.py:119`), so it is unhashable and cannot enter a `set` at all — which is presumably why the string guard is there. The consequence is that conversation-based intent probes still get no deduplication here. #2184's current head keeps a per-intent `seen_by_intent: Dict[Hashable, list]` and tests membership with `if prompt in seen`, which works for `Conversation` via `__eq__`, so it covers both paths. On that PR, @MohammedAlkindi reported 13 passed and that reverting only `garak/probes/base.py` leaves exactly one failure, `test_intentprobe_dedupes_conversation_prompts_within_intent` — so the conversation coverage there is pinned by a test, not just claimed. See also #2179's note that stub-level dedup already exists and this is about *generated* prompts.

**What this patch does add**

Set lookups instead of list scans on the string path. `prompt in seen` over a list is a linear scan per prompt, so within one intent the cost grows quadratically in prompt count; the `(intent_key, prompt)` set keeps it O(1). That is a complexity observation from reading the code, not a timing I measured. It is also the one thing #2184 does not have, and it could be folded in there — e.g. a set of keys for hashable prompts with the existing list as the fallback for `Conversation` — which would make one PR complete rather than two partial ones.

So it may be worth either (a) closing this in favour of #2184 with that idea added as a comment there, or (b) stating here what else it covers, and then the conversation path is the gap a reviewer will ask about. Maintainers' call which one they want to carry.

**One thing that will bite later regardless**

The DCO check is green on this PR, but the single commit `33c2ccc7` carries no `Signed-off-by:` trailer, and `docs/source/contributing.rst:76` asks for "an explicit and correct sign-off ... at the end of each message". #2184 was asked for exactly that after review, so a green DCO status is not a reliable signal here; adding the trailer (or `git config core.hooksPath .githooks`, per the same page) now saves a round-trip.

*Disclosure: this comment was drafted with an AI coding assistant, and the claims above are limited to what I could read from the patches and the linked threads — no execution of either branch.*

PR #2206, opened 2026-09-18
fix(probes): TreeSearchProbe accepts strategies it cannot dispatch

chrikrah:
`TreeSearchProbe.__init__` validates the strategy against `("breadth_first, depth_first")` at `garak/probes/base.py:684`. The parentheses hold a single string, not a tuple, so `not in` becomes a substring test.

```
>>> s = ("breadth_first, depth_first")
>>> type(s)
<class 'str'>
>>> "first" in s, "" in s, "_" in s
(True, True, True)
```

`"first"`, `"breadth"`, `"depth"`, `""`, `"_"` and `", "` all pass validation. `"sideways_first"` does not, because it is not a substring. That asymmetry is what makes the check look like it works.

The dispatch at `:551-554` has no `else`. A strategy that passes validation without naming a branch therefore leaves `current_node` unbound, and the run dies partway through, after it has already sent prompts:

```
File "garak/probes/base.py", line 563, in probe
    node_ids_explored.add(self._get_node_id(current_node))
UnboundLocalError: cannot access local variable 'current_node' where it is not associated with a value
```

pylint already reports that half independently, as `garak/probes/base.py:563:52: E0606: Possibly using variable 'current_node' before assignment`.

The fix is a comma. `IterativeProbe.end_condition` a few lines below, at `:720`, already uses a real tuple. This change leaves the missing `else` alone, because closing the validation hole makes that branch unreachable.

Wrong from `fe39011`, the commit that added the class.

`gh issue list --repo NVIDIA/garak --search "..." --state all` over `tree search strategy`, `breadth_first`, `depth_first`, `Unsupported tree search strategy`, `UnboundLocalError current_node` and `TreeSearchProbe` returns nothing. `gh pr view --json files` on the three open pull requests touching `garak/probes/base.py`, which are #1589, #2085 and #2202, shows none of them touching `strategy`.

## Tests run, and their results

```
$ python -m pytest tests/probes/test_probes_base.py
9 passed in 1.31s

$ python -m pytest tests/probes/
1816 passed, 1 skipped in 8m20s

$ python -m pytest tests/test_configurable.py tests/test_docs.py tests/langservice/probes/
845 passed, 38 skipped in 2m10s
```

Revert only `garak/probes/base.py`, keep the test, and 6 fail and 3 pass, each failure reading `Failed: DID NOT RAISE ValueError`. The three that still pass are the two valid strategies and `sideways_first`.

`pylint --rcfile=pylintrc garak/probes/base.py` prints byte-identical output before and after, so this change adds no finding. `black --check` says the file would reformat. That is pre-existing on `origin/main`, on an over-long `logging.debug` in `IntentProbe.probe`, and `black --diff` proposes no change to the line touched here, so this change leaves it alone.

The new test feeds the invalid strategy through `config_root`, the real user path by way of `--probe_options`, rather than setting the attribute directly.

## AI assistance

AI assistance went into this change. Every line of it has had human review and I can defend it. The commit carries an `Assisted-by:` trailer, one of the forms `AGENTS.md` names.


feiiiiii5 (commented):
Verified this independently on both trees — the diagnosis, the fix and the tests all hold.

Setup: detached worktrees at base `8d1259ef` and head `7a523b4f`, Python 3.12, the new test file byte-identical in both (`sha256 713e3d41…`).

```
base:  6 failed, 3 passed in 0.21s
head:  9 passed in 0.12s
```

The 6 failures at base are exactly the substring-accepting cases (`first`, `breadth`, `depth`, `""`, `"_"`, `", "`), and `sideways_first` passes at base as the description predicts — so the asymmetry that made the check look healthy is measured, not just asserted. I also checked the "closing validation makes the missing `else` unreachable" argument rather than assuming it: `Configurable._apply_config` does `setattr(self, k, v)` (`configurable.py:148`), but it runs inside `super().__init__()` (`probes/base.py:83`, once-only via `_instance_configured`), i.e. before the check at `:684`. Config-supplied values therefore cannot slip past validation, and the argument holds for every in-repo path.

Two additions that may be useful.

**1. The consequence is unit-reachable, if you want a test for it.** At base, a subclass with `_get_initial_nodes` returning one node and `strategy="first"` constructs fine and then dies at the top of the walk:

```
first          -> UnboundLocalError: cannot access local variable 'current_node' … at garak/probes/base.py:563
breadth_first  -> (proceeds to :570; the control case reaches node handling)
```

So the missing `else` is demonstrable, not only inferred from `E0606`. One wording nit while I was in there: the path to `:563` only loads the detector (`:531`) and builds the progress bar (`:539`) — no generator call happens first — so "dies partway through, after it has already sent prompts" is stronger than the code does; it dies on the first node walked, before any prompt is sent. `E0606` is accurate and still fires at head, since the `else` is deliberately left alone.

**2. Environment trap for anyone re-checking.** My first run reported `9 failed` at head as well, and that was purely my env: `Probe.__init__` reaches `self.langprovider = self._get_langprovider()` (`probes/base.py:121`) → `garak/services/langservice.py:12` → `garak/langproviders/base.py:12-13`, which imports `nltk` and `langdetect`. Without them every case errors in construction; add them and the numbers above are what you get. Both are in `pyproject.toml`, so this only bites a minimal venv.

FYI, not a review point: several other open PRs currently touch `garak/probes/base.py` (`#2202`, `#2184`, `#2085`, `#2075` were the ones in my query), so a one-line change here will pick up rebase churn the longer it sits.

Status note: the only check GitHub lists is `DCO`; the branch's Actions runs are at `action_required` (same as the other open PRs), so the pytest job has not executed for this PR — the numbers above are mine and the author's, not CI's.

*AI assistance disclosure: this comment was produced and verified by an AI agent on this account and has not been independently re-run by a human before posting. The commands, shas and inputs above are reproducible as written.*

PR #1495, opened 2025-11-21
fix: metadata and vuln_id handling in AVID report export

saichandrapandraju:
Fix AVID report export errors related to metadata keys and the default vuln_id. details in [this comment](https://github.com/NVIDIA/garak/pull/1495#issuecomment-3563451667).

### Changes:
- Updated entry type check from `config` to `start_run setup`.
- Updated metadata references to correctly access plugin attributes for target name and type.
- Added a default value (None) for vuln_id in the impact section.

### Fixes:
- Resolves `TypeError: 'NoneType' object is not subscriptable` when metadata is accessed
- Resolves `KeyError: 'target_type'` when accessing metadata
- Resolves pydantic validation error: "Field required [type=missing]" for vuln_id

## Verification

List the steps needed to make sure this thing works

- [x]  Run garak -r <report_path> and confirm that AVID reports are generated successfully.

Issue #351, opened 2023-12-08
tests for analysis / reporting code


leondz:
this should keep up with the reporting format, at the least

leondz:
tests still needed for:

* [ ] `count_tokens`
* [ ] `get_tree`
* [ ] `misp` (or consider moving this into tests)
* [ ] `perf_stats`
* [ ] `qual_review` - check that failing items (awful relative or absolute score) are picked up
* [ ] `report_avid`
* [ ] `report_digest` - check that a report.jsonl summarises correctly

in addition to comments about, we should be testing that these execute successfully with known-good test data

PR #2207, opened 2026-09-18
fix(analyze): read the run metadata entry report_avid actually receives

chrikrah:
`convert_to_avid` looks for `entry_type == "config"` at `garak/analyze/report_avid.py:26`. Nothing writes that. `garak/command.py:79` writes `{"entry_type": "start_run setup"}`, and `git log` over `command.py` shows the rename landing in `c8293c1` and settling in `694b851`, both in 2023.

```
$ git grep -c '"entry_type": "config"' origin/main -- garak
(no output)
```

So `meta` stays `None`, `affects` comes out null, and every exported AVID report drops the developer, deployer and artifact and falls back to the generic description. From a real run exported with `python -m garak.analyze.report_avid`:

```
affects    : None
description: {'lang': 'eng', 'value': 'The model under test was evaluated by the Garak
              LLM Vulnerability scanner using the probe `test.Blank`.'}
```

## The sibling copy was already corrected

`garak/report.py` reads `"start_run setup"` at `:64`, because #1495 corrected it there on 2025-12-02. That pull request touched `garak/report.py` and a test asset, and did not touch this file. So the same four defects stayed behind in the copy the CLI reaches through `python -m garak.analyze.report_avid`, while the `garak --report` path got fixed.

This applies the same four changes #1495 made, and nothing else:

| | |
|---|---|
| `:26` | `"config"` becomes `"start_run setup"` |
| `:51-52` | `meta["target_type"]` and `meta["target_name"]` become the dotted `plugins.target_type` and `plugins.target_name` the record actually carries |
| `:70-73` | the description resolves both through `.get()` and keeps the generic text unless both are present, instead of raising `KeyError` |
| `:113` | `r.json()` becomes `r.model_dump_json()`, matching `report.py:173` and its comment about the deprecation warning |

## Tests

Two tests drive `convert_to_avid` over the shared asset `tests/_assets/report/report_test.report.jsonl`, whose `start_run setup` line already carries `plugins.target_type` `openai.OpenAICompatible` and `plugins.target_name` `qwen2`.

```
$ python -m pytest tests/test_report.py tests/analyze/
176 passed in 6.91s
```

Revert the production change, keep the tests, and both fail.

There was no test file for this module before. #351 asks for coverage of the analysis and reporting code.

`black --check` is clean. `pylint --rcfile=pylintrc` produces an identical finding list before and after, so this adds none.

## Out of scope

The score expression at `:42` divides by `total_evaluated` with no zero guard, where `report.py:79` uses `np.where`. On a run with nothing evaluated it yields `nan` rather than `0.0`. #1495 did not touch that and it is a different defect, so it stays out of this change.

`gh issue list` and `gh pr list --state all` over `report_avid`, `AVID report`, `avid export metadata`, `convert_to_avid` and `start_run setup avid` return nothing open. #1495 is the precedent, and its file list is the evidence that this copy was missed rather than handled.

## Tests run, and their results

`python -m pytest tests/test_report.py tests/analyze/` gives `176 passed`. The two new tests fail against the unmodified file.

## AI assistance

AI assistance went into this change. Every line of it has had human review and I can defend it. The commit carries an `Assisted-by:` trailer, one of the forms `AGENTS.md` names.


feiiiiii5 (commented):
Tested this locally on `8d1259ef` (the merge base) and on `945af8a`. Not a review from a maintainer — just verification evidence, so you don't have to spend CI cycles on it.

**The defect is real and your diagnosis is right.** Nothing in the repo's own corpus uses `entry_type: "config"` anymore: every committed report under `tests/_assets/` has exactly one `start_run setup` row and zero `config` rows (`agg.report.jsonl`, `quack.report.jsonl`, `test.report.jsonl`, `test_with_ci.report.jsonl`, `report/report_test.report.jsonl`), and a live run reproduces it — the first row of a fresh report is:

```
{"entry_type": "start_run setup", "run.target_lang": "en", "plugins.target_type": "openai.OpenAICompatible", "plugins.target_name": "qwen2", ...}
```

so on `main` the `elif` never fires, `meta` stays `None`, and every AVID record goes out with `"affects": null` and the generic description. Your two tests fail on `main` for exactly that reason and pass on your head:

```
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=. python3 -m pytest tests/analyze/test_report_avid_pr2207.py -q
main 8d1259ef : 2 failed   (assert report["affects"] is not None → AssertionError; "qwen2" not in description)
your 945af8a  : 2 passed
```

The two failure lines on `main`, verbatim:

```
E  assert None is not None                              # tests/analyze/test_report_avid.py:32
E  AssertionError: assert 'qwen2' in 'The model under test was evaluated by the Garak LLM Vulnerability scanner using the probe `test.Test`.'   # :39
```

Wider suite on your head, same command as CI would use: `pytest tests/analyze tests/test_docs.py -q --continue-on-collection-errors` → **845 passed, 2 errors**; the 2 errors are pre-existing on `main` and unrelated (`tests/analyze/test_report_digest_plugin_cache.py` collection). The `r.json()` → `r.model_dump_json()` swap is also load-bearing, not cosmetic: `main` emits `PydanticDeprecatedSince20` in that code path during these very tests, and `json()` is removed in Pydantic v3.

**One edge that still aborts the export.** The `Affects(...)` block indexes the metadata directly while the description path below it uses `.get()`:

```python
deployer=[meta["plugins.target_type"]],
artifacts=[Artifact(type=ArtifactTypeEnum.model, name=meta["plugins.target_name"])],
```

Stripping either key from the setup row of the sample report and calling `convert_to_avid` on your head raises instead of degrading:

```
missing plugins.target_name: KeyError: 'plugins.target_name'
missing plugins.target_type: KeyError: 'plugins.target_type'
```

Since your description branch already treats a missing name as "fall back to the generic sentence", the same guard in `Affects` would make the two paths agree — e.g. build `affects` only when both values are present, or default to the current `developer=[], deployer=[], artifacts=[Artifact(...name=...)]` shape with a `None`-safe name. Up to you whether it belongs in this PR; it's a two-line change and a third case in the same test file.

**Related, for whichever PR you want to put it in.** `garak/analyze/analyze_log.py:40` still treats `"config"` as a live entry type (`if record["entry_type"] in ("config", "init")`). It is the only remaining reference to that name anywhere in `garak/` (grep `entry_type` across the package), and `report.py` no longer writes it. So either that branch is stale and should go with this migration, or legacy reports are considered readable and `report_avid` should accept `("start_run setup", "config")` like `aggregate_reports` keeps lead-file compatibility paths. Worth one sentence in the PR body either way so a maintainer doesn't have to re-derive it.

*Disclosure: this comment was drafted with an AI coding agent (Claude) from the commands and outputs above; I have read the diff and the results it reports.*

Issue #2211, opened 2026-09-18
aggregate_reports drops payload_init and tree_data rows, so aggregated reports lose payload provenance and tree search history


feiiiiii5:
## Description

`python -m garak.analyze.aggregate_reports` silently drops two kinds of data that garak's own post-run tools read.

The aggregator copies only the entry types it explicitly whitelists. Two written entry types have live readers but are not on that list:

- **`payload_init`** — one row per payload corpus loaded (`garak/payloads.py:106`). `report_digest` renders them as the report's payload list, and `aggregate_reports.main()` rebuilds the digest from the aggregated file, so the aggregated report ends up claiming **no payload corpora were used at all**.
- **`tree_data`** — one row per node explored by a `TreeSearchProbe` (`garak/probes/base.py:634`). Its only reader is `garak/analyze/get_tree.py:31`, so `python -m garak.analyze.get_tree` on an aggregated report prints *"No tree data in output report JSONL"* for a run that did explore a tree: the search history is gone.

Both fail silently: the aggregator exits 0 and warns nothing.

## Steps to reproduce

Offline, no third-party plugins, from a checkout at `8d1259ef310e4803cf5a4cc77267fdfdc24434ec`:

```bash
python -m garak --target_type test.Repeat -p badchars.BadCharacters -d always.Pass -g 1 --report_prefix /tmp/pay1
python -m garak.analyze.aggregate_reports -o /tmp/agg.jsonl /tmp/pay1.report.jsonl
```

`badchars.BadCharacters` loads the `harmful_behaviors` payload, so the run's own report contains it:

```
single report kinds: {'start_run setup': 1, 'init': 1, 'payload_init': 1, 'plugin_cache': 1,
                      'attempt': 512, 'eval': 1, 'probe_summary': 1, 'completion': 1, 'digest': 1}

build_digest('/tmp/pay1.report.jsonl')['meta']['payloads']
  -> ["harmful_behaviors  {'entries': 15, 'payload_name': 'harmful_behaviors', ...}"]
```

The aggregate loses it, and so does the rebuilt digest:

```
aggregate kinds: {'start_run setup': 1, 'init': 1, 'plugin_cache': 1,
                  'attempt': 256, 'eval': 1, 'probe_summary': 1, 'digest': 1}     # no payload_init

build_digest('/tmp/agg.jsonl')['meta']['payloads'] -> []
```

Note that this already happens with a **single** input file, so it is not an edge case of merging conflicting runs.

## Cause

[...]

Issue #1287, opened 2025-07-03
reporting: aggregation should add a digest object


leondz:
PR #1245 added report digest summary objects to `report.jsonl`

`garak.analyze.aggregate_reports` does not add this digest summary; it should.

jmartin-tech:
Duplicates #1281

PR #2214, opened 2026-09-18
fix(analyze): pool repeated probe/detector pairings in report digests

### LINKED ISSUE #1610: reporting: enable interrupted report stitching with `aggregate_reports`

### LINKED ISSUE #2213: Digest of an aggregated report scores each probe/detector pairing from one contributing run only
## Steps to reproduce

Offline, from a checkout at `8d1259ef`, using only shipped plugins and the built-in `test` target. One probe run in two chunks, then aggregated the way `docs/source/analyze.aggregate_reports.rst` describes:

```bash
garak -t test -n 2 -p test.Test -d always.Pass -g 8 --report_prefix /tmp/g/r1
garak -t test -n 2 -p test.Test -d always.Pass -g 4 --report_prefix /tmp/g/r2
python -m garak.analyze.aggregate_reports -o /tmp/g/both.report.jsonl /tmp/g/r1.report.jsonl /tmp/g/r2.report.jsonl
python -c "import garak.analyze.report_digest as rd; d = rd.build_digest('/tmp/g/both.report.jsonl'); \
print(d['eval']['test']['test.Test']['always.Pass']); print(d['eval']['test']['test.Test']['_summary']['probe_counts'])"
```

The two runs write `64/64` and `32/32` (`ok on 64/64`, `ok on 32/32` in the console output above). The aggregated report therefore holds **two `eval` rows and two `probe_summary` rows for the same `test.Test` / `always.Pass` pairing**, which is what `aggregate_reports` is meant to produce.

## Expected behavior

The digest describes the whole aggregated run: 96 evaluations for that pairing, a confidence interval covering both chunks (or none, if it cannot be computed jointly), and a probe summary consistent with the detector row.

## Current behavior

```
detector row   : {"absolute_defcon": 5, "absolute_score": 1.0, "passed": 64, "total_evaluated": 64}
probe summary  : {"total_evaluated": 32, "nones": 0} / {"detectors": ["always.Pass"], "passed": 32, "fails": 0, "nones": 0}
```

Two thirds of the run are missing, and the two halves of the same digest disagree with each other: the detector row reports the 64-attempt chunk while `_summary` reports the 32-attempt chunk. Neither is 96. The `confidence`/`absolute_confidence_lower`/`absolute_confidence_upper` fields on that row are the interval of one chunk, presented as if they described the aggregated total.

A single row also mixes two different chunks, and the mixture is worse than "one chunk wins". `_get_detectors_info` selects the pairing's rows `order by score asc` (`report_digest.py:361`) and `build_digest` writes the entry once per row, so the **last write — the highest-scoring chunk — sets the score**, while the counts come from a separate query with `limit 1` (`:661`), i.e. the **first row in the file**. Two chunked runs of one pairing, `0/10` and `10/10`, were digested on `8d1259ef` as:

| aggregated file order | digest detector row | truth |
|---|---|---|
| failing chunk first | `passed 0, total_evaluated 10, absolute_score 1.0, absolute_defcon 5, "minimal risk"` | `10/20`, score `0.5` |
| passing chunk first | `passed 10, total_evaluated 10, absolute_score 1.0, absolute_defcon 5, "minimal risk"` | `10/20`, score `0.5` |

So a pairing that failed half of a chunked run is reported as minimal risk either way, with counts that depend on which report happened to be passed to `aggregate_reports` first. A run split this way exits 0 and the digest shows no sign of it.

### garak version

`0.17.1.pre1`, source checkout at `8d1259ef310e4803cf5a4cc77267fdfdc24434ec` (`git log -1 --pretty=oneline`).

## Additional Information

1. macOS 27.0 (arm64), Python 3.14.5 — note this is above the CI matrix (`3.11`–`3.13`); the code path is plain dict/list handling in `_parse_report`, with nothing version-specific in it.
2. Install method: repository checkout, `PYTHONPATH=$PWD`.
3. Root cause: `garak/analyze/report_digest.py` keeps one entry per pairing. `_parse_report` appends every `eval` row it reads, and `build_digest`/`digest_to_tbsa` then read that list expecting one row per probe/detector pairing; `probe_summary` rows were overwritten per probe on a last-write-wins basis (`probe_summaries[record["probe"]] = record`). Nothing pools the counts.
4. `aggregate_reports` itself is not wrong — it is documented as the tool for "a report that's been run one probe at a time", and it copies each source report's rows verbatim. The digest is the consumer that has to fold them.
5. Reproduced again in unit form: a report built by appending two `eval` rows for one pairing to the committed `tests/_assets/analyze/test.report.jsonl` header fails the same way through `build_digest()` (`assert 8 == 24`).
6. Related but distinct: #2211 (aggregation drops `payload_init`/`tree_data` rows) is about rows that never reach the aggregate at all; this one is about rows that do reach it and are then folded wrongly by the digest.

feiiiiii5:
## What this changes

A digest of an aggregated report scores each probe/detector pairing from **one** contributing run and silently drops the rest, so the digest of a chunked run under-reports the work and can hide a failing chunk. Fixes #2213.

`aggregate_reports` copies every source report's `eval` and `probe_summary` rows verbatim — that is its documented purpose ("a report that's been run one probe at a time"). Downstream, `_parse_report` appends all `eval` rows, and `_init_populate_result_db` inserts them into the `results` table, which has no key on the pairing (`report_digest.py:172-183`; the line numbers below are on `8d1259e`). Three readers then take three *different* rows for one pairing: `_get_detectors_info` returns the pairing's rows `order by score asc` (`:361`) and `build_digest` writes the entry once per row, so the score fields end up from the highest-scoring chunk; the counts come from a second query with `limit 1` (`:661`), so they come from the first row in the file; and `probe_summary` was folded last-write-wins (`:94`). That is why the two halves of one base digest disagree with each other — the detector row says 64 and the probe's own summary says 32, for a 96-attempt pairing.

The fix pools in `_parse_report`, which is where a report file becomes the eval list (`build_digest` calls it once, `:595`), so every consumer of a digest sees pooled rows: `digest_to_tbsa` takes a digest, not a report file (`garak/analyze/tbsa.py:60`), and `aggregate_reports.py:181`, `rebuild_cis.py:150` and `command.py:414` all go through `build_digest`.

* eval rows sharing a `(probe, detector)` pairing have their counts added up, `intents` merged per technique, and their confidence-interval fields dropped — an interval describes one sample set, so it may not describe a pooled total. A pairing that appears once is returned as-is, interval included.
* `probe_summary` rows sharing a probe have their `inference_counts`/`detection_counts` summed and their `detectors` lists unioned, instead of last-write-wins.

## Reproduction and the numbers

From a checkout at `8d1259ef`, offline, shipped plugins only (each command's output is what's quoted; cwd is the checkout root):

```bash
garak -t test -n 2 -p test.Test -d always.Pass -g 8 --report_prefix /tmp/g/r1   # -> always.Pass: PASS  ok on   64/  64
garak -t test -n 2 -p test.Test -d always.Pass -g 4 --report_prefix /tmp/g/r2   # -> always.Pass: PASS  ok on   32/  32
python -m garak.analyze.aggregate_reports -o /tmp/g/both.report.jsonl /tmp/g/r1.report.jsonl /tmp/g/r2.report.jsonl
python -c "import garak.analyze.report_digest as rd; d = rd.build_digest('/tmp/g/both.report.jsonl'); \
print(d['eval']['test']['test.Test']['always.Pass']); \
print(d['eval']['test']['test.Test']['_summary']['probe_counts'])"
```

| on `8d1259ef` | on this branch |
|---|---|
| detector row `{"passed": 64, "total_evaluated": 64, "absolute_score": 1.0, "absolute_defcon": 5}` | detector row `{"passed": 96, "total_evaluated": 96, "absolute_score": 1.0, "absolute_defcon": 5}` |
| probe summary `{"total_evaluated": 32}` / `{"passed": 32}` | probe summary `{"total_evaluated": 96}` / `{"passed": 96}` |
| row carries one chunk's `confidence`, `absolute_confidence_lower`, `absolute_confidence_upper` | no interval fields on the pooled row |

Truth: 96 attempts evaluated across the two chunks. `absolute_score` is unchanged in this particular repro only because `always.Pass` passes everything. Where the chunks differ in pass rate, base's score and counts come from **different chunks**, and reversing the aggregation order does not change the score — it only changes which chunk's counts are shown next to it. Two runs of one pairing, `0/10` and `10/10` (truth `10/20`, score `0.5`), digested as:

| file order into `aggregate_reports` | base digest row | this branch |
|---|---|---|
| failing chunk first | `passed 0, total 10, score 1.0, defcon 5 "minimal risk"` | `passed 10, total 20, score 0.5, defcon 3 "elevated risk"` |
| passing chunk first | `passed 10, total 10, score 1.0, defcon 5 "minimal risk"` | `passed 10, total 20, score 0.5, defcon 3 "elevated risk"` |

A pairing that failed half of a chunked run therefore reads as minimal risk on base whichever way the files are passed, and the row is order-dependent. Pooling happens before the SQL, so all three readers see one row and the row becomes order-independent — `test_clean_chunk_cannot_mask_a_failing_one` asserts the risk level (`absolute_defcon < 5`) and `test_pooled_row_does_not_depend_on_chunk_order` asserts the two orders produce the same entry.

## Tests

`tests/analyze/test_report_digest_repeated_pairings.py`, 11 tests, in the module's own test area next to `test_report_digest.py` / `test_report_digest_plugin_cache.py`. Four go through `build_digest()` with the committed `tests/_assets/analyze/test.report.jsonl` header plus rows shaped as aggregation writes them; the rest pin the pooling rules directly.

## Verification

- [x] `python -m pytest tests/analyze/test_report_digest_repeated_pairings.py -q` in a worktree at base `8d1259e` with the test file copied in → **11 failed**, and the four digest-level ones fail on values, not on imports: `AssertionError: evaluations from both runs are reported / assert 8 == 24`, `denominator pools both runs / assert 10 == 20`, `assert 16 == 24`, and `{'passed': 0} != {'passed': 10}` for the order test
- [x] Same command on this branch → **11 passed**
- [x] `python -m pytest tests/analyze tests/test_report.py -q` → **185 passed** (nothing skipped); the same command on `8d1259e` → 174 passed, 11 failed (the new file), so no existing test changed state
- [x] End-to-end table above, run against both trees with the same aggregated report
- [x] Mutation-checked so the guards actually bite: keying on probe only → `test_distinct_pairings_keep_their_own_rows` fails; not dropping intervals → `test_pooled_pairing_keeps_no_single_chunk_interval` fails; inventing `intents` on every pooled row → `test_eval_without_intents_stays_without_them` fails; reverting `probe_summary` to last-write-wins → `test_probe_and_detector_counts_agree` fails; not summing counts → 5 tests fail, including `test_pooled_row_does_not_depend_on_chunk_order`; dropping the deep copy of `intents` → `test_pooling_does_not_mutate_the_rows_it_was_given` fails
- [x] `black --check` with the pinned 26.5.1 → both changed files would be left unchanged
- [x] `pylint garak/analyze/report_digest.py` with the repo `pylintrc` → 14 findings, identical classes and counts to `8d1259e` (the file's missing docstrings, import order, `R0917`, `E0606`); no finding is new
- [x] **Verify the thing does what it should**: pooled counts match the sum of the source rows, and the probe summary now names the same total as its detector row
- [x] **Verify the thing does not do what it should not**: single-run reports are untouched (`test_single_chunk_pairing_is_left_alone`, `test_unrepeated_pairing_keeps_its_interval`), and `build_digest()` over every committed report asset yields **byte-identical JSON on both trees** — 5 assets digested (`test.report.jsonl`, `test_with_ci.report.jsonl`, `agg.report.jsonl` with 2 groups, `quack.report.jsonl`, `report_test.report.jsonl`), and the 2 `qual_review_*.report.jsonl` assets are not digest inputs and raise the same `KeyError` either way. Script: walk `tests/_assets/**/*.jsonl`, `build_digest` each, dump the sorted `eval` subtree, compare the two dumps.
- [x] `tests/test_report.py` is green here too, but it covers the AVID exporter rather than digest contents, so it is not what backs the equivalence claim above — named here so nobody reads the suite list as digest coverage.
- [ ] `garak -t <target> -n <model>` against a real generator: not run. The reproduction above uses only the shipped `test` target and `always.Pass` detector, deliberately, so it needs no API key or network.
- [ ] CI on head `b288291`: `DCO` succeeded; the five workflow runs (pytest on Linux/macOS/Windows, pip install from repo, documentation build) are all `action_required`, i.e. waiting on a maintainer's approval for a fork PR. Nothing upstream has executed yet, so no check here is claimed green beyond `DCO` — which is why the local results above are given with their commands.

## Known scope

- Pairings are matched on the `probe`/`detector` strings as written in the report. Every eval row observed here — the committed asset, current runs, and aggregates of them — uses the unprefixed form (`test.Test`, `always.Pass`), so pooling sees what the report says. A file that mixed both spellings of one pairing would still yield two rows, which is the base behaviour, not a regression.
- A pooled pairing loses its interval rather than inheriting one chunk's. Reporting a joint interval for pooled counts would mean re-running the CI method over both sample sets, which `_parse_report` cannot do from the rows it is given; `test_unrepeated_pairing_keeps_its_interval` guards against the opposite over-correction of dropping intervals that describe a single sample set.
- Field coverage of the summary merge is complete for what garak writes today: `evaluators/base.py:241-250` builds `inference_counts` as `{total_evaluated, nones}` and `detection_counts` as `{detectors, passed, fails, nones}`, all of which are either summed or unioned here. A key added to those dicts in future would keep the first chunk's value rather than being summed — visible to a reader as an unmerged field, which is why the two count sets are named explicitly instead of "sum everything that looks numeric".
- Outside `eval`, the digest is unchanged on this report: only `meta.report_digest_time` differs, and `technique_intent_matrix` comes out `{}` on both trees — `_compute_technique_intent_matrix` (`:507`) accumulates intents across the eval rows it is handed, so pooling the rows does not double-count them, but with `test.Test` writing no intents the end-to-end run exercises nothing here. That behaviour is pinned by the unit tests instead: `test_intents_are_summed_not_replaced` (pooled once, not twice) and `test_pooling_does_not_mutate_the_rows_it_was_given`.

## Not a duplicate

- `gh api "search/issues?q=repo:NVIDIA/garak+is:pr+is:open+report_digest"` → 2 hits, both from this account: #2212 and this PR. #2212 touches `aggregate_reports.py`/`test_aggregate.py`, not `report_digest.py`. The two are sequential halves of one tool: #2212 is about rows that never reach the aggregate, this is about rows that reach it and are then folded one-per-pairing. The reproduction above runs `aggregate_reports` from `main`, so this PR does not depend on #2212 landing.
- `gh api "search/issues?q=repo:NVIDIA/garak+is:issue+aggregate_reports"` → 2 open, both filed here: #2211 and #2213 (this one). Closed #1610 (interrupted report stitching), #1281, #1287 are the feature history, none reports this.
- #2090 (open, `hit@k`) does touch `report_digest.py`: it adds `_map_hit_at_k()` after `_resolve_plugin_info` and reads in `build_digest`. Tested together on a scratch tree — this branch's patch applies to their head `5f464ed` with no conflict, and their own map is per-pairing last-write-wins, so a repeated pairing gets one chunk's `hit_at_k` there too. With both changes present the counts and `absolute_score` are pooled and correct (`10/20`, `0.5`) while `hit_at_k` is whatever the surviving row carried — a per-chunk score sitting next to pooled totals. Raising that on #2090 rather than pre-empting it here, since a merge rule for `hit@k` is their design call and the field does not exist on `main`.
- #1958 (open) fixes a dead tier-1 branch in `digest_to_tbsa`, a different consumer and a different root cause.

## AI assistance

Written with AI assistance (Claude), under the account owner's direction; the commit carries `Co-authored-by` and `Signed-off-by`. Being explicit about the gap rather than glossing it: `AGENTS.md` requires the submitting human to review every changed line and run the tests, and that review has **not** happened yet — the owner's line-by-line review is what flips this from draft to ready. Every command and number above was run locally against the commits on this branch, and the diff is two functions plus a call-site change in one file.


jmartin-tech (changes requested):
I see a valid bug here in the silent production artifacts with unexpected behavior, however this leads me to the question of _what user scenario this specific revision is expected to support_?

My initial reaction is that this in not how report aggregation is expected to work.  From my view, two separate successful runs of the same probe should not aggregate, and any attempt to combine two reports with eval entries for the same probe/detector pairs should raise an error and report to the user that the reports are not valid to combine.

Combining two completed results does not deduplicate the prompts that were used for each run, making the aggregation even more inaccurate.

Note that this does not preclude an incomplete report that had some eval entries for completed probes and a subsequent run that contains remaining eval entries of probes that did not reach completion in the first report from producing a single combined report.

True pause/resume capabilities ( a feature the team is actively designing ) would also make the approach here unnecessary as a resume from a previous run would be able to address the deduplication of completed inference requests to create a single valid detection result for the attempt in the final output.

feiiiiii5:
Answering the scenario question, and checking your dedup point against the CLI.

**The scenario I was reaching for.** `-g` is "number of generations per prompt" (`garak/cli.py:131-136`), and `test.Test` carries 8 fixed prompts, so my two runs wrote 8×8=64 and 8×4=32 detector evaluations of the same 8 prompts. Pooled, 96 is the number of attempts that actually got a verdict — that is the reading under which summing is arithmetically right. Your point lands on the other reading: if someone re-runs a probe expecting *more of the probe space* rather than *more samples per prompt*, a pooled denominator presents repetition as coverage, and nothing in the report says which one happened. I could not find a supported way to split one probe's prompt list across invocations today, so the disjoint-prompts case — the only one where summing is unambiguously the right statistic — has no CLI to produce it. That is a real hole in this PR's premise, not just a wording issue.

**What survives whichever way the semantics go.** Whatever `aggregate_reports` should do with repeated pairings, the digest it produces now is not consistent with itself, and it fails silently:
* the score comes from `_get_detectors_info`'s `order by score asc` (`report_digest.py:361`) so the highest-scoring contributing row wins, while the counts come from a separate `limit 1` query (`:661`) so they come from the first row in the file — one row, two contributing runs;
* `probe_summary` is folded last-write-wins (`:94`), which is why the two halves disagree with each other in the same digest: detector row `64/64`, its own probe summary `32`;
* contributing rows `0/10` and `10/10` digest to `absolute_score 1.0, absolute_defcon 5 "minimal risk"` in both file orders, and the run that produced them exits 0.

Under "raise an error" the third bullet is the bug being reported rather than fixed, so I'd want the error to be raised where those rows are written rather than left to the readers.

**Which do you want?** I can retarget this PR to `aggregate_reports` refusing to combine inputs that repeat a `(generator, probe, detector)` pairing — naming the offending pairing and both files, and leaving the documented one-probe-at-a-time case alone. One wrinkle worth deciding first, because it is your exception rather than mine: your incomplete→remaining case can repeat a pairing. If run 1 wrote eval rows for probe X and run 2 wrote X's remaining rows, the pairing is the same on both sides, so a pairing-level rejection would block precisely the combination you say should still work. Making that case survive looks like a decision at attempt granularity (which attempts each run contributed) rather than at pairing granularity — which is the same bookkeeping pause/resume needs, so it may be a reason to let this wait for that design rather than have me approximate it here.

If you would rather I go ahead with the pairing-level rejection and treat resume-overlap as a known gap, say the word and I will retarget this PR to that. Otherwise I will close #2213 and #2214 and leave the measurements on #2213 for the design work rather than keep a draft open.

AI-assisted: this reply and the PR were drafted with an AI coding agent (Claude, via Qoder CLI); the file:line references and the numbers above are from local runs against `8d1259ef`, and CI on this head has not executed — `DCO: success` is the only completed check, the fork workflow runs still await maintainer approval.

feiiiiii5:
Follow-up on the wrinkle raised in this thread, because it changes what is blocked on what.

**The resume exception does not repeat a pairing.** Checked against `8d1259ef`: an `eval` row is written only from `evaluators/base.py:189-222`, reached only via `evaluator.evaluate(attempt_results)` at `harnesses/base.py:286`, which runs after that probe's attempts are all written, inside the `for probe in probes:` loop — and there is no `try`/`except` between the two. So a probe that was interrupted contributes `attempt` rows and **no** `eval` row. The case described in review ("a subsequent run that contains remaining eval entries of probes that did not reach completion in the first report") is therefore disjoint at `(probe, detector)` granularity by construction, and a pairing-level rejection leaves it working.

**What it does hit is attempt accounting, which no pairing check can fix.** `_process_file_body` carries every `status == 2` attempt, so if run 1 died 30 attempts into a 40-attempt probe and run 2 re-probes it, the aggregate holds 70 completed attempts over 40 prompts, each with its own uuid. `analyze_log` counts those uuid sets (`analyze_log.py:44-50`, `99-102`), so its completion rate and prompt-level lists see the re-run prompts twice. That is the deduplication of completed inference requests that pause/resume has to settle; flagging it so the design isn't scoped to eval rows only.

**Measurements behind the earlier numbers, on real reports at `8d1259ef`.** Two runs of one pairing (`test.Test / always.Pass`, `-g 8` then `-g 4` → `64/64` and `32/32`), aggregated:

| input order | digest detector row | same probe's `_summary.probe_counts` |
|---|---|---|
| 64-chunk first | `passed 64, total_evaluated 64, absolute_score 1.0, absolute_defcon 5` | inference `32`, detection passed `32` |
| 32-chunk first | `passed 32, total_evaluated 32, absolute_score 1.0, absolute_defcon 5` | inference `64`, detection passed `64` |

Both exit 0. One further consequence, unrelated to which reading of aggregation you settle on: `main()` opens `-o` in `w+` at line 105, before the target/version compatibility checks at 169-176, so any rejected combination truncates whatever report was already at that path. Measured: a run that died on `assert target_name == setup["plugins.target_name"]` left 37,318 bytes of partial output where a pre-existing file had been. Validating before opening the output file avoids that regardless of how repeated pairings are decided.

**Status from our side:** holding #2214 as a draft until you choose. If it is the pairing-level rejection, that change is written and tested - `_check_repeated_pairings()` in `aggregate_reports`, called before `-o` is opened, naming the pairing and both contributing files; three tests (two rejection, one asserting the disjoint one-probe-per-job case still aggregates), failing at base on `assert 0 != 0` and passing at head; `tests/analyze` 165 passed locally; the mixed-CI aggregation test retargeted because its two inputs repeated a pairing; and `index_analyze.rst` + the "Probe aggregation" section of `faster.rst` updated to state the constraint. It can be opened as a PR in one step on your word.

AI-assisted: this reply was drafted with a coding agent (Claude, via Qoder CLI); the file:line references and both tables are from local runs against `8d1259ef` on this machine, and no CI has executed on any of it.

feiiiiii5:
Two additions from an independent run. Both cost you nothing to check: this reproduction uses only files already committed to the repo, no model calls and no generated reports.

`tests/_assets/analyze/test.report.jsonl` and `tests/_assets/analyze/test_with_ci.report.jsonl` already share one pairing — `test.Test` / `always.Pass`, 8/8 in the first and 37/50 **with** a bootstrap interval in the second. Aggregating them on `8d1259ef` in each order:

| input order | digest detector row | same probe's `_summary` | interval fields |
| --- | --- | --- | --- |
| `test` first | `absolute_score 1.0`, `passed 8`, `total_evaluated 8` | `probe_score 0.74` | `None`/`None` |
| `test_with_ci` first | `absolute_score 1.0`, `passed 37`, `total_evaluated 50` | `probe_score 0.74` | `None`/`None` |

So one pairing carries two different scores inside a single digest (the detector row and the probe summary), the counts follow the file order, and — the part I had not seen written down anywhere yet — **a confidence interval that exists in the input is dropped in both orders**, because the contributing row holding it is never the one whose values survive. Pooled truth would be 45/58 with some interval; no order gives that.

That is an argument for your direction rather than against it: an aggregated row has no well-defined interval, so any pooling implementation has to invent a rule (ours pops the field, which is why the head shows `45/58` with `None` bounds), whereas rejecting the input keeps every number in a digest traceable to one run.

To reproduce: `python -m garak.analyze.aggregate_reports -o /tmp/a.jsonl tests/_assets/analyze/test.report.jsonl tests/_assets/analyze/test_with_ci.report.jsonl`, then read the `eval` tree from the last line; swap the two file arguments for the second row. Run against a clean tree of `8d1259ef` and this head `b2882913`; no CI has executed on either.

*AI-assisted.*

feiiiiii5:
Superseded by #2216, which follows the direction requested in review: `aggregate_reports` now rejects reports that repeat a probe/detector pairing before any output is written, while the documented disjoint workflow still aggregates. Closing this draft in favour of #2216.

feiiiiii5:
Thanks - that answers the question I could not settle from the code, and I think it says my premise is wrong rather than my repro.

What I assumed is that a second `hit`/`probe/detector` entry for the same pair is worth pooling, because dropping it silently loses real results. You are describing the opposite contract: two completed runs of the same pair are not the same measurement, so pooling them is not aggregation but an invalid merge, and the only legitimate cross-run case is a partial report followed by the entries that did not reach completion. Since the prompt deduplication you mention is missing on my side, my numbers would be a weighted mixture of two runs rather than a digest of one - so I am not going to defend the current shape.

Two ways I can take this, your call:

1. Rework this PR to the narrower rule you described: raise when both colliding entries come from completed runs, keep the incomplete + follow-up case producing one combined digest, and name the offending `probe/detector` pair (and the two report paths) in the message so the user knows what to drop. No dedup logic needed, because the ambiguous case is never merged - a smaller diff than what is here now.
2. Close this and let the pause/resume work own the question, since resuming from a previous report answers the same user need without any merge semantics at all.

If a clear error on invalid merges is useful to you independent of pause/resume, I will do 1 - otherwise tell me to go with 2 and I will close it rather than leave it sitting in the queue. Keeping this in draft either way; nothing has been pushed since your review.

